PR #257, opened 2023-02-18
fix: DODO dex transaction volume cannot be obtained

yongjun925:


dtmkeng:
i can't use gh cli for checkout to check this pr do you have result after run test

dexs/dodo/index.ts
```diff
@@ -47,9 +50,21 @@ const getFetch = (chain: string): Fetch => async (timestamp: number) => {
...
   console.log("DODO volume debug:", JSON.stringify(dailyResponse, null, 2))
+  const dailyData = dailyResponse.data.dashboard_chain_day_data.list.find((item: any) => item.timestamp === dayTimestamp);
+  const dailyVolume = dailyData?.volume[chain];
+  if (dailyData && dailyVolume === '0') {
+    delete dailyData.volume['__typename'];
+    // If the transaction volume data of all chains on the current day is 0, it means that the data acquisition failed, 
+    // and the error status is returned directly, and the data will be retrieved after a period of time 
+    const checkResult = Object.values(dailyData.volume).every(volume => volume === '0');
+    if (checkResult) {
+      throw new Error(`Normal data of current time ${timestamp} is not obtained`)
```
dtmkeng, line 61: not sure if your added exception helps collect data dodo is not equal to 0, I think the problem is caused by the network when the lambda trigger collects data or dodo does not have already data when the lambda trigger, and we have a process for backfilling data

crystalshay2es: Ok what do we do ?

dtmkeng: not sure,
we trigger lambda at 0:00 UTC time I think API should have an already existing volume for yesterday

dtmkeng: we have process auto backfill volume, then I will close this pr, thank you for this PR

PR #334, opened 2023-03-27
fix Wombat Exchange trading volume

corey-wombat:
it is now double counted


dtmkeng:
@corey-wombat  ty for pr
after merge should I refill data Since when?

PR #372, opened 2023-04-03
Add fees and volume adapters for Morphex on Fantom

daedboi:
Morphex is already listed on DefiLlama, along with yield adapters. Protocol is based off of GMX, this PR just adds fees and volume adapters to display. Thank you!


dtmkeng:
@daedboi  hi could you check graph api see it can't return daily volume

PR #454, opened 2023-04-28
Add SMBSwap dimensions adapter

Cr3k:


dtmkeng:
hi, seem we have smbswap but does not include v3, stable swap, could you update this file https://github.com/DefiLlama/dimension-adapters/blob/master/dexs/smbswap/index.ts

dtmkeng:
hmmm ok i will switch for use code thx

PR #590, opened 2023-07-05
add bitkeep volume

bitkeep-tech:


dtmkeng:
volume is from other LP (not from bitkeep LP) ?

bitkeep-tech:
We are aggregators, not from bitkeep LP

dtmkeng:
ok

PR #982, opened 2023-11-15
Remove Derivio volume adapter temporarily

zimrain:
We found some inconsistencies/errors with how our subgraph counts volume, apologies. May you remove the adapter & data temporarily? Thank you


dtmkeng:
> hi @dtmkeng we haven't fixed some of the bugs yet, it's more subtle than we initially thought. May you disable Derivio's volume adapters until further notice?

oh sorry my bad will disable

PR #1124, opened 2024-01-10
add active addresses for squadefi

catcatcat8:
As far as I understand, it is necessary to add my protocol to RouterAddresses file so that the “Active Addresses” button appears on the frontend. If everything is correct, please add it.


dtmkeng:
@0xngmi

PR #1174, opened 2024-02-01
update: added the volume stats for ethereum chain

tharunhack:


dtmkeng:
Can you uncommit file lock?
The adapter type has a type that must be returned as follows. this
```
  dailyVolume?: string // | IJSON<string>;
  totalVolume?: string // | IJSON<string>;
```

PR #1185, opened 2024-02-05
Rename thorchain fees/dex

pluto9r:
1. THORSwap is not a "dex", as defined by Defillama. The liquidity exists on THORChain, which is a sovereign protocol. THORSwap is just one of many UIs building on top of THORChain. This PR renames the fees and dex adapters for THORSwap => THORChain, since the fees/volume being queried are that of THORChain as a protocol (sum of all volume/fees accrued by all UIs e.g. THORSwap, TrustWallet, THORWallet, etc.) not that of THORSwap solely (as a UI).

2. Remove THORWallet. Same as above. Having an adapter for dex volume / fees for a single wallet (THORWallet) that queries an entire protocol's volume / fees (THORChain) is improper methodology.

NOTE: UI updates are probably needed to ensure this doesn't break: https://defillama.com/dexs/thorchain


dtmkeng:
so sorry , i can't rename of file because it link id with our db

PR #1234, opened 2024-02-21
baseswap v3

0xShoegazer:
Brought in v3 data and removed old derivatives items


dtmkeng:
baseswap is not support derivative volume ?  seem you delete derivative volume and add v3

dtmkeng:
is there tvl for v3 ?

dtmkeng:
oh seem call baseX

PR #1262, opened 2024-03-04
fix hyperliquid

dtmkeng:


llamatester:
The hyperliquid adapter exports: 
        
 
 _______________________________________
    Dexs for 3/3/2024
    _______________________________________
    
    ARBITRUM 👇
    Backfill start time: 25/2/2023
    NO METHODOLOGY SPECIFIED
    Total volume: 79.60 B
    Daily volume: 1.60 B
    Timestamp: 1709424000 (2024-03-03T00:00:00.000Z)
    
    
    
    
    Done in 7.41s.

chameleon-jeff:
Hey, why was this change necessary? The stats page actually updates around 4 UTC every day and occasionally fails, which leads to some inaccurate numbers for Hyperliquid.

The globalStats endpoint updates every 5 minutes and is very reliable. Is it possible to continue only using that as before?

Thanks!

dtmkeng:
> Hey, why was this change necessary? The stats page actually updates around 4 UTC every day and occasionally fails, which leads to some inaccurate numbers for Hyperliquid.
> 
> The globalStats endpoint updates every 5 minutes and is very reliable. Is it possible to continue only using that as before?
> 
> Thanks!

ppl complaint our value don't match this page https://stats.hyperliquid.xyz/

chameleon-jeff:
> > Hey, why was this change necessary? The stats page actually updates around 4 UTC every day and occasionally fails, which leads to some inaccurate numbers for Hyperliquid.
> > The globalStats endpoint updates every 5 minutes and is very reliable. Is it possible to continue only using that as before?
> > Thanks!
> 
> ppl complaint our value don't match this page https://stats.hyperliquid.xyz/

I see. Do you have examples for the complaints? I wonder if they are using the non-HLP volume graph, which is two-sided but excludes HLP trades. Or maybe they looked before the stats page updated for that day.

The "total volume" graph on the stats page should match the globalStats endpoint for each date.

dtmkeng:
> > > Hey, why was this change necessary? The stats page actually updates around 4 UTC every day and occasionally fails, which leads to some inaccurate numbers for Hyperliquid.
> > > The globalStats endpoint updates every 5 minutes and is very reliable. Is it possible to continue only using that as before?
> > > Thanks!
> > 
> > 
> > ppl complaint our value don't match this page https://stats.hyperliquid.xyz/
> 
> I see. Do you have examples for the complaints? I wonder if they are using the non-HLP volume graph, which is two-sided but excludes HLP trades. Or maybe they looked before the stats page updated for that day.
> 
> The "total volume" graph on the stats page should match the globalStats endpoint for each date.

image
i can switch use prev version if you want

chameleon-jeff:
> > > > Hey, why was this change necessary? The stats page actually updates around 4 UTC every day and occasionally fails, which leads to some inaccurate numbers for Hyperliquid.
> > > > The globalStats endpoint updates every 5 minutes and is very reliable. Is it possible to continue only using that as before?
> > > > Thanks!
> > > 
> > > 
> > > ppl complaint our value don't match this page https://stats.hyperliquid.xyz/
> > 
> > 
> > I see. Do you have examples for the complaints? I wonder if they are using the non-HLP volume graph, which is two-sided but excludes HLP trades. Or maybe they looked before the stats page updated for that day.
> > The "total volume" graph on the stats page should match the globalStats endpoint for each date.
> 
> image i can switch use prev version if you want

Yeah, can you switch it back to the old endpoint? The old endpoint always returns rolling 24 hour volume from the time it is queried. As long as it's queried around 0 UTC it should be accurate.

dtmkeng:
okay thank you

lmlmt:
Hey @dtmkeng. What is the frequency that this adapter is called? Is it possible to have whatever calls the adapter to query Hyperliquid's rolling stats endpoint once per day, say around 0 UTC?  Not super familiar with the code that calls this adapter so figured I'd ask.

dtmkeng:
> Hey @dtmkeng. What is the frequency that this adapter is called? Is it possible to have whatever calls the adapter to query Hyperliquid's rolling stats endpoint once per day, say around 0 UTC? Not super familiar with the code that calls this adapter so figured I'd ask.

we change call every hour

chameleon-jeff:
Hey @dtmkeng thanks for the reply. We have users complaining that Hyperliquid stats are not comparable directly to other DEXs because most DEXs only update their volume endpoint once per day while Hyperliquid's endpoint is real-time 24 hour tracking. 

What's the intended return value of the 24 hour volume endpoint? We could change to last UTC date but I'm assuming that you guys want real-time volume because you're polling every hour?

chameleon-jeff:
> > Hey @dtmkeng thanks for the reply. We have users complaining that Hyperliquid stats are not comparable directly to other DEXs because most DEXs only update their volume endpoint once per day while Hyperliquid's endpoint is real-time 24 hour tracking.
> > What's the intended return value of the 24 hour volume endpoint? We could change to last UTC date but I'm assuming that you guys want real-time volume because you're polling every hour?
> 
> just run update value every hour i can set flag run one time if you need it will run at 0:00 UTC

Ok that sounds good. If you can set it to run once at 0:00 UTC for Hyperliquid every day that would be great since it will return previous UTC day volume. It would be nice to retry if the request fails but if it succeeds then keep that value for the UTC day.

Thanks again for your help!

dtmkeng:
okay will fix

PR #1273, opened 2024-03-07
fix: rename kizuna to kim and update subgraph url

agileurbanite:
fix: rename kizuna to kim and update subgraph url


dtmkeng:
kizuna is for v1 you should new adapter name kim for v2

agileurbanite:
https://defillama.com/protocol/kim-exchange

dtmkeng:
i mean current adapter you modify is v1 and new endpoint is for v2 you should create new adapter for v2 instead modify v1
image

agileurbanite:
I see okay will do that and close this PR

PR #1531, opened 2024-05-30
add bungee + a few jumper chains

Octave-byte:


Octave-byte:
Add Bungee in the bridge-aggregator section 
Add some Jumper chains (Manta, XDAI, Linea, Aurora)

PR #1573, opened 2024-06-11
Add maverick v2 and move maverick v1 off of subgraph data

gte620v:
Adds Maverick v2 and moves Maverick v1 fee and volume data to a getLogs paradigm.

The existing Maverick v1 integration may stop working soon given the The Graph is shifting to an API-key pay-per-call model.


gte620v:
I think there is some sort of rate limiting in the bsc rpc.  The call that errors out definitely works. 

It works on other chains and it works if I make the call individually on bsc.  Perhaps there is a rate limit on these public RPCs with the multicall?

e.g.
```
> curl --location 'https://bsc-dataseed1.defibit.io' \
--header 'Content-Type: application/json' \
--data '{"method":"eth_call","params":[{"from": null,"to":"0xf23e49fc9da1400a92901207358710f716a34590","data":"0x2d9d1ca60
00000000000000000000000000000001"},"latest"],"id":1,"jsonrpc":"2.0"}'

{"jsonrpc":"2.0","id":1,"result":"0x000000000000000000000000000000000000000000000000000009184e72a000"}%       
```

dtmkeng:
could you create new file for v2, and v3

gte620v:
instead of this breakdown adapter?

dtmkeng:
yes it good for backfill data

gte620v:
> yes it good for backfill data

done; still getting some bsc rpc errors, but it works locally.

PR #1588, opened 2024-06-17
add BladeSwap dex

Bladeswap:
Hello! this is our adapter for bladeswap. Thank you!


Bladeswap:
It was CL only. I've added amm volumes too. Thank you for reminding me of that.

PR #1589, opened 2024-06-17
Add volume for Aktionariat

rube-de:
Add volume for https://defillama.com/protocol/aktionariat

(Aktionariat is now also live on Polygon)


dtmkeng:
is it swap volume ?

rube-de:
yes, it's swap volume of the AMM pools

PR #1600, opened 2024-06-21
Hop Aggregator - updated url to https

bonkman22:
Hello, in our original PR for hop aggregator, our api url was in http and this is not letting it load on defi llama. We have changed the URL and it show now load. Please accept this PR


bonkman22:
Hey @dtmkeng, do you have any idea, why our stats are not loading?

https://defillama.com/protocol/hop-aggregator?denomination=USD#aggregators-volume

dtmkeng:
@bonkman22  could you check API seem error `Gateway Timeout`

PR #1603, opened 2024-06-21
feat: add milkyway fee adaper

manu0466:
Add [MilkyWay](https://www.milkyway.zone/) liquid staking protocol fee adapter

### DIFF
--- a/fees/milkyway.ts
+++ b/fees/milkyway.ts
@@ -37,7 +37,7 @@ const fetch: FetchV2 = async ({ startTimestamp, endTimestamp }) => {
 
 const adapter: Adapter = {
   adapter: {
-    celestia: {
+    osmosis: {
       fetch,
       runAtCurrTime: false,
       // 2024-04-30T17:55:14:00Z


fees/milkyway.ts
```diff
@@ -0,0 +1,59 @@
...
+    dailyFees: dailyStakingRewards,
+    dailyRevenue: dailyUserFees,
+    dailyProtocolRevenue: dailyUserFees,
+    dailySupplySideRevenue,
+  };
+};
+
+const adapter: Adapter = {
+  adapter: {
+    celestia: {
```
dtmkeng: why chain is not `Osmosis` like tvl ?

dtmkeng (changes requested):

fees/milkyway.ts
manu0466: Sorry, my bad I though that I should put the chain of the fee token instead of the one where the protocol is deployed. I'll update it.

waynebruce0x:
@manu0466 we're getting `Error getting fees: missing destination name amount in *postgres.receiveRewardsRow` from the endpoint in the milkyway adapter - is this still maintained? thanks

manu0466:
@waynebruce0x sorry for the delay, the api now don't return the error any more. I'm still investigating why is returning 0 fees and commissions. I'll let you know when is fixed.

manu0466:
@waynebruce0x The issue is fixed we are now reindexing the data from september 25th to october 7th

PR #1604, opened 2024-06-21
fixed timestamp errors

bonkman22:
I believe this is causing issues with loading data on the DefiLLama dashboard for Hop Aggregator. The fixed timestamps should now make it work properly on here:

https://defillama.com/protocol/hop-aggregator#aggregators-volume


dtmkeng:
will check

PR #1615, opened 2024-06-25
Add TLX Finance Volume

chase-manning:


dtmkeng:
event Minted and Redeemed it mean you user trade volume ? 
and seem txs is direct to synthetix ?

chase-manning:
> event Minted and Redeemed it mean you user trade volume ? and seem txs is direct to synthetix ?

@dtmkeng Yeah we are a Leveraged Token protocol. So we wrap the Synthetix Positions as ERC20s. When users mint a Leveraged Token, we asynchronously open a matching Synthetix Position. So for us, minting and redeeming is essentially opening and closing a position.

chase-manning:
Transactions are not direct to Synthetix, they are indirect to Synthetix. As it's done as a separate asynchronous step. This is because we want the minting and redeeming of leveraged tokens to be atomic. But transactions on Synthetix are not atomic.

PR #1619, opened 2024-06-26
fix(volume): adding timestamp in DUSA  volume api

ayabelarbi:
Adding timestamp in DUSA api, to allow the fetch of the right data


ayabelarbi:
i can't fix the small conflicts, as I don't have write access

dtmkeng:
fixed https://github.com/DefiLlama/dimension-adapters/commit/b3e4fe96016b707a6debe78c2233453b5cda7df4

PR #1627, opened 2024-07-01
add fee in JOJO

JoscelynFarr:
add fee collection in JOJO

### DIFF
--- a/fees/jojo/index.ts
+++ b/fees/jojo/index.ts
@@ -1,0 +1,31 @@
+import { FetchOptions, SimpleAdapter, FetchV2, FetchResultV2 } from '../../adapters/types'
+import { httpGet } from '../../utils/fetchURL'
+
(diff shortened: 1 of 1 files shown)


fees/jojo/index.ts
```diff
@@ -0,0 +1,31 @@
...
+    const dailyFees = item.flat().find((i) => i.day.split('T')[0] === dateStr)?.fee
+    return {
+        dailyFees: dailyFees ? dailyFees : undefined,
+    }
+}
+
+const adapter: SimpleAdapter = {
+    version: 2,
+    adapter: {
+        "jojo": {
```
dtmkeng, line 24: we don't have jojo chain which chain you need list ?

JoscelynFarr: base Thanks

PR #1595, opened 2024-06-19
Updated Rubicon Volume Adapter

bghughes:
Hey dev/admin - **I need a bit of help. How can I marry my two adapters? We have the old one, which was working, but cuts off around January 2 of this year, then I would like to transition our volume to the new one while keeping all the data from the old one in one continuous stream. Can you please provide guidance?**

This is incomplete in my mind and I wanted to ask to see if you had guidance on how to do this. Sorry if I am missing something and thank you in advance 🙏

PR #1631, opened 2024-07-01
Allow Rubicon Volume adapter to back fill

bghughes:
Hey @dtmkeng - would this fix our historical charts?

Cc #1595


bghughes:
Oh wow, now all of a sudden it is working in production. Going to close this then! I guess the problem resolved itself on its own. 😄 

Sorry for the confusion...

PR #1632, opened 2024-07-01
trying to fix the volume data issue 

ayabelarbi:


ayabelarbi:
as i don't have write access, i cannot fix the conflicts

dtmkeng:
i fixed support historical data already

ayabelarbi:
yes but, it's a different code. i've remove 24h to the time in order to make my volume api code works

dtmkeng:
we collect data every hour for current data when your data your update, i think some your data is misstake when we collect data

dtmkeng:
i refilled your data will okay next 2-3 hour

PR #1633, opened 2024-07-01
Update fee split

0xShad0w:
Hi please update fee split for all Ramses forks

Thank you


waynebruce0x:
@0xShad0w the subgraph endpoint in this file seems to be stale - is this still maintained? thanks

PR #1649, opened 2024-07-05
removing 24h to the timestamp to get the right data from the api 

ayabelarbi:


dtmkeng:
if you need change you should pull latest update at this repo

PR #1776, opened 2024-08-15
patch

waynebruce0x:


g1nt0ki (commented):
my recommendation would be to leave existing implementation as is. And create a new function for new implementation

Then, update each adapter to use the new function and test that it works, it is tedious but safer and one time change. This way, we can still support adapters that dont work with blocks

PR #1876, opened 2024-09-08
remove fees tracking for echelon

LawsonGraham:
we decided we don't want to dispaly fees for now - just removing the endpoint


dtmkeng:
why do you want to remove ?

PR #1932, opened 2024-09-20
add yellow router addresses

dimast-x:
Hi! We want to add Active Users metric to the Yellow project. I set up yellow router contract addresses. Please let us know if anything else is needed, and how we can test this

### DIFF
--- a/users/routers/routerAddresses.ts
+++ b/users/routers/routerAddresses.ts
@@ -2408,5 +2408,16 @@ export default (
           bsc: ["0x888888888889758F76e7103c6CbF23ABbF58F946"],
       }
   },
+  {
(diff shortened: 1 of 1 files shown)


users/routers/routerAddresses.ts
```diff
@@ -2408,5 +2408,16 @@ export default (
           bsc: ["0x888888888889758F76e7103c6CbF23ABbF58F946"],
       }
   },
+  {
+    id: "parent#yellow",
```
dtmkeng: could you change `parent#yellow` to  `5157`

dtmkeng (changes requested):

PR #2326, opened 2025-01-17
rebrand spark to v12

PaulZhemanov:


dtmkeng:
rebrand you can inform at llama discord, rename file is affect a lot

PR #2459, opened 2025-02-25
Update assets for LightLink network

AmpedFinance:


g1nt0ki (changes requested):
again, renaming chain breaks things. and adding LL here will have 0 effect. https://github.com/DefiLlama/DefiLlama-Adapters/discussions/13627

PR #2511, opened 2025-03-06
feat: add hyperfluid dimension adapter

alex4506:


alex4506:
There is a ACL rule, already fix, please run again

PR #2846, opened 2025-04-09
Adding Echo Lending/LSD fees

Yaron1902:
Echo lending fees in fees/echo-lending/index.ts
Echo LSD fees in fees/echo-lsd/index.ts


fees/echo-strategy/index.ts
```diff
@@ -0,0 +1,48 @@
...
+interface EchoStrategyStats {
+  data:{
+    fee: string;
+    revenue: string;
+  }
+}
+
+const methodology = {
+  Fees: 'Strategy rewards earned by all staked APT',
+  ProtocolRevenue: 'Strategy rewards',
```
g1nt0ki, line 16: hi @Yaron1902 
thanks for the PR, but what % of the fees go to the protocol in each of these listings? atm, it is 100% ?

g1nt0ki (changes requested):
.

fees/echo-strategy/index.ts
Yaron1902: @g1nt0ki  yes, currently it's 100%

g1nt0ki: sorry, missed the comment, so when someone deposits money for lending on Echo, all the borrow interest paid by the users go to the protocol and the lenders get nothing?

echo-pro-dev: @g1nt0ki Hi, the user who deposits on Echo earns the borrow interest absolutely. The fees we reported here are only the portion of the interest Echo charges from the borrow interest.

echo-pro-dev:
@g1nt0ki Hi, are there any other issues about this PR? We beg you, how could it be merged since this PR has been open for a long time? thanks!

g1nt0ki:
hi @echo-pro-dev 
is there a way to get the yield that goes to the user? in our terminology, fees = what borrowers pay, revenue = what goes to the protocol, supply side revenue = what goes to the lenders

image

noelportillo (approved):

Yaron1902:
@g1nt0ki  Is the only concern right now about Lending? LSD and Strategy are fine, correct? If so, we can resubmit a PR containing only LSD and Strategy to get it merged as quickly as possible.

g1nt0ki:
hi @Yaron1902 same concern for the others, when users deposit in echo strategy or echo lsd, they get no yield?

PR #2897, opened 2025-04-15
update d8x chains

m66260:
updating list of chains


g1nt0ki (changes requested):
sorry no, rejecting this, the data is correct, also if we merge this, we will still end up showing past data for these chains

PR #3205, opened 2025-05-18
feat: new fee wallets

henryzhang03:


treeoflife2:
@henryzhang03 does all fee goes to the axiom tresuary? as we also need to add dailyProtocolRevenue and dailyUserFees? 

treeoflife2:
> Yea it does but we give back a chunk to our users. Not sure what changes need to be made because all I'm doing is changing the fee wallet array.

Don't worry about the changes, i can make the changes. is there any way where we can track how much fees go back to users? so we have accurate information?

henryzhang03:
That info isn't available atm. We'd like to just add these fee wallets for now.

treeoflife2:
> That info isn't available atm. We'd like to just add these fee wallets for now.

then i will remove dailyRevenue as it's not accurate, as not all fee goes back to the protocol.

henryzhang03:
It's not a protocol... It's similar to photon and bullx so that's why we have it as such. Revenue is equivalent to fees, there just isn't any holder revenue.

treeoflife2:
> It's not a protocol... It's similar to photon and bullx so that's why we have it as such. Revenue is equivalent to fees, there just isn't any holder revenue.

Protocol as in "Axiom" not solana or anything else. 

Here's our revenue definition: 
dailyRevenue: (Required) The portion of dailyFees kept by the protocol entity itself, distributed either to the treasury (dailyProtocolRevenue) or governance token holders (dailyHoldersRevenue).

PR #3233, opened 2025-05-21
Update index.ts

Octave-byte:


llamabutler:
The lifi adapter exports:
        
 
 
    > adapters@1.0.0 test
    > ts-node --transpile-only cli/testAdapter.ts bridge-aggregators lifi
    
    🦙 Running LIFI adapter 🦙
    bridge-aggregators/lifi/index.ts(7,19): error TS1005: ',' expected.

noateden:
hi @Octave-byte, we're tracking all transfer on Lifi, why do we need to filter integrations?

Octave-byte:
Hey @noateden . In the same style than Socket/Bungee are doing we want to separate the volume coming from the B2C interface (which is Jumper) from all the B2B business (which LI.FI - Jumper), that's the reason.

Also do you know if LI.FI figures could be shown in the Defillama UI as well ?

treeoflife2:
> Hey @noateden . In the same style than Socket/Bungee are doing we want to separate the volume coming from the B2C interface (which is Jumper) from all the B2B business (which LI.FI - Jumper), that's the reason.
> 
> Also do you know if LI.FI figures could be shown in the Defillama UI as well ?

This makes sense, as it won't double count volumes

PR #3410, opened 2025-06-09
[BLUEFIN]: Add 7k aggregator under Bluefin and rename to Bluefin7k Aggregator

abdullahabro:
## Description 
- Hey team, we want to move the 7k aggregator to be under Bluefin as a parent project 
- Also, 7k Aggregator needs to be renamed to Bluefin7k Aggregator (https://x.com/bluefinapp/status/1929596937181032908)


noateden:
hi @abdullahabro, thanks for the PR, pls don't rename the file, it makes more complicated to update.
please revert the file name, we will support you to update protocol info

PR #3476, opened 2025-06-18
Add yala adapter

yala-hub:
sum all fees and revenue

### DIFF
--- a/fees/yala/index.ts
+++ b/fees/yala/index.ts
@@ -1,0 +1,58 @@
+import { Adapter, FetchOptions, FetchResultV2 } from "../../adapters/types";
+import { getConfig } from "../../helpers/cache";
+import { CHAIN } from "../../helpers/chains";
+import { addTokensReceived, nullAddress } from "../../helpers/token";
+
+const methodology = {
+  Fees: "Sum of all fees",
+  Revenue: "Sum of all revenue",
+};
+
+async function fetch(options: FetchOptions): Promise<FetchResultV2> {
+  const dailyFees = options.createBalances();
+
+  const res = await getConfig(
+    "yala/ethereum",
+    "https://raw.githubusercontent.com/yalaorg/yala-defillama/refs/heads/main/config.json"
+  );
+
+  const dailySupplySideRevenue = options.createBalances();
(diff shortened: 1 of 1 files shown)


treeoflife2:
@yala-hub can we also track bitcoin and solana fees too? if have any doubts let me know.

yala-hub:
> @yala-hub can we also track bitcoin and solana fees too? if have any doubts let me know.

Currently, the main logic of our platform is still based on the Ethereum platform.

On the Bitcoin side, it's just a cross-chain transfer process. The transaction fee is the fee for the transfer. It has nothing to do with the platform, so it was not included.

On the Solana side, for the time being, we also use LayerZero for cross-chain transfer of the stablecoin Yu. Then, we participate in yield on some cooperating platforms. The transaction fees are not related to the platforms. We are also developing native applications for Solana. Once they are launched, we will include these fees.

treeoflife2:
> > @yala-hub can we also track bitcoin and solana fees too? if have any doubts let me know.
> 
> Currently, the main logic of our platform is still based on the Ethereum platform.
> 
> On the Bitcoin side, it's just a cross-chain transfer process. The transaction fee is the fee for the transfer. It has nothing to do with the platform, so it was not included.
> 
> On the Solana side, for the time being, we also use LayerZero for cross-chain transfer of the stablecoin Yu. Then, we participate in yield on some cooperating platforms. The transaction fees are not related to the platforms. We are also developing native applications for Solana. Once they are launched, we will include these fees.

thanks for clarification, is there a token, if yes, is there buyback/burns from protocol revenue? as we also track other dimensions too like dailyFees, dailyRevenue, dailyProtocolRevenue, dailyHoldersRevenue

yala-hub:
> thanks for clarification, is there a token, if yes, is there buyback/burns from protocol revenue? as we also track other dimensions too like dailyFees, dailyRevenue, dailyProtocolRevenue, dailyHoldersRevenue

Not yet, but the related governance token are currently being designed. Later on, after it goes live, I will make further updates.

treeoflife2:
made changes in new PR(#3505), as edits are not allowed in this PR

PR #3482, opened 2025-06-18
dead project

treeoflife2:


g1nt0ki (commented):
I just fixed it, eth part still have 100k tvl and think legit volume

PR #3564, opened 2025-07-01
(tonco) filter out v1.5

lilchizh:


lilchizh:
> hi @lilchizh thanks for the PR. can you pls explain why are you filter out version of 1.5?

hi! 

It's not production ready yet, and is being used privately for testing, volumes in v1.5 so far don't represent real liquidity movements, so they are should not be exposed for now

PR #3643, opened 2025-07-09
Delete LiquidOps incorrect fees adapter

LorimerJenkins:
Hey!

This adapter is wrong, please can we delete the data from DeFi Llama too and delete this adapter. 

It is giving incorrect data saying the protocol made $200k this year.

Thanks!


treeoflife2:
@LorimerJenkins please add correct fee calculation, we will refill the historical data

PR #3675, opened 2025-07-13
fix: rollbit holder revenue logic as they have migrated from solana to eth

KPHEMRAJ:


g1nt0ki (changes requested):
sorry, no. we are not using this adapter, and we are not tracking rollbit revenue because we cant verify it.

PR #3678, opened 2025-07-14
Meso finance add revenue

jack-meso:


treeoflife2:
@jack-meso can you add the dailysupplysiderevenue too, you can see definition here. 

jack-meso:
@treeoflife2 Yes, we have updated

treeoflife2:
@jack-meso we are getting incorrect data from the API, which is inflating the fees data, 
https://api.meso.finance/api/v1/Tool/defillama/fees?timeframe=
please fix this otherwise we would need to make it onchain due to incorrect data from API.
i fixed on bug, but issue is with API data, current day and yesterday data gap looks huge.

PR #3696, opened 2025-07-14
added more chains tracking rabby fees

KPHEMRAJ:


treeoflife2:
@KPHEMRAJ fee wallets are same for all new chains?

PR #3717, opened 2025-07-16
Optfun: Add PUMP market

lukmaan-k:
Same methodology as existing BTC market


lukmaan-k:
Thanks, left a comment there

PR #3726, opened 2025-07-16
Update elys's total volume by adding perpetual volume

99adarsh:


treeoflife2:
@99adarsh we have to create separate listing to track perp volume, as dex volume and perp volume are track separately.

treeoflife2:
created new listing for perp at #3745 , closing this PR

PR #3761, opened 2025-07-18
Update Odos V3 Multiswap

mdeib:
Fix V3 multiswap event and adjust fees for V3

### DIFF
--- a/aggregators/odos/index.ts
+++ b/aggregators/odos/index.ts
@@ -8,7 +8,7 @@ const event_multiswap = 'event SwapMulti(address sender, uint256[] amountsIn, ad
 const ODOS_ROUTER_V3 = "0x0D05a7D3448512B78fa8A9e46c4872C88C4a0D05"
 
 const event_swap_v3 = 'event Swap (address sender, uint256 inputAmount, address inputToken, uint256 amountOut, address outputToken, int256 slippage, uint64 referralCode, uint64 referralFee, address referralFeeRecipient)';
-const event_multiswap_v3 = 'event SwapMulti(address sender, uint256[] amountsIn, address[] tokensIn, uint256[] amountsOut, address[] tokensOut, uint64 referralCode, uint64 referralFee, address referralFeeRecipient)';
+const event_multiswap_v3 = 'event SwapMulti(address sender, uint256[] amountsIn, address[] tokensIn, uint256[] amountsOut, address[] tokensOut, int256[] slippage, uint64 referralCode, uint64 referralFee, address referralFeeRecipient)';
 
 type TPool = {
   [c: string]: string[];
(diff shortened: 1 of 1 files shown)


aggregators/odos/index.ts
```diff
@@ -8,7 +8,7 @@ const event_multiswap = 'event SwapMulti(address sender, uint256[] amountsIn, ad
 const ODOS_ROUTER_V3 = "0x0D05a7D3448512B78fa8A9e46c4872C88C4a0D05"
 
 const event_swap_v3 = 'event Swap (address sender, uint256 inputAmount, address inputToken, uint256 amountOut, address outputToken, int256 slippage, uint64 referralCode, uint64 referralFee, address referralFeeRecipient)';
-const event_multiswap_v3 = 'event SwapMulti(address sender, uint256[] amountsIn, address[] tokensIn, uint256[] amountsOut, address[] tokensOut, uint64 referralCode, uint64 referralFee, address referralFeeRecipient)';
+const event_multiswap_v3 = 'event SwapMulti(address sender, uint256[] amountsIn, address[] tokensIn, uint256[] amountsOut, address[] tokensOut, int256[] slippage, uint64 referralCode, uint64 referralFee, address referralFeeRecipient)';
```
treeoflife2, line 11: was this incorrect event from the beginning? or is this new update? if it's new update then we have to track older one too, so in case of refill, it works correctly for historical data.

aggregators/odos/index.ts
```diff
@@ -58,7 +58,7 @@ async function fetch({ getLogs, createBalances, chain }: FetchOptions) {
 
   // add v3 fees
   logs_v3.forEach(i => dailyFees.add(i.outputToken, Number(i.slippage) > 0 ? i.slippage : 0))
-  multiswapLogs_v3.forEach(i => dailyFees.add(i.tokensOut, i.amountsOut.map((a: any) => Number(a) * .01/100))) // 0.01% fixed fee
+  multiswapLogs_v3.forEach(i => dailyFees.add(i.tokensOut, i.slippage.map((a: any) => Number(a) > 0 ? a : 0)))
```
treeoflife2, line 61: when did the fees changes, we have to track older fees using timestamp condition in codebase, so we have accurate data

treeoflife2 (changes requested):

aggregators/odos/index.ts
mdeib: This was incorrect in the original PR, the V3 address has only had this event and not the other one (V3 changes were just merged before this PR).

aggregators/odos/index.ts
mdeib: Similar to the above, this was just incorrect in the original PR that was just merged - multi fee functionality was removed in V3, now it is just positive slippage collection.

PR #3785, opened 2025-07-21
Fix Penpie revenue, and add protocol rev + bribes + holders revenue

treeoflife2:
https://docs.penpiexyz.io/fee-structure

### DIFF
--- a/fees/penpie.ts
+++ b/fees/penpie.ts
@@ -113,8 +113,8 @@ const meta = {
   methodology: {
     Fees: 'Total boosted PENDLE rewards from liquidity farming on Pendle Finance',
     Revenue: 'Protocol revenue from boosted PENDLE rewards',
-    ProtocolRevenue: '5% vlPNP holders + 5% operational expenses + bribes',
-    HoldersRevenue: '12% mPENDLE Staking pool + 20% PRT mechanism',
+    ProtocolRevenue: '5% operational expenses + bribes share of protocol',
+    HoldersRevenue: '5% vlPNP holders + 12% mPENDLE Staking pool + 20% PRT mechanism',
     SupplySideRevenue: '78% to liquidity providers',
     BribesRevenue: 'Bribes from voting incentives'
   }


fees/penpie.ts
```diff
@@ -29,69 +25,114 @@ const exclude: any = {
...
 }
 
 const meta = {
   methodology: {
-    Fees: 'Staking rewards collected from assets staked on Pendle',
-    Revenue: 'Staking rewards collected from assets staked on Pendle',
+    Fees: 'Total boosted PENDLE rewards from liquidity farming on Pendle Finance',
+    Revenue: 'Protocol revenue from boosted PENDLE rewards',
+    ProtocolRevenue: '5% vlPNP holders + 5% operational expenses + bribes',
+    HoldersRevenue: '12% mPENDLE Staking pool + 20% PRT mechanism',
```
noateden: should we count mPENDLE fees as supplySide revenue? since PENDLE is not project token

fees/penpie.ts
```diff
@@ -29,69 +25,114 @@ const exclude: any = {
...
+  };
 }
 
 const meta = {
   methodology: {
-    Fees: 'Staking rewards collected from assets staked on Pendle',
-    Revenue: 'Staking rewards collected from assets staked on Pendle',
+    Fees: 'Total boosted PENDLE rewards from liquidity farming on Pendle Finance',
+    Revenue: 'Protocol revenue from boosted PENDLE rewards',
+    ProtocolRevenue: '5% vlPNP holders + 5% operational expenses + bribes',
```
noateden: should we count vlPNP fees as holders revenue? since this token is from project

noateden (commented):
I'm confused fees for vlPNP and mPENDLE holders.

fees/penpie.ts
treeoflife2: agree, we should count it as holdersRevenue

fees/penpie.ts
treeoflife2: it's still stacking rewards, so we have to keep it as holdersRevenue, and i agree it's not project token, but we have to still count it as holderRevenue

noateden (commented):
PR is good now

PR #3802, opened 2025-07-23
fix: incorrect revenue of binance eth

KPHEMRAJ:
Source : https://defirate.com/staking/ (The same link which was already added in code comment)


treeoflife2:
@KPHEMRAJ can you verify if it was protocol revenue was changed from 25% to 10%? if that's the case then we have to handle it with the timestamp condition.

treeoflife2:
> > @KPHEMRAJ can you verify if protocol revenue was changed from 25% to 10%? if that's the case then we have to handle it with the timestamp condition.
> 
> @treeoflife2 imo revenue percentage keeps changing , so wouldnt it be better if we do evmReceivedGasAndTokens?

yes that would be better.

PR #3803, opened 2025-07-23
fix: jito fees

KPHEMRAJ:
Source : https://solanacompass.com/stake-pools/Jito4APyf642JPZPx3hGc6WWJ8zPKtRbRs4P815Awbb


treeoflife2:
@KPHEMRAJ i think new listing like jito-staked-sol would be better

PR #3839, opened 2025-07-26
fix: blazestake revenue

KPHEMRAJ:


treeoflife2:
Fixed in https://github.com/DefiLlama/dimension-adapters/pull/3842

PR #3855, opened 2025-07-28
fix: lis-bnb fees and revenue

KPHEMRAJ:
https://docs.bsc.lista.org/introduction/liquid-staking-slisbnb/rewards-and-fees

### DIFF
--- a/fees/lista-slisbnb/index.ts
+++ b/fees/lista-slisbnb/index.ts
@@ -21,28 +21,50 @@ const slisBNB = "0xb0b84d294e0c75a6abe60171b70edeb2efd14a1b";
 const fetch = async (options: FetchOptions) => {
   const dailyFees = options.createBalances();
 
-  const logs_reward = await options.getLogs({
-    target: ListaStakeManagerAddress,
-    eventAbi: "event RewardsCompounded(uint256 _amount)",
+  const slilsBnbSupplyBefore = await options.fromApi.call({
+    target: slisBNB,
+    abi: 'uint256:totalSupply',
   });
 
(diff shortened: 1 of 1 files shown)


fees/lista-slisbnb/index.ts
```diff
@@ -21,28 +21,50 @@ const slisBNB = "0xb0b84d294e0c75a6abe60171b70edeb2efd14a1b";
...
 
-  logs_reward.forEach((log) => {
-    const amount = log._amount;
-    dailyFees.add(slisBNB, amount);
+  const pooledBnbAfter = await options.toApi.call({
+    target: ListaStakeManagerAddress,
+    abi: 'uint256:getTotalPooledBnb',
   });
 
+  const dailyBnbYield = (pooledBnbAfter / slisBnbSupplyAfter - pooledBnbBefore / slilsBnbSupplyBefore) * (slisBnbSupplyAfter / 1e18);
```
noateden: hi, I believe that these yields are distributed to slisBnb holders only which means that protocol fees were  took already. Here is a same case you can check into:
https://github.com/DefiLlama/dimension-adapters/blob/master/fees/binance-staked-eth.ts#L33

KPHEMRAJ: > hi, I believe that these yields are distributed to slisBnb holders only which means that protocol fees were took already. Here is a same case you can check into: https://github.com/DefiLlama/dimension-adapters/blob/master/fees/binance-staked-eth.ts#L33

Hey @noateden  I have fixed accordingly

PR #3868, opened 2025-07-29
add: multipli revenue

SrivatsaJ:


treeoflife2:
@SrivatsaJ sorry for delay, but we won't be assuming dailyFees/dailyRevenue based on yield rates, we need accurate method to track the numbers, like total yield difference b/w two days and by doing subtraction we get any given day yield, or some other onchain way to track the yield.

PR #3875, opened 2025-07-30
Update to Amped Finance, seperate fees and revenue between supply and protocol

AmpedFinance:


treeoflife2:
@AmpedFinance when did the 30% revenue started flowing to the amped stakers(date)? does protocol earn anything, as 30% going to amped staker will be counted as the holder revenue as per our definitions.

treeoflife2:
merged in https://github.com/DefiLlama/dimension-adapters/pull/3879

PR #3898, opened 2025-07-31
Opensea Bridge-aggregator and Aggregator

treeoflife2:
https://opensea.notion.site/Identifying-OpenSea-Swaps-23a162eb75d9809f8ddac26fe3107800

### DIFF
--- a/aggregators/opensea/index.ts
+++ b/aggregators/opensea/index.ts
@@ -1,0 +1,58 @@
+import { FetchOptions } from "../../adapters/types";
+import { queryDuneSql } from "../../helpers/dune";
+import { CHAIN } from "../../helpers/chains";
+import { SimpleAdapter } from "../../adapters/types";
+
+const chainConfig = {
+	[CHAIN.ETHEREUM]: {start: 1720000000, dune_chain: 'ethereum'},
+	// [CHAIN.ABSTRACT]: {start: 1720000000, dune_chain: 'abstract'},
+	[CHAIN.APECHAIN]: {start: 1720000000, dune_chain: 'apechain'},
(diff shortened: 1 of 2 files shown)


noateden (changes requested):
hi tree, can you pls remove Dune and use our indexer instead? like we did with Metamask here
https://github.com/DefiLlama/dimension-adapters/blob/master/aggregators/metamask.ts

noateden:
> > hi tree, can you pls remove Dune and use our indexer instead? like we did with Metamask here https://github.com/DefiLlama/dimension-adapters/blob/master/aggregators/metamask.ts
> 
> opensea doesn't have any router Address, they simply use relay and add calldata at the end, more details here, https://opensea.notion.site/Identifying-OpenSea-Swaps-23a162eb75d9809f8ddac26fe3107800

but we can still detect by transactions data, right?

treeoflife2:
> > > hi tree, can you pls remove Dune and use our indexer instead? like we did with Metamask here https://github.com/DefiLlama/dimension-adapters/blob/master/aggregators/metamask.ts
> > 
> > 
> > opensea doesn't have any router Address, they simply use relay and add calldata at the end, more details here, https://opensea.notion.site/Identifying-OpenSea-Swaps-23a162eb75d9809f8ddac26fe3107800
> 
> but we can still detect by transactions data, right?

i think not possible to track all the txs for aggregator and check the calldata for all and finding correct amount for aggregator would be difficult 

Here's an example TX:
https://basescan.org/tx/0xd5f37bc535cd37a4bf72eb1f4ad33cf6360366365c0ea34f0ba057b398d701a1

You can see the tag data at the end of the calldata

PR #3937, opened 2025-08-04
stellaswap v4

atticuzor:


treeoflife2:
@atticuzor any info regarding how much fees goes to LP and how much to protocol?

PR #4032, opened 2025-08-14
Add WOOFi Pro perps volume

0xMerl:


g1nt0ki (changes requested):
sorry, need breakdown by chain

PR #4049, opened 2025-08-17
Adding hx.finance

wayne-o:


wayne-o:
I believe the check failure is related to a file outside of this PR

PR #4075, opened 2025-08-21
feat: Add hyperliquid source for derivatives volume and revenue

NotRithik:
Addressing the issue "Find source to refill hyperliquid derivatives volume and revenue(especially holder revenue)" https://github.com/DefiLlama/dimension-adapters/issues/3793.

### DIFF
--- a/dexs/hyperliquid-perp/index.ts
+++ b/dexs/hyperliquid-perp/index.ts
@@ -1,40 +1,94 @@
-import type { SimpleAdapter } from "../../adapters/types";
+import type { SimpleAdapter, FetchOptions } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
-import { getUniqStartOfTodayTimestamp } from "../../helpers/getUniSubgraphVolume";
-import { httpPost } from "../../utils/fetchURL";
+import { httpGet } from "../../utils/fetchURL";
 
-const URL = "https://api.hyperliquid.xyz/info";
+/**
+ * The stats API is the backend for the official Hyperliquid stats frontend.
+ * This API does not generate data in real-time. It serves data from a PostgreSQL database
+ * which is populated by a backend process that downloads historical data from a private AWS S3 archive.
+ *
+ * The Hyperliquid team only updates this S3 archive approximately once per month.
+ * This is the source of the significant data lag.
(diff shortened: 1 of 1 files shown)


NotRithik:
This adapter fetches daily volume and fee data for the Hyperliquid perpetuals exchange.

It works in three main steps:

1. Fetches Data from the Official Stats API: It uses the backend API that powers [stats.hyperliquid.xyz](https://stats.hyperliquid.xyz). This source provides accurate, pre-calculated historical data for a full UTC day.

2. Calculates Metrics:

    - Daily Volume: It calls the /total_volume endpoint and sums the volume for all markets for the requested day.
    - Daily Fees & Revenue: It calls the `/user_pnl` endpoint. The protocol's revenue is the inverse of the net profit/loss of all traders (`-1` * `total_user_pnl`), as `user_pnl` + `hlp_pnl` + `liquidator_pnl` = 0 for any trade. Fees paid by users are the same but are floored at zero (Since users can't pay a negative fee, which would be a profit. We are not measuring profit here).

3. Handles API Quirks:
    - Timestamp Correction: It adds two days to the requested date to correct for compounding off-by-one-day errors found in both the DefiLlama test runner and the Hyperliquid API, ensuring the data aligns with the official frontend.
    - Stale Data: The API's data source is known to lag by several weeks. To avoid showing incorrect $0 data for recent dates, the code intentionally throws an error if the requested date is recent but the API has no data for it.

noateden:
hi @NotRithik, thanks for the PR. Look like this API data is stale from 2025-07-12

treeoflife2:
> @NotRithik we can use allium query, as they have hyperliquid core trades data, https://docs.allium.so/historical-data/supported-blockchains/hyperliquid/dex/trades

ignore this as we don't have access to allium hyperliquid data

NotRithik:
If I can get a cloud instance with the required resources, I will run a hyperliquid node and access trade info in this format: https://hyperliquid.gitbook.io/hyperliquid-docs/for-developers/nodes/l1-data-schemas. I will then create an open-source public endpoint that indexes Hyperliquid order info by the hour: 
image

Then, I'll create an API endpoint that uses this order info to effectively create a new source for derivatives volume and revenue. This hyperliquid node can also serve as a data source for other deeper custom DefiLlama dashboards.

Would DefiLlama be open to this approach and funding the setup?

Issue #2939, opened 2025-04-16
track Tonstakers fees


noateden:

KPHEMRAJ:
https://tonstakers.com/faq/en/article/are-there-any-fees-for-liquid-staking

PR #4119, opened 2025-08-28
feat: tracking tonstakers lsd fee and revenue

KPHEMRAJ:
https://github.com/DefiLlama/dimension-adapters/issues/2939


fees/tonstakers-lsd/index.ts
```diff
@@ -0,0 +1,110 @@
...
+    seqno: blockNumber
+  };
+
+  const options = {
+    method: "POST",
+    headers: { "Content-Type": "application/json" },
+    body: JSON.stringify(payload),
+  };
+
+  try {
```
noateden, line 25: can you throw errors here instead of try/catch?

KPHEMRAJ: > can you throw errors here instead of try/catch?

done

PR #4173, opened 2025-09-05
tracking idex-v1 fee and volume

KPHEMRAJ:
Issue - https://github.com/DefiLlama/dimension-adapters/issues/4116

I understand we are reducing dependency on dune, but in this case , I believe we just backfill and mark as dead


g1nt0ki (approved):
thanks @KPHEMRAJ

PR #4177, opened 2025-09-05
feat: Add Virtus aggregator adapter using backend daily volume endpoint

cto-virtus:


cto-virtus:
I fixed calculations for bridge volume and dex agg as well, now values should be correct for both cases

PR #4206, opened 2025-09-11
myx-finance update fees and volume

ryanfanjc:


dexs/myx-finance/helpers..ts
```diff
@@ -69,11 +69,13 @@ export function getFetch(metric: 'volume' | 'fees'): FetchV2 {
       })
       for (const event of executeOrderV2Events) {
         const pairIndex = Number(event.pairIndex)
+
         if (pairIdsMaps[pairIndex]) {
-          dailyVolume.add(pairIdsMaps[pairIndex].indexToken, event.orderSize)
+          const decimals = await options.api.multiCall({ abi: 'erc20:decimals', calls: [pairIdsMaps[pairIndex].indexToken, pairIdsMaps[pairIndex].stableToken] })
+          dailyVolume.add(pairIdsMaps[pairIndex].indexToken, Number(event.executionSize) * Number(event.executionPrice) / (10 ** (30 + Number(decimals[0]) - Number(decimals[1]))))
```
noateden, line 75: you're adding token balance into `dailyVolume`, why are you multiplying with token price here?

ryanfanjc: I’ve been calculating the volume in USD here. If you’re calculating the volume in terms of the amount of tokens, could you clarify how you handle the precision when multiplying by the price? We don’t have visibility into the corresponding code, so we can’t locate the issue. Based on our data, the trading volume deviation is quite significant.

noateden: hi @ryanfanjc, We obtain the USD value by retrieving the token price at that time. I believe that you are over-reporting your volume and fees. For example, on 2025-09-10:
- We calculated $54K of volume using on-chain events (these codes).
- Your UI shows $300M in volume.

The precision of the token price is not the issue here.

PR #4210, opened 2025-09-12
Updated Fee and Revenue

neelra8n:


noateden:
hi @neelra8n, thanks for the PR. We don't fetch fees APY from API, I have updated adapter to query fees from smart contracts. Can you help to verify output numbers?

PR #4281, opened 2025-09-24
Update to the FullSail Finance fees adapter

AlexKomzzz:
PR Description:
This is an update to the FullSail Finance fees adapter. Changes include:
	1.	Updated protocol_revenue_usd — now includes not only swap fees (fees_usd) but also revenue from Rebase oSail.
	2.	Added new fields:
	•	holders_revenue_usd — includes all swap fees and Rebase oSail (Sail token buyback at half its price). As a result, this value will be higher than fees_usd.
	•	revenue_usd — total protocol revenue, calculated as the sum of protocol_revenue_usd and holders_revenue_usd.


noateden:
updated https://github.com/DefiLlama/dimension-adapters/pull/4307

PR #4313, opened 2025-09-29
track metamask usd fees

KPHEMRAJ:


treeoflife2:
@KPHEMRAJ any sources for the musd? is it only on two chains? and token holders doesn't earn any yield?

KPHEMRAJ:
> @KPHEMRAJ any source for the musd docs? is it only on two chains? and token holders doesn't earn any yield?

I couldn't find much docs than this
https://metamask.io/en-GB/news/introducing-metamask-usd-your-dollar-your-wallet.
 Yes , it is only on 2 chains as of now.
I don't think metamask usd is yield bearing, it's just pegged to usd

PR #4349, opened 2025-10-04
track rysk-v12 finance fees and volume

KPHEMRAJ:


dexs/rysk-v12/index.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    return {
+        dailyVolume,
+        dailyFees,
+        dailySupplySideRevenue: dailyFees
+    }
+}
+
+const methodology = {
+    DailyVolume: "Assets deposited by the users to be executed at pre-agreed price",
```
noateden: it should be Volume

dexs/rysk-v12/index.ts
```diff
@@ -0,0 +1,64 @@
...
+    DailyFees: "Option premium received by users of rysk",
```
noateden: the key should be Fees

dexs/rysk-v12/index.ts
```diff
@@ -0,0 +1,64 @@
...
+    });
+
+    premiumReceivedLogs.forEach((premium: any) => {
+        const { asset, amount } = premium;
+        dailyFees.add(asset, amount);
+    });
+
+    return {
+        dailyVolume,
+        dailyFees,
```
noateden, line 45: can you add dailyRevenue: 0 here, to show that protocol makes no revenue from fees?

noateden (changes requested):
thanks for PR

noateden:
hmm just checked rysk finance have two versions, if this adapter for both, can you please break it into two versions?

Issue #4352, opened 2025-10-04
Change of Chain name from Tomo Chain to Viction


okolicodes7:
In the https://github.com/DefiLlama/dimension-adapters/blob/master/helpers/chains.ts 
one of the chain still has the old name "Tomo Chain" and is supposed to be changed  to "Viction chain"
as can be seen here: https://blog.viction.xyz/introducing-viction/

PR #4353, opened 2025-10-04
Change of Tomo chain name to Viction chain 

okolicodes7:
I raised this issue here:  https://github.com/DefiLlama/dimension-adapters/issues/4352 and provided a fix. The chain has changed its name and is no more Tomo Chain but now Viction 
https://blog.viction.xyz/introducing-viction/


noateden:
sorry, it was chain id, we don't change the chain id. We still show Viction in display name

PR #4445, opened 2025-10-14
Feat: update volume for FlowX Perps

devflowx:
Hey guys,
We would like to update more volume for FlowX Perps.
We prefer this url like that: https://defillama.com/protocol/flowx-perp.
The perp volume should be in FlowX protocol: https://defillama.com/protocol/flowx-finance

With perp volume we use hyperliquid as liquidity source, so you don't need to include volume to total volume.

Thank for reviewing my PR <3

### DIFF
--- a/dexs/flowx-perps/index.ts
+++ b/dexs/flowx-perps/index.ts
@@ -3,15 +3,15 @@ import { CHAIN } from "../../helpers/chains";
 import fetchURL from "../../utils/fetchURL";
 
 const fetch = async ({ fromTimestamp, toTimestamp }: FetchOptions) => {
-  const volume = await fetchURL(`https://api.flowx.finance/flowx-be/api/perp-tracking/volume-in-timerange?startTime=${fromTimestamp * 1000}&endTime=${toTimestamp * 1000}`);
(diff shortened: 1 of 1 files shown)


devflowx:
@treeoflife2 
> so how does it work?

Each user trade on FlowX Finance we'll generate in backend a wallet. So we have wallet of user and tracking each user every minute, after that we have data to submit.

treeoflife2:
> @treeoflife2
> 
> > so how does it work?
> 
> Each user trade on FlowX Finance we'll generate in backend a wallet. So we have wallet of user and tracking each user every minute, after that we have data to submit.

it would be still interface category, and won't count as sui chain, as behind the scenes it still uses hyperliquid liquidity, there might be builder code, otherwise how you earn from it?

llamabutler:
The flowx-perps adapter exports:
        
 
 
    > adapters@1.0.0 test
    > ts-node --transpile-only cli/testAdapter.ts dexs flowx-perps
    
    🦙 Running FLOWX-PERPS adapter 🦙
    Start Date:	Thu, 16 Oct 2025 14:05:25 GMT
    End Date:	Fri, 17 Oct 2025 14:05:25 GMT
    
    HYPERLIQUID 👇
    Backfill start time: 1/10/2025
    Daily volume: 393.97 k
    End timestamp: 1760709924 (2025-10-17T14:05:24.000Z)

PR #4505, opened 2025-10-22
Add Brotocol volume data

lighthx:
https://brotocol.xyz/bridge/cross-bridge


noateden:
hi @lighthx thanks for the PR. Any way to track volume per chain?

lighthx:
We already broke down the volume per chain.

PR #4537, opened 2025-10-27
fix: clanker fee and holder revenue

KPHEMRAJ:
Clanker has made around 700k in fees yesterday, we were just reporting 250k .


treeoflife2:
@KPHEMRAJ nice catch, thanks

PR #4540, opened 2025-10-28
track aethir fees

KPHEMRAJ:
Numbers verification : https://dashboard.aethir.com/protocol/overview
---

### DIFF
--- a/fees/aethir/index.ts
+++ b/fees/aethir/index.ts
@@ -1,37 +1,21 @@
 import { FetchOptions, FetchResult, SimpleAdapter } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
+import fetchURL from "../../utils/fetchURL";
 
-const AETHIR_CORE = "0x226DC7D2AA1F9a565e82faf04772FDbBaF2da42d";
-const AETHIR_TOKEN = "0xc87B37a581ec3257B734886d9d3a581F5A9d056c";
+const AETHIR_ENDPOINT = "https://dashboard-api.aethir.com/protocol/demand-metrics/historical-network-revenue";
 
-const DEPOSIT_SERVICE_FEE_EVENT = "event DepositServiceFee (address indexed developer, uint64 nonce, uint256 amount)";
+async function fetch(_a: any, _b: any, options: FetchOptions): Promise<FetchResult> {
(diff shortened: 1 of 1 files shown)


fees/aethir/index.ts
```diff
@@ -0,0 +1,54 @@
...
+
+async function fetch(options: FetchOptions): Promise<FetchResult> {
+    const dailyFees = options.createBalances();
+
+    const serviceFeeDepositLogs = await options.getLogs({
+        eventAbi: DEPOSIT_SERVICE_FEE_EVENT,
+        target: AETHIR_CORE
+    });
+
+    const serviceFeeWithdrawalLogs = await options.getLogs({
```
treeoflife2: is this like credit system, where remaining funds withdraw back to the users?

fees/aethir/index.ts
```diff
@@ -0,0 +1,54 @@
...
+        eventAbi: WITHDRAW_SERVICE_FEE_EVENT,
+        target: AETHIR_CORE
+    });
+
+    serviceFeeDepositLogs.forEach((deposit: any) => dailyFees.add(AETHIR_TOKEN, deposit.amount));
+
+    serviceFeeWithdrawalLogs.forEach((withdraw: any) => dailyFees.subtractToken(AETHIR_TOKEN, withdraw.amount));
```
treeoflife2: is there any way to directly track the consumed/used aethir for fees? instead of tracking deposit - withdrawals? or that's the correct way, please elaborate. thanks

fees/aethir/index.ts
KPHEMRAJ: Yeah, quite similar to defillama subscription, but people rarely withdraw, but worth adding or else could lead to wrong cumulative numbers

fees/aethir/index.ts
KPHEMRAJ: I couldnt find a way to get exact daily usage, we can use their api instead , https://dashboard-api.aethir.com/protocol/demand-metrics/historical-network-revenue

treeoflife2: if api gives accurate numbers then we could use that.

KPHEMRAJ:
@treeoflife2  I have switched to api call

PR #4550, opened 2025-10-29
update fees calculation for privacy cash

SocialfiPanda:
Current fees are outdated for Privacy Cash.

If you go to their app (https://privacycash.org/), connect your wallet, enter an amount, and you'll clearly see the fees are (0.35%  + 0.006 SOL) on withdrawals.

Screenshot 2025-10-29 at 10 04 28 AM


SocialfiPanda:
Pretty early on. Maybe after a week? Would be OK just to ignore.

treeoflife2:
> @treeoflife2 has the change been deployed?
> 
> the stats still look way off from Dune data, which is correct.
> 
> https://dune.com/datadashboards/privacy-cash

just refilled should be live in few hours

PR #4602, opened 2025-11-05
Add Clickhouse Client

0xpeluche:


package-lock.json
g1nt0ki, line 1: maybe safe to delete this file?

helpers/env.ts
```diff
@@ -69,6 +69,11 @@ export const ENV_KEYS = new Set([
   'DUNE_BULK_MODE',
   'DUNE_BULK_MODE_BATCH_TIME',
   'LLAMA_HL_INDEXER',
+  'CLICKHOUSE_HOST',
```
g1nt0ki, line 72: I prefer a single env var, can be a json string, check what we are doing in server repo

helpers/clickhouse.ts
g1nt0ki, line 1: wondering if should move this to @defillama/sdk repo

this way, code is not repeated

helpers/clickhouse.ts
```diff
@@ -0,0 +1,65 @@
...
+    return await rs.json<T>();
+  } catch (error: any) {
+    if (error?.code === "ECONNRESET" || error?.code === "ECONNREFUSED") {
+      await disconnectClickhouse();
+    }
+    throw error;
+  }
+}
+
+export async function disconnectClickhouse() {
```
g1nt0ki, line 90: please add auto disconnect on exit event

PR #4608, opened 2025-11-06
track Rho-trading OI

treeoflife2:
https://github.com/DefiLlama/dimension-adapters/pull/4589


treeoflife2:
> @treeoflife2 Hi. In my first version, I added price conversion for better precision. [5ecc674#diff-4d8dbd5f94c0146894abd169742ead3463ee4a431892895b5e112ad91feb7fd1R95](https://github.com/DefiLlama/dimension-adapters/commit/5ecc6741b3cb1c9b65f743441d055496ae10b2e2#diff-4d8dbd5f94c0146894abd169742ead3463ee4a431892895b5e112ad91feb7fd1R95)
> 
> EXCHANGE_INFO_URL provides the actual active pairs, base currencies, and exchange rates. But if you prefer simplicity, stablecoins are close to a 1:1 rate. On the other hand, I can’t guarantee that we’ll support only stablecoins in the future.

but you do will still show the usd value of OI in exchange UI, so i think this is fine.

PR #4615, opened 2025-11-07
track skale(chain) fees

0saurabh0:
Fixes https://github.com/DefiLlama/dimension-adapters/issues/2605

Skale

### DIFF
--- a/fees/skale.ts
+++ b/fees/skale.ts
@@ -1,0 +1,46 @@
+import { Adapter, FetchOptions, ProtocolType } from "../adapters/types";
+import { CHAIN } from "../helpers/chains";
+
+const fetch = async (options: FetchOptions) => {
+  const dailyFees = options.createBalances();
+  const dailyRevenue = options.createBalances();
+  const dailyUserFees = options.createBalances();
+  const dailyProtocolRevenue = options.createBalances();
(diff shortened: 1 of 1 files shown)


fees/skale.ts
```diff
@@ -0,0 +1,46 @@
...
+  const dailyFees = options.createBalances();
+  const dailyRevenue = options.createBalances();
+  const dailyUserFees = options.createBalances();
+  const dailyProtocolRevenue = options.createBalances();
+  const dailySupplySideRevenue = options.createBalances();
+  
+  // Even though gas is consumed on-chain, sFUEL is free and has no value
+  // Users obtain sFUEL from faucets at no cost
+  // All fee metrics are 0 in USD terms since sFUEL = $0
+  // Network revenue comes from SKL token staking subscriptions, not transaction fees
```
treeoflife2, line 14: can we add network revenue?

0saurabh0: The network revenue they earn comes from the SKL staking/subscription model that happens off-chain; I couldn’t find any public API or data feed that reports those staking revenues on a per-day basis. If there’s a data source we can hook into, I’m happy to integrate it.
Otherwise, we gotta contact their team.

PR #4646, opened 2025-11-12
add historical data

ratex-tony:
We have added historical data starting from last year and fixed a bug in the API, which caused the data for November 9, 2025 and November 10, 2025 to be inaccurate. Could you please help re-fetch all the data? Thank you.


ratex-tony:
Hi @treeoflife2 ,we found that the historical data is already available on DefiLlama, but the values for November 9, 2025 and November 10, 2025 are still the previous incorrect data. Could you help resync the data for these two days? Thank you.

PR #4694, opened 2025-11-19
fix: Update Thorswap adapter for fees/revenues calculation

VAhafonov:


noateden:
@VAhafonov hmm I see. But now, all fees are on Ethereum chain only?

PR #4724, opened 2025-11-24
remove fee info, just show revenue info

Stephen1130:
I already added spaace fee info before, however there is some bugs on showing revenue and fee info.
we realized that we don't need to show fees info, just need to show revenue info so we removed fee related code.
Thanks


Stephen1130:
Yes, but the exact code lines are
https://github.com/DefiLlama/dimension-adapters/blob/master/fees/opensea/seaport.ts#L41

https://github.com/DefiLlama/dimension-adapters/blob/master/fees/opensea/seaport.ts#L43

Stephen1130:
And I cannot join the defillama discord server. I got an "Unable to accept
invite" error. How to fix this issue?

PR #4736, opened 2025-11-25
Track spaace

Stephen1130:
Added daily volume info to spaace info


noateden:
hmm I think there is something wrong, we don't track NFT trade volume as you can see on Opensea and Blur. Let's me check why on MagicEden

PR #4752, opened 2025-11-26
fix: helio fees

KPHEMRAJ:
Helio doesnt have a token

### DIFF
--- a/fees/helio.ts
+++ b/fees/helio.ts
@@ -8,7 +8,7 @@ import { FetchOptions, SimpleAdapter } from "../adapters/types";
 import { CHAIN } from "../helpers/chains";
 import { getSolanaReceived } from "../helpers/token";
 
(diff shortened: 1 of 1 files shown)


fees/helio.ts
```diff
@@ -9,18 +9,13 @@ import { CHAIN } from "../helpers/chains";
 import { getSolanaReceived } from "../helpers/token";
 
 const fetch: any = async (options: FetchOptions) => {
```
noateden: it should be
`const fetch: any = async (_a: any, _b: any, options: FetchOptions) =>`

noateden (changes requested):
.

noateden:
hi @KPHEMRAJ was it wrong from beginning or it was changed recently?

fees/helio.ts
KPHEMRAJ: fixed

PR #4754, opened 2025-11-26
track fragment fees

KPHEMRAJ:
---

### DIFF
--- a/fees/fragment/index.ts
+++ b/fees/fragment/index.ts
@@ -3,12 +3,19 @@ import { CHAIN } from "../../helpers/chains";
 import { queryDuneSql } from "../../helpers/dune";
 
 async function fetch(_a: any, _b: any, options: FetchOptions): Promise<FetchResult> {
-    const query = `select sum(value/1e9) as ton_received from ton.messages
-    where direction = 'in'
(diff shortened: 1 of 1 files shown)


fees/fragment/index.ts
```diff
@@ -0,0 +1,37 @@
...
+const methodology = {
+    Fees: "Includes NFT marketplace fees, fees to buy telegram ads , premium , stars, usernames, numbers etc ",
+    Revenue: "All fees are revenue",
+    ProtocolRevenue: "All the fees goes to protocol"
+};
+
+const adapter: SimpleAdapter = {
+    fetch,
+    chains: [CHAIN.TON],
+    start: '2022-10-26',
```
noateden: is this correct start date? I just think it's a bit old

KPHEMRAJ: Yes, I filtered by date, https://tonscan.org/address/UQBAjaOyi2wGWlk-EDkSabqqnF-MrrwMadnwqrurKpkla4QB and that was the day it received first txn

KPHEMRAJ:
> hi @KPHEMRAJ I think fees seem aren't correct. We're counting all TON transferred to the address. Can you double check their actually sell transactions?

Its fee recipient address. It gets fees of everything premium , stars , nft trade fee etc. Should be filter only nft trade fee?

PR #4757, opened 2025-11-26
Use on-chain pool fees for Blackhole AMM fee calculations

janhenkgerritsen:
The current Blackhole AMM implementation calculates fee and revenue metrics by applying a fixed 0.3% fee to all swaps. This is incorrect, as each Blackhole AMM pool has a specific fee configured on-chain. So the current implementation underestimates the fees for pools that have a higher fee and overestimates fees for pools that have a lower fee. This pull request fixes that by retrieving the actual on-chain pool fee and using that in the fee and revenue calculations.

### Changes
* Removed the `SwapFee` variable with the hardcoded 0.3% fee.
* Added a multicall to retrieve the `stable` status for each pool.
* Added a multicall to retrieve the pool fee for each pool by calling the `getFee(address,stable)` method on the Factory contract.
* Use the actual pool fee in the fee and revenue calculations. On-chain fees are scaled by 10_000, so the code takes that into account.
* Small style cleanup by changing a couple of `fetchOptions.api.multiCall` calls to `api.multiCall`.

### Tests
I used the following test command to get the numbers for a couple of different days and compared them to the numbers mentioned on the Blackhole statistics page:

`ts-node --transpile-only cli/testAdapter.ts dexs blackhole 2025-11-23`

The output of this command before applying my fix:
```
> adapters@1.0.0 test
> ts-node --transpile-only cli/testAdapter.ts dexs blackhole 2025-11-23

🦙 Running BLACKHOLE adapter 🦙
---------------------------------------------------
Start Date:	Sat, 22 Nov 2025 00:00:00 GMT
End Date:	Sun, 23 Nov 2025 00:00:00 GMT
---------------------------------------------------

AVAX 👇
Daily volume: 1.26 M
Daily fees: 3.77 k
Daily user fees: 3.77 k
Daily revenue: 3.74 k
Daily supply side revenue: 27.00
Daily protocol revenue: 0.00
Daily holders revenue: 3.74 k
End timestamp: 1763855999 (2025-11-22T23:59:59.000Z)
```

And the output of this command after my fix:

```
> adapters@1.0.0 test
> ts-node --transpile-only cli/testAdapter.ts dexs blackhole 2025-11-23

🦙 Running BLACKHOLE adapter 🦙
---------------------------------------------------
Start Date:	Sat, 22 Nov 2025 00:00:00 GMT
End Date:	Sun, 23 Nov 2025 00:00:00 GMT
---------------------------------------------------

AVAX 👇
Daily volume: 1.24 M
[...]

PR #4877, opened 2025-12-01
Improve Blackhole adapter

janhenkgerritsen:
This PR updates the Blackhole basic pools adapter to use `Fees` events for fee and revenue calculation.
This makes the adapter fully aligned with how Blackhole measures swap fees.

### Background
After my previous PR (https://github.com/DefiLlama/dimension-adapters/pull/4757), which replaced the hardcoded 0.3% fee with the on-chain `getFee()` value, I contacted the Blackhole team to share the improvement. I also asked them how they calculate the fees on their own stats page. They told me that `Fees` events are the correct and most accurate method of tracking swap fees for Blackhole basic pools, as pool fees can change over time and the `Fees` event includes the actual fees charged for each swap.

### Changes

* Use `Fees` events to compute actual fees per swap (covers dynamic fee changes and internal deductions).
* Split fees between unstaked LPs (supply-side revenue) and veBLACK voters (holders revenue) using the current gaugeSupply / totalLpSupply ratio.
* Volume remains based on `Swap` events.
* Remove outdated multicalls related to fee percentages.
* Minor code-style cleanups (consistent formatting, `forEach` instead of `map`, comments).

### Result
Accurate and time-consistent fee and revenue tracking that now matches how the Blackhole protocol itself measures fees.

### DIFF
--- a/dexs/blackhole.ts
+++ b/dexs/blackhole.ts
@@ -1,12 +1,16 @@
-import { CHAIN } from "../helpers/chains";
-import { FetchV2, IJSON, SimpleAdapter } from "../adapters/types";
-import { cache } from "@defillama/sdk";
-import { filterPools } from "../helpers/uniswap";
-import { addOneToken } from "../helpers/prices";
+import { CHAIN } from "../helpers/chains"
+import { FetchV2, IJSON, SimpleAdapter } from "../adapters/types"
+import { cache } from "@defillama/sdk"
+import { filterPools } from "../helpers/uniswap"
+import { addOneToken } from "../helpers/prices"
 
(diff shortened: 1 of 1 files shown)


janhenkgerritsen:
@noateden The `Fees` events are emitted for each swap, so for each swap you'll see both a `Swap` and a `Fees` event onchain. This has always been the case, it just wasn't used in the adapter.

The implemenation in this PR replaces two multicalls for `state` and `getFee` with a `getLogs` call. I looked through the `getLogs` source code, but it does not seem possible to get logs for multiple events with a single call. So that's why I've added a second `getLogs` call.

If there is a better approach to achieve the same result, let me know. I'm just getting familiar with your codebase, so I'm happy to revise this PR as needed!

PR #4884, opened 2025-12-02
Add Boson Protocol fees and volume

zajck:
Add two metrics to Boson Protocol Adapter:
- protocol fees: fees taken by the protocol on every successful exchange
- daily volume: sum of funds deposited into the protocol and value of newly created offers


treeoflife2:
> @treeoflife2 I removed the dex adapter.
> 
> Just a note - Boson Protocol is miscategorised. It is a RWA protocol and not a farm. We'll soon open a PR in https://github.com/DefiLlama/defillama-server to fix that. I have 2 additional questions:
> 
> 1. Are you planning to add support for RWA volume? We believe it gives a useful insight about the protocol activity.

we are planning to add more metrics for RWA category soon

> 2. In https://github.com/DefiLlama/defillama-server I see that some projects have `dimensions` property. Is this something we should update or do you do it once this PR is merged?

we will do that you don't have to worry about it

treeoflife2:
@zajck intesting find, i will check and fix the issue

PR #4940, opened 2025-12-06
Add Stabull fee, revenue and volume

0x4r45h:
I could calculate fees and revenue based on volume, but I thought tracking the treasury address might be a better approach, in case they add new sources of revenue in the future.

### DIFF
--- a/dexs/stabull/index.ts
+++ b/dexs/stabull/index.ts
@@ -1,9 +1,6 @@
 import {Adapter, FetchOptions, FetchResultV2} from "../../adapters/types";
 import {CHAIN} from "../../helpers/chains";
-import { addTokensReceived } from '../../helpers/token';
-import {Balances} from "@defillama/sdk";
 import {addOneToken} from "../../helpers/prices";
-
 
 const events = {
   trade: 'event Trade( address indexed trader, address indexed origin, address indexed target, uint256 originAmount, uint256 targetAmount, int128 rawProtocolFee)',
@@ -15,18 +12,15 @@ const events = {
 const configs: Record<string, any> = {
   [CHAIN.ETHEREUM]: {
     factory: { address: "0x2e9E34b5Af24b66F12721113C1C8FFcbB7Bc8051", fromBlock: 19773852 },
-    treasury: '0xaBd582EFa8ff22D5258FAFf769082002B3a1d919',
-    startDate: '2024-04-29',
+    start: '2024-04-29',
   },
   [CHAIN.POLYGON]: {
     factory: { address: "0x3c60234db40e6e5b57504e401b1cdc79d91faf89", fromBlock: 56377840 },
-    treasury: '0x2ABbb079e97bd30605CcD65b8aDFB756AFE4f511',
-    startDate: '2024-04-29',
+    start: '2024-04-29',
   },
   [CHAIN.BASE]: {
     factory: { address: "0x86Ba17ebf8819f7fd32Cf1A43AbCaAe541A5BEbf", fromBlock: 32584321 },
-    treasury: '0x1ea84Ba9598AA1083D7F4096E06B532Da2c3dEf5',
-    startDate: '2025-07-08',
+    start: '2025-07-08',
   },
 };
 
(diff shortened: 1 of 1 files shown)


llamabutler:
The stabull adapter exports:
        
 
 
    > adapters@1.0.0 test
    > ts-node --transpile-only cli/testAdapter.ts dexs stabull
    
    🦙 Running STABULL adapter 🦙
    Start Date:	Fri, 05 Dec 2025 02:11:55 GMT
    End Date:	Sat, 06 Dec 2025 02:11:55 GMT
    
    Token transfers: Failed to use indexer, falling back to logs base Llama Indexer v2 URL/api key is not set
    Token transfers: Failed to use indexer, falling back to logs ethereum Llama Indexer v2 URL/api key is not set
    Token transfers: Failed to use indexer, falling back to logs polygon Llama Indexer v2 URL/api key is not set
    chain     | Daily volume | Daily revenue | Daily fees | Start Time
    ---       | ---          | ---           | ---        | ---       
    ethereum  | 72.00        | 0.00          | 0.00       | 29/4/2024 
    polygon   | 11.62 k      | 5.00          | 17.00      | 29/4/2024 
    base      | 8.25 k       | 4.00          | 12.00      | 8/7/2025  
    Aggregate | 19.94 k      | 9.00          | 29.00      |

dexs/stabull/index.ts
```diff
@@ -0,0 +1,99 @@
...
+    dailyVolume.addToken(event.origin, event.originAmount)
+  }
+}
+
+const fetchRevenue = async (options: FetchOptions, fromAdddesses: string[], dailyFees: Balances) => {
+  await addTokensReceived({
+    options,
+    balances: dailyFees,
+    targets: [configs[options.chain].treasury], // Treasury address
+    fromAdddesses,
```
g1nt0ki: even though you are restricting the source here, it still has room for bugs and a bit inefficient as we are pulling two sets of logs when trade log already has fee info, also what happens when they change the treasury address?

dexs/stabull/index.ts
```diff
@@ -0,0 +1,99 @@
...
+}
+
+const fetchVolume = async (options: FetchOptions, targets: string[], dailyVolume: Balances) => {
+  const tradeEvents: Array<any> = await options.getLogs({
+    targets: targets,
+    eventAbi: events.trade,
+  })
+
+  for (const event of tradeEvents) {
+    dailyVolume.addToken(event.origin, event.originAmount)
```
g1nt0ki: can you use the [addOneToken](https://github.com/DefiLlama/dimension-adapters/blob/master/helpers/prices.ts#L12) here, not strictly necessary, what it does is, if one of the token is well known like eth or a stablecoin, uses that for fees, this way, we are more likely to have price support for the given token

dexs/stabull/index.ts
```diff
@@ -0,0 +1,99 @@
...
+  },
+};
+
+// This method dynamically fetches pool addresses, this way newly created pools will be counted automatically in the future
+const getPools = async (options: FetchOptions): Promise<string[]> => {
+  const {address, fromBlock} = configs[options.chain].factory
+  const logs = await options.getLogs({
+    target: address,
+    eventAbi: events.newCurve,
+    onlyArgs: true,
```
g1nt0ki: can you add `cacheInCloud: true` here, this caches the log response

dexs/stabull/index.ts
0x4r45h: the feeData on the event is raw and needs some calculations based on Oracle data at that exact moment to figure out the exact fee amount , unfortunately there is no event to read the calculated amount.
here is the respective code:
https://basescan.org/address/0x1246B19c59FfF6A92E875b57402743cf576C86bB#code#L2184


for the treasury address i can dynamically read it from curve contracts 
https://basescan.org/address/0x86Ba17ebf8819f7fd32Cf1A43AbCaAe541A5BEbf#readContract#F10


if you still think this way of calculation of fee/revenue is inefficient, i can discard all of these and calculate fees based on volume. since the rates are fixed i think it will achieve the same result

dexs/stabull/index.ts
0x4r45h: ah i was looking for a way to cache this data, didn't know about this. thanks

dexs/stabull/index.ts
```diff
@@ -0,0 +1,101 @@
...
+    return {
+      ...acc,
+      [chain]: {
+        fetch,
+        start: configs[chain].startDate,
+      }
+    }
+  }, {}),
+  methodology: {
+    Fees: 'Stabull charges a flat rate of 0.15% per swap per pool',
```
noateden, line 70: if it was flat fee of 0.15%, why don't you just count fees from volume and don't need to fetch token received revenue?

PR #4957, opened 2025-12-08
fix fragment fees

KPHEMRAJ:


fees/fragment/index.ts
```diff
@@ -4,37 +4,64 @@ import { queryDuneSql } from "../../helpers/dune";
...
 
     const queryResults = await queryDuneSql(options, query);
     
-    if (!queryResults[0] || !queryResults[0].ton_received) {
-      throw Error('query Dune return null result');
+    if (!queryResults[0] || !queryResults[0].ton_received ||!queryResults[0].ton_sent) {
+      throw new Error('query Dune return null result');
     }
+    if(queryResults[0].ton_sent>queryResults[0].ton_received)
+      throw new Error("Rewards exceed payments, possibly due to more onchain redeems of offchain purchased stars");
```
noateden, line 41: so in this case, should we consider revenue to be negative?

KPHEMRAJ: I dont think so , thats why we are throwing error , ignoring that day. Bcz as of now no way to track off chain revenue

Issue #4817, opened 2025-12-01
Track  xStocks Fee/Revenue


treeoflife2:
https://defillama.com/protocol/xstocks

PR #4964, opened 2025-12-09
add xStocks fee and revenue

0x4r45h:
closes https://github.com/DefiLlama/dimension-adapters/issues/4817

### DIFF
--- a/fees/xstocks/index.ts
+++ b/fees/xstocks/index.ts
@@ -1,0 +1,141 @@
+import {Adapter, Dependencies, FetchOptions, FetchResultFees} from "../../adapters/types";
+import {CHAIN} from "../../helpers/chains";
+import {queryDuneSql} from "../../helpers/dune";
+// source : https://xstocks.fi/products
+import PRODUCTS from './xstocks_products.json'
+
+const evmFeeEvents = {
+  transferEvent: 'event Transfer(address indexed from, address indexed to, uint256 value)',
+}
+
+const fetchEthereum: any = async (options: FetchOptions): Promise<FetchResultFees> => {
+  const ethereum_tokens = PRODUCTS.filter(i => i.ethereum_address).flatMap(i => i.ethereum_address);
+  const dailyMintAndBurns = options.createBalances()
+
+  // fetch mintEvents
+  const mintEvents: Array<any> = await options.getLogs({
+    targets: ethereum_tokens,
+    eventAbi: evmFeeEvents.transferEvent,
+    entireLog: true,
+    topics: [
+      '0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef',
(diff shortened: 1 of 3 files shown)


fees/xstocks/index.ts
```diff
@@ -0,0 +1,141 @@
...
+        fetch: fetchEthereum,
+        start: '2024-05-02',
+      },
+      [CHAIN.SOLANA]: {
+        fetch: fetchSolana,
+        start: '2025-05-25',
+      },
+    },
+  methodology: {
+    Fees: "Up to 0.50% of your investment's value is charged when entering and exiting the investment",
```
treeoflife2, line 137: who charges this fee backedFi or xStocks?
https://docs.xstocks.fi/overview/issuance-and-redemption
https://docs.backed.fi/frequently-asked-questions#how-does-backed-make-money

fees/xstocks/xstocks_products.json
```diff
@@ -0,0 +1,398 @@
+[
+  {
+    "name": "ABTx Abbott xStock",
```
treeoflife2, line 3: is there some way to make it dynamic? or reuse the tvl adapter, 

https://github.com/DefiLlama/DefiLlama-Adapters/blob/main/projects/xstocks/index.js

as that will be upto date

fees/xstocks/index.ts
treeoflife2: we also track backedFi, so we can route the fees going to backedFi in their own adapter, and xStocks fees in their own adapter

0x4r45h: Hey @treeoflife2, good catch – I did not notice that you are already tracking backedFi.
I think xStocks is just a product line of backedFi and it does not take any direct fees on-chain. So everything should be moved to the backedFi adapter.

0x4r45h: @treeoflife2  backedFi has a larger set of products including the one listed on xStocks. if i update the backedFi adapter to include all these tokens for TVL counting, what happens to overlapping one with xStocks? should we remote xStocks entirely?

xstocks : https://xstocks.com/products
backedfi : https://assets.backed.fi/legal-documentation/product-database

fees/xstocks/xstocks_products.json
0x4r45h: i'll look to see if i can find an API endpoint or something. 

btw, is there any way to read a json file from adapters repository inside the dimension repo? so i can reuse the list

treeoflife2: you can do the simple http request, add try catch block which throw error incase anything changes

fees/xstocks/index.ts
treeoflife2: yes if xStocks doesn't take any fees, then let's track the backedFi fees and you can remove the xStocks adapter

Issue #4806, opened 2025-12-01
Track Theo Fee/Revenue


treeoflife2:
https://defillama.com/protocol/theo

PR #4989, opened 2025-12-10
track theo fees

KPHEMRAJ:
Closes https://github.com/DefiLlama/dimension-adapters/issues/4806

### DIFF
--- a/fees/theo/index.ts
+++ b/fees/theo/index.ts
@@ -25,7 +25,7 @@ async function prefetch(options: FetchOptions): Promise<any> {
     return (priceToday - priceYesterday) / 1e6;
 }
 
-async function fetch(_a: any, _b: any, options: FetchOptions): Promise<FetchResult> {
+async function fetch(options: FetchOptions): Promise<FetchResult> {
     const dailyFees = options.createBalances();
     const dailyPriceChange = options.preFetchedResults;
 


fees/theo/index.ts
```diff
@@ -0,0 +1,82 @@
...
+        totalSupply = maxTotalSupply - oftLockedSupply;
+    }
+    else {
+        totalSupply = await options.api.call({
+            target: thBillOFT,
+            abi: 'uint256:totalSupply'
+        });
+    }
+
+    dailyFees.addUSDValue(dailyPriceChange * totalSupply / 1e6),METRIC.ASSETS_YIELDS;
```
noateden, line 52: I think multiply with `totalSupply` is not correct here, we should multiply with total underlying assets

fees/theo/index.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+const methodology = {
+    Fees: "Treasury yields from underlying",
+    Revenue: "No revenue",
+    SupplySideRevenue:"Treasury yields received by users"
+};
+
+const adapter: SimpleAdapter = {
+    version: 1,
```
noateden: can you use version 2 adapter?

fees/theo/index.ts
KPHEMRAJ: Hi @noateden ,
My understanding is this:
We are using ```convertToAssets``` implies  dailyPriceChange would mean how much asset has been increased/ how much dollar value has been increased per single thBill , so I feel multiplying by total number of thBill is correct.
Instead if we has ```convertToShares``` in that case I agree that we should multiply with total assets.

fees/theo/index.ts
KPHEMRAJ: done

PR #4998, opened 2025-12-10
Hanji: added DEX adapter with V1/V2 factory support

jack-squirrel:
Hanji has been deployed on Monad and Base. Added multiple chains support for Hanji adapter using events instead of graph queries.


jack-squirrel:
Hey guys! I have 2 issues:
1. Looks like there are limitations when requesting event logs for Etherlink chain, I can't get all logs in requested period, just last ~700-800 blocks. What is recommended way to fix it?
2. Test script is stuck on Monad chain (it is commented in config at the moment). How can I check, if it is RPC issue or some problems with the adapter code?

PR #5001, opened 2025-12-10
Track hanji multichain

noateden:
from https://github.com/DefiLlama/dimension-adapters/pull/4998


jack-squirrel:
Thanks for refactoring! 
It is still incorrect daily volume for etherlink chain, should be around 200k for last 24 hours. Is it possible to request OrderPlaced events without limitation or add batching/pagination?

PR #5062, opened 2025-12-14
fix: update bmx base subgraph to v0.0.2

kanerep:
resolves: #5044 

Subgraph was returning 404 error because it was outdated. I tested both base and mode subgraph versions, it appears only base has been updated to `0.0.2`

**Before:**
Screenshot 2025-12-14 at 10 35 10

**After:**
Screenshot 2025-12-14 at 10 37 04


noateden (changes requested):
.

PR #5142, opened 2025-12-17
Kgen: convert to dune adapter for historical support

treeoflife2:


treeoflife2:
> Hi @treeoflife2 is there something needed to be done by our side as It shows one of the tests as failed.

you can ignore that, i've refilled last 2 months data, should be reflected in few hours

PR #5271, opened 2025-12-19
Track Axelar Fees 

PatelPrinci:
Addresses issue: https://github.com/DefiLlama/dimension-adapters/issues/2594

### DIFF
--- a/fees/axelar/index.ts
+++ b/fees/axelar/index.ts
@@ -28,10 +28,12 @@ const ABI = [
   "event GasPaidForExpressCall(address indexed sourceAddress, string destinationChain, string destinationAddress, bytes32 indexed payloadHash, address gasToken, uint256 gasFeeAmount, address refundAddress)",
   "event NativeGasPaidForExpressCall(address indexed sourceAddress, string destinationChain, string destinationAddress, bytes32 indexed payloadHash, uint256 nativeGasAmount, address refundAddress)",
   "event GasAdded(bytes32 indexed txHash, uint256 indexed logIndex, address gasToken, uint256 gasFeeAmount, address refundAddress)",
-  "event NativeGasAdded(bytes32 indexed txHash, uint256 indexed logIndex, uint256 gasFeeAmount, address refundAddress)"
+  "event NativeGasAdded(bytes32 indexed txHash, uint256 indexed logIndex, uint256 gasFeeAmount, address refundAddress)",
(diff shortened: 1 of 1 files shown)


fees/axelar/index.ts
```diff
@@ -0,0 +1,124 @@
...
+  "event GasPaidForContractCallWithToken(address indexed sourceAddress, string destinationChain, string destinationAddress, bytes32 indexed payloadHash, string symbol, uint256 amount, address gasToken, uint256 gasFeeAmount, address refundAddress)",
+  "event NativeGasPaidForContractCall(address indexed sourceAddress, string destinationChain, string destinationAddress, bytes32 indexed payloadHash, uint256 nativeGasAmount, address refundAddress)",
+  "event GasPaidForExpressCall(address indexed sourceAddress, string destinationChain, string destinationAddress, bytes32 indexed payloadHash, address gasToken, uint256 gasFeeAmount, address refundAddress)",
+  "event NativeGasPaidForExpressCall(address indexed sourceAddress, string destinationChain, string destinationAddress, bytes32 indexed payloadHash, uint256 nativeGasAmount, address refundAddress)",
+  "event GasAdded(bytes32 indexed txHash, uint256 indexed logIndex, address gasToken, uint256 gasFeeAmount, address refundAddress)",
+  "event NativeGasAdded(bytes32 indexed txHash, uint256 indexed logIndex, uint256 gasFeeAmount, address refundAddress)"
+];
+
+const iface = new ethers.Interface(ABI);
+const TOPICS = {
```
treeoflife2, line 37: can you also add the 'event Refunded' as i think axelar gas service contract refunds if charged gas fee was more than required. so we have to substract those fees

PatelPrinci: Sure, i'll implement that!!

treeoflife2: does each chain has the chain fees in the same gas token(Ethereum), let's verify if we are not mixing the decimal/gas tokens in calculation.

PatelPrinci: Each chain emits gas fees in its own native gas token and in its native smallest unit. In the adapter, fees are aggregated per chain and recorded using the standard zero address for native tokens. Even though the address is the same, it is resolved on a per-chain basis by DefiLlama, so decimals and pricing are applied correctly. This ensures we are not mixing gas tokens or decimals across chains during fee aggregation or USD conversion.

PR #5277, opened 2025-12-19
Updated fee token for Duck-chain,CAPX,Sanko,EARNM and Adventure Layer.

Smriti-08:


treeoflife2:
@Smriti-08 correct CGToken is 'the-open-network' for Toncoin, others are wrong too, can you update them?

image

PR #5251, opened 2025-12-18
fix: Makina fee adapter

uwezukwechibuzor:
**NOTE**

#### Please enable "Allow edits by maintainers" while putting up the PR.

---

> - If you would like to add a `tvl` adapter please submit the PR [here](https://github.com/DefiLlama/DefiLlama-Adapters).

1. Once your adapter has been merged, it takes time to show on the UI. If more than 24 hours have passed, please let us know in Discord.
2. Please fill the form below **only if the PR is for listing a new protocol** else it can be ignored/replaced with reason/details about the PR
3. **For updating listing info** It is a different repo, you can find your listing in this file: https://github.com/DefiLlama/defillama-server/blob/master/defi/src/protocols/data4.ts, you can edit it there and put up a PR
4. Do not edit/push `package.json/package-lock.json` file as part of your changes
5. No need to go to our discord/other channel and announce that you've created a PR, we monitor all PRs and will review it asap

---

PR #5279, opened 2025-12-19
fix: aligns Makina fees and yield with DeFiLlama standards and corrects calcs

ColinPlatt:
corrects #5251

### DIFF
--- a/fees/makina/index.ts
+++ b/fees/makina/index.ts
@@ -148,14 +148,19 @@ async function fetch(options: FetchOptions): Promise<FetchResultV2> {
       shareTokens: [machine.shareToken],
       machines: [machine.machine],
     });
+    
+    dailyFees.add(dailyNetYield, METRIC.ASSETS_YIELDS);
+    dailySupplySideRevenue.add(dailyNetYield, METRIC.ASSETS_YIELDS);
 
-    dailyRevenue.add(machine.shareToken, totalFeeShares);
-    dailyProtocolRevenue.add(totalProtocolRevenueShares);
-    dailySupplySideRevenue.add(dailyNetYield);
+    dailyFees.add(totalProtocolRevenueShares, METRIC.PROTOCOL_FEES);
+    dailyRevenue.add(totalProtocolRevenueShares, METRIC.PROTOCOL_FEES);
+    dailyProtocolRevenue.add(totalProtocolRevenueShares, METRIC.PROTOCOL_FEES);
 
-    //Daily fees are net yield + total fees minted
-    dailyFees.add(dailyNetYield);
-    dailyFees.add(machine.shareToken, totalFeeShares);
+    const operatorsRevenue = dailyRevenue.clone(1)
+    operatorsRevenue.subtract(dailyProtocolRevenue)
+    
+    dailyFees.addBalances(operatorsRevenue, METRIC.OPERATORS_FEES);
+    dailySupplySideRevenue.addBalances(operatorsRevenue, METRIC.OPERATORS_FEES);
   }
 
   return {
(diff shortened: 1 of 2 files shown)


fees/makina/index.ts
```diff
@@ -1,68 +1,132 @@
...
-  feeManager: string;
   machine: string;
+  protocolReceiver: string; // where protocol revenue is sent, as we cannot depend on the order to be consistent
 };
 
+// modified from helpers/erc4626.ts
+async function getMachineYield({
+  options,
+  shareTokens,
+  machines,
```
Hugoo, line 26: this function is always used with ONE machine at at a time (as the name suggests: getMachineYield - singular), i suggest to convert this to a string instead of string[]


```suggestion
  machine,
```

fees/makina/index.ts
```diff
@@ -1,68 +1,132 @@
...
+  assetAbi = "address:accountingToken",
+  valueAbi = "uint256:totalSupply",
+  convertAbi = "function convertToAssets(uint256) view returns (uint256)",
+}: {
+  options: FetchOptions;
+  shareTokens: string[];
+  machines: string[];
```
Hugoo, line 33:
```suggestion
  machine: string;
```

fees/makina/index.ts
```diff
@@ -1,68 +1,132 @@
...
+  assetAbi?: string;
+  valueAbi?: string;
+  convertAbi?: string;
+}) {
+  const assets = await options.api.multiCall({
+    abi: assetAbi,
+    calls: machines,
```
Hugoo, line 40:
```suggestion
    calls: [machine],
```

but here maybe multicall is not needed as we have only one machine to check

fees/makina/index.ts
```diff
@@ -1,68 +1,132 @@
...
-  sharePriceOracle: string;
-  feeManager: string;
   machine: string;
+  protocolReceiver: string; // where protocol revenue is sent, as we cannot depend on the order to be consistent
 };
 
+// modified from helpers/erc4626.ts
+async function getMachineYield({
+  options,
+  shareTokens,
```
Hugoo, line 25: same as for machine below, this is used only for one machine at a time, so let's reduce confusion and keep this as a "single machine yield" call:


```suggestion
  shareToken,
```

fees/makina/index.ts
```diff
@@ -1,68 +1,132 @@
...
+  machines,
+  assetAbi = "address:accountingToken",
+  valueAbi = "uint256:totalSupply",
+  convertAbi = "function convertToAssets(uint256) view returns (uint256)",
+}: {
+  options: FetchOptions;
+  shareTokens: string[];
```
Hugoo, line 32:
```suggestion
  shareToken: string;
```

fees/makina/index.ts
```diff
@@ -1,68 +1,132 @@
...
+  convertAbi?: string;
+}) {
+  const assets = await options.api.multiCall({
+    abi: assetAbi,
+    calls: machines,
+    permitFailure: true,
+  });
+  const values = await options.api.multiCall({
+    abi: valueAbi,
+    calls: shareTokens,
```
Hugoo, line 45:
```suggestion
    calls: [shareToken],
```

again, now we make it obvious we always call it with one item so maybe multicall is not necessary

fees/makina/index.ts
```diff
@@ -1,68 +1,132 @@
...
+    permitFailure: true,
+  });
+  const decimals = await options.api.multiCall({
+    abi: "uint8:decimals",
+    calls: shareTokens,
```
Hugoo, line 50:
```suggestion
    calls: [shareToken],
```

fees/makina/index.ts
```diff
@@ -1,68 +1,132 @@
...
+    permitFailure: true,
+  });
+  const convertCalls = machines.map((machine, index) => {
+    return {
+      target: machine,
+      params: [String(10 ** Number(decimals[index]))],
+    };
+  });
```
Hugoo, line 58:
```suggestion
  const convertCalls = [{
      target: machine,
      params: [String(10 ** Number(decimals[0]))],
  }];
```

if possible, using viem format function would be safer imo, but not sure if avail. in that repo.

fees/makina/index.ts
```diff
@@ -1,68 +1,132 @@
...
+
+  const cumulativeIndexAfter = await options.toApi.multiCall({
+    abi: convertAbi,
+    calls: convertCalls,
+    permitFailure: true,
+  });
+
+  const balances = options.createBalances();
+
+  for (let i = 0; i < assets.length; i++) {
```
Hugoo, line 73: as i suggest to convert the function to only use 1 machine at a time (which is how it is being used) this assets will always be of length 1

fees/makina/index.ts
```diff
@@ -71,83 +137,32 @@ async function fetch(options: FetchOptions): Promise<FetchResultV2> {
...
-        target: machine.feeManager,
-        abi: ABI.sharePriceWatermark,
-      }),
-    ]);
-
-    // Fetch previous and new share price
-    const prevTokenPrice = await options.fromApi.call({
-      target: machine.sharePriceOracle,
-      abi: ABI.getSharePrice,
+    const totalProtocolRevenueShares = await addTokensReceived({
```
noateden, line 140: can you calculate revenue from `dailyNetYield`? We don't like idea of use difference method for fees and revenue

ColinPlatt: no that is not possible, price accumulation (netYield) is delivered through price accumulation, whereas fees are integrated as newly minted tokens. Both occur in the same call (so they're not really different methods)

https://github.com/MakinaHQ/makina-core/blob/2d470e117022b980729cf03fdbb7e605f6a16417/src/machine/Machine.sol#L398-L410

fees are split between the protocol and the operator (curator) by the fee manager
https://github.com/MakinaHQ/makina-periphery/blob/c8b8df1aaff00053a94ce395e593cc97957b5b4f/src/fee-managers/WatermarkFeeManager.sol#L181-L220

noateden:
hi @ColinPlatt I did refactor a bit, count operators fees to supplySideRevenue

llamabutler:
The makina adapter exports:
        
 
 
    > adapters@1.0.0 test
    > ts-node --transpile-only cli/testAdapter.ts fees makina
    
    🦙 Running MAKINA adapter 🦙
    Start Date:	Thu, 18 Dec 2025 13:59:51 GMT
    End Date:	Fri, 19 Dec 2025 13:59:51 GMT
    
    Token transfers: Failed to use indexer, falling back to logs ethereum Llama Indexer v2 URL/api key is not set
    Token transfers: Failed to use indexer, falling back to logs ethereum Llama Indexer v2 URL/api key is not set
    Token transfers: Failed to use indexer, falling back to logs ethereum Llama Indexer v2 URL/api key is not set
    Token transfers: Failed to use indexer, falling back to logs ethereum Llama Indexer v2 URL/api key is not set
    Token transfers: Failed to use indexer, falling back to logs ethereum Llama Indexer v2 URL/api key is not set
    Token transfers: Failed to use indexer, falling back to logs ethereum Llama Indexer v2 URL/api key is not set
    ETHEREUM 👇
    Backfill start time: 25/10/2025
    Daily fees: 14.56 k
    Daily revenue: 94.00
    Daily protocol revenue: 94.00
    Daily supply side revenue: 14.46 k
    End timestamp: 1766152790 (2025-12-19T13:59:50.000Z)

ColinPlatt:
thanks, @noateden should we include something like HoldersRevenue that removes the Operators? along the lines of the  the Euler methodology

noateden:
hi @ColinPlatt Makina hasn't a token yet, so we don't include HoldersRevenue here

ColinPlatt:
> hi @ColinPlatt Makina hasn't a token yet, so we don't include HoldersRevenue here

sorry, had interpreted it as Vault token holder

>@ColinPlatt For future conversations, should we make a group chat?

yes please. Emailed you

Issue #5164, opened 2025-12-17
Track aera-v2 Fee/Revenue


treeoflife2:
https://defillama.com/protocol/aera-v2

PR #5332, opened 2025-12-22
track aera v2 fees

KPHEMRAJ:
closes https://github.com/DefiLlama/dimension-adapters/issues/5164

### DIFF
--- a/fees/aera-v2/index.ts
+++ b/fees/aera-v2/index.ts
@@ -2,6 +2,7 @@ import { FetchOptions, FetchResult, SimpleAdapter } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
 import pLimit from "p-limit";
 import fetchURL, { postURL } from "../../utils/fetchURL";
+import { METRIC } from "../../helpers/metrics";
 
 const CHAIN_CONFIG: Record<string, any> = {
     [CHAIN.ETHEREUM]: {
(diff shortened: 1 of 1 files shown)


fees/aera-v2/index.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    const feeTokenPrice = await options.api.multiCall({
+        calls: vaults,
+        abi: ABIs.feeTokenPrice,
+        permitFailure: true
+    });
+
+    for (const [index, vaultDetail] of vaultDetails.entries()) {
+        const currentTvlInUsd = vaultValueMap.get(vaultDetail.vaultAddress) || 0;
+        const currentApy = vaultDetail.summary.apy.value;
```
treeoflife2, line 98: is there a way to track fees by event logs like aera v3? so we can refill instead of current apy

KPHEMRAJ: I dont think there is any event to get yields onchain, for example https://etherscan.io/address/0x3d6eef6a92b15361697698695334e98c5db91d6b#events is the contract of vault with 3% apy, I couldnt find any event that has either emitted yield data or fee data.
Though we have claimed event, I think its mostly for fees but never has been emitted cz totalfee collected function returns 0 for all vaults(any ways we are using function call in case its collected in future)

fees/aera-v2/index.ts
```diff
@@ -0,0 +1,150 @@
...
+const ABIs = {
+    vaultCreated: "event VaultCreated (address indexed vault, address assetRegistry, address hooks, address indexed owner,address indexed guardian, address feeRecipient, uint256 fee, string description, address wrappedNativeToken)",
+    feeTotal: "uint256:feeTotal",
+    feeTokenPrice: "uint256:lastFeeTokenPrice"
+}
+
+const limit = pLimit(5);
+
+async function prefetch() {
+    return await postURL("https://app.aera.finance/api/metric/v1", {
```
g1nt0ki, line 47: would be good to turn this into a fetcher and cache in `getConfig`, this way, we are good even when the api becomes unreachable

fees/aera-v2/index.ts
```diff
@@ -0,0 +1,150 @@
...
+    const vaultValueMap = new Map(vaultsOfCurrentChain.map((vaultDetail: any) => [vaultDetail.label.vault_address, vaultDetail.series[0].value]));
+
+    const dailyFees = options.createBalances();
+    const dailyRevenue = options.createBalances();
+    const dailySupplySideRevenue = options.createBalances();
+
+    const vaultCreationLogs = await options.getLogs({
+        eventAbi: ABIs.vaultCreated,
+        targets: CHAIN_CONFIG[options.chain].factoryAddresses,
+        fromBlock: CHAIN_CONFIG[options.chain].startBlock,
```
g1nt0ki, line 68: need to include cacheInCloud flag, so we dont fetch the start each time

PR #5337, opened 2025-12-22
Fix Felix USDHL metrics

FelixBruguera:
Replaced dailyHoldersRevenue with dailySupplySideRevenue

### DIFF
--- a/fees/felix-usdhl/index.ts
+++ b/fees/felix-usdhl/index.ts
@@ -19,7 +19,7 @@ const fetch = async (options: FetchOptions) => {
 
   return {
     dailyFees,
-    dailyRevenue: dailyFees,
+    dailyRevenue: 0,
     dailySupplySideRevenue: dailyFees,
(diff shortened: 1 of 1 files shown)


fees/felix-usdhl/index.ts
```diff
@@ -20,7 +20,7 @@ const fetch = async (options: FetchOptions) => {
   return {
     dailyFees,
     dailyRevenue: dailyFees,
```
g1nt0ki: need to set this as 0 as well?

FelixBruguera: done

FelixBruguera:
> @FelixBruguera i tried to fetch fees for many days, but didn't got any fees. if we don't get distribution created event then we could use the claim event too. https://hyperevmscan.io/address/0xbD647dbcCae38A49149C4f3152b26d2a6Ba1EE6E#events

Try 2025-11-25 or 2025-10-24

Issue #5222, opened 2025-12-18
Track Serum Fee/Revenue


treeoflife2:
https://defillama.com/protocol/serum

bheluga:
Duplicate of https://github.com/DefiLlama/dimension-adapters/issues/5476

PR #5447, opened 2026-01-01
feat: add Serum fees/revenue adapter

Jason4276:
## Add Serum Fees Adapter

Resolves [#5222](https://github.com/DefiLlama/dimension-adapters/issues/5222)

### Summary
Adds historical fees tracking for Serum DEX using Vybe API volume data.

### Implementation Details
- Uses existing Vybe API (same as volume adapter in `dexs/serum.ts`)
- Calculates fees based on Serum/OpenBook's documented fee structure:
  - Taker fee: 4 bps (0.04%)
  - Maker rebate: 2 bps (0.02%)
- Protocol marked as dead from 2023-09-12 (post-FTX collapse)

### Note for Reviewers
- Requires `PROD_VYBE_API_KEY` environment variable
- Local testing without the key will show 403 errors, which is expected
- The adapter gracefully handles missing authentication

### Methodology
- **UserFees**: Taker fees (4 bps on 50% of volume)
- **Fees**: Net fees after maker rebates (2 bps average)
- **Revenue**: ~20% of net fees (UI providers)
- **SupplySideRevenue**: Maker rebates (2 bps on 50% of volume)

### DIFF
--- a/fees/serum/index.ts
+++ b/fees/serum/index.ts
@@ -1,0 +1,155 @@
+import { SimpleAdapter } from "../../adapters/types";
+import { CHAIN } from "../../helpers/chains";
+import { getUniqStartOfTodayTimestamp } from "../../helpers/getUniSubgraphVolume";
+import { gql, GraphQLClient } from "graphql-request";
+import { getEnv } from "../../helpers/env";
+
+/**
+ * Serum  Fees Adapter
+ * Uses Vybe API to fetch historical volume & calculate fee metrics.
+ * Falls back to zero output if no API key or no matching volume.
+ */
+
+const endpoint = "https://api.vybenetwork.com/v1/graphql";
+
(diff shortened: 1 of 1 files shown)


Jason4276:
@bheluga Thanks for the review!

You’re right, relying on Vybe or Bitquery requires API keys, which won’t be available in DefiLlama infrastructure. I also tried using Bitquery as an alternative, but it still depends on an API key and returns zero values across all days.

I wasn’t able to find any reliable public, on-chain–only data source to compute historical Serum fees, especially after activity migrated to OpenBook. Additionally, Serum has been effectively inactive since 2023-09-12, so daily fees are zero after that date.

If required, I can update the adapter to return zero values with a clear methodology explanation and keep deadFrom: 2023-09-12, consistent with other deprecated protocols.

Issue #5237, opened 2025-12-18
Track DFlow Fee/Revenue


treeoflife2:
[DFlow](https://defillama.com/protocol/dflow)

bheluga:
Dflow has no fees for now

PR #5456, opened 2026-01-01
track dflow fees

urvi-bala:
Addresses issue: https://github.com/DefiLlama/dimension-adapters/issues/5237

### DIFF
--- a/fees/dflow/index.ts
+++ b/fees/dflow/index.ts
@@ -1,0 +1,59 @@
+import { CHAIN } from "../../helpers/chains";
+import {
+  Dependencies,
+  FetchOptions,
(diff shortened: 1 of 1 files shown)


urvi-bala:
> @urvi-bala Thanks for the PR. Dflow is dex aggregator and not dex. So fees of trade going through dflow will go to underlying dex and not dflow aggregator

Thanks for the clarification. Since DFlow functions as a DEX aggregator and does not capture fees and there are no details in DFlow’s documentation indicating an explicit fee mechanism should this adapter report protocol fees and revenue as zero?

bheluga:
Dflow doesnt have fees for now

PR #5463, opened 2026-01-02
Fix Aave fee split precision using BigInt arithmetic

davidzk3:
**Summary**
This PR fixes precision issues in Aave fee calculations caused by unsafe `Number` conversions when handling very large values derived from liquidity index math.

**What changed**
- Use pure `BigInt` arithmetic for liquidity index calculations (1e27 scale)
- Prevent silent rounding when splitting interest into protocol vs supply-side revenue
- Preserve existing business logic and fee outputs

**Why this matters**
Aave V3 pools can accumulate very large interest deltas. Converting these values to `Number` can introduce silent precision loss, leading to small but systematic discrepancies in fee attribution.

**Validation**
Manually verified via local adapter tests:
- `pnpm test fees aave-v2 2025-06-01`
- `pnpm test fees aave-v3 2025-06-01`

Results match expected aggregates.

### DIFF
--- a/helpers/aave/index.ts
+++ b/helpers/aave/index.ts
@@ -1,9 +1,9 @@
 import { BaseAdapter, FetchOptions, IStartTimestamp } from "../../adapters/types";
 import * as sdk from "@defillama/sdk";
-import AaveAbis from './abi';
-import {decodeReserveConfig} from "./helper";
+import AaveAbis from "./abi";
+import { decodeReserveConfig } from "./helper";
 import { normalizeAddress } from "@defillama/sdk/build/util";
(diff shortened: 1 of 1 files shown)


davidzk3:
> hi @davidzk3 thanks but I don't think these changes are needed

@noateden 
Thanks for taking a look.

Understood, I’ll close this out as not needed.

Is there any particular area or type of issue you’d prefer focus on? 
I have a couple of PRs in progress already but happy to reprioritize based on what’s most useful.

Cheers

davidzk3:
@noateden 
great, thanks for the pointer.

Issue #5224, opened 2025-12-18
Track Knightrade Fee/Revenue


treeoflife2:
[Knightrade](https://defillama.com/protocol/knightrade)

PR #5465, opened 2026-01-02
Add Knightrade fee & revenue tracking (Solana)

Jason4276:
## Summary

This PR adds **fee and revenue tracking for Knightrade on Solana** using the Dimension v2 adapter format.  
It tracks the Knightrade fee wallet and correctly separates **fees** from **revenue**, including yield generated via Jupiter Lend.

---

## What’s Included

- Added Knightrade fee & revenue adapter (Solana)
- Tracks the official Knightrade fee wallet
- Supports USDC and USDT fee inflows
- Includes jlUSDC and jlUSDT as yield-based revenue
- Uses `createBalances()` (Dimension v2 compatible)
- Safe handling for local testing and RPC/API failures

---

### Fee Wallet

BKVWqzbwXGFqQvnNVfGiM2kSrWiR88fYhFNmJDX5ccyv

### Fees

- USDC + USDT inflows to the fee wallet
- Represents management fees, performance fees, and platform fees

### Revenue

- USDC + USDT (realized fees)
- jlUSDC + jlUSDT (yield tokens from Jupiter Lend)
- Includes both direct fee collection and yield generated from deployed capital

---

## Notes on Local Testing

- During local testing, token balance fetches may return zero due to RPC/API limitations.
- This behavior is expected and handled safely with fallback logic.
- Production runs will use proper RPC access and historical balance resolution.

---

## Related Issue

- https://github.com/DefiLlama/dimension-adapters/issues/5224

---

## Checklist

- [x] Adapter follows Dimension v2 structure  
- [x] Clear separation of fees and revenue  
- [x] Safe error handling for local testing  
- [x] Tested locally using `npm run test fees knightrade`

### DIFF
--- a/fees/knightrade/index.ts
+++ b/fees/knightrade/index.ts
@@ -1,79 +1,137 @@
 import { FetchOptions, SimpleAdapter } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
-import { getTokenBalance } from "../../helpers/solana";
+import { httpGet } from "../../utils/fetchURL";
 
-const FEE_WALLET = "BKVWqzbwXGFqQvnNVfGiM2kSrWiR88fYhFNmJDX5ccyv";
+// -----------------------------
+// Config
+// -----------------------------
 
-const TOKENS = {
-  USDC: "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v",
-  USDT: "Es9vMFrzaCERz9rjXezwPqydat9qYjCCj2LRrPaLd4i8",
-  jlUSDC: "9BEcn9aPEmhSPbPQeFGjidRiEKki46fVQDyPpSQXPA2D",
-  jlUSDT: "3Xd1xYhGJCPZc1oHffLaRFCob4u8FvnZDJHWi5hWzw3y",
-};
+// Knightrade Drift Vault
+const VAULT_ADDRESS = "CqLEEKfZwzp9BLp8kYsnPCFFSUPqMx5wRHJn8GRR9eUj";
 
-// Safe wrapper to handle API failures during local testing
-async function getBalanceSafe(token: string, account: string): Promise<number> {
+// Drift vault historical snapshots (public, no API key)
+const DRIFT_S3_BASE =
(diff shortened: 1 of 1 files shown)


bheluga:
> Hi @bheluga 👋,
> 
> I did a deeper on-chain investigation into Knightrade’s Drift vault fees to make sure we track them correctly.
> 
> ### Findings
> * Knightrade’s Drift vaults charge a **20% performance fee** (management fee is **0%**).
> * The performance fee is applied via Drift’s internal **`profitShare` accounting**, which is emitted in vault events.
> * I scanned recent vault transactions and could not find any `managerWithdraw`, `withdrawProfitShare`, or similar instructions that result in **actual token transfers** to a manager or fee wallet.
> * I also ran an on-chain script over ~200 recent vault transactions to detect outgoing token flows attributable to fee withdrawals, and none were found.
> * This indicates that fees are currently **retained as vault shares inside Drift’s accounting** and have **not yet been materialized as token transfers on-chain**.
> 
> ### Question
> Would it be acceptable to track Knightrade fees using **event-based accounting**, i.e. extracting the `profitShare` value from Drift vault events (and converting it to USD), even though the fees are not yet withdrawn as tokens?
> 
> Thanks for the guidance!

You can simply do this,
Check its drift vaults daily performance, consider all as fees , 20% of it as revenue and 80% as supply side revenue.
To calculate daily performance, just find its vault price of yesterday and today and subtract

fees/knightrade/index.ts
```diff
@@ -0,0 +1,154 @@
...
+// -----------------------------
+
+async function getVaultSnapshots(
+  year: number,
+  month: number
+): Promise<VaultSnapshot[]> {
+  try {
+    const url = `${DRIFT_S3_BASE}/vault/${year}/${String(month).padStart(2, "0")}`;
+    const csv = await httpGet(url);
+
```
bheluga, line 42: @Jason4276 this URL is always throwing an error, so fee is 0 everyday
Do you have anyday where this URL doesnt throw an error?

Jason4276:
@bheluga, I checked multiple months/dates and couldn’t find a single instance where Drift has published S3 vault snapshots for this Knightrade vault. The URL consistently returns NoSuchKey, which is why fees are always zero.

Issue #2824, opened 2025-04-08
List of Broken Fees Adapter


treeoflife2:
## Adapters Requiring Data Source Update
- QuickSwap - Working but data is incorrect (total volume/fees are wrong)
- Reserve Protocol - Indexer stopped working, needs alternate solution
- UniDex - Subgraph stopped, replace with Dune/Allium (https://dune.com/supakawaiidesu/unidex-molten-stats) or other
- KyberSwap - All subgraphs are down
- Buffer Finance - Need to use new subgraph at https://stats.buffer.finance/arbitrum
- ZetaChain - Need new solution for fees
- Angle - Need to use Dune dashboard (https://dune.com/SebVentures/Angle-Dashboard)
- Hydradex - Convert from API to GraphQL at https://info.hydradex.org/graphql (504 error)
- Xena Finance - Replace GraphQL with Dune query (https://dune.com/xena/analytics)
- Moonbeam/Moonriver - need to find new datasource as etherscan api is blocked in server

## API/Service Issues
- JOJO - Fee API is down
- SUITE - Fee API is down
- Firebird - Fee API is down
- MilkyWay Zone - Fee API stopped working
- PLEXUS - API and website are down (possibly deprecated)
- Torch Stable Swap - API down

PR #5525, opened 2026-01-06
Fixs Angle Fees adapter

PatelPrinci:
Addresses issue: https://github.com/DefiLlama/dimension-adapters/issues/2824

## Summary by CodeRabbit

* **Refactor**
  * Simplified fee calculation methodology for Angle protocol with streamlined data collection
  * Updated supported networks to Ethereum and Polygon only (removed Arbitrum and Avalanche)
  * Historical data start date set to August 8, 2023 for both networks

### DIFF
--- a/fees/angle/index.ts
+++ b/fees/angle/index.ts
@@ -2,16 +2,29 @@ import { FetchOptions, SimpleAdapter } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
 import { addTokensReceived } from "../../helpers/token";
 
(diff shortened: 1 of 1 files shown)


noateden (changes requested):
hi @PatelPrinci thanks for the PR. But I think the methodology is wrong. You're counting all tokens transferred to address `0x00253582b2a3FE112feEC532221d9708c64cEFAb` while I saw there are all tokens came from swaps. can you check again and update?

PR #5645, opened 2026-01-17
AdapterFactory and Registry for common helpers + dailyNormalizedVolume

treeoflife2:
## Summary by CodeRabbit

* **New Features**
  * Added a daily normalized volume metric to volume results and a corresponding normalized-volume adapter type.
  * Added factory-provided normalized volume adapters to increase protocol coverage.

* **Chores**
  * Added a registry and helper discovery to enable dynamic adapter fallback and loading.
  * Test/run flows updated to prefer registry-sourced adapters when available.
  * Updated ignore rules to exclude generated/auxiliary build artifacts.

### DIFF
--- a/factory/registry.ts
+++ b/factory/registry.ts
@@ -26,7 +26,7 @@ export function createFactoryExports<T extends { [key: string]: SimpleAdapter }>
 // Legacy helpers are stored in helpers/{filename}.ts (marked with 'helpers/' prefix)
 const factoriesByAdapterType: { [adapterType: string]: string[] } = {
   'fees': [
-    // 'helpers/liquity',  // Legacy helper in helpers/ folder
+    'helpers/liquity',  // Legacy helper in helpers/ folder
   ],
   'dexs': [
     // 'normalizedVolume', // Factory in factory/ folder
--- a/fees/felix.ts
+++ b/fees/felix.ts
@@ -1,6 +1,0 @@
-import { CHAIN } from "../helpers/chains"
-import { liquityV2Exports } from "../helpers/liquity"
-
-export default liquityV2Exports({
-  [CHAIN.HYPERLIQUID]: { collateralRegistry: '0x9De1e57049c475736289Cb006212F3E1DCe4711B', stableTokenAbi: "address:feUSDToken", stabilityPoolRatio: 1, start: '2025-03-14' }
-})
--- a/helpers/liquity.ts
+++ b/helpers/liquity.ts
@@ -1,4 +1,6 @@
 import { BaseAdapter, FetchV2, IJSON, SimpleAdapter } from "../adapters/types";
+import { createFactoryExports } from "../factory/registry";
+import { CHAIN } from "./chains";
 import { addTokensReceived, nullAddress } from "./token";
 import { METRIC } from "./metrics";
 
@@ -279,3 +281,15 @@ export function liquityV1Exports(config: IJSON<LiquityV1Config>) {
     },
   } as SimpleAdapter
 }
+
+
+
+
+// Define all protocols
+const protocols = {
+  'felix': liquityV2Exports({
+    [CHAIN.HYPERLIQUID]: { collateralRegistry: '0x9De1e57049c475736289Cb006212F3E1DCe4711B', stableTokenAbi: "address:feUSDToken", stabilityPoolRatio: 1, start: '2025-03-14' }
+  }),
+} as const;
+
+export const { protocolList, getAdapter } = createFactoryExports(protocols);


adapters/types.ts
```diff
@@ -144,6 +144,7 @@ export type FetchResultVolume = FetchResultBase & {
   openInterestAtEnd?: FetchResponseValue
   dailyBridgeVolume?: FetchResponseValue
   totalBridgeVolume?: FetchResponseValue
+  dailyNormalizedVolume?: FetchResponseValue
```
g1nt0ki, line 147: one issue I am seeing here is, we have dailyNotionalVolume for options, so their keys might clash (dnv), not really an issue, we could use `dnvol` here

cli/testAdapter.ts
```diff
@@ -40,42 +41,74 @@ function toTimestamp(timeArg: string) {
...
+    // File doesn't exist, try to find it in helper registry
+    const result = getAdapterFromHelpers(adapterType, moduleArg);
+    
+    if (!result) {
+      // Only show error if not found in registry either
+      console.error(`❌ Protocol "${moduleArg}" not found in ${adapterType}/ or factory registry`);
+      
+      // Show available protocols in helpers for this adapter type
+      const helperProtocols = listHelperProtocols(adapterType);
+      if (helperProtocols.length > 0) {
```
g1nt0ki: not sure if this is a good idea, we will fill the page with the lists?

factory/normalizedVolume.ts
```diff
@@ -0,0 +1,51 @@
...
+  return {
+    version: 2,
+    runAtCurrTime: true,
+    adapter
+  };
+}
+
+// Define all protocols
+const protocols = {
+  'abc': dailyNormalizedVolumeAdapter({
```
g1nt0ki: better to keep only config here, and on the next step do

`Object.entries(protocols).forEach(([k, v]) => protocols[k] = dailyNormalizedVolumeAdapter(v) )`


or  `... createFactoryExports(protocols, dailyNormalizedVolumeAdapter)`

adapters/types.ts
```diff
@@ -212,7 +213,7 @@ export type FetchResult = FetchResultVolume & FetchResultFees & FetchResultAggre
 export const whitelistedDimensionKeys = new Set([
   'startTimestamp', 'chain', 'timestamp', 'block',
 
-  'dailyVolume', 'totalVolume', 'shortOpenInterestAtEnd', 'longOpenInterestAtEnd', 'openInterestAtEnd', 'dailyBridgeVolume', 'totalBridgeVolume',
+  'dailyVolume', 'totalVolume', 'shortOpenInterestAtEnd', 'longOpenInterestAtEnd', 'openInterestAtEnd', 'dailyBridgeVolume', 'totalBridgeVolume', 'dailyNormalizedVolume',
```
g1nt0ki, line 217: need to add a new AdapterType?

cli/testAdapter.ts
```diff
@@ -40,42 +41,74 @@ function toTimestamp(timeArg: string) {
...
+(async () => {
+  const file = `${adapterType}/${moduleArg}`
+  const passedFile = path.resolve(process.cwd(), `./${file}`);
+  
   // throw error if module doesnt start with lowercase letters
   if (!/^[a-z]/.test(moduleArg)) {
     throw new Error("Module name should start with a lowercase letter: " + moduleArg);
   }
+  
+  try {
```
g1nt0ki, line 59: on second thought, can you extract this to a new file? this way, we have a common import function that can be used here and in the server repo

factory/normalizedVolume.ts
treeoflife2: current config is better as if we have slight variation in the AdapterFactory, we can use different once easily
like 
const protocols = {
  'abc': dailyNormalizedVolumeAdapter({
    protocolName: 'abc',
    chains: [CHAIN.ETHEREUM],
    start: '2021-05-05'
  }),
  'abc2': dailyNormalizedVolumeAdapter2({
    protocolName: 'abc2',
    chains: [CHAIN.ETHEREUM],
    start: '2021-05-05'
  }),

} as const;

factory/normalizedVolume.ts
```diff
@@ -0,0 +1,51 @@
...
+  return {
+    version: 2,
+    runAtCurrTime: true,
+    adapter
+  };
+}
+
+// Define all protocols
+const protocols = {
+  'hyperliquid': dailyNormalizedVolumeAdapter({
```
g1nt0ki, line 44: this is still the same

factory/normalizedVolume.ts
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Adapter factory for dailyNormalizedVolume metric
+ */
+function dailyNormalizedVolumeAdapter(config: NormalizedVolumeConfig): SimpleAdapter {
+  const { protocolName, chains, start } = config;
+
+  const adapter: any = {};
+
+  chains.forEach(chain => {
```
g1nt0ki, line 28: this not needed, final response can include chains array and start at the root (will be auto applied to all the chains)

but I am guessing the code here is dummy atm

PR #5666, opened 2026-01-19
chore: hydrex fee adapter

larrettgee:
Hydrex already exists on DeFi Llama with the TVL adapter. Adding in fee adapter here. LMK if you need anything else!

## Summary by CodeRabbit

* **New Features**
  * Added a Hydrex adapter on BASE that provides daily fees, daily revenue, daily balances, and detailed bribe tracking (external vs. internal), with a public fetch entry and start date for historical data.

### DIFF
--- a/fees/hydrex/index.ts
+++ b/fees/hydrex/index.ts
@@ -12,16 +12,20 @@ const event_reward_added = 'event RewardAdded(address indexed rewardToken, uint2
 const fetch = async (options: FetchOptions) => {
   const { createBalances, getLogs, getToBlock } = options;
 
-  const dailyRevenue = createBalances();
+  const dailyFees = createBalances();
+  const dailyProtocolRevenue = createBalances();
+  const dailyHoldersRevenue = createBalances();
+  const dailyBribesRevenue = createBalances();
 
-  // 1. Get options revenue from Option Exercise events (goes to SPR)
+  // 1. Get options revenue from Option Exercise events (goes to Strategic Protocol Reserve/treasury)
   const exerciseLogs = await getLogs({
     target: OPTION_EXERCISE_CONTRACT,
     eventAbi: event_exercise,
   });
 
   exerciseLogs.forEach((log: any) => {
-    dailyRevenue.add(ADDRESSES.base.USDC, log.paymentAmount);
+    dailyFees.add(ADDRESSES.base.USDC, log.paymentAmount);
+    dailyProtocolRevenue.add(ADDRESSES.base.USDC, log.paymentAmount);
   });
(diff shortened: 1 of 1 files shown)


fees/hydrex/index.ts
```diff
@@ -0,0 +1,70 @@
...
+  version: 2,
+  adapter: {
+    [CHAIN.BASE]: {
+      fetch,
+      start: '2025-09-08',
+    },
+  },
+  methodology: {
+    Fees: "Total fees from DEX fees, option exercises (SPR), Omni Liquidity Fees, and bribe rewards",
+    Revenue: "Total revenue from DEX fees, option exercises (SPR), Omni Liquidity Fees, and bribe rewards",
```
treeoflife2: what's bribe rewards in this? we track dailyBribeRevenue separtely, and bribes paid to pools won't be included in the dailyRevenue

fees/hydrex/index.ts
```diff
@@ -0,0 +1,70 @@
...
+const adapter: SimpleAdapter = {
+  version: 2,
+  adapter: {
+    [CHAIN.BASE]: {
+      fetch,
+      start: '2025-09-08',
+    },
+  },
+  methodology: {
+    Fees: "Total fees from DEX fees, option exercises (SPR), Omni Liquidity Fees, and bribe rewards",
```
treeoflife2: what's SPR, here? we should also track yield generated by the pool as dailyFees, and dailyRevenue(part of fees going back to token holders if any + fees going to protocol treasury)

fees/hydrex/index.ts
larrettgee: Both DEX fees & bribes are called "Bribes" in our contracts (internal vs external bribes). I've updated the adapter to better fit your definitions and track the distinction here.

fees/hydrex/index.ts
larrettgee: Thank you, updated based on your recommendation. SPR = Strategic Protocol Reserve which is a protocol treasury. Updated accordingly

treeoflife2:
@larrettgee created new PR as can't edit this one, https://github.com/DefiLlama/dimension-adapters/pull/5666
fixed many issues like there is no such things as internal/external bribes. and we don't count bribes as fees and don't add it in revenue or holderrevenue too, we have toggle in defillama UI at the top right corner to include bribes in revenue(so its as per user preference). and is there a way we could track bribes and revenue for hydrex-omni and hydrex-intergral separately? closing this for now, you can re-open if have any doubts and want to make changes, otherwise you can create new PR. thanks

PR #5301, opened 2025-12-20
feat: track termmax fees/rev 

kanerep:
resolves: #5150 

### Summary 
Adds a fee adapter for TermMax protocol. Tracks protocol revenue by monitoring tokens received by the treasury address across Ethereum, Arb, and BSC.

**TermMax is a DeFi protocol offering:**
- Fixed-rate lending/borrowing through range orders
- Gearing Token (GT) minting for leveraged positions
- Vault-based yield strategies with curator performance fees

**Fee Sources:**
- Swap/trading fees (taker portion →  treasury, maker portion → LPs)
- GT minting fees (100% → treasury)
- Vault performance fees (when withdrawn by curators)

#### Resources
- https://docs.ts.finance/
- https://github.com/term-structure/termmax-contract-v2

PR #5794, opened 2026-02-02
[TermMax] fee adapter

tkspring-henry:
This PR is intended to replace [https://github.com/DefiLlama/dimension-adapters/pull/5301](https://github.com/DefiLlama/dimension-adapters/pull/5301).

The previous implementation only tracked the yield generated by the vault itself, but the calculation method was incorrect.

TermMax currently has three revenue streams:

- Protocol fee
- Liquidation penalty
- Performance fee

TermMax charges fees in the form of FT (Fixed-rate Token). However, DefiLlama currently does not provide an oracle for FT, so the real value of the fees would be underestimated. These FT tokens are redeemable 1:1 into the underlying asset at maturity. To simplify the calculation (since the exact underlying asset price at redemption cannot be predicted), we treat the FT’s value as the underlying asset price at the time the fee is charged. Because FT trades at a discount, this approach still underestimates the fee value, but it is closer to the real fee value than treating FT as having zero value (as before).

If a borrower cannot repay the debt before maturity, the liquidation penalty at maturity will also flow into the treasury in the form of collateral, which is also revenue for TermMax.

The vault’s idle funds are deployed into pools and parked on Aave / Morpho to earn yield. After that yield is distributed according to TermMax’s documented rules, a designated wallet transfers TermMax’s allocated share of the revenue into the treasury.

## Summary by CodeRabbit

* **New Chains**
  * Added support for Berachain network.


fees/termmax.ts
```diff
@@ -1,108 +1,219 @@
...
+      // _4: Underlying Token (debt) - we use this for protocol fee valuation
+      abi: "function tokens() view returns (address, address, address, address, address)",
+      calls: tuples.map((t) => t.marketAddress),
+      permitFailure: true,
+    });
+    for (let i = 0; i < tuples.length; i++) {
+      const { log } = tuples[i];
+      const tokens = allTokens[i];
+      if (tokens && tokens[4]) {
+        // FT contract -> Protocol fee (valued in underlying token)
```
treeoflife2: FT trades at a discount. The actual value at fee collection time is lower than the underlying, this would be incorrect as price could move drastic until it unlocks, so it would be wrong to price it as 1:1 even tho it's discounted, maybe we could track the fees at when it matures?

treeoflife2 (changes requested):
.

llamabutler:
The termmax.ts adapter exports:
        
 
 
    > adapters@1.0.0 test
    > ts-node --transpile-only cli/testAdapter.ts fees termmax.ts
    
    🦙 Running TERMMAX.TS adapter 🦙
    Start Date:	Mon, 02 Mar 2026 02:43:36 GMT
    End Date:	Tue, 03 Mar 2026 02:43:36 GMT
    
    ------ ERROR ------
    TypeError: Cannot read properties of undefined (reading 'success')
        at /home/runner/work/dimension-adapters/dimension-adapters/node_modules/.pnpm/@defillama+sdk@5.0.208_typescript@5.9.3/node_modules/@defillama/sdk/build/abi/index.js:205:54
        at Array.filter (<anonymous>)
    Cannot read properties of undefined (reading 'success')

llamabutler:
The termmax.ts adapter exports:
        
 
 
    > adapters@1.0.0 test
    > ts-node --transpile-only cli/testAdapter.ts fees termmax.ts
    
    🦙 Running TERMMAX.TS adapter 🦙
    Start Date:	Mon, 02 Mar 2026 04:13:06 GMT
    End Date:	Tue, 03 Mar 2026 04:13:06 GMT
    
    chain     | Daily user fees | Daily fees | Daily revenue | Daily protocol revenue | Start Time
    ---       | ---             | ---        | ---           | ---                    | ---       
    ethereum  | 6.61 k          | 6.61 k     | 6.61 k        | 6.61 k                 | 27/3/2025 
    arbitrum  | 0.00            | 0.00       | 0.00          | 0.00                   | 27/3/2025 
    bsc       | 133.00          | 133.00     | 133.00        | 133.00                 | 28/5/2025 
    berachain | 0.00            | 0.00       | 0.00          | 0.00                   | 8/7/2025  
    bsquared  | 0.00            | 0.00       | 0.00          | 0.00                   | 15/2/2026 
    Aggregate | 6.75 k          | 6.75 k     | 6.75 k        | 6.75 k                 |

fees/termmax.ts
tkspring-henry: Good point — we've addressed this in the latest revision. FT fees are no longer counted 1:1.

We now discount FT fees using the actual swap price from the on-chain Swap event in the same transaction, reflecting the realized market value at trade time. For cases without a matching swap (e.g. `issueFt` mints), a conservative 5% annual discount is applied based on time-to-maturity.

PR #5830, opened 2026-02-06
updated fee & revenue

neelra8n:
## Summary by CodeRabbit

* **Refactor**
  * Reworked yield fee calculation to aggregate on-chain events across multiple contracts, replacing legacy ERC4626 logic with an event-driven, per-vault rate-difference model.
* **New Features**
  * Added multi-chain adapter with centralized contract/ABI configuration, async log-driven fetch, expanded return fields (dailyFees, dailyRevenue, dailySupplySideRevenue, dailyProtocolRevenue) and allowNegativeValue support.
* **Documentation**
  * Updated methodology texts to describe combined management and performance fee computation.


bheluga (changes requested):
Hey @neelra8n  thanks for the PR
I have refactored the code, can you verify?
Also according to your code, fees and revenue is calculated only on eth and bsc, so whats the point of so many chains?

neelra8n:
> Hey @neelra8n thanks for the PR I have refactored the code, can you verify? Also according to your code, fees and revenue is calculated only on eth and bsc, so whats the point of so many chains?

hey @bheluga thanks for reviewing and refactoring my PR, 
sure we can remove the unused chain, also the numbers looks a little off, let me update the changes 
thanks

PR #5921, opened 2026-02-16
fix pacifica perp volume double counting

bheluga:
## Summary by CodeRabbit

* **New Features**
  * Pacifica DEX now aggregates data across multiple tickers and adds a dedicated open-interest report.
  * Default dataset now includes a start date for historical fetches.
* **Bug Fixes**
  * Adjusted daily volume calculation (volumes recalibrated to avoid double-counting).

### DIFF
--- a/dexs/pacifica/index.ts
+++ b/dexs/pacifica/index.ts
@@ -19,7 +19,7 @@ const fetch = async (_a: any, _b: any, options: FetchOptions) => {
       const todaysData = data.data.filter((kline: any) => kline.t == options.startOfDay * 1000);
       const volume = (todaysData[0].v * +todaysData[0].c) / 2; // They include taker + maker in ohlcv candles
       dailyVolume += volume;
-      await new Promise(r => setTimeout(r, 500));
+      await new Promise(r => setTimeout(r, 1000));
     })
 
   if (errors.length > 0) {


dexs/pacifica/index.ts
```diff
@@ -8,7 +8,7 @@ const fetch = async (_a: any) => {
   const oi = data.reduce((a: number, b: { open_interest: string, mark: string }) => a + (Number(b.open_interest) * Number(b.mark)), 0)
 
   return {
-    dailyVolume: volume,
+    dailyVolume: volume/2,
```
treeoflife2: @bheluga can you use their kline api, so we have backfilling support? https://api.pacifica.fi/api/v1/klines, first use info endpoint to get all active perp markets and then use kline api for all markets with their marketId. 
and also let's separate the open-interest in it's own folder, so it doesn't rewrite older data.

ops-pacifica: @bheluga @treeoflife2 this /2 is wrong, maker and taker both contribute the fees are volume should be counted this way. This is the standard for any exchange out there. We are reverting this PR. Mistakes like this shouldn't be merged to DefiLlama unchecked. This hurts the exchange reputation immensely, and we are taking this very seriously.

ops-pacifica: Also, doing it via kline for all symbols is not scalable and reliable, this adds unnecessary traffic to both DefiLlama and Pacifica's loads. Will revert this PR entirely.

treeoflife2: > @bheluga @treeoflife2 this /2 is wrong, maker and taker both contribute the fees are volume should be counted this way. This is the standard for any exchange out there. We are reverting this PR. Mistakes like this shouldn't be merged to DefiLlama unchecked. This hurts the exchange reputation immensely, and we are taking this very seriously. 

That doesn't matter if both are paying fees, we don't count maker + taker volume for any perp exchanges

treeoflife2: > Also, doing it via kline for all symbols is not scalable and reliable, this adds loads to both DefiLlama and Pacifica's loads. Will revert this PR entirely. 

Why is that, we have added rate limits to prevent the issue of missing any market, in case of 429 we fail the adapter and retry next hour

PR #5976, opened 2026-02-23
Track Humidifi revenue

FelixBruguera:


fees/humidifi.ts
```diff
@@ -0,0 +1,66 @@
...
+        SELECT
+            token,
+            COALESCE(SUM(amount_sold), 0) AS amount_sold,
+            COALESCE(SUM(amount_bought), 0) AS amount_bought
+        FROM (
+            SELECT
+                token_sold_mint_address AS token,
+                token_sold_amount_raw AS amount_sold,
+                0 AS amount_bought
+            FROM dex_solana.trades
```
treeoflife2, line 24: instead of tracking trades? can we track inflow/outflow to their address and check the token increase after adjusting normal non-dex transfers inflow/outflows, to get better picture

FelixBruguera: But that's what the query is doing: 
amount_sold = outflows
amount_bought = inflows
and the dex_solana.trades table only has dex transfers

PR #5982, opened 2026-02-24
Add GateSwap multi-chain volume adapter

defioasis-eth:
## GateSwap Volume Adapter

This PR adds volume tracking for GateSwap across multiple EVM chains.

### Methodology

- Gate Layer: volume derived from factory pairs and Swap events.
- Other EVM chains: volume derived strictly from router execution events emitted by GateSwap routers.
- First matching ABI per router is used to avoid overlapping event variants.
- Underlying pool events are not counted on non-GateLayer chains to prevent double counting.

### Routers

- New main router: 0x00000000d4c4a3528948586472f519d5fa2200ff
- Legacy router: 0x0000000067a38ab30ba8db9eebf5722c50d8efc0

### Notes

Router ABI extraction uses a best-effort candidate set.
If a more precise event signature is identified, ROUTER_SWAP_EVENT_ABIS will be updated accordingly.

## Release Notes

* **New Features**
  * Expanded multi-chain support including Gate Layer and multiple EVM networks
  * Improved volume calculation accuracy with enhanced router architecture

* **Refactor**
  * Restructured adapter for improved scalability and robustness with better error handling across chains

### DIFF
--- a/dexs/gateswap/index.ts
+++ b/dexs/gateswap/index.ts
@@ -7,7 +7,19 @@ import { filterPools } from '../../helpers/uniswap'
 const FACTORY_ADDRESS = '0xaD8d59f3e026c02Aed0DAdFB46Ceca127030DFa2'
 const SWAP_EVENT = 'event Swap(address indexed sender, uint amount0In, uint amount1In, uint amount0Out, uint amount1Out, address indexed to)'
 
-const fetch: FetchV2 = async (fetchOptions) => {
+const NEW_MAIN_ROUTER = '0x00000000d4c4a3528948586472f519d5fa2200ff'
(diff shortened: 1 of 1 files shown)


defioasis-eth:
Hi, thanks for the feedback.

I’ve reviewed the comments and agree that since the EVM routers are part of the Gate DEX aggregator, it makes more sense to list them as a separate aggregator under aggregators/gate.

Additionally, I identified the reason why EVM chains were returning 0 volume. The router contract is not verified, so the previous implementation relying on event ABI matching was unreliable. I will refactor the logic to use topic0-based log filtering with manual decoding to ensure accurate volume extraction across chains.

Given that this involves:

Moving the EVM logic into a new aggregator listing

Refactoring the log parsing approach

Updating the structure to align with the aggregator template

I will close this PR and reopen a clean one with the updated implementation.

Thanks again for the guidance.

PR #5860, opened 2026-02-10
Add NanoPort aggregator adapter

0xerodev:
## Protocol Description

NanoPort is a liquidity execution fabric / meta-aggregator that routes swaps through multiple DEX aggregators (ParaSwap, Odos, KyberSwap) to find optimal execution for on-chain trades.

- **Website:** https://nano-labs.io
- **Category:** Aggregator
- **Chains:** Ethereum, Arbitrum, Optimism, Base, Polygon, BSC, Avalanche
- **Executor wallet:** `0xF0E9286CfCB75c94ac19E99bCD93D814da55e304`

## Methodology

Volume is tracked via Dune SQL by summing all DEX trades originating from the NanoPort executor wallet (`tx_from`). This captures all swaps routed through any underlying DEX aggregator.

## Adapter Details

- **Type:** Dune SQL based (prefetch pattern)
- **Query:** `dex.trades WHERE tx_from = executor`
- **Multi-chain:** Yes, grouped by blockchain in single prefetch query

## Checklist

- [x] Adapter follows existing patterns (modeled after dflow/enso)
- [x] `Allow edits by maintainers` enabled
- [x] No changes to package-lock.json

## Summary by CodeRabbit

* **New Features**
  * Added NanoPort aggregator support for tracking and reporting daily USD trading volume across Ethereum, Arbitrum, Optimism, Base, Polygon, BSC, and Avalanche C-Chain; includes a start date and automated daily volume calculation using on-chain/external query results.

PR #6011, opened 2026-02-27
Remove NanoPort aggregator adapter

0xerodev:
The NanoPort aggregator adapter (merged in #5860) is showing zero volume across all chains and all time because the Dune SQL query pattern does not match our actual execution flow.

NanoStack is a cross-chain execution protocol — our executor wallet does not appear in `dex.trades` as a standard DEX aggregator. The adapter produces inaccurate data (all zeros) which misrepresents the protocol on DeFiLlama.

Requesting removal rather than a bad listing.


treeoflife2:
@0xerodev can you explain how it works, i could try to fix this?

PR #6238, opened 2026-03-21
optimize the flashbot dune query

FelixBruguera:
## Summary by CodeRabbit

* **Refactor**
  * Queries now use a unified, parameterized time-window for all date filtering, allowing consistent, flexible range queries.
  * Final aggregations simplified to compute daily proposer revenue directly from fee and MEV sums.

* **Chores**
  * Removed several intermediate staking and supply metrics from the pipeline; these derived metrics are no longer produced.

### DIFF
--- a/helpers/queries/flashbots.sql
+++ b/helpers/queries/flashbots.sql
@@ -8,11 +8,10 @@ with
         from
             ethereum.transactions t
             join ethereum.blocks b on block_number="number"
+            and b."date" >= cast(from_unixtime({{start}}) as date)
+            and b."date" < cast(from_unixtime({{end}}) as date)
         where
-            (
-                block_date between date('2020-01-01') and date(now())
-            )
-            and (date between date('2020-01-01') and date(now()))
+            TIME_RANGE
         group by
             1,
             2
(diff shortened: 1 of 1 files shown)


ErebusCipher:
Noticed Flashbots fees have shown $0 since Mar 20, right after this query optimization merged.

I checked whether the query itself was broken. The current SQL returns non-zero when run directly on Dune: for 2026-05-28, ethereum.transactions alone shows 215.35 ETH in priority fees before MEV/coinbase transfers. I also ran the current adapter locally with DUNE_API_KEYS set, and it returned ~$204.6k daily fees for 2026-05-29.

So the SQL and adapter path appear to work locally; the production $0 looks more like an execution/runtime issue for the expensive Dune adapter than a query-level bug. Sharing in case useful for debugging the live data.

FelixBruguera:
> Noticed Flashbots fees have shown $0 since Mar 20, right after this query optimization merged.
> 
> I checked whether the query itself was broken. The current SQL returns non-zero when run directly on Dune: for 2026-05-28, ethereum.transactions alone shows 215.35 ETH in priority fees before MEV/coinbase transfers. I also ran the current adapter locally with DUNE_API_KEYS set, and it returned ~$204.6k daily fees for 2026-05-29.
> 
> So the SQL and adapter path appear to work locally; the production $0 looks more like an execution/runtime issue for the expensive Dune adapter than a query-level bug. Sharing in case useful for debugging the live data.

Hey, thanks for flagging this, I will take a look

PR #6261, opened 2026-03-24
fix(pyth-network): update all EVM chain fees from OP-PIP-93 : move calculation onchain

pythia-assistant:
## Problem
The Pyth Core (pyth-network) adapter was showing near-zero fees because most chains used `DEFAULT_FEE = 1n` (1 wei) instead of the actual governance-approved fee values.

**Before:** ~$54/day across 70+ chains (almost all zeros)
**After:** Should reflect actual protocol revenue

## Solution
Updated `feePerUpdateByChain` with all correct fee values from:
- **Source:** [Pyth Docs - Current Fees](https://docs.pyth.network/price-feeds/core/current-fees)
- **Governance:** [OP-PIP-93](https://forum.pyth.network/t/passed-op-pip-93-q1-2026-pyth-core-fee-implementation-evm-chains/2346)

## Fee Updates (examples)
| Chain | Old Fee | New Fee |
|-------|---------|---------|
| Ethereum | 1 wei | 0.000003 ETH |
| Arbitrum | 1 wei | 0.000003 ETH |
| Base | 1 wei | 0.000015 ETH |
| Polygon | 1 wei | 0.1 POL |
| BNB | 1 wei | 0.0000125 BNB |
| Avalanche | (had value) | 0.0005 AVAX |

**Total:** 60+ chains now have correct fee values configured.

## Notes
- Fee values are in native token wei units (18 decimals for most chains)
- Fees are set by Pyth DAO governance and may change quarterly
- Source of truth: `apps/developer-hub/content/docs/price-feeds/core/current-fees.mdx` in pyth-crosschain repo

## Summary by CodeRabbit

* **Updates**
  * Fees are now obtained via live on-chain lookup per chain instead of a static mapping.
  * Fee totals are computed using the fetched per-update fee, falling back to a minimal default when unavailable.
  * Methodology text updated to reflect governance-set fees and revenue attribution to the protocol treasury.
  * No public API or adapter shape changes.

### DIFF
--- a/fees/pyth-network/index.ts
+++ b/fees/pyth-network/index.ts
@@ -313,8 +313,7 @@ async function fetchEvm(
     })
 
     const updateCount = updateLogs.length;
-    const feePerUpdate = updateFee || DEFAULT_FEE;
-    dailyFees.addGasToken(BigInt(feePerUpdate) * BigInt(updateCount));
+    dailyFees.addGasToken(BigInt(updateFee) * BigInt(updateCount));
   }
 
   return {


fees/pyth-network/index.ts
```diff
@@ -277,22 +277,92 @@ const evmChainConfig: Record<string, { start: string; contract: string }> = {
...
+  [CHAIN.LIGHTLINK_PHOENIX]: 1_000_000_000_000_000_000n, // 1 LL
+  [CHAIN.MERLIN]: 100_000_000_000n, // 0.0000001 BTC
+  [CHAIN.METER]: 40_000_000_000_000_000n, // 0.04 MTR
+  [CHAIN.MEZO]: 100_000_000_000n, // 0.0000001 BTC
+  [CHAIN.MONAD]: 5_000_000_000_000_000n, // 0.005 MON
+  [CHAIN.NEON]: 150_000_000_000_000_000n, // 0.15 NEON
+  [CHAIN.PLASMA]: 50_000_000_000_000_000n, // 0.05 XPL
+  [CHAIN.SHIMMER_EVM]: 100_000_000_000_000_000_000n, // 100 SMR
+  [CHAIN.STORY]: 6_000_000_000_000_000n, // 0.006 IP
+  [CHAIN.WEMIX]: 30_000_000_000_000_000n, // 0.03 WEMIX
```
bheluga: @pythia-assistant is there a way we get it through api? so we dont miss new updates and also since the list is big?
If not, you can still keep it manual but time based, say it was this much before this timestamp and new fees after some timestamp, so we dont use same fees for all periods while refilling

bheluga: @pythia-assistant its even better if a function of a contract returns these value, instead of hardcoding

pythia-assistant:
@bheluga Thanks for the feedback! I've updated the adapter with a hybrid approach:

## Changes

### 1. Dynamic Fee Fetching from Contract
For recent data (last 7 days), the adapter now calls `singleUpdateFeeInWei()` directly on each chain's Pyth contract:

```typescript
const fee = await options.api.call({
  target: contract,
  abi: "function singleUpdateFeeInWei() view returns (uint256)",
});
```

This means **new fee changes via governance will be picked up automatically** without code updates.

### 2. Timestamp-Based Historical Fees
For historical backfills, calling the contract would return today's fee (not the fee at the historical block). To handle this correctly:

- Added `OP_PIP_93_EFFECTIVE_TIMESTAMP` (Feb 25, 2026) marking when current fees became active
- Historical queries before this date use the appropriate fee for that period
- Historical queries after this date use the fee table from OP-PIP-93

### 3. Warning When Fallback Used
Per CodeRabbit's suggestion, added `console.warn()` when a chain falls back to DEFAULT_FEE (1 wei), making it easier to spot missing configurations.

## Why This Approach?

| Query Type | Fee Source | Why |
|------------|------------|-----|
| Recent (< 7 days) | Contract call | Always current, auto-updates with governance |
| Historical (post OP-PIP-93) | Fee table | Contract returns current fee, not historical |
| Historical (pre OP-PIP-93) | DEFAULT_FEE (1 wei) | Matches on-chain reality at the time |

This addresses both concerns:
- ✅ **No stale fees** — recent data fetches live from contract
- ✅ **Correct historical backfills** — uses fee table with timestamps

Let me know if you'd like any adjustments!

PR #6358, opened 2026-04-08
feat: refactor lite vault fees and add lite-usd vault fee calculations

SamarendraGouda:


bheluga (changes requested):
@SamarendraGouda thanks for the PR.
Can you include yields generated in fees too?
You can add it as both dailySupplySideRevenue and dailyFees 
It would also be great if you can breakdown revenue into withdraw fees and perfomance fees instead of just Lite Vaults Fees.
Should be easy , 
dailyFees = all asset yields ( (exchange rate after - before) *  supply) + withdraw/exit fees
dailyRevenue = 0.2 * asset yields + withdraw/exit fees
dailySupplySideRevenue = 0.8 * assetyields

bheluga:
@SamarendraGouda thanks for the PR
Added [here](https://github.com/DefiLlama/dimension-adapters/pull/8549)

PR #6398, opened 2026-04-14
dedup ll fees

lordshisho:
simply adding a cutoff timestamp and condition where fee events are used properly as the protocol changed to avoid any duplication

also wondering if its possible to have liquidlaunch fees added to the liquid labs parent similarly to how it was done here: https://github.com/DefiLlama/defillama-server/pull/11575

thanks


bheluga:
@lordshisho thanks for the PR.
Can you explain a little whats the cutoff time for and what changed after cutoff? and what you are trying to achieve

lordshisho:
> @lordshisho thanks for the PR. Can you explain a little whats the cutoff time for and what changed after cutoff? and what you are trying to achieve

gm, after cutoff time we no longer bond to external dex and the fee claiming becomes available at token creation. this means that the 1% fee on buys and sells is duplicated with the fee claim event after the cutoff.

the goal is to no longer double count these fees after the cutoff time where the fee claim event is the source of truth

PR #6399, opened 2026-04-14
fix(kasu): correct fees magnitude, multi-deployment per chain, add XDC USDC pool

kivanov82:
## Summary
- **Fix Base fees magnitude (~10x overstated):** `dailyFees` now reports the actual performance fee captured on-chain (`ecosystemFee + protocolFee`) rather than the back-calculated gross interest. Previously Base was reporting ~$42k/day vs the real ~$4k.
- **Support multiple deployments per chain:** refactored per-chain config to `Record<chain, Deployment[]>` (`{feeManager, systemVariables, paymentToken}`).
- **Add XDC USDC deployment** alongside the existing AUDD one.
- **Fix XDC AUDD token:** was incorrectly set to USDC.e; now uses the real AUDD address (`0x9fe4e632...`). AUDD is the Australian dollar stablecoin (~$0.66), so the prior config was mispricing XDC fee values.
- Adds `dailySupplySideRevenue` (lender share of gross interest, back-calculated from on-chain `performanceFee`).
- Adds `breakdownMethodology`.

## Test plan
- [ ] `npx tsx adapters-tools/cli/runAdapter.ts fees/kasu.ts` runs cleanly across Base / Plume / XDC
- [ ] Base `dailyFees` aligns with `dailyRevenue` (Kasu remits 100% of performance fees to protocol+ecosystem)
- [ ] XDC: both AUDD and USDC pools surface in fee output


kivanov82:
Quick note on the CI bot output showing all-zero slices:

`FeesEmitted` on Kasu only fires at pool settlement, not continuously. On Base there have been ~2 events in the last ~4.6 days (last one at Apr 12 23:50 UTC, block 44624233). The CI's default 24h test window (Apr 13 17:00 → Apr 14 17:00 UTC) genuinely contained zero events — so zero is the correct output for that window.

Running the adapter locally against a window that includes a real event (e.g. `npm test -- fees kasu.ts 2026-04-13`) produces:

```
base | Daily fees: 4.26k | Daily revenue: 4.26k | Daily supply side: 38.37k
```

which matches on-chain reality.

Happy to remove `pullHourly: true` (upstream added it in #5995) if it makes the test output less noisy, but that wouldn't change the underlying "zero on days without settlement" behaviour — it's just how Kasu's fee-emission cadence works.

bheluga (commented):
@kivanov82 thanks for the PR.
We do consider interests under fees.
so fees = revenue + supplySideRevenue.
You can refer https://docs.llama.fi/list-your-project/other-dashboards and https://docs.llama.fi/analysts/data-definitions for more information

kivanov82:
Thanks @bheluga — updated. `dailyFees` now reports gross borrower interest (`totalFee + supplySide`), restoring `dailyFees == dailyRevenue + dailySupplySideRevenue`. Fees breakdown now uses `METRIC.BORROW_INTEREST`.

PR #6415, opened 2026-04-17
edit priority trade's adapter

thomasxpx:
hi, sorry for another close pull request to last one i didn't test well enough and the fees are wrongly calculated

also removed volume and the unclean topic code


bheluga:
@thomasxpx thanks for the pr.
I don't get the reason behind this PR.
The adapter which was added before has clean logs from event abi including volume.
aAso treasury address hasn't received any transactions from router , so i don't think this is the best approach over the previous one

thomasxpx:
> @thomasxpx thanks for the pr. I don't get the reason behind this PR. The adapter which was added before has clean logs from event abi including volume. aAso treasury address hasn't received any transactions from router , so i don't think this is the best approach over the previous one

oh my bad i thought you didn't like so much the previous version. 
the fees are still miss calculated let me revert with only the fees correction

bheluga:
> > @thomasxpx thanks for the pr. I don't get the reason behind this PR. The adapter which was added before has clean logs from event abi including volume. aAso treasury address hasn't received any transactions from router , so i don't think this is the best approach over the previous one
> 
> oh my bad i thought you didn't like so much the previous version. the fees are still miss calculated let me revert with only the fees correction

Can you resolve merge conflicts?
also try to use addTokensReceived helper instead of allium
also what exactly was wrong with fee calculation? isn't it 1% of volume?

thomasxpx:
> > > @thomasxpx thanks for the pr. I don't get the reason behind this PR. The adapter which was added before has clean logs from event abi including volume. aAso treasury address hasn't received any transactions from router , so i don't think this is the best approach over the previous one
> > 
> > 
> > oh my bad i thought you didn't like so much the previous version. the fees are still miss calculated let me revert with only the fees correction
> 
> Can you resolve merge conflicts? also try to use addTokensReceived helper instead of allium also what exactly was wrong with fee calculation? isn't it 1% of volume?

resolved
i saw with devs and the fees error comes from they changed the router contract after a couple of days. miss communication between us.
yes it's 1% of volume

sorry for the disturbing, everything is correct on what's currently on defillama i'll close the pr

PR #6418, opened 2026-04-17
fix: calc fix, new tranches add

0xnaman1:
#### Name (to be shown on DefiLlama)

Strata

#### Changes
Fixed performance fee calculation + added 2 new tranche contracts.

### DIFF
--- a/fees/strata-markets/index.ts
+++ b/fees/strata-markets/index.ts
@@ -100,7 +100,6 @@ async function computeMetrics(options: FetchOptions, config: TokenConfig) {
   // note: distributeReserve does not change accounting.nav (reserve moves to tranches within accounting)
   const deltaNav = Number(navEnd) - Number(navStart);
   const strategyYield = deltaNav - netUserFlows + totalReserveReductionsBaseAssets;
-  const positiveYield = strategyYield > 0 ? strategyYield : 0;
 
   // Performance fee via reserve-diff:
   // reserveNavEnd = reserveNavStart + perfFee + redemptionFeesToReserve
@@ -111,10 +110,10 @@ async function computeMetrics(options: FetchOptions, config: TokenConfig) {
     deltaReserveNav + totalReserveReductionsBaseAssets + distributeReserveTotal - redemptionFeesReserve
   );
 
-  const totalFees = positiveYield + redemptionFeesTotal;
+  const totalFees = strategyYield + redemptionFeesTotal;
   const protocolRevenue = perfFee + redemptionFeesReserve;
   const totalRevenue = perfFee + redemptionFeesReserve;
-  const supplySideRevenue = positiveYield - perfFee + redemptionFeesTranche;
+  const supplySideRevenue = strategyYield - perfFee + redemptionFeesTranche;
 
   return { totalFees, totalRevenue, protocolRevenue, supplySideRevenue };
 }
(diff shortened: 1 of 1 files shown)


fees/strata-markets/index.ts
```diff
@@ -26,80 +37,84 @@ interface TokenConfig {
...
-  const totalFees = grossYield + redemptionFeesTotal;
-  const protocolRevenue = performanceFee + redemptionFeesReserve;
-  const totalRevenue = performanceFee + redemptionFeesReserve;
-  const supplySideRevenue = grossYield - performanceFee + redemptionFeesTranche;
+  // Strategy yield:
+  // navEnd = navStart + strategyYield + deposits - withdrawals(net) - reserveReductionsToTreasury
+  // note: distributeReserve does not change accounting.nav (reserve moves to tranches within accounting)
+  const deltaNav = Number(navEnd) - Number(navStart);
+  const strategyYield = deltaNav - netUserFlows + totalReserveReductionsBaseAssets;
+  const positiveYield = strategyYield > 0 ? strategyYield : 0;
```
bheluga: we want to allow this to be negative since there are ups and downs
or else cumulative fees would be inflated

0xnaman1: if the protocol yield is negative, we cover the losses using the liquidity from the tranches and the reserve, hence the gain never goes below zero. 

for reference: [strata gain calculation](https://github.com/Strata-Markets/contracts/blob/f25b4a7ad3a898c64530101119d5b2ca8f642005/contracts/tranches/Accounting.sol#L282)

bheluga: but junior tranche holders(supply side) would realize losses right?

0xnaman1: yes. 

earlier, i was directly using strategyYield but for few days, the llama started showing negative fees on charts. plz check [here](https://defillama.com/protocol/strata) for 7 April. upon reset, other days got fixed while 7 April still shows negative. 

clamping to zero seemed like a reasonable thing to do to. 

or could there be another likely reason for the negative fee?

bheluga: 7th april around 7-8  am UTC, 
these were the numbers:
{ grossYield : -4.3301866709840965e+22 
deltaNav: 143558389340236150000 
netUserFlows: 4.34454250991812e+22 
}, so fees was -42k for that hour.
any idea what happened around that time?

0xnaman1: let me check and get back

0xnaman1: i just ran the already existing script on llama locally. 

both for 6-7 April, 7-8 April. 
now i didn't get a single negative value in fees while earlier i was getting a lot of negative values in between hours. 

same script, no change. 
thoughts ?

bheluga: @0xnaman1 i have refilled that date and should be better now.
Can you remove this line

0xnaman1: i've removed the positiveYield yes

0xnaman1:
hey @bheluga , could you plz also reset the fees for 2, 3 and 9th April. 

they seem incorrect on the chart. 

2,3 April fees shows +$100K
9 April fees shows only $3K

bheluga:
> hey @bheluga , could you plz also reset the fees for 2, 3 and 9th April.
> 
> they seem incorrect on the chart.
> 
> 2,3 April fees shows +$100K 9 April fees shows only $3K

I still get higher fees on april 2nd and 3rd
I dont think const grossYield = deltaNav - netUserFlows + (totalReserveReductionsUSD * WAD); is correct?
Sometimes there is huge difference between deltaNav and netUserFlows, try running ``` yarn test fees strata-markets 2026-04-04```

PR #6457, opened 2026-04-22
Add Flying Tulip Lend fees adapter

patcito:
## Summary

Adds \`fees/flying-tulip-lend.ts\` covering daily borrow interest on Flying Tulip Lend (Sonic).

Daily fees per reserve: \`borrows * IRM.irmSampleAPR(util) * windowSeconds / (1e18 * 365d)\`. All borrower interest accrues to the protocol \`reserves\` accumulator, so \`dailyFees == dailyRevenue == dailyProtocolRevenue\`. Supply-side revenue on this protocol comes from a separate FT epoch-rewards stream and is not tracked in this adapter.

Flying Tulip Lend is a custom lending market (\`PositionsManager\` + \`LendingLens\`) and is not Aave V3-compatible, so the Aave helper is not reused.

## Companion PRs

- DefiLlama-Adapters: TVL adapters for ftUSD + Lend
- defillama-server: protocol entries + parent

## Test plan

- [x] \`npm run test fees flying-tulip-lend\` completes without throwing
- [x] Verified per-reserve daily interest math against borrows/APR from \`api.flyingtulip.com/mm/lend\`
- [x] Addresses lowercased, BigInt literals used for WAD/year scaling

Sibling of https://defillama.com/protocol/flying-tulip

### DIFF
--- a/fees/flying-tulip-lend.ts
+++ b/fees/flying-tulip-lend.ts
@@ -1,0 +1,113 @@
+import { CHAIN } from '../helpers/chains'
+import { FetchOptions, SimpleAdapter } from '../adapters/types'
+
+const LENDING_LENS = '0x3682168023e6ba8d1f995fda1d920827c5a8a43e'
+
+const RESERVES: Record<string, string[]> = {
+  [CHAIN.SONIC]: [
+    '0x29219dd400f2bf60e5a23d13be72b486d4038894', // USDC
+    '0x039e2fb66102314ce7b64ce5ce3e5183bc94ad38', // wS
+    '0x5dd1a7a369e8273371d2dbf9d83356057088082c', // FT (address LendingLens is configured to key on)
+    '0xe5da20f15420ad15de0fa650600afc998bbe3955', // stS
+    '0xf7d85ec4e7710f71992752eac2111312e73e9c9c', // ftUSD
+    '0x50c42deacd8fc9773493ed674b675be577f2634b', // WETH
+    '0x0555e30da8f98308edb960aa94c0db47230d2b9c', // WBTC
+  ],
+}
+
+const ASSET_STATE_ABI =
+  'function assetState(address) view returns (uint256 cash, uint256 borrows, uint256 reserves, uint256 utilWad)'
+const ASSET_CFG_ABI =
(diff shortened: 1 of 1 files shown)


fees/flying-tulip-lend.ts
```diff
@@ -0,0 +1,113 @@
...
+  }
+}
+
+const methodology = {
+  Fees: 'Interest paid by borrowers estimated as borrows * borrowAPR * windowSeconds / year, read via LendingLens.assetState and IRM.irmSampleAPR.',
+  Revenue:
+    'All borrower interest accrues to the protocol reserves accumulator before being redistributed; tracked equal to Fees.',
+  ProtocolRevenue: 'Same as Revenue.',
+  SupplySideRevenue:
+    'Not tracked here. Suppliers earn via FT epoch rewards and via idle liquidity deployed to Aave; those are measured separately.',
```
bheluga: @patcito can you explain a little how the income flow works here

patcito: Hey, the flow is a bit different from Aave or Compound.

Borrowers pay interest and 100% of that interest accrues into a reserves accumulator on the PositionsManager. There is no reserve factor split on chain so suppliers do not receive a cut of the interest directly.

Separately, any idle cash in each reserve gets automatically deployed to Aave through a ftYieldWrapper and AaveStrategy setup. The yield from that Aave position goes into the same bucket.

Suppliers earn through a different path. FT has a fixed total supply so nothing is minted. The protocol uses the fees it collects to buy FT on the market, and those tokens are then distributed to suppliers quarterly through the EpochRewardsVault. That is what shows up for them as supply side APY on the frontend. Those FT emissions are not captured in this fees adapter because they come out of a separate rewards stream, not directly from borrower interest flowing through the adapter.

So in this adapter, dailyFees is total borrower interest, dailyProtocolRevenue is the same number because all of it flows to the protocol reserves, and dailySupplySideRevenue is 0 on purpose. Happy to wire up a second source reading EpochSettled events from the EpochRewardsVault if you want FT distributions tracked too.

```
  Borrowers            Idle cash in reserves
      |                         |
      | interest                | deployed to Aave
      v                         v
      +=====>  Protocol reserves  <=====+
                      |
                      |  buys FT on open market
                      v
                EpochRewardsVault
                      |
                      |  quarterly distribution
                      v
                  Suppliers
              (receive FT, fixed supply)
```

PR #6461, opened 2026-04-23
fix(standx): temporarily remove fee metrics to prevent user confusion

Alansx8:
Hi DefiLlama team! 👋

This PR temporarily removes the `Fees` and `Revenue` metrics from the StandX adapter. 

- The current metrics track **DUSD yield claims and withdrawal fees**.
- On our [default dashboard](https://defillama.com/protocol/standx?tvl=false&events=false&perpVolume=true), users are confusing this with our **Perps trading fees**. Because the magnitude of these two fees is completely different, it's causing significant misunderstanding among our users.
- We are currently building the data pipeline for our actual Perps trading fees. We will submit a new PR to update and re-add the exact perps fees once it's ready.

Thanks for the quick review and merge! 🦙

image


bheluga (changes requested):
@Alansx8 thanks for the PR.
We don't usually remove the fees 
Why don't you raise a PR to add perp fees? If you still need time to build robust api for fees data, you can temporarily use either of these methods to calculate fees

1. Calculate through onchain fee paid events
2. If fee is flat multiple of volume, you can simply calculate by multiplying fee percentage with volume

PR #6468, opened 2026-04-23
Fix/thorwallet fees revenue split

CopsLikeDonuts:
It's a fix for recently merged [PR](https://github.com/DefiLlama/dimension-adapters/pull/6353) 

Summary
                                                                                                                                                                                                                 
  The previous adapter recomputed the revenue split client-side as fees × 0.45 / 0.50 / 0.05 across all fee sources. That assumption is wrong for ThorWallet: only THORChain, Maya, and Near Intents fees are    
  split 45/50/5 between protocol treasury / $TITN stakers / raffle pot. Fees from 1inch, Unizen, Chainflip, Harbor, and perpetual close fees go 100% to the protocol treasury for legal/compliance reasons. The
  old math misattributed non-shareable fees to stakers and the raffle pot.                                                                                                                                       
                                                         
  The /defillama/fees endpoint now returns the correctly bucketed dailyProtocolRevenue, dailyHoldersRevenue, and dailyRaffleRevenue fields, so the adapter just passes them through.                             
   
  Changes                                                                                                                                                                                                        
                                                         
  - fees/thorwallet/index.ts: replace dailyFees.clone(0.45/0.5/0.05, …) with direct reads of data.dailyProtocolRevenue, data.dailyHoldersRevenue, data.dailyRaffleRevenue.                                       
  - Update methodology and breakdownMethodology strings to reflect which fee sources are shared vs. 100%-to-treasury.
  - Fall back to '0' for dailyRaffleRevenue until the API change ships to prod — without it, parseFloat(undefined) is NaN and the SDK rejects it. The fallback becomes a no-op once the field is populated.

### DIFF
--- a/fees/thorwallet/index.ts
+++ b/fees/thorwallet/index.ts
@@ -18,7 +18,7 @@ const fetch = async (options: FetchOptions) => {
     dailyHoldersRevenue.addUSDValue(parseFloat(data.dailyHoldersRevenue), METRIC.STAKING_REWARDS);
 
     const dailySupplySideRevenue = options.createBalances();
-    dailySupplySideRevenue.addUSDValue(parseFloat(data.dailyRaffleRevenue), 'Raffle Pot Rewards');
+    dailySupplySideRevenue.addUSDValue(parseFloat(data.dailyRaffleRevenue ?? '0'), 'Raffle Pot Rewards');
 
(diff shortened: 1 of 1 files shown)


CopsLikeDonuts:
tried to re-request coderabby review but it didn't work

PR #6477, opened 2026-04-24
refactor: standx fee adapters

Alansx8:
Hi DefiLlama team! 👋

This PR introduces the new standx-perps fee adapter and refactors the existing standx-dusd adapter to ensure data accuracy and transparency.

- Added standx-perps Adapter

Estimated trading volume fees.

The current metric should include fees (previously not present), which track **Perps trading fees** 
image

The current metrics track **DUSD withdrawal fees** 
image

The current metrics track **Perps trading fees** + **DUSD withdrawal fee** (total fees)
image

- Refactored standx-dusd Adapter:
1. Optimized dailyFees logic: Improved the accuracy of fee data collection for DUSD.
2. Temporarily disabled Revenue fields (dailyRevenue, dailyProtocolRevenue, dailySupplySideRevenue )
The current calculation logic for these fields deviates from the project's actual operations. To avoid user confusion, we have decided to temporarily remove these metrics for methodology recalibration.

### DIFF
--- a/dexs/standx/index.ts
+++ b/dexs/standx/index.ts
@@ -21,10 +21,13 @@ interface MarketInfo {
   v: number[];
 }
 
-const fetch = async (_a: any, _b: any, options: FetchOptions): Promise<FetchResultVolume> => {
-  const dailyVolume = options.createBalances();
+// NOTE: Take MM0 fee rate as reference to estimate trading fees
+export const PERPS_FEE_RATE = 0.0003;
+
+export async function getDailyVolume(options: FetchOptions): Promise<number> {
   const symbolsResponse: SymbolInfo[] = await fetchURL(`${apiEndpoint}/query_symbol_info`);
   const symbols = symbolsResponse.map((item) => item.symbol);
+  let totalVolUsd = 0;
 
   await PromisePool.withConcurrency(1).for(symbols).process(async (symbol) => {
     const marketInfo: MarketInfo = await fetchURLAutoHandleRateLimit(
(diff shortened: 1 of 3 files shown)


bheluga (changes requested):
@Alansx8 thanks for the PR.
Any reason why you have removed yield claimed events?
Also for fees instead of duplicating same code as of volume, you can simply calculate fees in /dexs/standx/index.ts file and then return fees along with volume

Alansx8:
> @Alansx8 thanks for the PR. Any reason why you have removed yield claimed events? Also for fees instead of duplicating same code as of volume, you can simply calculate fees in /dexs/standx/index.ts file and then return fees along with volume

@bheluga 
The yield claimed data used for calculating dailyFees and revenue is currently inaccurate. To avoid misleading users, we have temporarily removed it and will implement a more precise adapter in a future iteration.

The common function getDailyVolume，getDailyFees has been extracted to /dexs/standx/index.ts.

fees/standx-perps/index.ts
```diff
@@ -0,0 +1,63 @@
...
+import { CHAIN } from "../../helpers/chains";
+import { PromisePool } from "@supercharge/promise-pool";
+import { sleep } from "../../utils/utils";
+import fetchURL from "../../utils/fetchURL";
+import { METRIC } from "../../helpers/metrics";
+
+const apiEndpoint = "https://perps.standx.com/api";
+
+// NOTE: Take MM0 fee rate as reference to estimate trading fees
+const PERPS_FEE_RATE = 0.0003;
```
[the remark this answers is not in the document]
Alansx8: This is an estimated value.
https://github.com/DefiLlama/dimension-adapters/pull/6461#pullrequestreview-4164465534

dexs/standx/index.ts
```diff
@@ -21,10 +21,13 @@ interface MarketInfo {
   v: number[];
 }
 
-const fetch = async (_a: any, _b: any, options: FetchOptions): Promise<FetchResultVolume> => {
-  const dailyVolume = options.createBalances();
+// NOTE: Take MM0 fee rate as reference to estimate trading fees
+export const PERPS_FEE_RATE = 0.0003;
```
bheluga, line 25: your website mentions different number https://docs.standx.com/docs/stand-x-perps-solutions/trading-fee

Alansx8: > your website mentions different number https://docs.standx.com/docs/stand-x-perps-solutions/trading-fee

Hi @bheluga, thanks for reviewing!
- You are right about the base trading fee
- However, as you suggested earlier, we are trying to provide an estimate as a quick fix, since the actual fee is quite challenging because it varies significantly by user tier, so we use MM0 rate as a reference for now, as a quick fix 
  - refer: https://docs.standx.com/docs/stand-x-campaigns/market-maker-uptime-program#mm-fee-tiers
  - This is exactly why we initially proposed removing the metric — we want to maintain objective accuracy and avoid misleading users
- Since an estimate is the suggestion for now, we believe the MM0 rate is the most responsible proxy
- For StandX DUSD, only withdraw fee is available for now, so removed revenue metric

bheluga: cant we calculate using onchain logs?
also how long do you think it will take to build a proper fee calculation and how off might be current estimation from actual number

Alansx8: - The logs on BSC & Solana are only about DUSD withdrawals, not the perps fees.
- Timeline & Variance: We are undergoing major architectural upgrades (called "network yielding" in my best knowledge), so building a robust fee API will take some time. I will update you once it goes live. However, the current live numbers are definitely very far away from the actual figures.

This fix at least for now is a much-needed stopgap to prevent misleading users with the DUSD withdrawal numbers and perps fee. 

We deeply appreciate your flexibility and understanding, since we are trying to explore something new in this space.

bheluga:
@Alansx8 we have temporarily disabled standx DUSD fees , https://github.com/DefiLlama/defillama-server/pull/11824
I am closing the PR

Issue #6067, opened 2026-03-05
Track bluemove dex volume & fees on sui


g1nt0ki:
https://defillama.com/protocol/bluemove-dex

PR #6486, opened 2026-04-25
bluemove: track volume, fees

uncoolzero:
tracks volume and fees for Bluemove on SUI
aptos api endpoint is dead, added deadFrom date
update queryEvents SUI helper to accept custom filters

related issue: #6067


bheluga:
@uncoolzero thanks for the PR. Has bluemove stopped operating on aptos , if yes you can uncomment and add approx deadFrom dates
Can you use same logic and calculate volume here https://github.com/DefiLlama/dimension-adapters/blob/master/dexs/bluemove.ts

PR #6556, opened 2026-04-28
[fix] aries-markets - encode rate history input

### LINKED ISSUE #6526: fix aries markets fees
https://defillama.com/protocol/aries-markets

Genmin:
Fixes DefiLlama/dimension-adapters#6526

## Summary
- URL-encode the full Aries `reserve.rateHistory` JSON input instead of manually encoding only quotes
- switch the adapter fetch to the correct v2 `FetchOptions` signature
- use the start day of the actual 24h run window for rate history so no-date CI/current runs do not query an incomplete current UTC day
- guard missing rate/reserve fields so a partial API response cannot produce `NaN` balances

## Context
`pnpm test fees aries-markets 2026-04-26` failed with `Request failed with status code 400`. The Aries rate-history endpoint still returns data when `input` is a fully URL-encoded JSON object, while the previous partially encoded query now fails parsing.

GitHub Actions also runs adapter tests without an explicit date. In that mode, v2 adapters use a rolling 24h window, so the rate-history lookup must align to the start of that window rather than the current UTC day.

## Validation
- `pnpm test fees aries-markets`
- `pnpm test fees aries-markets 2026-04-26`
- `pnpm test fees aries-markets 2026-03-02`
- `pnpm run ts-check`

### DIFF
--- a/fees/aries-markets/index.ts
+++ b/fees/aries-markets/index.ts
@@ -20,7 +20,7 @@ const getRateHistoryURL = (timestamp: number, reserveKey: string) => {
   return `${rateURL}${input}`;
 };
 
-const fetch = async (options: FetchOptions) => {
+const fetch = async (_a: any, _b: any, options: FetchOptions) => {
   const reserves = (await fetchURL(reserveURL)).result.data.stats;
 
   let df = 0;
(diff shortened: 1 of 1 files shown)


bheluga:
@Genmin thanks for the PR

PR #6577, opened 2026-04-28
fix: repair DuckChain chain fees adapter

### LINKED ISSUE #6500: fix duckchain fees
https://defillama.com/chain/duckchain

lukasa1993:
```markdown
## Summary

Fixes #6500.

This PR fixes DuckChain chain fee tracking in the generic Blockscout fee adapter.

## Root cause


DuckChain is already wired through the Blockscout chain-fees factory, but the adapter was failing because DuckChain's Blockscout stats endpoints currently return Cloudflare `522` plain text instead of fee JSON.

The broken endpoint was:

```text
https://scan.duckchain.io/api?module=stats&action=totalfees&date=2026-04-25
```

DuckChain's public RPC is available and supports `eth_getBlockReceipts`, so DuckChain now uses receipt-based gas fee summing while other Blockscout chains keep the existing stats API path.

## Changes

- Add an opt-in RPC receipt fee path for DuckChain
- Sum `gasUsed * effectiveGasPrice` from `eth_getBlockReceipts`
- Keep DuckChain mapped through the existing `factory/blockscout.ts` registry
- Preserve existing Blockscout fee behavior for other chains
- No new dependencies

### DIFF
--- a/factory/blockscout.ts
+++ b/factory/blockscout.ts
@@ -63,7 +63,6 @@ const protocolChainMap: Record<string, string> = {
   "adventure-layer": CHAIN.ADVENTURE_LAYER,
   "deri-chain": CHAIN.DERI_CHAIN,
   "earnm": CHAIN.EARNM,
-  "duck-chain": CHAIN.DUCKCHAIN,
   "edu-chain": CHAIN.EDU_CHAIN,
   "ethereal": CHAIN.ETHEREAL,
   "eventum": CHAIN.EVENTUM,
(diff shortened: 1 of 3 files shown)


bheluga (changes requested):
@lukasa1993 thanks for the PR.
Since duckchain is the only one which uses RPC, can you remove it from the  helper and create an isolated file with the same name and implement the logic?

lukasa1993:
> @lukasa1993 thanks for the PR.
> 
> Since duckchain is the only one which uses RPC, can you remove it from the  helper and create an isolated file with the same name and implement the logic?

okay will try

lukasa1993:
@bheluga hope this is better 🙏

g1nt0ki:
@lukasa1993 thanks for the PR

PR #6584, opened 2026-04-28
[fix] wavex - end inactive fees adapter

ducnmm:
## Summary
- Mark Wavex fees adapter as inactive from 2026-01-03, when the API starts returning null data.
- Keeps historical fetches before the cutoff working while avoiding current adapter failures.

## Tests
- pnpm test fees wavex
- pnpm test fees wavex 2026-01-02
- pnpm test fees wavex 2025-01-15
- pnpm run ts-check


ducnmm:
Closing because the same Wavex inactive adapter fix was already merged in #6578.

PR #6586, opened 2026-04-28
[fix] xchain - end inactive fees adapter

ducnmm:
## Summary
- Mark the XChain fees adapter inactive from 2026-01-24, after the chain fee chart drops to zero and the Blockscout endpoint stops serving valid API responses.
- Prevents current runs from failing on the broken explorer TLS/API host.

## Tests
- pnpm test fees xchain
- pnpm run ts-check

## Notes
- Historical requests against the old explorer still fail because `xchain-explorer.kuma.bid` now resolves to a Conduit host with a mismatched certificate and the direct Conduit host returns 502/404 for the API.


ducnmm:
Closing this branch because master moved and caused a conflict; recreated the same fix on an updated branch.

PR #6590, opened 2026-04-28
[fix] get-protocol - end inactive fees adapter

ducnmm:
## Summary
- Mark GET Protocol's fees adapter inactive from 2025-08-01, after the live chart stopped receiving non-zero data and current runs began failing on the paid Graph gateway.
- Fix staking fee aggregation so multiple reward types are accumulated instead of overwriting the previous accumulator value.
- Correct staking reward decimal conversion from 1e19 to 1e18.

## Tests
- pnpm test fees get-protocol
- pnpm test fees get-protocol 2025-08-03
- pnpm run ts-check


bheluga:
@ducnmm thanks for the PR

PR #6604, opened 2026-04-28
[fix] get-protocol - end unavailable fees source

Genmin:
## Summary
- mark the GET Protocol Polygon fees source dead from 2025-08-01
- DefiLlama daily fees stop on 2025-07-31
- the current Graph endpoint returns an auth/payment-required error, so current adapter runs should skip instead of failing

## Validation
- npm test -- fees/get-protocol
- npm run ts-check


bheluga:
@Genmin thanks for the PR
Closing since its duplicate of https://github.com/DefiLlama/dimension-adapters/pull/6590

PR #6609, opened 2026-04-28
[fix] illuvium - end unavailable fees source

Genmin:
## Summary
- mark the Illuvium ImmutableX fees source dead from 2026-03-10
- DefiLlama daily fees are zero through 2026-03-09
- the current Immutable transfers endpoint returns HTTP 404 with an empty body, causing current adapter runs to throw while parsing JSON

## Validation
- npm test -- fees/illuvium
- npm run ts-check

### DIFF
--- a/fees/illuvium.ts
+++ b/fees/illuvium.ts
@@ -28,6 +28,7 @@ const adapter: Adapter = {
   adapter: {
     [CHAIN.IMMUTABLEX]: {
       fetch: fetchFees,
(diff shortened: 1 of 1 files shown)


treeoflife2:
@Genmin only mark adapter dead if project is dead. also cite the source for the claim

bheluga:
@Genmin thanks for the PR.
Closing since the protocol is still active

PR #6610, opened 2026-04-28
[fix] nile-cl - stop current fees while bribes source is unavailable

Genmin:
Stops current Nile CL fee reporting while the bribes subgraph deployment is unavailable, rather than reporting partial swap-fee data without bribes.\n\nValidation:\n- npm run test fees nile-exchange\n- npm run ts-check -- --pretty false

### DIFF
--- a/fees/nile-exchange/index.ts
+++ b/fees/nile-exchange/index.ts
@@ -61,16 +61,14 @@ const adapter: Adapter = {
         const adapter = getUniV3LogAdapter({ factory: "0xAAA32926fcE6bE95ea2c51cB4Fcb60836D320C42", revenueRatio: 1, userFeesRatio: 1, protocolRevenueRatio: 0.08, holdersRevenueRatio: 0.92 })
         const response = await adapter(options)
 
-        try {
-          const bribesResult = await getBribes(options);
-          response.dailyBribesRevenue = bribesResult.dailyBribesRevenue;
-        } catch {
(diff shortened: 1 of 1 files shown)


fees/nile-exchange/index.ts
```diff
@@ -61,8 +61,12 @@ const adapter: Adapter = {
         const adapter = getUniV3LogAdapter({ factory: "0xAAA32926fcE6bE95ea2c51cB4Fcb60836D320C42", revenueRatio: 1, userFeesRatio: 1, protocolRevenueRatio: 0.08, holdersRevenueRatio: 0.92 })
         const response = await adapter(options)
 
-        const bribesResult = await getBribes(options);
-        response.dailyBribesRevenue = bribesResult.dailyBribesRevenue;
+        try {
+          const bribesResult = await getBribes(options);
+          response.dailyBribesRevenue = bribesResult.dailyBribesRevenue;
```
treeoflife2: we prefer not showing the data then incorrect data.

Genmin:
Updated this after the maintainer feedback. The PR no longer reports partial swap-fee-only data while bribes are unavailable; it now stops current Nile CL fee reporting at the missing bribes source window.\n\nValidated with:\n- `npm run test fees nile-exchange`\n- `npm run ts-check -- --pretty false`

treeoflife2:
@Genmin closing the pr as nile-cl exchange is not dead, so marking it dead doesn't make sense

PR #6580, opened 2026-04-28
fix(fees/buffer): set deadFrom for DNS-dead Satsuma subgraph endpoint

AimenHallou:
Closes #6503.

The Buffer Finance fees adapter calls a Satsuma-hosted subgraph at `subgraph.satsuma-prod.com/.../bufferfinance/v2.5-arbitrum-mainnet/api` whose hostname no longer resolves DNS. `pnpm test fees buffer` fails with `ENOTFOUND` for every fetch; nothing has been ingested since mid-2025.

Setting `deadFrom: '2025-07-17'` retires the adapter from current fetches while preserving the historical chart on `defillama.com/protocol/buffer-finance`. Same playbook as the recently-merged Ribbon (#6551), Premia v2 (#6559), Collex (#6565), and wavex (#6578) fixes.

The chosen date is the day after the last non-zero fee point on the protocol's historical chart (last value `2025-07-16` per `https://api.llama.fi/summary/fees/buffer-finance`).

PR #6565, opened 2026-04-28
[fix] collex - end inactive fees adapter

HarshS1611:
Fixes #6521 

**Summary**
- Added `deadFrom: '2026-03-13'` since the Collex API (`api.collex.fun`) returns 502 and the last charted fee day on DefiLlama was March 12, 2026

**Validation**
```bash
pnpm test fees collex
# Skipping collex because the adapter ended at Fri, 13 Mar 2026 00:00:00 GMT
```

PR #6611, opened 2026-04-28
fix(fees/yakafinance): set deadFrom for dead SEI blocks subgraph

### LINKED ISSUE #6512: Fix yaka v3 fees
https://defillama.com/protocol/yaka-v3

AimenHallou:
Closes #6512.

The Yaka Finance fees adapter resolves block numbers via the SEI blocks subgraph at `api.studio.thegraph.com/query/82132/sei-blocks`, which has been deleted (queries return *"deployment u82132/s80587/latest does not exist"*). Even if that were swapped out, public SEI EVM RPCs only serve blocks from `79123881` onward, so the adapter's `2024-07-01` start date can no longer be backfilled from chain data either. Both the volume subgraph and the block lookup paths fail in `pnpm test fees yakafinance`.

Setting `deadFrom: '2026-03-05'` (day after the last non-zero point on `defillama.com/protocol/yaka-finance`) retires the adapter from current fetches while preserving the historical chart. Same playbook as the recently-merged Buffer (#6580), Ribbon (#6551), Premia v2 (#6559), Collex (#6565), and wavex (#6578) fixes.

Note: this adapter is `version: 2` with per-chain config, so `deadFrom` sits inside the SEI chain block rather than at the adapter root.

### DIFF
--- a/fees/yakafinance/index.ts
+++ b/fees/yakafinance/index.ts
@@ -9,6 +9,7 @@ const adapter: Adapter = {
         [CHAIN.SEI]: {
             fetch: getFees as any,
             start: '2024-07-01',
+            deadFrom: '2026-03-05',
         },
     },
 };


AimenHallou:
@coderabbitai review

AimenHallou:
@coderabbitai full review

treeoflife2:
@AimenHallou we only mark adapter dead if project is dead. not if source is not working/changed.

bheluga:
@AimenHallou thanks for the PR.
Closing since the protocol still seems to be active https://yaka.finance/analytics
In this case if a data source is broken , you need to find another data source

PR #6616, opened 2026-04-29
fix(solidly-v3): correct invalid Fantom start date

ducnmm:
## Summary

The Solidly V3 Fantom chain start date is set to `'2023-25-12'` (month 25), which is invalid and causes the adapter to skip Fantom entirely:

```
Skipping fantom because the configured start time is Invalid Date
```

### Fix

Changed start date from `'2023-25-12'` to `'2023-12-25'` (December 25, 2023).

This is a date format typo - day and month fields were swapped.

### DIFF
--- a/fees/solidly-v3.ts
+++ b/fees/solidly-v3.ts
@@ -92,7 +92,7 @@ const adapter: Adapter = {
     },
     [CHAIN.FANTOM]: {
       fetch: fetch(CHAIN.FANTOM),
-      start: '2023-25-12',
+      start: '2023-12-25',
     }
   }, 
 }


ducnmm:
Reproduced the failing check locally with `pnpm test fees solidly-v3.ts`.

The failure appears unrelated to this start-date fix:

```text
auth error: payment required for subsequent requests for this API key
```

I also checked the current `solidly-v3` subgraph endpoints and they are all returning the same The Graph gateway paywall/auth error, so this looks like an upstream endpoint issue rather than a regression from changing Fantom's `start` from `2023-25-12` to `2023-12-25`.

If there is a preferred replacement endpoint for this adapter, I can follow up with a separate fix.

PR #6648, opened 2026-04-30
Track Ubeswap V3 volume and fees

uwezukwechibuzor:
https://defillama.com/protocol/ubeswap-v3


bheluga (dismissed):
Can you please add it to factory instead? https://github.com/DefiLlama/dimension-adapters/blob/master/factory/uniV3.ts

Issue #6334, opened 2026-04-04
[Pendle] Protocol revenue incorrectly reported before September 2025 + double counting in revenue figures


Labasca:
Two issues with the Pendle fees adapter:

**1. Protocol revenue reported before September 2025**

The adapter is currently returning protocol revenue for dates prior to September 2025. According to the Pendle team, there was no protocol-level revenue before that date, these values should be zero. The current data shows protocol revenue going back to December 2023, totalling ~$8.3M in incorrectly attributed protocol revenue across 21 months.

**Expected behaviour:** `protocolRevenue` should return `0` / `null` for all dates before September 2025.

**2. Double counting in revenue figures**

There also appears to be double counting in the overall revenue numbers. Revenue components (holders revenue, protocol revenue, supply side revenue) don't consistently sum to the reported total, in Pendles team Dune suggesting certain fee flows are being counted more than once.

**Reference sources from the Pendle team:**
- Dune dashboard with full revenue calculation methodology: https://dune.com/p_team/pendle-tvl-and-revenue
- V2 revenue calculation instructions for third-party data platforms: https://pendle.notion.site/3rd-party-Data-Platform-Pendle-V2-Revenue-Calculation-Instruction-224567a21d3780afac7ef668866b4d71

**Adapter file:** https://github.com/DefiLlama/dimension-adapters/blob/master/fees/pendle.ts

pDubey7:
After reviewing the current adapter at fees/pendle.ts, both issues appear 
to already be addressed in the codebase:

1. protocolRevenue is already gated by a September 2025 start date 
   (lines 150-153) returns 0 for all dates before that.
2. The fee flow follows the documented methodology with separate tracking 
   for dailyFees, dailyRevenue, dailySupplySideRevenue, dailyHoldersRevenue, 
   and dailyProtocolRevenue.

This issue may be ready to close if the fix was already merged. @bheluga

bheluga:
https://github.com/DefiLlama/dimension-adapters/commit/e10c9b6e8a3b682100145579ff18d5958a127138

PR #6649, opened 2026-04-30
[fix] pendle - remove duplicate asset whitelist fetch

Genmin:
## Summary

Removes an extra `getWhitelistedAssets()` call from the Pendle fees adapter. The result was discarded and the same helper is called again immediately afterward to get `markets`, `sys`, and `marketToSy`, so each adapter run was doing duplicate config/API work before processing logs.

This does not change fee classification or reported values; it just avoids the redundant lookup.

## Validation

- `npm test -- fees pendle 2025-10-01` attempted, but the current adapter run fails after startup in the SDK ABI/log fallback path with `Cannot read properties of undefined (reading 'success')` when Llama Indexer env is unavailable. The change is before that path and only removes a discarded duplicate call.

Related: #6334 already appears addressed on master for the September 2025 protocol revenue gate; this is a small cleanup found while auditing that issue.

### DIFF
--- a/fees/pendle.ts
+++ b/fees/pendle.ts
@@ -85,7 +85,6 @@ const chainConfig: IConfig = {
 
 const fetch = async (options: FetchOptions) => {
   const { chain } = options;
-  await getWhitelistedAssets(options.api);
   const { api, getLogs } = options;
 
   const dailyFees = options.createBalances()


Genmin:
The failing adapter check appears to have been a transient RPC/indexer fallback issue. I reran the exact check locally with the current pnpm lockfile and it now passes:\n\n- `pnpm test fees pendle.ts`\n\nI refreshed the branch SHA without changing the diff so GitHub Actions can rerun the same Pendle adapter test.

Genmin:
Retried the adapter check with an empty commit after the previous run failed on BSC public RPC throttling (`getLogs` 429 / limit exceeded). No code diff changed; the branch still only removes the duplicate whitelist fetch.

PR #6645, opened 2026-04-29
Track Midas RWA fees and revenue

uwezukwechibuzor:
closes  #6640

PR #6650, opened 2026-04-30
[fees] add midas-rwa adapter

Genmin:
Adds a fees adapter for Midas mTokens, starting with the two products where the fee/revenue treatment is clear from public docs and observable NAV oracles: mTBILL and mBASIS.

What changed:
- tracks mTBILL and mBASIS across Ethereum, Base, Plume, and Etherlink from the official Midas smart contract registry
- computes daily fees from positive daily NAV-per-token appreciation multiplied by current token supply
- reports mToken holder yield as supply-side revenue
- reconstructs mBASIS gross yield from net holder yield using the published 20% performance-fee model, and reports the retained share as revenue/protocol revenue
- uses current-time execution because several public RPC/indexer paths used by the local adapter runner do not expose reliable archival state across all covered chains

Validation:
- `npm test -- fees midas-rwa`

### DIFF
--- a/fees/midas-rwa/index.ts
+++ b/fees/midas-rwa/index.ts
@@ -1,0 +1,203 @@
+import BigNumber from "bignumber.js";
+import { FetchOptions, SimpleAdapter } from "../../adapters/types";
+import { CHAIN } from "../../helpers/chains";
+import { METRIC } from "../../helpers/metrics";
+
+type ProductConfig = {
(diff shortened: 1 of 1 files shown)


bheluga:
@Genmin thanks for the PR
Closing as duplicate of https://github.com/DefiLlama/dimension-adapters/pull/6645

Issue #6293, opened 2026-03-30
Track meteora DLMM/DAMM v2/DAMM v2 Fees/Revenue onchain


treeoflife2:

pDubey7:
Hi @treeoflife2 ! I see you self assigned this back in March. 
Are you still working on it, or would it be helpful if I took a 
shot at the on-chain approach for DLMM/DAMM v2 fees?

treeoflife2:
> Hi [@treeoflife2](https://github.com/treeoflife2) ! I see you self assigned this back in March. Are you still working on it, or would it be helpful if I took a shot at the on-chain approach for DLMM/DAMM v2 fees?

yes give it a shot, what i wanted to do was create the fee adapter for each meteora products which has complete historical fee/revenue data. and fully onchain(you could use dune as allium is enterprise only), one of the edgecase i wanted to cover was trying to keep the realistic revenue, for example meteora earns lots of memecoins, and memecoins lose value in seconds, so meteora can't realize those gains in time, so some filter to remove those memecoins is needed. and i think you can start with DLMM first, 

pr created by @Genmin is using their API, so let's not do that.

pDubey7:
Thanks @treeoflife2! I'll start with DLMM on chain fees using Dune. 
A few quick questions to make sure I get the approach right:

1. For the memecoin filter , would you prefer filtering by minimum 
   liquidity threshold, or by token age, or by whether the token 
   has a price in the coins API?
2. For revenue split , is Meteora DLMM's protocol fee percentage 
   available on chain per pool, or do we need to hardcode it?
3. Should this go in fees/meteora-dlmm.ts as a new file?

treeoflife2:
> Thanks [@treeoflife2](https://github.com/treeoflife2)! I'll start with DLMM on chain fees using Dune. A few quick questions to make sure I get the approach right:
> 
> 1. For the memecoin filter , would you prefer filtering by minimum
>    liquidity threshold, or by token age, or by whether the token
>    has a price in the coins API?

coins api pricing is better

> 2. For revenue split , is Meteora DLMM's protocol fee percentage
>    available on chain per pool, or do we need to hardcode it?

that need to be researched, and adjuste them if it was changed in past using the timestamp conditions

> 3. Should this go in fees/meteora-dlmm.ts as a new file?

yes update the existing one.

pDubey7:
Hii @treeoflife2 ! I Researched the DLMM on-chain approach. Found that 
[...]

PR #6651, opened 2026-04-30
[fees] add meteora fees adapter

Genmin:
Adds an initial Meteora fees adapter for the fee dashboard using the same public Meteora data sources already used by the existing DEX adapters.

Scope:
- DLMM fees, protocol revenue, and LP supply-side revenue from `dlmm.datapi.meteora.ag`
- DAMM v2 fees split by each pool's exposed protocol-fee percentage from `damm-v2.datapi.meteora.ag`
- DAMM v1 fees from `damm-api.meteora.ag`, reported as unclassified swap fees because that API exposes fee volume but not the LP/protocol split
- the same basic blacklisted/wash-like filters used by the DEX adapters

This is deliberately not marked as fixing #6293 because that issue asks for a full onchain implementation. This PR gives the dashboard a maintainable baseline while keeping the methodology explicit about what is API-derived and what cannot be split yet.

Validation:
- `npm test -- fees meteora`

### DIFF
--- a/fees/meteora.ts
+++ b/fees/meteora.ts
@@ -1,0 +1,211 @@
+import { FetchOptions, SimpleAdapter } from "../adapters/types";
+import { CHAIN } from "../helpers/chains";
+import { METRIC } from "../helpers/metrics";
+import { httpGet } from "../utils/fetchURL";
+import { sleep } from "../utils/utils";
+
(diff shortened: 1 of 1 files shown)


treeoflife2:
@Genmin we already have this https://github.com/DefiLlama/dimension-adapters/blob/master/dexs/meteora-damm-v2/index.ts

PR #6660, opened 2026-04-30
feat(fees): track FrenFlow Polymarket builder-fee distributions

frenflow:
## Summary

Extends the existing FrenFlow service-fee adapter (PR #6481) with a second on-chain revenue stream: **builder-fee distributions** Polymarket pays to FrenFlow's builder profile wallet for trades carrying our `builderCode`.

## What changed

- Adds `Builder Fees` breakdown alongside the existing `Service Fees`.
- Watches incoming `Transfer(_, to=BUILDER_PAYOUT, _)` events on three tokens (pUSD, USDC.e, USDC native) into FrenFlow's builder profile wallet `0x58715321c2c6a216d1259f368c34f987a4a26b64`.
- Updates `methodology` and `breakdownMethodology` so DefiLlama displays the two streams separately.

## Why now

The first builder payout to any Polymarket builder happened on **2026-04-30** (tx [`0x4e0e7e42…`](https://polygonscan.com/tx/0x4e0e7e4296167ea28fb92f8058409514459c8a4d5c994511b22e6e913010d5c3), block `86195685`). It was a single batched distribution (~\$3,087 pUSD across 27 builders) routed through a disperse-style contract; FrenFlow received \$0.32 in pUSD. Without this change that payout — and every future one — would be invisible in DefiLlama.

The pUSD denomination matched the existing `factory/polymarket.ts` builder-volume entry and our internal CLOB builder-trades aggregate (`clob.polymarket.com/builder/trades?builder_code=0x1a907452…`) to the cent.

## Why three tokens

Polymarket has only paid in pUSD so far, but their builder fee accrues in an internal treasury (`0x115f48dc…`) that holds value in pUSD. They could switch to USDC.e or USDC native without notice. The three-token watch costs nothing extra (empty `getLogs` results when those tokens aren't used) and prevents silent revenue gaps if the settlement currency changes.

## Why a recipient-only filter

The middle topic is `null` because builder distributions can come from any sender (today: an EOA via a disperse contract; future: possibly a different contract). Filtering only by the indexed `to` topic keeps the adapter robust to any sender rotation.

## Verification

- On-chain check on the pUSD payout block range returns the expected single log:
  `eth_getLogs target=pUSD topics=[Transfer, null, padAddress(BUILDER_PAYOUT)] → 1 log, value 317400 (= \$0.3174 pUSD)`.
- `Balances.getUSDJSONs()` resolves it to **\$0.324** at the historical price (pUSD price \$1.00 confidence 0.99 in DefiLlama price oracle).
- `labelBreakdown` returns `{"Builder Fees": 0.324}` cleanly alongside existing `Service Fees`.

## Test plan

- [ ] DefiLlama CI runs `npm test fees frenflow` against a recent backfill window covering 2026-04-30 (the first payout day) and confirms a non-zero `Builder Fees` row in the breakdown.
- [ ] Inspect the breakdown on the live page once the adapter rolls out: https://defillama.com/protocol/frenflow

### DIFF
--- a/fees/frenflow.ts
+++ b/fees/frenflow.ts
@@ -1,25 +1,34 @@
 import { Adapter, FetchOptions, FetchV2 } from "../adapters/types";
 import { CHAIN } from "../helpers/chains";
+import { ethers } from "ethers";
 
 /**
  * FrenFlow — social copytrading + trading UI for prediction markets.
  * https://frenflow.com  ·  https://x.com/frenflow_
  *
- * Revenue model: a 1% service fee on manual trades routed through
- * FrenFlow's Safe wallet infrastructure on Polymarket. At trade
- * settlement, the FeeCollector contract pulls the fee atomically
- * (user → treasury) and emits a `FeeCollected` event with the exact
- * feeAmount. We sum those events for the period.
+ * Two on-chain revenue streams on Polygon, both summed into `dailyFees`:
  *
- * Volume for FrenFlow as a Polymarket builder is tracked separately
- * through `factory/polymarket.ts` (Polymarket's official builder
- * volume API). This file covers only the service-fee revenue.
+ *   1. Service Fees — a 1% fee pulled atomically (user → treasury) by
+ *      FrenFlow's FeeCollector contract at Polymarket trade settlement.
+ *      Tracked via the `FeeCollected` event. Denominated in USDC.e.
+ *
+ *   2. Builder Fees — Polymarket pays builders a per-fill commission for
+ *      trades carrying their `builderCode`. PM accrues these in an
+ *      internal treasury and periodically distributes them on-chain to
+ *      each builder's profile wallet (typically batched through a
+ *      "disperse"-style contract). FrenFlow's builder profile wallet is
+ *      `0x58715321c2c6a216d1259f368c34f987a4a26b64`. We sum incoming
+ *      pUSD / USDC.e / USDC (native) Transfer events to that wallet.
+ *
+ * Volume (notional) for FrenFlow as a Polymarket builder is tracked
+ * separately in `factory/polymarket.ts`.
  *
  * Income-statement mapping (per GUIDELINES):
- *   dailyFees            — gross protocol revenue (all sources)
- *   dailyUserFees        — portion directly paid by end-users
- *                          (100% here: the fee is pulled from the
- *                          user's wallet at trade settlement)
+ *   dailyFees            — gross protocol revenue (Service + Builder)
+ *   dailyUserFees        — portion paid directly by end-users (100%:
+ *                          service fee is pulled from the trader's
+ *                          wallet at fill; builder fee is funded by
+ *                          Polymarket out of the trader's order amount)
  *   dailyRevenue         — gross profit (no supply-side to reimburse)
  *   dailyProtocolRevenue — portion allocated to treasury (100%)
  */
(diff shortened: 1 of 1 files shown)


frenflow:
Thanks for the review @coderabbitai. Pushed `75717a19f` addressing all three points:

**1. Builder fees out of `dailyUserFees`** — split into separate `dailyFees` / `dailyUserFees` / `dailyRevenue` Balances. Service Fees go into all three (atomic, user-paid at fill). Builder Fees go into `dailyFees` and `dailyRevenue` only — not `dailyUserFees`, since Polymarket pays them on its own settlement cadence and the distribution day rarely matches the original trade day. Methodology + breakdown text updated to reflect this.

**2. Recipient-only filter could overcount** — added `KNOWN_PM_BUILDER_PAYOUT_SENDERS` allowlist (currently `0xd7a0535c…`, the EOA Polymarket used for the 2026-04-30 first builder distribution). Topic-level filter still keeps the indexed `to` constraint for cheap RPC pre-filtering; senders are then checked in JS against the allowlist. Comment in code documents that the list needs updating if PM rotates its hot wallet.

**3. `Promise.all` swallowing the service-fee read** — service-fee `getLogs` now awaits independently and the three optional builder-token probes run inside `Promise.allSettled`. Failed probes log the error and yield zero for that token without dropping the required service-fee stream.

Note on the failing `github-actions` test run earlier — it sweeps slices over the past 24h hourly and reports `0.00` because:
- The first builder payout block is `86195685` (`2026-04-30T01:10:14Z`), only ~6h before the CI ran. DefiLlama's Llama Indexer may not have picked up the block yet at test time.
- I verified locally with a wider window that the adapter correctly returns 1 `FeeCollected` log + 1 pUSD `Transfer` from `0xd7a0535c…` and resolves to `~$0.32` USD via `Balances.getUSDJSONs()`. Re-running the CI now should pick up both streams.

frenflow:
Pushed `994b94da7` — addresses the follow-up.

Unknown senders are now collected into a `Set<string>` during the slice and emitted as a single `console.error` after the inflow loop, e.g.:

```
frenflow: skipped builder inflows from unknown senders (consider adding to KNOWN_PM_BUILDER_PAYOUT_SENDERS): 0x…
```

So if Polymarket rotates the EOA, the undercount becomes a one-line, idempotent log entry per slice rather than vanishing silently. We can then update `KNOWN_PM_BUILDER_PAYOUT_SENDERS` and the next backfill picks up the missing fees.

(Used `Array.from(set)` instead of spread to stay TS5/ES5 lib-target-friendly.)

frenflow:
Heads-up on the `github-actions[bot]` `0.00` output: that appears to be a **display quirk in `cli/testAdapter.ts`**, not a real failure.

I reproduced the exact same hourly sweep locally on this branch (`pullHourly: true`, no-arg run, 24 slices) with `console.error` diagnostics injected at the end of the adapter's fetch:

```
>>> probes returned: 0xd7a053/0xc011a7=27 0xd7a053/0x2791Bc=0 0xd7a053/0x3c499c=0 fee_logs=0
>>> MATCH token=0xc011a7E1 value=317400
```

The slice covering 2026-04-30 01:00 → 02:00 UTC correctly returns the 27 sibling Transfer events from the disperse-style payout, the JS `to`-check matches the single transfer to FrenFlow's builder wallet, and `Balances.add(...)` is called with `value=317400` (= \$0.317 pUSD ≈ \$0.32 USD). I also confirmed `Balances.getUSDJSONs()` resolves to:

```
dailyFees=$0.3220 dailyRevenue=$0.3220
breakdown={"Service Fees":0.0048,"Builder Fees":0.3173}
```

But the same run still prints `Daily fees: 0.00` in the test runner output — the display path doesn't seem to surface `labelBreakdown` aggregates correctly when balances span multiple tokens (pUSD + USDC.e here). The data IS in the returned `Balances` instances; the line printed by `printVolumes` just doesn't reflect it.

So the CI sweep `0.00` is misleading. The actual production ingestion (which I assume uses `getUSDJSONs` directly, not the CLI display) should pick up the \$0.32 cleanly once it runs against this branch.

Happy to drop in a `console.log(await dailyFees.getUSDJSONs())` somewhere if it would help confirm during merge — let me know.

bheluga:
@frenflow thanks for the PR

PR #6661, opened 2026-04-30
Add bedrock fees and revenue adapter

Shaileshkhote:
Fixes fee tracking for https://defillama.com/protocol/bedrock

## Summary

Adds a Bedrock fees adapter covering:

- uniETH staking rewards from `RevenueAccounted`
- uniETH protocol revenue using `managerFeeShare`
- uniETH supply-side revenue to holders
- uniBTC delayed redemption fees from `DelayedRedeemCreated.redeemFee`

## Methodology

- `dailyFees`: gross uniETH rewards plus uniBTC redemption fees
- `dailyUserFees`: uniBTC redemption fees paid by users
- `dailyRevenue`: Bedrock-kept revenue
- `dailyProtocolRevenue`: same protocol-kept revenue, explicitly tracked
- `dailySupplySideRevenue`: uniETH rewards passed through to uniETH holders

For uniETH, `managerFeeShare` is read at each `RevenueAccounted` event block so historical splits are not affected by later fee-share changes.

## Chains

- Ethereum
- Base
- Merlin
- Bitlayer
- ZetaChain


Shaileshkhote:
@coderabbitai review

fees/bedrock/index.ts
```diff
@@ -0,0 +1,170 @@
...
+    redeemRouter: "0xe001Ce855F9e964e5243F0Ff11f2353dC371e810",
+  },
+};
+
+async function addUniETHFees(options: FetchOptions, config: ChainConfig, balances: BedrockBalances) {
+  if (!config.uniETHStaking) return;
+
+  const logs = await options.getLogs({
+    target: config.uniETHStaking,
+    eventAbi: UNIETH_REVENUE_ACCOUNTED_EVENT,
```
bheluga: i believe this is what the team makes, bcz https://etherscan.io/tx/0x8f47ba7b65d5024269e76669e7a218317e4008f19a213a3095d5c0f23aad569e#eventlog , revenue accounted and manager revenue compounded are same.
ig you can track yields through exchangeRatio function of the same contract

Shaileshkhote: looking into it

bheluga (dismissed):
@Shaileshkhote thanks for the pr
ideally we should be tracking as separate adapters, as tvl is separate https://defillama.com/protocol/bedrock

Shaileshkhote:
> @Shaileshkhote thanks for the pr ideally we should be tracking as separate adapters, as tvl is separate https://defillama.com/protocol/bedrock

looking into calculation one more time and splitting this into 3 adapters

fees/bedrock/index.ts
Shaileshkhote: I checked the tx and contract state. `RevenueAccounted` is gross rewards. In `_distributeRewards`, the contract does `fee = rewards * managerFeeShare / 1000`, adds that to `accountedManagerRevenue`, adds the rest to `accountedUserRevenue`, then emits `RevenueAccounted(rewards)`.

`ManagerRevenueCompounded` can equal `RevenueAccounted` because it compounds accumulated manager revenue, not only the fee from the current tx. In that tx there was already a large `accountedManagerRevenue` backlog, so `_compoundManagerRevenue()` compounded the full available ETH.

So the right source is still `RevenueAccounted` split by `managerFeeShare` at that block.

here are diff tnx where the fees are without backlogs 
https://etherscan.io/tx/0x4a9eb0a9f11c48fefb253026797efcf25fb05be355603e8b58b40ebf9016a759#eventlog
https://etherscan.io/tx/0x179ffa98e2c84c00f8723987747f695565c4489edc97524253359a03991ea7e7#eventlog

Shaileshkhote: I checked the contract source: `exchangeRatio()` tracks uniETH holder yield because it includes `accountedUserRevenue` and excludes `accountedManagerRevenue`.

I can update the adapter to use exchangeRatio delta for supply-side revenue, then add Bedrock’s manager fee on top for protocol revenue. Since docs state a 10% commission:

SupplySideRevenue = totalSupply * exchangeRatioDelta / 1e18
ProtocolRevenue = SupplySideRevenue * 10 / 90

Alternatively, protocol revenue can stay as `RevenueAccounted * managerFeeShare / 1000` to handle fee-rate changes exactly.

bheluga: > I checked the tx and contract state. `RevenueAccounted` is gross rewards. In `_distributeRewards`, the contract does `fee = rewards * managerFeeShare / 1000`, adds that to `accountedManagerRevenue`, adds the rest to `accountedUserRevenue`, then emits `RevenueAccounted(rewards)`.
> 
> `ManagerRevenueCompounded` can equal `RevenueAccounted` because it compounds accumulated manager revenue, not only the fee from the current tx. In that tx there was already a large `accountedManagerRevenue` backlog, so `_compoundManagerRevenue()` compounded the full available ETH.
> 
> So the right source is still `RevenueAccounted` split by `managerFeeShare` at that block.
> 
> here are diff tnx where the fees are without backlogs https://etherscan.io/tx/0x4a9eb0a9f11c48fefb253026797efcf25fb05be355603e8b58b40ebf9016a759#eventlog https://etherscan.io/tx/0x179ffa98e2c84c00f8723987747f695565c4489edc97524253359a03991ea7e7#eventlog

makes sense
how about we calculate sum of revenueAccounted as fees
sum of managerRevenueAccounted as revenue and difference as supply side? will this be correct?

Shaileshkhote: There isn’t a `managerRevenueAccounted` event on uniETH. The closest event is `ManagerRevenueCompounded`, but that’s not the same thing because it compounds accumulated manager revenue and can include backlog from previous periods.

So the cleaner options are either:
1. use `RevenueAccounted` as gross fees, then split by `managerFeeShare / 1000`, or
2. use `exchangeRatio` delta for supply-side revenue and derive protocol revenue as `supplySideRevenue * 10 / 90`.

The exchangeRatio method should give the correct economic split, but it may not always line up 1:1 with the accounting events in a specific tx because compounding/settlement can happen later.

Please suggest how we can move forward

bheluga: i would go with 2nd approach

Shaileshkhote: Got it, Applying 2nd approach

PR #6666, opened 2026-04-30
fix(fragment): report user fees

ohld:
## Summary

Adds `dailyUserFees` to the Fragment fees adapter and documents it in the methodology.

Fragment fees are user payments to Fragment wallets on TON, so this makes the adapter expose the user-fee dimension explicitly while keeping the existing fees/revenue SQL unchanged.

## Changes

- Return `dailyUserFees: dailyFees`
- Add `UserFees` methodology entry

No methodology change to `dailyFees`, `dailyRevenue`, or `dailySupplySideRevenue`.

### DIFF
--- a/fees/fragment/index.ts
+++ b/fees/fragment/index.ts
@@ -86,6 +86,7 @@ async function fetch(_a: any, _b: any, options: FetchOptions): Promise<FetchResu
 
   return {
       dailyFees,
+      dailyUserFees: dailyFees,
       dailyRevenue,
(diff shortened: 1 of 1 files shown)


ohld:
CI note: `Check ts` passed. `test-adapter` fails because fork PRs do not get `DUNE_API_KEYS` in Actions; the adapter itself is unchanged except exposing `dailyUserFees` and methodology.

bheluga:
@ohld thanks for the PR

PR #6735, opened 2026-05-01
fix etherfi liquid fees

### LINKED ISSUE #6704: fix etherfi liquid fees
https://defillama.com/protocol/ether.fi-liquid

uwezukwechibuzor:
### Summary
                                                                                                              
  Bug: The ether.fi liquid fees adapter was failing because the Dune query referenced a private dataset
  (dune.ether_fi.result_aum) inaccessible. Since restaking rewards and buybacks were
  combined in a single Dune query, the private dataset failure also prevented buybacks from being fetched.
                                                                                                                
  Fix:
  - Compute restaking rewards on-chain using totalSteth (already fetched by the adapter) with the same formula: 
  (3.5% * stETH_balance * 3.8%) / 365                                                                          
  - Separate buybacks into its own Dune query using only the public dex_aggregator.trades table 

Fixes  #6704

### DIFF
--- a/fees/ether-fi/index.ts
+++ b/fees/ether-fi/index.ts
@@ -210,63 +210,47 @@ const getMiscStakingRevenue = async (options: FetchOptions) => {
 
 const getAdditionalRevenueStreams = async (options: FetchOptions) => {
   const query = `
-    with 
-    
-    -- Restaking rewards from Eigenlayer via restaker contract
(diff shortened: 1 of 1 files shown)


bheluga:
@uwezukwechibuzor thanks for the PR. Fixed [here](https://github.com/DefiLlama/dimension-adapters/pull/6813)

PR #6736, opened 2026-05-01
fix(fees/ston): correct lp_fee denomination and add per-op sanity caps

### LINKED ISSUE #6707: fix ston.fi fees
https://defillama.com/protocol/ston.fi

AimenHallou:
Closes #6707.

## Summary

`pnpm test fees ston` currently crashes:

```
Error: value is too damn high | ton-dailyUserFees: 9584504856481046
```

The root cause is a denomination bug: `lp_fee_amount` is denominated in the swap's **input** asset, while `protocol_fee_amount` and `referral_fee_amount` are denominated in the swap's **output** asset (which is what `operation.fee_asset_address` points to). The current code multiplies all three by the **output** asset's price, so any cross-pair swap with very different per-unit prices (memecoin to TON, etc.) produces garbage USD values orders of magnitude above reality.

## Fix

In `fees/ston/index.ts`:

1. **Trusted-pool gate.** Only sum fees from pools with `lp_total_supply_usd >= 1000` (same threshold the existing price code already uses; previously not applied to fee summation).
2. **Correct fee denominations.**
   - `lp_fee_amount` x price of the swap's **input** asset
   - `protocol_fee_amount` x price of the swap's **output** asset
   - `referral_fee_amount` x price of the swap's **output** asset

   Input/output are derived from the signed `asset0_amount` / `asset1_amount` deltas in the operation (positive = into pool = input; negative = out of pool = output). `fee_asset_address` matches the output side.
3. **Per-op sanity caps.** A real LP fee is 0.2-0.3% of swap input; cap at 5% of input as the corrupt-data threshold. Same on the output side for protocol+referral. Operations exceeding either cap are dropped entirely.

## Test

```
TON
Backfill start time: 14/11/2023
Daily user fees: 3.53 k
Daily fees: 3.53 k
Daily supply side revenue: 2.40 k
Daily revenue: 1.13 k
```

Cross-check against the (working) volume adapter: `api.llama.fi/summary/dexs/ston.fi` reports **$2.07M** USD/day. With LP fees of ~0.2-0.3%, expected daily fees are roughly **$4.1k-$6.2k**. Our result ($3.53k) lands just under that range, which is consistent with the trusted-pool gate excluding low-liquidity / corrupt-fee operations.

The `dailyRevenue / dailyFees` ratio is 1.13k / 3.53k = **0.32**, matching the documented "Protocol receives 1/3 of fees" methodology almost exactly (the other 2/3 goes to LP + referral on the supply side).

## Methodology unchanged

The accounting still produces `Fees = LP + protocol + referral`, `Revenue = protocol`, `SupplySideRevenue = LP + referral`. The methodology block is unchanged. The only changes are which asset's price each component is multiplied by, plus the trusted-pool gate and per-operation sanity caps.

### DIFF
--- a/fees/ston/index.ts
+++ b/fees/ston/index.ts
@@ -6,34 +6,40 @@ import { FetchOptions, SimpleAdapter } from "../../adapters/types";
 
 const endpoint = "https://api.ston.fi/v1/stats/operations?";
 
+const TRUSTED_POOL_MIN_TVL_USD = 1000;
+// LP fee or protocol+referral fee should never exceed 5% of the side it's taken from.
+// Typical LP fee is 0.2-0.3%; 5% is wildly generous, so anything above that is corrupt.
+const MAX_FEE_FRACTION = 0.05;
+
 const fetchFees = async (options: FetchOptions) => {
   const pool_list = (await fetchURL("https://api.ston.fi/v1/pools")).pool_list;
-  // store pools info for each asset to calculate weigthed price later
-  const asset2pools = {};
-  const add_pool = (address: string, tvl: number, reserve: number) => {
-    if (!asset2pools[address]) {
-      asset2pools[address] = [];
-    }
-    asset2pools[address].push({ tvl, reserve });
-  };
+
+  const asset2pools: Record<string, Array<{ tvl: number; reserve: number }>> = {};
+  const trustedPools = new Set<string>();
+
   for (const pool of pool_list) {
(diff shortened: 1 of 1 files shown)


bheluga:
@AimenHallou thanks for the PR
Ston team has resolved this bug internally https://github.com/DefiLlama/dimension-adapters/pull/6829#issuecomment-4386459226

PR #6782, opened 2026-05-03
Add stUSDT fees and revenue adapter

Shaileshkhote:
## Summary

Fixes fee tracking for [stUSDT](https://defillama.com/protocol/stusdt).

Adds a v2 fees adapter that tracks stUSDT on Tron and Ethereum using on-chain logs.

## What Changed

- Tracks unstaking fees from `WithdrawalClaimed.fee`
- Tracks positive holder yield from `IncreaseBase`
- Tracks negative holder yield from `DecreaseBase`
- Uses `options.getLogs` for both Tron and Ethereum
- Reports:
  - `dailyFees`
  - `dailyRevenue`
  - `dailyProtocolRevenue`
  - `dailySupplySideRevenue`

## Methodology

- Unstaking fees are counted as `Deposit/Withdraw Fees`
- Unstaking fees are retained by the protocol, so they are counted as revenue and protocol revenue
- stUSDT rebases are counted as `Assets Yields`
- `IncreaseBase` is treated as positive supply-side yield to stUSDT holders
- `DecreaseBase` is treated as negative supply-side yield

### DIFF
--- a/fees/stUSDT/index.ts
+++ b/fees/stUSDT/index.ts
@@ -1,0 +1,119 @@
+import { FetchOptions, SimpleAdapter } from "../../adapters/types";
+import { CHAIN } from "../../helpers/chains";
+import { METRIC } from "../../helpers/metrics";
+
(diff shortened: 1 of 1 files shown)


bheluga:

PR #6801, opened 2026-05-04
fix: account for katana perps exchange contract upgrade

dileepfrog:
The Katana Perps Exchange contract was upgraded at Apr-27-2026 02:30:39 PM +UTC. This PR fixes the adapter to source events from the correct Exchange contract version based on requested timestamp


bheluga:
@dileepfrog thanks for the PR

dileepfrog:
@bheluga Would it be possible to re-run the backfill from April 27 to today? It looks like that period is still reporting blank per https://defillama.com/protocol/katana-perps?tvl=false&events=false&perpVolume=true

bheluga:
@dileepfrog should be good now

PR #6806, opened 2026-05-04
feat: Reservoir Protocol fees & revenue adapter

eukadish:
## Summary

- Replaces stub wsrUSD-only implementation with comprehensive on-chain yield tracking
- Captures yield from 4 sources: srUSD savings (SavingModule `currentPrice()`), trUSD term yield (MintTerm events on TermIssuer), ERC4626 vault positions held across treasury funds (Morpho MetaMorpho, Euler v2, Fluid, Cap, Ethena, InfiniFi, Stargate, IPOR, Hyperithm), and Aave v3 rebasing aToken positions
- Fixed two bugs in the previous implementation: wrong rUSD contract address and wrong yield calculation methodology
- All yield is classified as `dailySupplySideRevenue` (flows to rUSD/srUSD/wsrUSD/trUSD holders)

## Protocol Overview

Reservoir Protocol is a stablecoin system on Ethereum. It deploys capital into external DeFi yield sources and passes yield through to rUSD holders via srUSD (savings) and trUSD (fixed-term) instruments. Treasury funds hold positions across 27 ERC4626 vaults and 3 Aave aToken positions.

## Methodology

- **srUSD savings**: `srUSD.totalSupply × (SavingModule.currentPrice_end − currentPrice_start) / 1e8`
- **trUSD term yield**: `sum(principle − cost)` from MintTerm events (future interest owed to holders)
- **ERC4626 vaults**: `shares_start × (convertToAssets_end − convertToAssets_start) / 1e18` — using start shares isolates yield from same-day capital inflows
- **Aave aTokens**: balance diff between start and end of day (rebasing yield)

## Test plan

- [ ] Run `npm test fees reservoir-protocol` against recent dates to confirm output is non-zero and reasonable
- [ ] Verify no regressions on existing adapter structure (TypeScript compiles cleanly)

### DIFF
--- a/fees/reservoir-protocol.ts
+++ b/fees/reservoir-protocol.ts
@@ -1,63 +1,145 @@
 import { FetchOptions, SimpleAdapter } from '../adapters/types';
 import { CHAIN } from '../helpers/chains';
 
-const WSRUSD_CONTRACT = '0xd3fd63209fa2d55b07a0f6db36c2f43900be3094';
-const RUSD_CONTRACT = '0x31Eae643b679A84b37E3d0B4Bd4f5dA90fB04a61';
+// ─── Reservoir core
+const RUSD          = '0x09D4214C03D01F49544C0448DBE3A27f768F2b34';
+const SRUSD         = '0x738d1115B90efa71AE468F1287fc864775e23a31';
+const SAVING_MODULE = '0x5475611Dffb8ef4d697Ae39df9395513b6E947d7';
+const TERM_ISSUER   = '0x128D86A9e854a709Df06b884f81EeE7240F6cCf7';
 
-const abis = {
-  convertToAssets:
-    'function convertToAssets(uint256 shares) view returns (uint256)',
-  totalSupply: 'uint256:totalSupply',
-};
+// ─── Treasury holder addresses (mirrors TVL adapter)
+const FUND_A = '0x289C204B35859bFb924B9C0759A4FE80f610671c';
+const FUND_B = '0x3063C5907FAa10c01B242181Aa689bEb23D2BD65';
+const FUND_C = '0x841DB2cA7E8A8C2fb06128e8c58AA162de0CfCbC';
+const FUND_D = '0xA100A910A30b745064d7174863B730AD6d92Fe64';
+const FUND_E = '0xb595ba80d38b8e4c9894a6734a1b9a7b198870a2';
+const FUND_F = '0x5563CDA70F7aA8b6C00C52CB3B9f0f45831a22b1';
+const FUND_G = '0xE94fc572b5E5Abe38F326F7DeDfe4f0Df9851d2A';
+const FUND_J = '0x2adf038b67a8a29cda82f0eceb1ff0dba704b98d';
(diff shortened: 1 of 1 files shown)


bheluga (commented):
@eukadish thanks for the PR.
2 Issues:
1. Yield from mint term logs should be considered at maturity and not at issuance
2. Wont all these defi yields be realised in srUsd price? or are these yields different? and will they be distributed to  rUSD holders?

eukadish:
Thanks for the review @bheluga! Pushed a fix in the latest commit. Here's what changed and why:

**1. MintTerm timing → removed for now**

You're right. Looking at the `TermIssuer` source, the correct event to use is `RedeemTerm(from, to, termId, amount, timestamp)` — that fires when a holder redeems at maturity, which is when `rusd.mint(to, amount)` actually executes and the yield becomes real. The `MintTerm` event fires at issuance when rUSD is *burned*, not when yield is *created*.

However, the deployed `TermIssuer` has no `RedeemTerm` events yet (the contract only has admin role events — trUSD hasn't launched on this deployment). Rather than track yield at the wrong point, I've removed this section entirely. Once trUSD is live we can re-add it using `RedeemTerm` events and match against the original `MintTerm` cost for the same `termId`.

**2. ERC4626/Aave vs srUSD price — removed srUSD section**

Good catch. The architecture is:
- The treasury Funds (FUND_A through FUND_J) hold ERC4626 and Aave positions — these are the **upstream income sources**
- The `SavingModule.currentPrice()` reflects yield being **distributed downstream** to srUSD stakers

Measuring srUSD price appreciation *and* the ERC4626/Aave yields would be tracking the same money at two different stages of the value chain. The adapter now only counts the upstream DeFi yields (ERC4626 vault price appreciation + Aave rebasing), which is where the fees actually originate. All of it flows to rUSD/srUSD holders (`dailyRevenue` = 0).

Also added the `dailyRevenue` field and `METRIC.ASSETS_YIELDS` breakdown labels that CodeRabbit flagged.

bheluga:
@eukadish isnt it simpler to calculate price appreciation through contract call.
in this approach we need to keep changing the code everytime your strategy changes / you invest in a new fund.
do you have any specific reason why you feel this approach is better than the previous?

eukadish:
You're right — pushed a simplified version in the latest commit.

The adapter now uses a single `SavingModule.currentPrice()` delta × `SRUSD.totalSupply()` call. This:
- Captures **all yield automatically** — DeFi vaults, RWA, any future strategy change — without needing to maintain an enumerated position list
- Reports more accurate numbers (the previous ERC4626 list was only seeing ~$4K/day of the ~$25K/day actual yield, since RWA yield isn't directly queryable on-chain)
- Resolves the double-counting concern from the previous review — the srUSD price IS the downstream aggregate of all upstream yield sources

Also updated the methodology to:
- **TVL**: Total outstanding stablecoins minted (rUSD, srUSD, wsrUSD, and trUSD)
- **Fees**: Total yield and rewards from assets.
- **Revenue**: Protocol revenue consists of yield generated on assets.

PR #6823, opened 2026-05-05
Add saturn protocol fees and revenue adapter

Shaileshkhote:
Tracks fees and revenue for saturn-credit https://defillama.com/protocol/saturn

## Summary

Adds the Saturn Protocol fee adapter on Ethereum.

Tracks:
- sUSDat deposit fees, reconstructed from ERC4626 `Deposit.assets` and `depositFeeBps`
- sUSDat STRC rewards, valued through the STRC oracle
- Saturn’s 10% performance fee on gross sUSDat STRC rewards
- USDat reserve yield claimed by Saturn through `YieldClaimed`

## Methodology

sUSDat deposit fees are calculated from the net ERC4626 deposit amount because `Deposit.assets` is emitted after fees.

sUSDat STRC rewards are treated as the amount received by vault holders. The adapter values those rewards using the STRC oracle, then grosses them up to account for Saturn’s 10% performance fee.

USDat reserve yield is counted from the `YieldClaimed(uint256 amount)` event on the USDat contract and attributed to protocol revenue.

### DIFF
--- a/fees/saturn-protocol/index.ts
+++ b/fees/saturn-protocol/index.ts
@@ -1,0 +1,110 @@
+import { Adapter, FetchOptions } from "../../adapters/types";
+import { CHAIN } from "../../helpers/chains";
+import { METRIC } from "../../helpers/metrics";
+
+const USDAT = "0x23238f20b894f29041f48D88eE91131C395Aaa71";
+const sUSDat = "0xD166337499E176bbC38a1FBd113Ab144e5bd2Df7";
+const STRC_ORACLE = "0x5f7eCD0D045c393da6cb6c933c671AC305A871BF";
+const BPS = 10000n;
+const PERFORMANCE_FEE_BPS = 1000n;
+
+// Sources:
+// - USDat / sUSDat docs: https://saturncredit.gitbook.io/saturn-docs/solution
+// - Fee and reserve docs: https://saturncredit.gitbook.io/saturn-docs/operations-and-governance/protocol-fee-and-risk-reserve
+// - Verified contracts: https://etherscan.io/address/0xD166337499E176bbC38a1FBd113Ab144e5bd2Df7#code
+// - M0 JMI yield model: https://docs.m0.org/build/models/treasury/jmi/overview/
+const Event = {
+  Deposit: "event Deposit(address indexed sender,address indexed owner,uint256 assets,uint256 shares)",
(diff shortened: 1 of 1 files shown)


bheluga:

Shaileshkhote:
Hey @bheluga! What was the problem with the event-based implementation? Would love to understand it better and improve my PRs going forward.

Shaileshkhote:
@bheluga The USDAT yield is unaccounted in the current code 
I traced the fun and the USDAT yields goes to the protocol wallet by directly claiming it from the M0 protocol and then added as LP to USDAT,
I believe the yield is kept as protocol revenue

bheluga:
> @bheluga The USDAT yield is unaccounted in the current code I traced the fun and the USDAT yields goes to the protocol wallet by directly claiming it from the M0 protocol and then added as LP to USDAT, I believe the yield is kept as protocol revenue

Hey, thanks for letting me know.
I have raised PR to fix [here](https://github.com/DefiLlama/dimension-adapters/pull/6864)
usually there is nothing wrong with events, but I wanted to avoid it here as yields were claimed on random days, which would lead to 0 fees on most of the days and sudden spike on few days similar to https://defillama.com/protocol/ethena
but I am using the same in the new PR, as there is no better alternative.
we could still get M0 yields through earnerRate function and STRC yields from somewhere (11.5% now)  and keep yields uniform for every hour, but STRC yields aren't guaranteed and they can default anytime

Shaileshkhote:
> > @bheluga The USDAT yield is unaccounted in the current code I traced the fun and the USDAT yields goes to the protocol wallet by directly claiming it from the M0 protocol and then added as LP to USDAT, I believe the yield is kept as protocol revenue
> 
> Hey, thanks for letting me know. I have raised PR to fix [here](https://github.com/DefiLlama/dimension-adapters/pull/6864) usually there is nothing wrong with events, but I wanted to avoid it here as yields were claimed on random days, which would lead to 0 fees on most of the days and sudden spike on few days similar to https://defillama.com/protocol/ethena but I am using the same in the new PR, as there is no better alternative. we could still get M0 yields through earnerRate function and STRC yields from somewhere (11.5% now) and keep yields uniform for every hour, but STRC yields aren't guaranteed and they can default anytime

makes sense, I'll note this down and keep it in mind for upcoming PRs. thanks for the detailed explanation!

PR #6829, opened 2026-05-06
fix: STON - include referral swaps in fees

Nathan22177:
## Reason/details

This is not a new protocol listing.

This PR updates the existing STON.fi fees adapter to include referral swap operations in fee calculations.

STON.fi exposes referral swaps as successful swap operations with:

- `operation_type = "swap"`
- `success = true`
- `exit_code = "swap_ok_ref"`

These operations include the same fee fields used by the adapter:

- `lp_fee_amount`
- `protocol_fee_amount`
- `referral_fee_amount`
- `fee_asset_address`

Previously the adapter only counted `swap_ok`, which skipped `swap_ok_ref` operations and undercounted fees, including referral fees.

## Change

Adds an explicit allowed exit-code set:

```ts
const swapExitCodes = new Set(["swap_ok", "swap_ok_ref"]);
```
and uses it when counting successful swap operations.

## Tests

- `npm run test -- fees ston 2026-04-23`

### DIFF
--- a/fees/ston/index.ts
+++ b/fees/ston/index.ts
@@ -5,6 +5,7 @@ import fetchURL from "../../utils/fetchURL";
 import { FetchOptions, SimpleAdapter } from "../../adapters/types";
 
 const endpoint = "https://api.ston.fi/v1/stats/operations?";
+const swapExitCodes = new Set(["swap_ok", "swap_ok_ref"]);
 
 const fetchFees = async (options: FetchOptions) => {
   const pool_list = (await fetchURL("https://api.ston.fi/v1/pools")).pool_list;
@@ -48,7 +49,7 @@ const fetchFees = async (options: FetchOptions) => {
   // go through all operations and calculate fees based on the current prices
   for (const item of res["operations"]) {
     const operation = item.operation;
-    if (operation.success && operation.operation_type == "swap" && operation.exit_code == "swap_ok") {
+    if (operation.success && operation.operation_type == "swap" && swapExitCodes.has(operation.exit_code)) {
       if (operation.fee_asset_address in asset_prices) {
         const price = asset_prices[operation.fee_asset_address];
         total_lp_fees += operation.lp_fee_amount * price;


bheluga:
@Nathan22177 thanks for the PR.
Is the fee number correct? Ston has like 35M in volume and half of its volume comes from TON/USDT pair which has fee tier  of 0.07%, I dont get how the fees is 130k

Nathan22177:
We found a bug on our side while looking into fees disappearing from Llama website entirely and deployed a fix about 2 hours ago.
So I think the 130k is from 24h/daily window after our fix went live. 

While investigating, we also noticed that the adapter was not counting swaps with referrals, hence the PR.

The bug with fees reporting for non-referral swaps, if you plan to resync the historic data, went live on April 21st.
Swaps with referrals were never counted since the introduction of the adapter, as far as I can tell.

bheluga:
> We found a bug on our side while looking into fees disappearing from Llama website entirely and deployed a fix about 2 hours ago. So I think the 130k is from 24h/daily window after our fix went live.
> 
> While investigating, we also noticed that the adapter was not counting swaps with referrals, hence the PR.
> 
> The bug with fees reporting for non-referral swaps, if you plan to resync the historic data, went live on April 21st. Swaps with referrals were never counted since the introduction of the adapter, as far as I can tell.

The reason why we have blank data for past few days is because your api returned too high fee numbers, which were auto blocked.
Also coming to historic resync, it is not possible as you return only last 24h data, if your api supports querying historic data similar to your volume adapter, we can refill

Nathan22177:
> your api returned too high fee numbers

Yeah root cause for that behavior is what we've addressed today with the fix I mentioned. 

> if your api supports querying historic data similar to your volume adapter, we can refill

Yeah you can pass startTime/endTime as adapter does, if you could refill we would greatly appreciate it!

bheluga:
> > your api returned too high fee numbers
> 
> Yeah root cause for that behavior is what we've addressed today with the fix I mentioned.
> 
> > if your api supports querying historic data similar to your volume adapter, we can refill
> 
> Yeah you can pass startTime/endTime as adapter does, if you could refill we would greatly appreciate it!

Hey, thanks . I had actually missed that part which does allow backfilling and will surely fill.
Any idea what exactly changed on may 4th onwards, there were multiple fold increase in volumes?
2M -> 15M -> 40M https://defillama.com/protocol/ston.fi

Nathan22177:
Pavel Durov made some announcements. Speed-up of the blockchain, 6 times less blockchain fees, Telegram team overtaking TON foundation as leading validator and driver of the blockchain

bheluga:
@Nathan22177 looks like the bug is back

Nathan22177:
yeah we've rolled back the api and are now fixing the data, ETA 1h

Nathan22177:
should be fixed irreversibly now

bheluga:
@Nathan22177 thanks for the PR

PR #6835, opened 2026-05-06
Track mezo chain fees

uwezukwechibuzor:
https://defillama.com/chain/mezo


bheluga:

PR #6840, opened 2026-05-06
fix(jade): set Revenue to multisig inflow only; correct methodology

gordy-ftw:
Follow-up to #6836. Two issues with the merged adapter:

1. `dailyRevenue` was set to `dailyFees`, so DefiLlama renders "Fees 7d" and "Revenue 7d" identically. For Jade, only the 1% admin fee retained by the team multisig is protocol revenue — the rest flows to JADE holders via on-chain buyback-and-burn. `dailyRevenue` is now `dailyProtocolRevenue`.

2. Methodology mis-described both streams (wrong percentages, missing the Treasury PDA's internal 90/10 split between buyback and the deep-vein lottery, no mention of keeper gas reimbursement).

The adapter measures two wallets instead of using the single-wallet + `.clone(0.x)` pattern of `fees/ore` / `fees/godl` / `fees/orb` / `fees/coal-works`, because Jade's admin fee and vault flow to two genuinely separate on-chain accounts at every round reset:

- Fee-collector multisig `FW4UFt5nDKE2DLVNj979rXjFkjCdmzs9344JetX8hY9P` receives the 1% admin fee.
- Treasury PDA `F4Uwd5sQT8go5r6iiejrVc2iurYSc2dA4RKvX3cLB8e3` receives the vault (10% of winnings, or ~99% of total_deployed in no-winner rounds), later spent on JADE buyback-and-burn.

`isExpensiveAdapter: true` is dropped to match the rest of the Ore-family adapters.

### DIFF
--- a/fees/jade/index.ts
+++ b/fees/jade/index.ts
@@ -2,22 +2,26 @@ import { Dependencies, FetchOptions, SimpleAdapter } from "../../adapters/types"
 import { CHAIN } from "../../helpers/chains";
 import { oreHelperCountSolBalanceDiff } from "../ore";
 
-const fetch: any = async (_a: any, _b: any, options: FetchOptions) => {
-  const dailyFees = await oreHelperCountSolBalanceDiff(
-    options,
-    "F4Uwd5sQT8go5r6iiejrVc2iurYSc2dA4RKvX3cLB8e3"
-  );
+// Team multisig that receives the 1% admin fee on every round's total_deployed.
+const FEE_COLLECTOR = "FW4UFt5nDKE2DLVNj979rXjFkjCdmzs9344JetX8hY9P";
 
-  const dailyRevenue = dailyFees.clone(0.01);
-  const dailyProtocolRevenue = dailyFees.clone(0.01);
-  const dailySupplySideRevenue = dailyFees.clone(0.99);
-  const dailyHoldersRevenue = dailyFees.clone(0.99);
+// Treasury PDA that receives the vault portion of each round (~9% of winnings,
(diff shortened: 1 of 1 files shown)


bheluga:
@gordy-ftw thanks for the PR
dailyRevenue = dailyProtocolRevenue + dailyHoldersRevenue, i think its already correct

bheluga:
@gordy-ftw can i close the PR or you meant to change something else too?

gordy-ftw:
> @gordy-ftw thanks for the PR dailyRevenue = dailyProtocolRevenue + dailyHoldersRevenue, i think its already correct

image


These shouldn't match though correct?

bheluga:
> > @gordy-ftw thanks for the PR dailyRevenue = dailyProtocolRevenue + dailyHoldersRevenue, i think its already correct
> 
> image
> These shouldn't match though correct?

all the fees goes to either buyback and burn, token stakers or protocol right?
If yes, fees = revenue, fees wont be equal to revenue only if some part of fees goes to LPs , 3rd party etc

gordy-ftw:
> > > @gordy-ftw thanks for the PR dailyRevenue = dailyProtocolRevenue + dailyHoldersRevenue, i think its already correct
> > 
> > 
> > image
> > These shouldn't match though correct?
> 
> all the fees goes to either buyback and burn, token stakers or protocol right? If yes, fees = revenue, fees wont be equal to revenue only if some part of fees goes to LPs , 3rd party etc


ah ok, that make sense so its correct! thanks for the speedy help. I will close this.

PR #6848, opened 2026-05-06
[feat] Risex: add risex perps volume and oi adapters

Shaileshkhote:
Fixes #6845 
https://defillama.com/protocol/risex

## Summary
- Adds RiseX perpetuals adapters for DEX volume and open interest.
- Uses Rise Trade API endpoints to report current 24h perps volume and current market open interest.

## Changes
- Added `dexs/risex-perps` for `dailyVolume` on `CHAIN.OFF_CHAIN`.
- Added `open-interest/risex-perps-oi.ts` for `openInterestAtEnd` on `CHAIN.OFF_CHAIN`.
- Added explicit missing-data checks so API shape issues fail loudly instead of returning silent zeroes.

## Methodology
- Volume is sourced from `https://api.rise.trade/v1/stats/volume?include_bots=true` using `data.total_volume`.
- Open interest is calculated from `https://api.rise.trade/v1/markets` as `sum(open_interest * mark_price)` across visible/available markets, falling back to `index_price` or `last_price` if needed.
- Adapters use default v1/current-time behavior because the Rise Trade endpoints return current API aggregates rather than timestamp-filtered historical data.

PR #6852, opened 2026-05-07
feat: add RISEx fees adapter (missing from #6848)

pDubey7:
## Summary

Adds fees tracking for RISEx perpetual DEX on RISE Chain, which was 
missing from the merged PR #6848 (volume + OI only).

## Approach

Uses the same `/v1/markets` API endpoint as the OI adapter, applying 
a blended 2bps fee rate (Tier 1: 3bps taker + 1bps maker per docs).

## Fee Structure (from docs.rise.trade/docs/risex/fees)

| Tier | 14d Volume | Taker | Maker |
|------|-----------|-------|-------|
| Tier 1 | $0 | 3bps | 1bps |
| Tier 2 | $5M | 2.5bps | 0.75bps |

Most users are Tier 1, so blended 2bps is used.

## Verified Numbers

- Total 24h volume: ~$38M
- Daily fees (2bps): ~$7.6K ✅ (manual calc matches adapter output)

## Notes

- Maker rebate program not yet live per docs
- `runAtCurrTime: true` since API only returns current 24h data

### DIFF
--- a/fees/risex-perps.ts
+++ b/fees/risex-perps.ts
@@ -1,0 +1,43 @@
+import { FetchOptions, SimpleAdapter } from "../adapters/types";
+import { CHAIN } from "../helpers/chains";
+import fetchURL from "../utils/fetchURL";
+
+const MARKETS_API = "https://api.rise.trade/v1/markets";
+
+// Fee schedule (Tier 1): Taker 3bps, Maker 1bps — blended 2bps
+const BLENDED_FEE_BPS = 2;
+
+const fetch = async (options: FetchOptions) => {
+  const response = await fetchURL(MARKETS_API);
+  const markets = response.data?.markets;
+
+  if (!markets?.length) throw new Error("RiseX markets data missing");
+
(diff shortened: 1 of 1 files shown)


pDubey7:
@bheluga fees adapter is now ready in separate PR uses blended 2bps (Tier 1: 3bps taker + 1bps maker) applied to 24h volume. Daily fees: ~$7.6K. All findings of coderabbit addressed.                                                                               1. Wrapped fetch in try/catch returns zeroed balances on API outage
2. Added breakdown label 'trading_fees' to both addUSDValue calls
3. Added breakdownMethodology mapping

bheluga (changes requested):
@pDubey7 thanks for the PR
Here we are assuming that all/majority of the volume is from Tier 1, which can be wrong as a large portion of the volume may come from higher-tier traders. Try to track it on-chain if the data is available.

bheluga:
[this pr](https://github.com/DefiLlama/dimension-adapters/pull/6853) uses onchain logs

PR #6861, opened 2026-05-08
Track sierra protocol fees and revenue

uwezukwechibuzor:
https://defillama.com/protocol/sierra-protocol


bheluga:

PR #6876, opened 2026-05-08
[feat ]active-users: track user stats for sui and aptos 

Shaileshkhote:
## Summary
- Adds chain-level active user adapters for Sui and Aptos.
- Reports daily active users, daily transaction count, and daily new users from Dune-backed chain transaction tables.

## Changes
- Added `active-users/sui.ts`.
- Added `active-users/aptos.ts`.
- Marked both adapters with `dependencies: [Dependencies.DUNE]`.
- Uses exact timestamp windows for daily aggregation.
- Sui filters successful non-system transactions from `sui.transactions`.
- Aptos filters successful user transactions from `aptos.user_transactions`.

## Methodology
- `dailyActiveUsers`: distinct transaction senders in the daily window.
- `dailyTransactionsCount`: successful transaction count in the daily window.
- `dailyNewUsers`:
  - Sui: first successful non-system transaction timestamp for wallets active in the daily window.
  - Aptos: distinct senders with `sequence_number = 0` in the daily window.

## Tested Output
- `active-users sui 2026-05-08`
  - Daily active users: `193.81k`
  - Daily transactions count: `3.27M`
  - Daily new users: `9.46k`
- `active-users aptos 2026-05-08`
  - Daily active users: `404.96k`
  - Daily transactions count: `9.64M`
  - Daily new users: `17.77k`

### DIFF
--- a/active-users/aptos.ts
+++ b/active-users/aptos.ts
@@ -1,0 +1,44 @@
+import { Dependencies, FetchOptions, SimpleAdapter } from "../adapters/types";
+import { CHAIN } from "../helpers/chains";
+import { queryDuneSql } from "../helpers/dune";
+
+type Row = {
+  dau: number;
+  new_users: number;
+  tx_count: number;
+}
+
+const fetch = async (options: FetchOptions) => {
+  const [row]: Row[] = await queryDuneSql(options, `
+    SELECT
+      COUNT(*) AS tx_count,
+      COUNT(DISTINCT sender) AS dau,
+      COUNT(DISTINCT CASE WHEN sequence_number = 0 THEN sender END) AS new_users
+    FROM aptos.user_transactions
+    WHERE success = true
+      AND sender IS NOT NULL
+      AND block_date >= CAST(from_unixtime(${options.startTimestamp}) AS date)
+      AND block_date <= CAST(from_unixtime(${options.endTimestamp}) AS date)
+      AND block_time > from_unixtime(${options.startTimestamp})
+      AND block_time <= from_unixtime(${options.endTimestamp})
+  `);
+
+  return {
+    dailyActiveUsers: row?.dau ?? 0,
+    dailyTransactionsCount: row?.tx_count ?? 0,
+    dailyNewUsers: row?.new_users ?? 0,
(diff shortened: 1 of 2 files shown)


bheluga:
@Shaileshkhote thanks for the pr. we use allium wherever possible as its cheaper over dune

Shaileshkhote:
> @Shaileshkhote thanks for the pr. we use allium wherever possible as its cheaper over dune

Got it, sir. We’ll use Allium where possible since it’s cheaper than Dune Analytics.
I don’t have access to test Allium queries (enterprise-only), so should I prototype on Dune and then port ?

Shaileshkhote:
> @Shaileshkhote thanks for the pr. we use allium wherever possible as its cheaper over dune

Also, I think we should create a separate folder structure for new users, for example new-users/sui.ts. The server expects a different data shape for dailyNewUsers than it does for dailyActiveUsers, so keeping them separate should make the adapters cleaner and avoid mixing two different response formats.

bheluga:
> > @Shaileshkhote thanks for the pr. we use allium wherever possible as its cheaper over dune
> 
> Got it, sir. We’ll use Allium where possible since it’s cheaper than Dune Analytics. I don’t have access to test Allium queries (enterprise-only), so should I prototype on Dune and then port ?

you have schemas [here](https://docs.allium.so/historical-data/overview) so chances of making mistakes are very less, you can create PR with that and i will modify if anything breaks (sometimes they mention wrong column in schema)

Shaileshkhote:
> > > @Shaileshkhote thanks for the pr. we use allium wherever possible as its cheaper over dune
> > 
> > 
> > Got it, sir. We’ll use Allium where possible since it’s cheaper than Dune Analytics. I don’t have access to test Allium queries (enterprise-only), so should I prototype on Dune and then port ?
> 
> you have schemas [here](https://docs.allium.so/historical-data/overview) so chances of making mistakes are very less, you can create PR with that and i will modify if anything breaks (sometimes they mention wrong column in schema)

Got it, I’ll use the Allium schemas from the docs and proceed with the PR accordingly.

PR #6884, opened 2026-05-09
fix(arweave): replace broken ViewBlock API with Goldsky GraphQL

tarun-khatri:
## Summary

- Replaces the dead ViewBlock API (returns HTTP 400 on all requests) with [Goldsky's Arweave GraphQL index](https://arweave-search.goldsky.com/graphql)
- Adds `bundledIn: null` filter to capture only **base-layer fee-paying transactions** — bundled data items (Irys/Turbo bundles) carry a zero fee individually, so omitting this filter produces near-zero results
- Paginates across 5 parallel block sub-ranges for speed (~200 pages/day in ~28s total)
- Upgrades from `version: 1` / `Adapter` to `version: 2` / `SimpleAdapter` pattern

## Why `bundledIn: null` matters

~95% of Arweave user activity goes through Irys/Turbo bundlers. Each bundle is one parent transaction with a real fee; the individual items inside carry `fee.winston = "0"`. Without this filter, the query sums thousands of zero-fee items and yields ~$1/day. With the filter, results align with the expected ~$200-350/day.

## Test output

```
ARWEAVE 👇
Backfill start time: 8/6/2018
Daily fees: 257.00 AR
End timestamp: 1778317837 (2026-05-09T09:10:37.000Z)
```

Closes the data gap introduced when the ViewBlock endpoint stopped responding.

### DIFF
--- a/fees/arweave/index.ts
+++ b/fees/arweave/index.ts
@@ -1,72 +1,27 @@
 import { SimpleAdapter, FetchOptions, ProtocolType } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
-import { httpGet, httpPost } from "../../utils/fetchURL";
+import { httpGet } from "../../utils/fetchURL";
 
-const GOLDSKY_GQL = "https://arweave-search.goldsky.com/graphql";
-const AVG_BLOCK_TIME_S = 130; // ~2 min average Arweave block time
-const PAGE_SIZE = 100;
-const MAX_PAGES_PER_RANGE = 300;
-const PARALLEL_RANGES = 5;
-const WINSTON_PER_AR = 1e12;
-
-const estimateBlock = (targetTs: number, tipHeight: number, tipTs: number): number =>
-  Math.max(0, tipHeight - Math.round((tipTs - targetTs) / AVG_BLOCK_TIME_S));
-
-const fetchRangeWinston = async (minBlock: number, maxBlock: number): Promise<number> => {
-  let totalWinston = 0;
-  let cursor: string | undefined;
-
-  for (let page = 0; page < MAX_PAGES_PER_RANGE; page++) {
-    const query = `{
-      transactions(
-        block: { min: ${minBlock}, max: ${maxBlock} }
-        bundledIn: null
-        first: ${PAGE_SIZE}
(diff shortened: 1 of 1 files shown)


bheluga (changes requested):
Viewblock still works for me ?
also this seem to give wrong data for some older days , for eg: in peak of may 2025, https://viewblock.io/arweave/stat/txFees says around 20k USD (https://defillama.com/chain/arweave?chainFees=true) and this code give around 5-6k

tarun-khatri:
@bheluga thanks for the catch — you're right on both counts. Pivoted the PR (commit 9a9b5b6ac):

**1. ViewBlock is the source again.** Verified it returns HTTP 200 with the original `origin: arscan.io` header. The original adapter wasn't broken because of API access — it was breaking because `findIndex` returns -1 on days where the ViewBlock series lacks a point, which threw and aborted the whole backfill. Now treats missing days as `0` so backfill can complete.

**2. The "5–6k vs 20k" gap was a wrong-block-range bug, not a methodology gap.** My earlier replacement used linear interpolation against a fixed avg block time to map timestamps → block heights, which drifts massively for older periods (block 1,525,900 actually sits in Oct 2024, not May 2025). That's why the numbers came out off. With ViewBlock back as the source, the values match exactly.

**Validation across history (raw AR, ViewBlock vs adapter output):**

| Date       | ViewBlock UI |  Adapter | Match |
|------------|-------------:|---------:|------|
| 2025-05-08 |    3,864 AR  | 3,864 AR | ✅   |
| 2025-05-16 |      984 AR  |   984 AR | ✅   |
| 2025-09-15 |      374 AR  |   374 AR | ✅   |
| 2026-02-15 |      325 AR  |   325 AR | ✅   |
| 2026-05-08 |      135 AR  |   135 AR | ✅   |

(USD values shown in test output are CG-priced — `addCGToken("arweave", N)` × historical AR price.)

**Net diff vs main**: +18 / -62. Just removes the throw on missing days, keeps `dailySupplySideRevenue` + labeled `addCGToken` + `breakdownMethodology` for income-statement compliance, and upgrades to v2 SimpleAdapter shape.

Ready for re-review when you have a moment.

bheluga:
Thanks for the PR
Closing as there is no change compared to existing code

PR #6890, opened 2026-05-09
feat: MetaLeX - Track fees across Ethereum, Base, and Arbitrum

Cryptofowl:
Adds a MetaLeX fees adapter for Ethereum, Base, and Arbitrum.

The adapter uses the same production factory scope as the [MetaLeX TVL ](https://github.com/DefiLlama/DefiLlama-Adapters/pull/19136) adapter to discover `DealManager` and `RoundManager` contracts, then tracks `FeeDistributed` emitted during escrow finalization.

It reports `dailyFees`, `dailyRevenue`, and `dailyProtocolRevenue`. For MetaLeX, these values are identical because 100% of the current 0.3% service fee is retained entirely by the protocol.

Validated locally with:

```bash
pnpm test fees metalex 2025-12-17
```

Comparing cumulative FeeDistributed handling against raw event totals across all supported chains also match exactly


fees/metalex/index.ts
```diff
@@ -0,0 +1,119 @@
...
+    const managers = new Set(await discoverManagers(options));
+    if (!managers.size)
+      return {
+        dailyFees,
+        dailyRevenue: dailyFees,
+        dailyProtocolRevenue: dailyFees,
+      };
+
+    const feeLogs = await options.getLogs({
+      noTarget: true,
```
noateden: you can use `targets: managers` here and don't need to filter later, `noTarget` flag will make it query all logs from the chain, i's heavy

noateden (changes requested):
can you check again if protocol has no fees? because the tvl PR was closed?

fees/metalex/index.ts
Cryptofowl: Thanks, added in 74d24ce

Cryptofowl:
Yes, protocol has fees. 

Earliest examples per chain:
https://arbiscan.io/tx/0x1826fcf5d19818a416fce6162ef6e7626ea3e5b8623dfb877be19b41ec57fff6
https://basescan.org/tx/0x2b694858ff703cd8848ea93ce96f233f2ad2936ad8568e4690e28c78848c20c6
https://etherscan.io/tx/0x03baa9fedcec2f7ff6f3026b2ed848fe1248918d60821f3d7b9dbf28b2f617e7

I've shortened the start date to match, fee rate was set to 0.3% on December 15th, 2025. 

The protocol is on-chain tokenized securities infra, fees are collected at escrow close. The TVL is naturally going to fluctuate as a result of normal usage of escrow contracts when closing deals.

noateden (approved):

PR #6891, opened 2026-05-09
[feat] usdd: add usdd fees and revenue adapter 

Shaileshkhote:
Smart Allocator not added, it does not have protocol revenue and data is scattered with no confirm source.

https://defillama.com/protocol/usdd


fees/usdd/index.ts
```diff
@@ -0,0 +1,162 @@
...
+const blockOf = ({ blockNumber, block, block_number }: any) => Number(blockNumber ?? block ?? block_number);
+
+const tronRate = (markets: any[]) => {
+  let debt = 0;
+  let fees = 0;
+
+  markets.forEach(({ curMinted, stabilityFee }) => {
+    const minted = Number(curMinted ?? 0);
+    debt += minted;
+    fees += minted * Number(stabilityFee ?? 0);
```
noateden, line 56: why don't you simply count fees by sum `minted * stabilityFee` here?

fees/usdd/index.ts
```diff
@@ -0,0 +1,162 @@
...
+    dailyRevenue.addUSDValue(fee, METRIC.BORROW_INTEREST);
+
+    dailyProtocolRevenue.addBalances(dailyRevenue);
+    return { dailyFees, dailyRevenue, dailyProtocolRevenue };
+  }
+
+  const config = chainConfig[options.chain] as any;
+
+  const [folds, sells, buys, barks] = await Promise.all([
+    options.getLogs({ target: config.vat, eventAbi: ABI.fold, entireLog: true }),
```
noateden: remove `entireLog` flag here, and just need to use `log .rate later`

noateden (dismissed):
please fix some issues I commented above

fees/usdd/index.ts
Shaileshkhote: sum(curMinted * stabilityFee) would only be correct for current-time runs. Since this adapter supports historical dates, we use the snapshot only to derive a weighted average rate, then apply it to the historical debt. If we set runAtCurrTime, we can simplify to sum(curMinted * stabilityFee) / 365

fees/usdd/index.ts
Shaileshkhote: Thanks for flagging, i removed it

Shaileshkhote:
@noateden I fixed the logs issue.

For the stability fee calculation, we’d need to switch this to runAtCurrTime if we want to use the current sum(curMinted * stabilityFee) approach directly.

Also, I haven’t included Smart Allocator yields here. Those are generated from idle assets, but I couldn’t find all the relevant addresses, and I don’t see any protocol fee tied to that yield from docs.

noateden:
@Shaileshkhote you're using version, you should calculate fees based on from and to timestamps
```
const YEAR = 365 * 24 * 3600;
const timeframe = options.toTimestamp - options.fromTimestamp;
const fee = tronDebt(history.data?.items ?? [], options.fromTimestamp) * tronRate(collaterals.data?.items ?? []) * timeframe / YEAR;
```
I fixed and merged

Shaileshkhote:
> @Shaileshkhote you're using version, you should calculate fees based on from and to timestamps
> 
> ```
> const YEAR = 365 * 24 * 3600;
> const timeframe = options.toTimestamp - options.fromTimestamp;
> const fee = tronDebt(history.data?.items ?? [], options.fromTimestamp) * tronRate(collaterals.data?.items ?? []) * timeframe / YEAR;
> ```
> 
> I fixed and merged

Got it, will keep this in mind for future PRs. Thanks!

PR #6896, opened 2026-05-10
[feat] afiprotocol: add fees and revenue adapter for afiUSD

Shaileshkhote:
Tracks fees and revenue for https://defillama.com/protocol/afi-protocol

## Summary
- Adds the AFI Protocol fees adapter for afiUSD on Ethereum.
- Tracks net vault yield, treasury performance fees, and holder supply-side yield from the AFI `Yield.DistributeYield` event.

## Changes
- Added `fees/afiprotocol/index.ts`.
- Uses the AFI Yield contract on Ethereum: `0xb82b080791dFA4aa6Cac8c3f9c0fcb4471C9FEaD`.
- Reports:
  - `dailyFees` as net holder yield plus treasury performance fees.
  - `dailySupplySideRevenue` as net yield credited to afiUSD holders.
  - `dailyRevenue` and `dailyProtocolRevenue` as performance fees minted to or burned from the AFI treasury.
- Enables `allowNegativeValue` so loss distributions are reflected instead of dropped.

## Methodology
- Reads `DistributeYield(address asset, address receiver, uint256 amount, uint256 feeAmount, bool profit)` events.
- For profit distributions, adds `amount` as holder yield and `feeAmount` as treasury performance fees.
- For loss distributions, subtracts `amount` and `feeAmount` using the same labels.

### DIFF
--- a/fees/afiprotocol/index.ts
+++ b/fees/afiprotocol/index.ts
@@ -1,12 +1,10 @@
 import { FetchOptions, SimpleAdapter } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
-import coreAssets from "../../helpers/coreAssets.json";
 import { METRIC } from "../../helpers/metrics";
 
-const chainConfig: Record<string, { start: string; yield: string, yieldToken: string }> = {
+const chainConfig: Record<string, { start: string; yield: string }> = {
   [CHAIN.ETHEREUM]: {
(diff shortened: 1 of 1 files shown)


Shaileshkhote:
Skipping `afi-rwaUSDi` — its Ethereum, Base, and Monad vault contracts only expose deposit/withdrawal fees, which have been 0 since inception with no events to index. It's also institution-only and not publicly accessible, so there's nothing to backtest against. Happy to add it directly if needed, but we'd have no historical data.

PR #6902, opened 2026-05-10
[fix] renzo: double counting daily fee buckets

Shaileshkhote:
## Summary
- Fixes Renzo fees aggregation to avoid counting the next day’s midnight subgraph bucket in the prior day.
- Keeps `pullHourly` enabled while making Renzo’s daily bucket queries use exclusive end timestamps.

## Changes
- Updated `fees/renzo/common.ts` to use `timestamp_lt: $end` instead of `timestamp_lte: $end` across Renzo fee and earnings subgraph queries.
- Prevents the final hourly slice of a UTC day from including the next day’s `00:00` daily bucket.

### DIFF
--- a/fees/renzo/common.ts
+++ b/fees/renzo/common.ts
@@ -98,28 +98,28 @@ const ethFeesQuery = gql`
   query RenzoETHFeesQuery($start: Timestamp!, $end: Timestamp!) {
     stakingConsensusProtocolFeeStats(
       interval: day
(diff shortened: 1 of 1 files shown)


Shaileshkhote:
@noateden can you please check the server issue i raised about renzo?
renzo needs proper backfill from march 7 2026 since the pullyHourly was updated last

PR #6927, opened 2026-05-12
[feat] ready-cards - update revenue methodology to include web2 spin credits

hattonjey-256:
Follow-up to #6807.

This update keeps Ready Cards' existing TCG-style methodology and only adds Stripe-funded spin-credit rips into the existing Pack Purchases line:

- Pack Purchases now include normal on-chain pack payments plus `READY|WEB2` memo USD value
- Card Buybacks are limited to outbound USDT from the Ready treasury
- Fees / Revenue remain purchases minus buybacks, consistent with the current Ready Cards adapter
- Spin-credit revenue is included in Pack Purchases, not shown as a separate public line
- `READY|WEB2` memos are scoped to transactions touching the Ready treasury before being counted

### DIFF
--- a/fees/ready-cards.ts
+++ b/fees/ready-cards.ts
@@ -1,42 +1,43 @@
 import { Dependencies, FetchOptions, SimpleAdapter } from "../adapters/types";
 import { CHAIN } from "../helpers/chains";
 import { queryAllium } from "../helpers/allium";
-import { METRIC } from "../helpers/metrics";
 
 const READY_CARDS_TREASURY = "bvT9KFrAqmRpnb6AsuaJzdVKEVuT5jAVYt3N5CyGvkV";
-
-const READY_MINT = "HKJHsYJHMVK5VRyHHk5GhvzY9tBAAtPvDkZfDH6RLDTd";
+const MEMO_PROGRAM = "MemoSq4gqABAXKb96qnH8TysNcWxMyWCqXgDLGmfcHr";
 
 const PAYMENT_MINTS = [
-    "So11111111111111111111111111111111111111112", // SOL / wSOL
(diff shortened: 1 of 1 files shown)


noateden (changes requested):
thanks for the PR. can you update to count web2_memos fees into `Pack Purchases` only and don't update others? for ready-cards, we want to count fees/revenue as `purchases - buybacks`, same as others TCG protocols

hattonjey-256:
> thanks for the PR. can you update to count web2_memos fees into `Pack Purchases` only and don't update others? for ready-cards, we want to count fees/revenue as `purchases - buybacks`, same as others TCG protocols

thanks for the review - you guys are always great to work with 

we reverted the broader income-statement restructure and kept the existing Ready Cards / TCG-style methodology: fees and revenue remain purchases - buybacks.

This commit only adds READY|WEB2 memo USD values into the existing Pack Purchases bucket. It also keeps Card Buybacks limited to outbound USDT from the Ready treasury as before, and requires WEB2 memo transactions to be tied to Ready treasury activity before counting them as revenue 

QUESTION FOR METHODOLOGY:

Would it be possible to avoid showing a separate Earnings category here? Additionally, looking at other TCG-style protocols, it seems cleaner for Ready Cards to show gross pack-rip sales as Gross Protocol Revenue, then show Gross Profit as the delta between Pack Purchases and Card Buybacks.

Our concern is that revenue is normally understood as incoming funds, not the difference after buybacks. So the preferred display would be:

- Gross Protocol Revenue = gross Pack Purchases / pack-rip sales
- Card Buybacks = outbound USDT buybacks
- Gross Profit = Pack Purchases - Card Buybacks

noateden (approved):

PR #6933, opened 2026-05-13
Track Lorenzo sUSD1 fees and revenue

uwezukwechibuzor:
https://defillama.com/protocol/lorenzo-susd1+
https://lorenzo-protocol.gitbook.io/docs

### DIFF
--- a/fees/lorenzo-susd1/index.ts
+++ b/fees/lorenzo-susd1/index.ts
@@ -37,6 +37,7 @@ const fetch = async (options: FetchOptions) => {
 
   return {
     dailyFees,
+    dailyRevenue: 0,
     dailySupplySideRevenue,
   };
 };
--- a/fees/lorenzo-susd1/index.ts
+++ b/fees/lorenzo-susd1/index.ts
@@ -1,0 +1,72 @@
+import { FetchOptions, SimpleAdapter } from "../../adapters/types";
+import { CHAIN } from "../../helpers/chains";
+import { METRIC } from "../../helpers/metrics";
+
+// https://lorenzo-protocol.gitbook.io/docs
+// https://medium.com/@lorenzoprotocol/usd1-mainnet-launch-72550abac2ed
+const vaults: Record<string, string> = {
+  [CHAIN.BSC]: "0x4f2760b32720f013e900dc92f65480137391199b",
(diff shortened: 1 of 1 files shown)


noateden (changes requested):
please add `dailyRevenue`, set it even it's zero

noateden:
hi @uwezukwechibuzor, do you know how much were cut by protocol? You'are setting revenue = 0 but the methodology says otherwise
```
Net yield to sUSD1+ holders via NAV appreciation, after off-chain protocol and execution fees.
```

uwezukwechibuzor:
> hi @uwezukwechibuzor, do you know how much were cut by protocol? You'are setting revenue = 0 but the methodology says otherwise
> 
> ```
> Net yield to sUSD1+ holders via NAV appreciation, after off-chain protocol and execution fees.
> ```

@noateden  Yeah you're right.
So Lorenzo takes their fees off-chain before the unitNAV gets updated,  meaning what we see on-chain is already the net yield that goes to holders. There's no way to get the gross yield or fee rate from the contract, so we can't calculate what the protocol actually keeps, that's why dailyRevenue is 0. Updated the methodology to explain this properly and removed the confusing wording.  Let me know if this looks good or if there's a better way to handle it. Thank you!

noateden:
> > hi @uwezukwechibuzor, do you know how much were cut by protocol? You'are setting revenue = 0 but the methodology says otherwise
> > ```
> > Net yield to sUSD1+ holders via NAV appreciation, after off-chain protocol and execution fees.
> > ```
> 
> @noateden Yeah you're right. So Lorenzo takes their fees off-chain before the unitNAV gets updated, meaning what we see on-chain is already the net yield that goes to holders. There's no way to get the gross yield or fee rate from the contract, so we can't calculate what the protocol actually keeps, that's why dailyRevenue is 0. Updated the methodology to explain this properly and removed the confusing wording. Let me know if this looks good or if there's a better way to handle it. Thank you!

I will try to ask Lorenzo team

noateden:
hi @uwezukwechibuzor, Lorenzo takes 5% on positive yields, can you update?

uwezukwechibuzor:
> hi @uwezukwechibuzor, Lorenzo takes 5% on positive yields, can you update?

@noateden Updated! Added the 5% performance fee calculation. I was able to test this earlier, and it returned data, but now BSC public RPCs are returning errors for historical state queries. Can you take a look and let me know if everything looks good?

noateden (approved):

PR #6951, opened 2026-05-13
fix(users): price dailyGasUsed via native gas token

treeoflife2:
## Summary
- `users/utils/countUsers.ts` was dividing the per-tx fee sum by `1e18` and the resulting native-token quantity was passed straight to `dailyGasUsed`, which the dimension framework treats as a USD value. Result: 1 unit of gas token counted as 1 USD — inflating chains where the gas token is worth < $1.
- Concrete impact on Polymarket (Polygon-only): MATIC ≈ $0.09–0.08, so reported `dailyGasUsed` was roughly **10× the real USD cost**. The same code path also *under*-reports for ETH-priced chains (1 ETH shown as $1).
- Fix keeps the sum in wei (cast to string via `TO_VARCHAR` to avoid JS `Number.MAX_SAFE_INTEGER` truncation — Allium stores `gas_price` / `receipt_gas_used` as `NUMBER(38,0)`, so products of a busy day blow past 2^53), then wraps it in `createBalances().addGasToken(...)` so the framework prices it via each chain's native gas token. This mirrors the pattern used in `fees/chainlink-vrf-v1.ts`.

## Validation
- Allium docs confirm units across Ethereum / Polygon / HyperEVM `raw.transactions`:
  - `gas_price`, `receipt_effective_gas_price` → wei per gas unit
  - `receipt_gas_used` → gas units
  - All stored as `NUMBER(38,0)`
- `gas_price * receipt_gas_used` → wei, exactly what `addGasToken` expects.

### DIFF
--- a/users/list.ts
+++ b/users/list.ts
@@ -38,16 +38,23 @@ function getProtocolActiveUsersAdapter(item: typeof routers[0]): Adapter {
     return parseUserResponse(data, item.chains);
   }
 
-  async function fetch(_: any, _1: any, { chain, preFetchedResults, }: FetchOptions) {
+  async function fetch(_: any, _1: any, { chain, preFetchedResults, createBalances }: FetchOptions) {
     if (chain === CHAIN.CHAIN_GLOBAL)
       return {
         dailyActiveUsers: preFetchedResults?.all.users
       }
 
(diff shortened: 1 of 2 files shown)


treeoflife2:
we will be handling it at the frontend level, as this is deprecated feature, so no point in changing and we will convert it to the usd value based on gas token price in UI to make it accurate.

PR #6957, opened 2026-05-13
fix: typo, 'styart' should be 'start' in somnex-perps adapter

gultekinmakif:
## Summary

fixed a typo in the somnex-perps adapter so the start date is recognised.


bheluga:
@gultekinmakif thanks for the PR

PR #6970, opened 2026-05-14
fix(solidly-v3): correct invalid Fantom start date

DaniiRix:
Fix: Corrected an invalid start date for the Fantom chain in dexs/solidly-v3/index.ts. The value '2023-25-12' used month 25, which does not exist. Corrected to '2023-12-25' (December 25, 2023).


bheluga:
@DaniiRix thanks for the PR

PR #6980, opened 2026-05-15
remove duplicate fee additions

DaniiRix:
- w-dex was counting `protocolFees` and `infraFees` twice, first through `addBalances()`, then again through `.add()`, which already routes back to `addBalances()`
- This inflated live metrics: `dailyFees` by ~15% and doubled both `dailyRevenue` and `dailyProtocolRevenue`. The revenue/fees ratio showed ~23.5% instead of the expected 13.5%.
- Fixed by consolidating each destination into a single `addBalances()` call. `InfraFee` still remains under `dailySupplySideRevenue`


bheluga:
@DaniiRix thanks for the PR

PR #6981, opened 2026-05-15
fix: update subgraph endpoints

DaniiRix:
- Maverick v2: updated scroll endpoint
- Swaap v2: fixed subgraph url for mantle that was previously pointing to linea


bheluga:
@DaniiRix thanks for the PR

PR #6995, opened 2026-05-15
fix(flashtrade): include Equity.1 pool in volume + fees adapters

FrostDev7:
The Equity.1 pool (NVDA, AMZN, TSLA, AAPL, AMD, SPY perps) is missing from the dex and fees adapter pool lists.


bheluga:
@FrostDev7 thanks for the PR

PR #7008, opened 2026-05-16
fix: correct protocol/holders fee split and replace dead fallback

DaniiRix:
* Replace hardcoded fee split with dynamic toGaugeRate from FeesVaultFactory. Adapter now reports:

  * `dailyProtocolRevenue = fees × (toGaugeRate / 10_000)
  * `dailyHoldersRevenue = fees × (1 − toGaugeRate / 10_000)
  * `dailySupplySideRevenue = 0` (LPs earn no swap fees; incentives come from NEST emissions)

* Removed broken fallback path (`fetch24hMetricsFromV3Pools`):

  * `api.nest.aegas.it/graph/v3-pools` returns HTTP 404
  * `dayData.date * 86400` was incorrect since `poolDayData.date` is already a Unix timestamp in seconds, causing all snapshots to be filtered out
  * Endpoint was V3-only despite “all pools” methodology
  * Uses `/pools/liquidity-stats` on the Blaze host with the same `from/to` params

### DIFF
--- a/dexs/nest/index.ts
+++ b/dexs/nest/index.ts
@@ -183,7 +183,7 @@ const makeReturn = (volume: number, fees: number, gaugeRate: number) => {
     dailyFees: dailyFees.toString(),
     dailyRevenue: dailyFees.toString(),
     dailyProtocolRevenue: (dailyFees * (1 - gaugeRate)).toString(),
-    dailySupplySideRevenue: 0,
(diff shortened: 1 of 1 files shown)


noateden (changes requested):
thanks for the PR, can you dynamic fetch `toGaugeRate` from contract?

noateden (approved):

PR #7021, opened 2026-05-16
fix: dragonswap-sei v1 & v3

DaniiRix:
dragonswap: fixed missing revenue fields and metric labels in both `dragonswap-sei` and `dragonswap-sei-v3`.

* `dailySupplySideRevenue` was computed but not returned
* `dailyProtocolRevenue` was missing entirely
* Added missing methodology docs for V3


bheluga:
@DaniiRix thanks for the PR

PR #7024, opened 2026-05-16
Add PrimeFi fees adapter to aave factory and add permitFailure to hel…

uwezukwechibuzor:
## Summary    
  - Adds PrimeFi (Aave V2 fork on Base, Hyperliquid, XDC) to the aave factory registry                        
  - Adds `permitFailure: true` and null checks to aave helper multicalls to handle reserves that may not be
  available at queried block heights 
https://defillama.com/protocol/primefi


bheluga:

PR #7043, opened 2026-05-18
[fix] Beezie: undefined adddress bug (onlyArgs: false)

Shaileshkhote:
## Results with only args true 
Screenshot 2026-05-18 at 1 34 08 PM

## Result with only args false 
Screenshot 2026-05-18 at 1 35 43 PM

@noateden keeping onlyArgs true will undercount the fees as the addresses will be undefined hence the loop skips the claw machines


noateden:
ah yeah, my bad, thanks for the fix

PR #7029, opened 2026-05-17
fix(pika-protocol-v4): correct fix split

DaniiRix:
- Fixes Pika v4 fee/revenue breakdow: 50% to LPs, 30% to stakers, 20% to protocol treasury
- Adds `dailySupplySideRevenue` and `dailyHoldersRevenue`
- Adds methodology descriptions

PR #7046, opened 2026-05-18
fix(pika-protocol-v4): correct fee split to 50/50 vault/protocol

DaniiRix:
Follows up on #7029. The previously merged split (50/30/20) was based on the original docs, but the live contract has pikaRewardRatio=0 and protocolRewardRatio=5000 — PIKA staking was disabled when the token was retired. https://docs.pikaprotocol.com/pika-token

Actual split: 50% vault (LPs) / 50% protocol treasury.

Verified via:

cast call on 0x8c9b6a4a4e61f4635e8e375e05ff98db5516d25e (Optimism)
_updatePendingRewards in contract source: https://sourcify.dev/server/repository/contracts/full_match/10/0x8c9b6a4a4e61f4635e8e375e05ff98db5516d25e/sources/contracts/perp/PikaPerpV4.sol

### DIFF
--- a/dexs/pika-protocol-v4/index.ts
+++ b/dexs/pika-protocol-v4/index.ts
@@ -29,27 +29,25 @@ const fetch = async (options: FetchOptions) => {
     dailyFees.addUSDValue(Number(log.fee) / 1e8);
   }
 
-  // Fee split sources:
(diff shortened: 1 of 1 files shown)


bheluga:
@DaniiRix thanks for the PR

PR #7051, opened 2026-05-18
[fix] Aegis-yusd: allow negative value for erc4626 vault

Shaileshkhote:
@bheluga Added allow negative value for yusd too as its a erc4626 vault and can accure negative yields.


bheluga:

Shaileshkhote:
> @Shaileshkhote thanks for the PR

Why BSC keeps throwing errors ?
Is it due to archival node not having old contract state?

bheluga:
> > @Shaileshkhote thanks for the PR
> 
> Why BSC keeps throwing errors ? Is it due to archival node not having old contract state?

https://chainlist.org/chain/56 ig most of their RPC's are unstable

PR #7060, opened 2026-05-19
Track UltraYield Vaults fees and revenue

uwezukwechibuzor:
https://defillama.com/protocol/ultrayield-vaults


bheluga:

PR #7061, opened 2026-05-19
feat(dexr): add HyperEVM support

aum2024:


bheluga:
@aum2024 thanks for the PR

PR #7064, opened 2026-05-19
fix: add deadFrom to kiloex on manta

DaniiRix:


bheluga:
@DaniiRix thanks for the PR

PR #7069, opened 2026-05-19
fix: mira-ly graph endpoint

DaniiRix:
v2 indexer is unreachable, replaced it with v3


bheluga:
@DaniiRix thanks for the PR

PR #7062, opened 2026-05-19
trojan: net cashback / referral payouts from revenue

tarun-khatri:
Closes part of #6739 — the Trojan side. Axiom (#6789, merged) and Padre (#6997) have already landed the same pattern for their bots; this PR brings Trojan in line.

## Summary

The Trojan bot fee wallet (`9yMwSPk9mrXSN7yDHUuZurAh1sjbJsfpUqjZ7SvVtdco`) plays a dual role:
- **Inflow side** — collects per-trade fees on every swap routed through the Telegram bot.
- **Outflow side** — twice a day (`~00:02` and `~12:02 UTC`) it signs `SystemProgram` transfer batches that pay cashback + referral rewards out to user trading wallets.

The current adapter only counts the inflow side and treats every lamport the bot wallet receives as protocol revenue. That over-states `dailyRevenue` and leaves `dailySupplySideRevenue` empty, even though the cashback program is documented at [docs.trojanonsolana.com](https://docs.trojanonsolana.com/) (up to 35% direct referral revenue share, up to 47.5% at higher Honors ranks, plus 20–45% cashback on trading fees, min `0.005 SOL` per claim, twice a day).

This PR:
1. Adds the cashback / referral outflow to `dailySupplySideRevenue`.
2. Subtracts it from `dailyRevenue`, split between the Bot and Terminal products in proportion to gross-fee share.
3. Excludes the periodic internal treasury sweeps from the cashback figure — those are protocol-internal capital moves, not user distributions.
4. Switches the dune spell dependency from `ALLIUM` to `DUNE` and tightens the fee query so only DEX-trade fees are counted (joins `solana.account_activity` to `dex_solana.trades`).

## How the classification works

The bot wallet's outflows are a mix of two distinct things, and they have to be told apart before either can be billed correctly:

| Tx type | Recipients per tx | Sweep dest present? | Where the SOL goes |
|---|---|---|---|
| **Cashback / referral batch** | 4–15 user trading wallets | No | individual users |
| **Treasury sweep** | exactly 1 | Yes (`DrXMnP...`) | Squads multisig → wSOL treasury |

The SQL adds a `treasurySweepTxs` CTE that enumerates all outflow `tx_id`s in which `DrXMnPrFSiHA4JKKrSktTbAFrfvQCixHKvD7zGHxkzJP` receives a positive `balance_change`. The `cashbackPayouts` CTE then sums every bot-wallet outflow whose `tx_id` is **not** in that set. Clean, no heuristics on amount or recipient count.

[...]

PR #7077, opened 2026-05-19
dedust: switch volume source from deprecated v3 GraphQL to v4 REST

tarun-khatri:
## Summary

`dexs/dedust/index.ts` has been reporting **$0 daily volume since ~2026-04-21** because DeDust deprecated their v3 GraphQL endpoint. This PR switches to the v4 REST endpoint that `fees/dedust/index.ts` is already using.

## Bug verification

**1. Production state (DefiLlama API):** `/overview/dexs/ton` shows DeDust 24h/7d/30d volume = $0. Pre-cliff (April 19-20) the adapter reported $200-350k/day consistently.

**2. The v3 GraphQL endpoint is empty.** Probing the exact query the adapter sends:
```
$ curl -sX POST https://api.dedust.io/v3/graphql \
    -d '{"query":"query GetPools { pools(filter: {}) { address volume } }"}'
{"data":{"pools":[]}}
```
The same for `assets`. Schema introspection confirms `pools(filter: PoolsFiltersInput) -> [Pool]` still exists and the `Pool` type still has a `volume` field — the schema looks alive, but no rows are returned regardless of filter shape (tried `{}`, `{ assets: { in: [\"native\"] } }`, etc.).

**3. The chain is healthy.** TonAPI confirms live JettonSwap events on the canonical TON/USDT pool `EQA-X_yo3fzzbDbJ_0bzFWKqtRuZFIRa1sJsveZJ1YpViO3r` in the last minute, and the pool has live reserves (191,403 TON / 380,387 USDT). GeckoTerminal independently shows the top-10 DeDust pools doing roughly $865k+/day. The protocol is active; only the adapter's data source was stale.

## The fix

DeDust's canonical source is `mainnet.api.dedust.io/v4/api/get_pools` — the same endpoint already used by [`fees/dedust/index.ts`](https://github.com/DefiLlama/dimension-adapters/blob/master/fees/dedust/index.ts). It returns per-pool rows with a pre-computed `volume_24h_usd` field, so no asset-price lookup is needed.

```ts
const apiResponse = await httpPost(DEDUST_API, {
  offset, limit: 100,
  sort_by: \"volume_24h\",
  sort_direction: \"desc\",
  filter_by_type: [\"cpmm_v2\", \"stable\", \"cpmm_v1\"],
});
for (const poolRow of apiResponse.pool_rows) {
  dailyVolume += Number(poolRow.volume_24h_usd ?? 0);
}
```

Pagination mirrors the fees adapter exactly — sort by volume descending, stop when the tail falls below $1, otherwise paginate by 100 with a 3s sleep.

## Test plan

- [x] `pnpm run ts-check` clean
- [x] Local adapter run for 2026-05-18 → 2026-05-19: **Daily volume: 1.37 M** (production currently reports $0)
- [x] On-chain cross-check via TonAPI: TON/USDT pool emitting JettonSwap actions in the last minute
- [x] Third-party cross-check via GeckoTerminal: top-10 DeDust pools ~$865k/day, consistent with our $1.37M total across 1,838 pools

## Diff stats

`+34 / -63` — net smaller because the v4 endpoint returns pre-USD volume so we no longer need the GraphQL asset fetch + manual `price * raw_volume / 10^decimals` math.

## Rule-out: protocol-team fix

DeDust's API breakage is upstream of the adapter. The fix here uses the alternative v4 endpoint DeDust themselves recommend (and which their fees adapter already uses), so this doesn't depend on the protocol team patching v3.

### DIFF
--- a/dexs/dedust/index.ts
+++ b/dexs/dedust/index.ts
@@ -14,7 +14,6 @@ const fetch = async (options: FetchOptions) => {
   let dailyVolume = 0;
   let offset = 0;
 
-  try {
     while (true) {
       const apiResponse = await httpPost(DEDUST_API, {
         offset,
@@ -43,10 +42,6 @@ const fetch = async (options: FetchOptions) => {
       if (offset >= totalPools) break;
       await sleep(3000);
     }
-  } catch (error) {
-    console.error(`dedust dailyVolume fetch failed (offset=${offset}):`, error);
-    return { dailyVolume: 0 };
-  }
 
   return {
     dailyVolume,
--- a/dexs/dedust/index.ts
+++ b/dexs/dedust/index.ts
@@ -1,85 +1,64 @@
-import { postURL } from "../../utils/fetchURL"
 import { CHAIN } from "../../helpers/chains";
 import { FetchOptions, SimpleAdapter } from "../../adapters/types";
+import { httpPost } from "../../utils/fetchURL";
+import { sleep } from "../../utils/utils";
 
-const GRAPHQL_ENDPOINT = 'https://api.dedust.io/v3/graphql';
-
-const POOLS_QUERY = `
-query GetPools($filter: PoolsFiltersInput) {
-    pools(filter: $filter) {
-      address
-      totalSupply
-      type
-      tradeFee
-      assets
-      reserves
-      fees
-      volume
-    }
-  }
-`;
-
-const ASSETS_QUERY = `
-query GetAssets {
-    assets {
-      type
-      address
-      symbol
-      decimals
-      price
-    }
-  }
-`;
-
+// DeDust deprecated the v3 GraphQL `pools { volume }` schema; the endpoint
+// still responds but returns an empty `pools` array regardless of filter,
+// so this adapter was reporting $0 volume since ~2026-04-21. The v4 REST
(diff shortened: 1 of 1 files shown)


tarun-khatri:
Fixed two things in the latest push:

**1. Dropped an unrelated commit that leaked in by accident.** The PR was opened from a branch that had been forked off `fix-trojan-cashback-tracking` (my #7062) instead of `master`, so it picked up the Trojan commit. The DeDust branch has now been rebased onto current `upstream/master` and contains only the one `dexs/dedust/index.ts` change. Apologies for the noise.

**2. Applied both CodeRabbit suggestions:**

- Wrapped the entire paginate-and-fetch loop in `try/catch`. Recoverable failures now log via `console.error` with the current `offset` and return `{ dailyVolume: 0 }`, matching the team's "never swallow errors silently" guideline.
- Added shape validation on the API response: `Array.isArray(apiResponse?.pool_rows)` guards the iteration, `Number.isFinite(rowVolume)` skips any non-numeric `volume_24h_usd`, and `total_count` is parsed via `Number(... ?? 0)` with a fall-back exit if it's zero or non-finite.

Local test on the new diff: `Daily volume: 1.38 M` (~24h window 2026-05-18 → 2026-05-19). Production still showing \$0.

@coderabbitai review

bheluga:

PR #7078, opened 2026-05-19
update sundaeswap volume adapter

whankinsiv:
This PR updates the endpoint used to fetch SundaeSwap daily volume.


bheluga (commented):
@whankinsiv thanks for the PR
This just gives last 24h volume data right? Cant we get historic volumes?

bheluga:
@whankinsiv thanks for the PR

PR #7079, opened 2026-05-19
Change dailyVolume to use volumeQuote instead of volume

eugene-chen:
USD volume is in `volumeQuote` field, not `volume` field


bheluga:
@eugene-chen thanks for the PR

PR #7099, opened 2026-05-20
[fix] ElfomoFi fees and revenue

olegggatttor:
We've just launched ElfomoFi vaults and we log fee data for tracking. This PR adds correct calculation of fees and revenue based on the irregular VaultMetrics event (but at least once a week). The previous version seems to have been showing only the last event's data but I am not sure why, noticied it on the UI. 

Could you help verify that this one works correctly now and picks up all the events for the time period, evenly splitting accrued fees and revenue across the period? Thanks!

cc @bheluga

### DIFF
--- a/dexs/elfomofi/index.ts
+++ b/dexs/elfomofi/index.ts
@@ -1,23 +1,24 @@
-import { Dependencies, FetchOptions, FetchResult, SimpleAdapter } from "../../adapters/types";
+import { FetchOptions, FetchResult, SimpleAdapter } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
 import { METRIC } from "../../helpers/metrics";
-import { queryDuneResult } from "../../helpers/dune";
 
 
 const SwapEvent = "event ElfomoTrade(uint256 indexed quoteId, uint256 indexed partnerId, address executor, address receiver, address fromToken, address toToken, uint256 fromAmount, uint256 toAmount)";
-const ELFOMOFI_SWAP_ADDRESS = "0xf0f0F0F0FB0d738452EfD03A28e8be14C76d5f73"
+const VaultMetricsEvent = "event VaultMetrics(uint256 indexed vaultId, address indexed refTokenAddress, uint256 prevRate, uint256 newRate, uint256 newPreFeeRate, uint256 newHighWaterMarkRate, uint256 prevTimestamp, uint256 newTimestamp, int256 lastPeriodPreFeeAPRInBps, uint256 feeProtocolInRefToken, uint256 feeCuratorInRefToken, uint256 yieldToLPsInRefToken, uint256 protocolSharesMinted, uint256 curatorSharesMinted, uint256 totalAssetsInRefToken)";
+const ELFOMOFI_SWAP_ADDRESS = "0xf0f0F0F0FB0d738452EfD03A28e8be14C76d5f73";
+const ELFOMOFI_VAULTS_MANAGER = "0xE34CD3682AF9C04303386499FBa215B38Eff6106";
 
-// Dune query that decodes the VaultMetrics events and spreads each report's fees/yield evenly across
-// its accrual period, producing one row per (day, ref_token) with amortized amounts in token base
-// units. https://dune.com/queries/7560737
-const DUNE_QUERY_ID = "7560737"
+const DAY = 86400;
+const DISTRIBUTE_DAYS = 5;
+const DISTRIBUTE_SECONDS = DISTRIBUTE_DAYS * DAY;
+const LOOKBACK_SECONDS = 15 * DAY;
 
-interface IDuneRow {
-    day: string;
-    ref_token: string;
-    fee_protocol: string | number;
-    fee_curator: string | number;
-    yield_lps: string | number;
+interface ISettlement {
+    t: number;
+    token: string;
+    feeProtocol: bigint;
+    feeCurator: bigint;
+    yieldLps: bigint;
 }
 
 const fetch = async (options: FetchOptions): Promise<FetchResult> => {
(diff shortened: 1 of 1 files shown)


bheluga:
I have refilled fees and it looks like 
image
is this what you wanted or you want to split evenly across days

olegggatttor:
Yeah I was trying to split evenly across days. How do I do it? @bheluga

olegggatttor:
Oh and the data seems to be synced now on the UI so all the events are captured. But still wondering if it is possible to split fees and revenue evenly over the time range

olegggatttor:
Is it possible to do it like this? @bheluga

bheluga:
> Is it possible to do it like this? @bheluga

yes it is possible, i ll look into this after some time.
do you think it would be much cleaner to track yields from getRate functions? fromApi and toApi (yield is the diff * supply) so that you dont need to do extra calculations?
image

olegggatttor:
> > Is it possible to do it like this? @bheluga
> 
> yes it is possible, i ll look into this after some time. do you think it would be much cleaner to track yields from getRate functions? fromApi and toApi (yield is the diff * supply) so that you dont need to do extra calculations? image

I think it is better to fetch from events as we log all the required data including timestamps and the actual protocol fee/curator fee/LPs fee split so it really helps

olegggatttor:
> > > Is it possible to do it like this? @bheluga
> > 
> > 
> > yes it is possible, i ll look into this after some time. do you think it would be much cleaner to track yields from getRate functions? fromApi and toApi (yield is the diff * supply) so that you dont need to do extra calculations? image
> 
> I think it is better to fetch from events as we log all the required data including timestamps and the actual protocol fee/curator fee/LPs fee split so it really helps

And thanks! Looking forward to it

dexs/elfomofi/index.ts
```diff
@@ -29,21 +30,45 @@ const fetch = async (options: FetchOptions): Promise<FetchResult> => {
...
+    const toBlock = await options.getToBlock();
+    const blockTime = (options.toTimestamp - options.fromTimestamp) / (toBlock - fromBlock);
+    const now = Math.floor(Date.now() / 1000);
+    const effectiveLookahead = Math.max(0, Math.min(VAULT_METRICS_LOOKAHEAD_SECONDS, now - options.toTimestamp));
+    const lookaheadBlocks = Math.ceil(effectiveLookahead / blockTime);
+
     const vaultLogs = await options.getLogs({
         target: ELFOMOFI_VAULTS_MANAGER,
         eventAbi: VaultMetricsEvent,
+        toBlock: toBlock + lookaheadBlocks,
```
bheluga: lookahead wont work, as we wont be running historical dates repeatedly.
either you can use fromBlock: lookbehind or you can use getVaultSnapshot function that seems to have all data including rate change, curator fee, protocol fee, supply etc, only thing which is missing imo is list of all vaults, which should still be possible from some other methods.
I prefer the 2nd approach, as its clean

olegggatttor: @bheluga and how do we split the fees evenly through the period? Maybe there is an example

bheluga: ok looks like the price function doesnt update in streamed manner, so it wont help.
Either you can look back the same way which you were looking ahead  or you can leave it as it is , fees on uneven days are still fine

olegggatttor: Yeah I think we won't look ahead, so stick with the current impl basically. But we still want to spread fees/revenue evenly between the days, is it possible?

olegggatttor: Or what do you mean by looking back? I think we look back now anyway no?

bheluga: hmm, i thought we could do something , but i cant really think for any solution rn, i dont think any other adapter splits the numbers.
probably its good to leave as it is ? anyway cumulative numbers are correct and things might get even after introduction of few more vaults?

olegggatttor:
@bheluga hey, I've added a Dune query that splits fees and revenue evenly, and now extracting data from it here. Seems to be working just fine

bheluga:
@olegggatttor thanks for the effort, but i believe this still has the same problem that i mentioned before
To explain: we run adapter for particular timeframe only.
dune has latest date as 19th may, because that was the latest yield date
say if next yield is distributed on 26th may, we get new entries 20th - 26th may, but on 26th may when the adapter is run, it only considers 26th's yield because it as no ability to modify previous day's data. 
So we would have yield split data on 19th , 0s from 20th-25th, and again yield on 26th

olegggatttor:
> @olegggatttor thanks for the effort, but i believe this still has the same problem that i mentioned before
> 
> To explain: we run adapter for particular timeframe only.
> 
> dune has latest date as 19th may, because that was the latest yield date
> 
> say if next yield is distributed on 26th may, we get new entries 20th - 26th may, but on 26th may when the adapter is run, it only considers 26th's yield because it as no ability to modify previous day's data. 
> 
> So we would have yield split data on 19th , 0s from 20th-25th, and again yield on 26th

Is it possible to just pull data from dune for the whole chart? Not like recalculating for the timeframe, but getting the entire chart result

olegggatttor:
> > @olegggatttor thanks for the effort, but i believe this still has the same problem that i mentioned before
> > 
> > To explain: we run adapter for particular timeframe only.
> > 
> > dune has latest date as 19th may, because that was the latest yield date
> > 
> > say if next yield is distributed on 26th may, we get new entries 20th - 26th may, but on 26th may when the adapter is run, it only considers 26th's yield because it as no ability to modify previous day's data. 
> > 
> > So we would have yield split data on 19th , 0s from 20th-25th, and again yield on 26th
> 
> Is it possible to just pull data from dune for the whole chart? Not like recalculating for the timeframe, but getting the entire chart result

Fees are updated rarely so there is no need to execute actual dune request. There won't be any overhead here for just pulling the data to plot

bheluga:
> > @olegggatttor thanks for the effort, but i believe this still has the same problem that i mentioned before
> > To explain: we run adapter for particular timeframe only.
> > dune has latest date as 19th may, because that was the latest yield date
> > say if next yield is distributed on 26th may, we get new entries 20th - 26th may, but on 26th may when the adapter is run, it only considers 26th's yield because it as no ability to modify previous day's data.
> > So we would have yield split data on 19th , 0s from 20th-25th, and again yield on 26th
> 
> Is it possible to just pull data from dune for the whole chart? Not like recalculating for the timeframe, but getting the entire chart result

the thing is we run adapter once every day and store that day's data in db and display on ui
we sometimes do backfills if data is missing, but in your case it needs to be done everytime, which wont be feasible

olegggatttor:
@bheluga I got you, thanks for pointing that out. I switched from backward amortization to forward distribution: no need for backfilling now, just distributing forward till the next event evenly. We will settle epochs every 3-4 days and the remainder will land fully just before the new event

PR #7100, opened 2026-05-21
Track Lynx volume, fees and revenue

uwezukwechibuzor:
https://defillama.com/protocol/lynx


bheluga:

PR #7110, opened 2026-05-21
fix(ekubo-evm): migrate endpoint and support historical date queries

DaniiRix:
- Migrates from the deprecated eth-mainnet-api.ekubo.org endpoint  to prod-api.ekubo.org/overview/volume
- Switches from volumeByToken_24h (always-current snapshot) to volumeByTokenByDate filtered by options.dateString, enabling proper historical backfill
- Adds methodology fields


bheluga:
@DaniiRix thanks for the PR

PR #7111, opened 2026-05-21
fix: aave v3 chain config

DaniiRix:
Corrected mantle chain config


bheluga:
@DaniiRix thanks for the PR

PR #7112, opened 2026-05-21
Add DEX factories for Aerodrome-compatible V2/V3 pools

jlws:
## Overview

This PR introduces factory implementations for Aerodrome-compatible V2/V3 pools and refactors the existing integration approach for Aerodrome-based protocols. Adapters were implemented on top of a reusable helper layer following a modular architecture, where each function is responsible for extracting or computing a specific subset of pool data and metrics. This design improves reusability across similar integrations and serves as a foundation for predecessors such as Velodrome or Solidly, as well as other implementations that remain compatible with the same set of functions.

#### Factories return the following metrics:
- volume, calculated from swap events
- fees, calculated per swap using the fee rate derived from each pool
- swap fees distributed to holders, derived from staked LP positions in the gauge
- gauge emissions, calculated using rate-based accounting over the selected time period
- bribes, calculated from `NotifyReward` events and included in `holdersRevenue`

This approach improves consistency across integrations, reduces duplication, and simplifies future protocol onboarding.

## Changes

- Added factories for Aerodrome-compatible V2/V3 pools
- Added a reusable helper layer for extracting pool data and computing metrics
- Moved duplicated logic into shared factory-based implementations
- Standardized metric calculation across Aerodrome-compatible integrations
- Added gauge emissions calculation using rate-based accounting over the selected time period

## Why

The existing codebase only contained a Solidly-based [factory](https://github.com/DefiLlama/dimension-adapters/blob/master/helpers/solidly.ts) implementation, which did not fully cover all required metrics, and differed in design from Aerodrome-compatible integrations. As a result, each new Aerodrome-compatible protocol required re-implementing similar logic, leading to duplication and increased maintenance overhead. There were also gaps and inconsistencies in metric calculations across protocols, particularly around fees, gauge emissions, and bribe accounting. Introducing a dedicated factory layer for Aerodrome-compatible V2/V3 pools centralizes shared logic, enforces consistency in metric computation, and significantly reduces duplication when onboarding new protocols.

### Newly added pools
- Mezo Tigris V2 addc3cab5888aebd090a4e6b00bf0a92702ba2c0
- Mezo Tigris V3 24a9315bce354b7e05badfdc43415da1b823c188

### Migrated to factory
- Aerodrome 4e0ec19bd42c1b856b0fae768398d15476739cdc
- Aerodrome Slipstream a079097efd4fab40e2c7cac2c55c6f5f38828fb6
- Aborean 69938a7dd0a7f2489c4faa7c46d9be3aa4645da9
- Aborean CL 30a9bc234d1ed9123097331fef303100e623c9ff


treeoflife2:
@jlws thanks for the pr, but there is not much benefit as i see only 1 fork project for aerodrome, so let's keep it simpler, it only increase number of lines. if there is issue in logic then you can create PR, but let's skip the factory implementation for aero

PR #7115, opened 2026-05-21
Track pred volume

uwezukwechibuzor:
https://defillama.com/protocol/pred


bheluga:

PR #7124, opened 2026-05-22
feat: 🎸 add USDat tranche

0xnaman1:
Name: **Strata Markets**

Change: Added USDat tranches in fees adapter


bheluga:
@0xnaman1 thanks for the PR

PR #7136, opened 2026-05-22
add meteora protocol fee ratio

DaniiRix:
source: https://docs.meteora.ag/user-guide/guides/how-to-use-damm-v2#protocol-fee


bheluga (dismissed):
@DaniiRix thanks for the PR
This assumes that entire 20% goes to the protocol, but doesn't consider the fact that some part of that goes to referrals and $MET buybacks

DaniiRix:
> @DaniiRix thanks for the PR This assumes that entire 20% goes to the protocol, but doesn't consider the fact that some part of that goes to referrals and $MET buybacks

fair, thanks for it. for meteora v1 damm pools added stable vs other pools protocol share split and for v2, handled both the buybacks and the referral

PR #7142, opened 2026-05-23
Track 3F fees and revenue

uwezukwechibuzor:
### Summary

  - Add fees and revenue adapter for 3F Protocol
  - 3F is a vault protocol that provides leveraged exposure to tokenized RWAs on Ethereum
  - Tracks management fees (time-based on AUM), performance fees (on net gains), and supply side yield from
  PositionManager vaults

  Links

  - Docs: https://docs.3f.xyz/protocol/overview/
  - Website: https://www.3f.xyz/

### DIFF
--- a/fees/3f.ts
+++ b/fees/3f.ts
@@ -7,7 +7,7 @@ const FACTORY_DEPLOY_BLOCK = 24844184;
 
 const Abis = {
   PositionManagerCreated: "event PositionManagerCreated(address indexed positionManager, address indexed owner, address indexed collateralAsset, address debtAsset, uint256 ltv, address transferGuard)",
-  totalAssets: "int256:totalAssets",
+  totalAssets: "uint256:totalAssets",
   assets: "function assets() view returns (address collateralAsset, address debtAsset)",
   feeData: "function feeData() view returns (address feeRecipient, uint24 managementFee, uint24 performanceFee, uint256 lastTotalAssets, uint40 lastFeeAccrualTimestamp)",
 };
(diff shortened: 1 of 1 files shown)


fees/3f.ts
```diff
@@ -0,0 +1,140 @@
...
+      const gains = BigInt(totalAssetsAfter) - BigInt(totalAssetsBefore);
+
+      if (performanceFee > BigInt(0) && gains > managementFeeAmount) {
+        performanceFeeAmount = (gains - managementFeeAmount) * performanceFee / BPS;
+        dailyFees.add(debtAsset, performanceFeeAmount, METRIC.PERFORMANCE_FEES);
+        dailyRevenue.add(debtAsset, performanceFeeAmount, METRIC.PERFORMANCE_FEES);
+      }
+
+      // supply side = gains - management fee - performance fee (can be negative)
+      const supplySide = gains - managementFeeAmount - performanceFeeAmount;
```
noateden: why are you subtracting `managementFeeAmount` from gains here? are management fees additional fees that charged by protocol, not from yields/gains?

noateden (changes requested):
verify management fees

fees/3f.ts
uwezukwechibuzor: > why are you subtracting `managementFeeAmount` from gains here? are management fees additional fees that charged by protocol, not from yields/gains?

Thanks for the review! Yes, management fees are AUM-based and not from gains. The subtraction here is only for defining the performance fee base, it matches the contract's _pendingFees() which excludes managementFeeAssets from gains before calculating performance fees, to avoid charging performance fees on value already claimed as management fees.
Contract source: https://etherscan.io/address/0x8e0667429d1717b3e5fe783a6c472d6d901fe5fa#code

noateden (approved):

PR #7147, opened 2026-05-23
fix(nado-spot): correct fee accounting

DaniiRix:
fixes spot volume by removing the duplicate margined_products path that reused spot products and double-counted volume


bheluga:
@DaniiRix thanks for the PR

PR #7153, opened 2026-05-23
fix(convex): replace dead subgraph with on-chain CRV transfer tracking

ErebusCipher:
## Root Cause

The Convex fees adapter has reported $0 since The Graph's hosted service was deprecated. The adapter points to subgraph `86irRVuFotfaCFwtFxiSaJ76Y8pxfU3xX91gU6CoYTvi`, which now either returns "deployment does not exist" or requires a paid API key.

The silent failure mode: `graphFetch` calls `Object.keys(graphRes.dailyRevenueSnapshot)` where `dailyRevenueSnapshot` is `null` when the subgraph request fails — JavaScript throws `TypeError: Cannot convert undefined or null to object`, which the adapter catches and returns 0 for all fields.

## Fix

Replace the subgraph dependency entirely with on-chain data sources.

### 1. CRV fee revenue (new)

Convex's Booster distributes CRV to specific fee recipient contracts whenever `earmarkRewards` is called (~250 times/day). Using `addTokensReceived` with `fromAddressFilter = BOOSTER`:

| Recipient | Address | Role |
|---|---|---|
| `CVXCRV_STAKING` | `0x3Fe65692bfCD0e6CF84cB1E7d24108E434A7587e` | Receives `lockIncentive` % of CRV → cvxCRV stakers |
| `CVX_LOCKER_REWARDS` | `0xcf50b810e57ac33b91dcf525c6ddd9881b139332` | Receives `stakerIncentive` % of CRV → CVX lockers |

These addresses were verified by querying Booster→CRV Transfer events over a 7-day window. Both receive CRV daily with zero gaps.

### 2. Supply-side LP CRV (extrapolated)

The adapter reads `lockIncentive`, `stakerIncentive`, `earmarkIncentive`, and `platformFee` directly from the Booster contract. Since `CVXCRV_STAKING` always receives exactly `lockIncentive` basis points of total CRV:

```
LP CRV = CVXCRV_STAKING_CRV × (10_000 − totalProtocolFeeBps) ÷ lockIncentive
```

This avoids scanning the hundreds of individual pool reward contracts that the subgraph indexed.

### 3. Retained from previous adapter
- **reUSD revenue** — on-chain `RewardPaid` events on the registry staker contract (unchanged)
- **Votium bribes** — `llama.airforce` API (unchanged)

### Out of scope
FXS revenue is not tracked in this PR. Frax pool adoption has declined and the FXS fee routing changed; this is a separate investigation.

## Verification

Manually confirmed via Dune query on `convex_ethereum.booster_call_earmarkrewards` + `erc20_ethereum.evt_Transfer`:
- 212 earmarks on 2026-05-23 alone
- `CVXCRV_STAKING` appears as the single largest consistent CRV recipient from the Booster every day for the past 7 days, consistent with the 10% lockIncentive split

## Test plan
- [ ] Confirm `addTokensReceived` returns non-zero CRV balances for a recent date
- [ ] Confirm `lockIncentive()` call on Booster returns `1000` (10%)
- [ ] Verify daily fees reappear on DeFiLlama after merge

### DIFF
--- a/fees/convex.ts
+++ b/fees/convex.ts
@@ -191,34 +191,32 @@ const fetch = async (options: FetchOptions) => {
     supplySideCRV.add(CRV_TOKEN, lpCRVAmount, "CRV Revenue");
   }
 
-  // ── FXS revenue via Convex's Frax-side fee contracts ──────────────────────
-  // The original adapter tracked 5 FXS sub-flows via a Convex subgraph:
-  //   fxsRevenueToLpProviders, fxsRevenueToCvxFxsStakers,
-  //   fxsRevenueToCvxStakers, fxsRevenueToCallers, fxsRevenueToPlatform
-  // Without the subgraph's internal accounting we cannot split these cleanly,
-  // so we track the two on-chain recipient contracts directly:
-  //   STKFXS_STAKING   → stkCvxFxs staking rewards (cvxFXS stakers + CVX lockers)
-  //   FXS_FEE_DISTRIB  → fee distributor (LP providers + platform)
-  // Both flows are aggregated under 'FXS Revenue'. Post-2024 Frax gauge
-  // wind-down means these return ~$0 on recent dates, but the buckets are
-  // live and will capture any renewed FXS distribution automatically.
-  const fxsStakingRevenue  = createBalances();
-  const fxsDistribRevenue  = createBalances();
-
-  await Promise.all([
-    addTokensReceived({
-      token: FXS_TOKEN,
-      target: STKFXS_STAKING,
-      options,
-      balances: fxsStakingRevenue,
-    }),
-    addTokensReceived({
-      token: FXS_TOKEN,
+  // ── FXS revenue via on-chain distribution events ──────────────────────────
+  // Previous approach tracked inbound FXS transfers to these contracts, which
+  // misrepresents daily revenue (FXS arrives in batches, not per-distribution).
+  // noateden: use reward distribution/claim events instead.
+  //
+  //   FXS_FEE_DISTRIB → RewardDistributed(gauge_address, reward_amount)
+  //     FXS pushed out to each gauge → supply-side (LP providers)
+  //   STKFXS_STAKING  → RewardPaid(_user, _rewardsToken, _reward) filtered by FXS
+  //     FXS claimed by cvxFXS stakers / CVX lockers → holders revenue
+  const [fxsDistribLogs, fxsStakingLogs] = await Promise.all([
+    options.getLogs({
       target: FXS_FEE_DISTRIB,
-      options,
-      balances: fxsDistribRevenue,
+      eventAbi: "event RewardDistributed(address indexed gauge_address, uint256 reward_amount)",
+    }),
+    options.getLogs({
+      target: STKFXS_STAKING,
+      eventAbi: "event RewardPaid(address indexed _user, address indexed _rewardsToken, uint256 _reward)",
     }),
   ]);
+
+  const fxsDistribAmount = fxsDistribLogs.reduce(
+    (acc: bigint, log: any) => acc + BigInt(log.reward_amount), 0n
+  );
+  const fxsStakingAmount = fxsStakingLogs
+    .filter((log: any) => log._rewardsToken.toLowerCase() === FXS_TOKEN.toLowerCase())
+    .reduce((acc: bigint, log: any) => acc + BigInt(log._reward), 0n);
 
   // ── reUSD on-chain revenue (existing logic) ────────────────────────────────
   let reUSDRevenue = new BigNumber(0);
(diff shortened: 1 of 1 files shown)


ErebusCipher:
Quick status note for reviewers:

This PR is ready for review. The earlier failing bot run was from a prior commit where the LP CRV extrapolation read back from SDK balances and hit a `BigInt()` parse issue on scientific notation.

The current implementation avoids that path entirely: it fetches CRV `Transfer` logs directly with topic filters and sums `log.value` as raw `BigInt`, so the supply-side CRV derivation stays precision-safe end-to-end.

The latest visible adapter run after the main fix returned non-zero Convex output:
- dailyFees: 9.53k
- dailyRevenue: 9.53k
- dailyHoldersRevenue: 9.53k
- dailySupplySideRevenue: 54.52k

CodeRabbit appears rate-limited, so there has not been an automated review yet. Happy to adjust the accounting split if maintainers prefer a different classification.

noateden (changes requested):
thanks for the PR. it's missing `FXS Revenue` and `Others Revenue`, can you add?

ErebusCipher:
Thanks for the review. I've pushed an update for this.

FXS Revenue is now restored as a named breakdown bucket. The old subgraph exposed five FXS sub-flows (`fxsRevenueToLpProviders`, `fxsRevenueToCvxFxsStakers`, `fxsRevenueToCvxStakers`, `fxsRevenueToCallers`, `fxsRevenueToPlatform`). Since the subgraph is gone, this version tracks the on-chain FXS recipient contracts directly:
- `stkCvxFxs` rewards for staker/locker flows
- the FXS fee distributor for LP/platform-style flows

Recent FXS flows appear to be near-zero because Frax gauge emissions have mostly wound down, so the bucket may show ~$0 on current test dates, but the tracking path is live.

Others Revenue is also emitted as a named breakdown bucket. The old `otherRevenue` field was an opaque subgraph catch-all, so this replacement tracks the identifiable non-CRV components directly: Votium bribes via llama.airforce and reUSD locker rewards via `RewardPaid` events. I verified the label appears in the breakdown on a Votium epoch day.

ErebusCipher:
Addressed the CodeRabbit comments:

1. **Bribe API fallback** — wrapped `fetchBribesUSDForDay` in try/catch; returns 0 if llama.airforce is unavailable rather than crashing the adapter.

2. **Timestamp comment** — fixed: `1711152000` is 2024-03-23, not 2025.

3. **Missing breakdown labels** — added `"CRV Revenue"` and `"CVX Revenue"` labels to all `addBalances` calls so every bucket named in `breakdownMethodology` is mechanically matched by its emitted label path. Verified with `DEBUG_BREAKDOWN_FEES` — CRV Revenue now shows across Fees, Revenue, HoldersRevenue, and SupplySideRevenue columns.

4. **FXS double-count** — removed `fxsDistribRevenue` from `dailyProtocolRevenue`. The FXS fee distributor's primary purpose is LP provider incentives, so the entire flow is classified as supply-side. Without the original subgraph's internal LP/platform/caller split this is the most defensible single classification.

5. **`dailyBribesRevenue`** — keeping this field. It is not deprecated: it appears in `adapters/types.ts` with no deprecation notice and is actively used in `fees/curve.ts`, `fees/penpie.ts`, `fees/lynex.ts`, and `fees/lens-protocol.ts`.

ErebusCipher:
Added the error log in the catch block — `console.error` before returning 0. That's the only new flag from the second CodeRabbit pass.

ErebusCipher:
One more follow-up from the second CodeRabbit pass:

- **Methodology wording** — `methodology.Fees` and `breakdownMethodology.Fees["Others Revenue"]` both referenced Votium bribes as part of `dailyFees`, but bribes are only emitted under `dailyRevenue` / `dailyProtocolRevenue` / `dailyBribesRevenue`. Bribes are external protocol income, not gross user fees. Updated both strings to reflect that.

noateden:
hi @eugene-chen I think you're counting FXS revenue wrong. can you get rewards were distributed or claimed by contract events?
https://etherscan.io/tx/0xef3fb8be2c132c96487dfe61b22287b408e845e8f99ecd820154ac457a97dc1d#eventlog
https://etherscan.io/tx/0xec2fdc41ab55167075d05e3d6c782b21c84c6db100beac6774af6d6895c8749f#eventlog

ErebusCipher:
Thanks, that makes sense. I updated the FXS path to use the distribution/claim events instead of inbound token transfers.

Specifically:
- `FXS_FEE_DISTRIB`: now reads `RewardDistributed(address indexed gauge_address, uint256 reward_amount)`
- `STKFXS_STAKING`: now reads `RewardPaid(address indexed _user, address indexed _rewardsToken, uint256 _reward)` filtered to FXS

This avoids counting batch transfers into the contracts and instead attributes FXS when it is distributed/claimed. I also tested a historical active FXS date (`2023-03-01`) with `DEBUG_BREAKDOWN_FEES=true`; the `FXS Revenue` bucket emits ~$30.4k, while recent dates remain near-zero as expected.

ErebusCipher:
Addressed the final CodeRabbit consistency nit by passing the explicit "CRV Revenue" label on the supply-side CRV addBalances call as well.

noateden (approved):

PR #7159, opened 2026-05-23
fix(yield-basis): correct fee computation

DaniiRix:
* Fixed `TokenExchange` event ABI: replaced incorrect `price_scale` field with `price_oracle`, verified against deployed LEVAMM Vyper source on Sourcify.
* Fixed fee calculation: `TokenExchange.fee` is the swap fee rate (`self.fee`, 1e18 precision), not the fee amount. Previous logic added the raw rate to balances, producing invalid values. Actual fee is now computed as: `tokens_bought × feeRate / (1e18 − feeRate)`.
* Fixed fee token attribution: swap fees are deducted from the output token and now correctly assigned to `coins[bought_id]` instead of `coins[sold_id]`.
* Removed redundant `fee` multicall: the fee rate already exists in each `TokenExchange` log, so fetching `self.fee` separately was unnecessary and incorrect for per-swap accounting.
* Re-enabled and returned `dailyFees`, `dailyRevenue`, and `dailySupplySideRevenue`; also added `dailyUserFees` and `dailyHoldersRevenue`.
* LEVAMM does not implement `admin_fee` (all swap fees go to LPs). The `permitFailure: true` fallback to `0n` is intentional, so `dailyRevenue` and `dailyHoldersRevenue` remain zero unless a protocol fee split is added later.

### DIFF
--- a/dexs/yield-basis.ts
+++ b/dexs/yield-basis.ts
@@ -1,13 +1,13 @@
 import { FetchOptions, SimpleAdapter } from "../adapters/types";
 import { CHAIN } from "../helpers/chains";
+import { METRIC } from "../helpers/metrics";
 
 const ABI = {
   market_count: "uint256:market_count",
-  fee: "uint256:fee",
   admin_fee: "uint256:admin_fee",
   coins: "function coins(uint256) view returns(address)",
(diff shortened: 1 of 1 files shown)


bheluga:
@DaniiRix thanks for the PR
We already track fees in separate file [here](https://github.com/DefiLlama/dimension-adapters/blob/master/fees/yield-basis/index.ts)

PR #7204, opened 2026-05-25
fees/lido: split protocol take into DAO treasury vs node operator commission

tarun-khatri:
## Summary

The Lido fees adapter lumps the full 10% protocol take into a single Revenue bucket, but the methodology field already promises a split: *\"Lido applies a 10% fee on staking rewards that are split between node operators and the DAO Treasury.\"* The breakdownMethodology doesn't currently honor that split. This PR surfaces the two components separately by reading the live aggregate ratio from \`StakingRouter.getStakingFeeAggregateDistributionE4Precision()\` at the window's end block.

Totals (Fees, Revenue, ProtocolRevenue, SupplySideRevenue) are unchanged. Only the breakdown granularity increases.

## On-chain verification (three independent angles)

1. **Source.** Lido StakingRouter implementation \`0x226f9265cbc37231882b7409658c18bb7738173a\` (proxy at \`0xFdDf38947aFB03C621C71b06C9C70bce73f12999\`) declares the function:
   \`\`\`solidity
   function getStakingFeeAggregateDistributionE4Precision()
     external view returns (uint16 modulesFee, uint16 treasuryFee)
   \`\`\`
   where both components are validator-share-weighted aggregates across active staking modules.

2. **Live values today.**
   \`\`\`
   modulesFee  = 387   (3.87% of total staking rewards → node operator commissions)
   treasuryFee = 612   (6.12% of total staking rewards → DAO treasury)
   sum         = 999   (~10%, matches the existing PROTOCOL_FEE_RATIO hardcoded constant)
   \`\`\`

3. **Historical rate-change events.** \`StakingModuleFeesSet\` emissions on the StakingRouter prove the split has changed via governance over time. Notable: Module 1 (Curated Node Operators) changed from \`stakingModuleFee=500, treasuryFee=500\` (5% / 5%) to \`stakingModuleFee=350, treasuryFee=650\` (3.5% / 6.5%) on **2025-12-24** (block 24083230, tx \`0x470e74a0a11adaa9f1c44f06a8b792d743fe36cbe8f69160f8bd6ad2461adf67\`). Sister modules (Simple DVT, Community Staking) hold 8/2 and 6/4 splits respectively. Reading the live aggregate ensures the adapter follows future governance changes without code edits.

## Diff shape

- 1 file changed, +39 / -13.
- New \`METRIC.OPERATORS_FEES\` entries in Revenue / ProtocolRevenue breakdowns.
- Existing \`METRIC.STAKING_REWARDS\` and \`METRIC.MEV_REWARDS\` Revenue entries are now the *treasury share only* (instead of the full 10%). Total Revenue is unchanged: \`stakingTreasury + mevTreasury + operatorsAcrossBoth = dailyProtocolRevenueUSD\`.

## Test plan

- [x] \`pnpm run ts-check\` clean.
- [x] On-chain ABI verified at the implementation contract via Sourcify.
- [x] Live values cross-checked against eth_call to the proxy.
- [x] Event history sourced from \`ethereum.logs\` filtered by \`StakingModuleFeesSet\` topic over the StakingRouter's lifetime (13 emissions across 3 modules between 2023-05 and 2026-04).

### DIFF
--- a/fees/lido.ts
+++ b/fees/lido.ts
@@ -82,15 +82,17 @@ const fetch = async (timestamp: number, _a: any, options: FetchOptions) => {
   dailyFees.addUSDValue(dailyTotalRevenueUSD - totalMevFees, METRIC.STAKING_REWARDS)
   dailyFees.addUSDValue(totalMevFees, METRIC.MEV_REWARDS)
 
-  dailySupplySideRevenue.addUSDValue(stakingSupplySide, METRIC.STAKING_REWARDS)
-  dailySupplySideRevenue.addUSDValue(supplySideMevFees, METRIC.MEV_REWARDS)
+  dailySupplySideRevenue.addUSDValue(stakingSupplySide, 'Staking rewards to ETH stakers')
+  dailySupplySideRevenue.addUSDValue(supplySideMevFees, 'MEV rewards to ETH stakers')
 
   // Split the protocol take into the DAO-treasury share (kept by the Lido DAO) and the
   // module-operator share (paid to node operators). Totals are preserved; only the breakdown
   // gains granularity. Operator and treasury shares are summed across staking + MEV sources.
   dailyRevenue.addUSDValue(stakingProtocolFees * treasuryShare, METRIC.STAKING_REWARDS)
   dailyRevenue.addUSDValue(protocolMevFees * treasuryShare, METRIC.MEV_REWARDS)
-  dailyRevenue.addUSDValue((stakingProtocolFees + protocolMevFees) * operatorShare, METRIC.OPERATORS_FEES)
+
+  dailySupplySideRevenue.addUSDValue(stakingProtocolFees * operatorShare, 'Staking rewards to node operators')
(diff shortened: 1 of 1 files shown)


tarun-khatri:
Pushed `bc44e5e11` addressing the one substantive CodeRabbit point.

The StakingRouter zero-total fallback is fixed by defaulting `treasuryShare` to `1` instead of `0` when `modulesFeeBp + treasuryFeeBp == 0`. With `operatorShare = 0` and `treasuryShare = 1`, the whole 10% protocol take falls into the Treasury bucket (matching the pre-refinement behavior of attributing it to a single bucket), so total Revenue equals `dailyProtocolRevenueUSD` even under degraded reads. The only thing lost in that window is the operator/treasury breakdown granularity.

I did not wrap the `options.toApi.call` in a try/catch — per the team's pattern (commit `b02258910` on #7077 by bheluga removing the same try/catch CodeRabbit had suggested), errors should propagate to the monitoring pipeline rather than be silenced. The zero-total guard is the right shape because it preserves the Revenue total while letting any genuine call failure bubble up.

bheluga:

PR #7212, opened 2026-05-26
fix(txflow-perps-oi): query end-of-day OI instead of start-of-day

DaniiRix:
The WHERE clause used `options.startOfDay` (midnight UTC), so `openInterestAtEnd` was reporting the prior day's closing value. Changed to `hour_ts <= options.endTimestamp ORDER BY hour_ts DESC LIMIT 1` to pick the last hourly record within the reporting period.


bheluga:
@DaniiRix thanks for the PR

PR #7216, opened 2026-05-26
fees/kyberswap-aggregator: capture native gas-token fees alongside ERC20

tarun-khatri:
## Summary

The adapter at \`fees/kyberswap-aggregator.ts\` only credited the protocol fee collector \`0x4f82e73edb06d29ff62c91ec8f5ff06571bdeb29\` via \`addTokensReceived\`, which reads ERC20 \`Transfer\` events. KyberSwap's MetaAggregationRouterV2 also forwards the protocol fee as the chain's native gas token whenever the trade pays in ETH / BNB / etc., and those inflows arrive as internal value calls with no \`Transfer\` event. The previous version of the file carried a \`// MISSING INTERNAL ETH TRANSFERS!\` self-note (added in commit \`d2fbde5fd1\`) acknowledging the gap; this PR closes it.

## On-chain verification

Internal native-token traces landing on \`0x4f82e7…\` over the 7-day window ending 2026-05-26 ([Dune query 7579643](https://dune.com/queries/7579643)):

| Chain | Internal traces | Native received | Dominant sender |
|---|---:|---:|---|
| Ethereum | 786 | 1.06 ETH | \`0x8f10b468…\` (784 traces, 0.97 ETH) |
| Base | 3,586 | 2.29 ETH | \`0x8f10b468…\` (3,586 traces, 2.29 ETH) |
| BSC | 786 | 0.29 BNB | \`0x8f10b468…\` (784 traces, 0.29 BNB) |
| Arbitrum | 161 | 0.0051 ETH | \`0x8f10b468…\` |

Total ≈ \$12k of native fees in seven days, or roughly \$50k/30d that the current adapter would silently drop. The dominant sender on every chain is the router \`0x8f10b468b06c6fd214b65f87778827f7d113f996\`, which is the CREATE2-deployed MetaAggregationRouterV2 (verified by reading \`owner()\` on Ethereum: \`0x697161a9ece8afd6ae30946d790077050f993cbc\`, a known KyberSwap multisig). The same router address is used on every supported chain.

## Diff shape

- \`getETHReceived\` is called after the existing \`addTokensReceived\`, accumulating native inflows into the same \`dailyFees\` balances object.
- Call is gated on a small whitelist of chains whose entries exist in \`helpers/allium.ts\` \`ALLIUM_CHAIN_MAP\` (Ethereum, Arbitrum, Avax, BSC, Optimism, Polygon, Scroll, Base, Plasma, Berachain, Unichain, Monad, ERA, Hyperliquid). Outside this set the helper would resolve to a non-existent schema and throw, so the existing ERC20-only path stays in effect there. No try/catch — failures continue to propagate.
- Stale commented-out debug print block removed.
- \`methodology\` block added. Production \`/summary/fees/kyberswap-aggregator\` currently returns \`Fees=?, Revenue=?\` because the adapter never declared a methodology; this fills that.

## Test plan

- [x] \`pnpm run ts-check\` clean
- [x] Branch cut fresh from upstream master at \`409ca6d04\`
- [x] \`git diff --stat upstream/master..HEAD\` = \`1 file changed, +33/-13\`
- [x] Sender identity verified by reading \`owner()\` on \`0x8f10b468…\` via 1rpc.io eth_call
- [x] Fee collector code length confirmed \`0x\` (EOA) so the helper choice (Allium traces, not SafeReceived) is correct

### DIFF
--- a/fees/kyberswap-aggregator.ts
+++ b/fees/kyberswap-aggregator.ts
@@ -1,5 +1,5 @@
 import { CHAIN } from "../helpers/chains"
-import { FetchOptions, SimpleAdapter } from "../adapters/types";
+import { Dependencies, FetchOptions, SimpleAdapter } from "../adapters/types";
 import { addTokensReceived, getETHReceived } from "../helpers/token";
 import { getDefaultDexTokensBlacklisted } from "../helpers/lists";
 
@@ -95,6 +95,7 @@ const methodology = {
 const adapter: SimpleAdapter = {
   version: 2,
   pullHourly: true,
+  dependencies: [Dependencies.ALLIUM],
   fetch,
   adapter: chainConfig,
   methodology,
--- a/fees/kyberswap-aggregator.ts
+++ b/fees/kyberswap-aggregator.ts
@@ -1,6 +1,6 @@
 import { CHAIN } from "../helpers/chains"
 import { FetchOptions, SimpleAdapter } from "../adapters/types";
-import { addTokensReceived } from "../helpers/token";
+import { addTokensReceived, getETHReceived } from "../helpers/token";
 import { getDefaultDexTokensBlacklisted } from "../helpers/lists";
 
 const chainConfig: Record<string, { id: number, start: string }> = {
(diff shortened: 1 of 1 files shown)


noateden (changes requested):
can you add allium dependencies flag? like this: https://github.com/DefiLlama/dimension-adapters/pull/7220/changes

tarun-khatri:
Pushed `532741957` adding `dependencies: [Dependencies.ALLIUM]` to the adapter export, mirroring the trust-wallet follow-up at #7220.

On the two CodeRabbit inline comments:

1. **try/catch around `getETHReceived`** — leaving it as-is. Per the team convention surfaced in #7077 (commit `b02258910` titled "dont swallow errors"), the helper is meant to propagate failures to the runAdapter framework so monitoring sees them, instead of silently returning 0. A narrow `catch { return 0 }` here would mask Allium outages as "kyberswap had no native fees today," which is the failure mode the existing convention is meant to prevent.

2. **`breakdownMethodology`** — not adding it. The adapter accumulates ERC20 and native-token fees into a single undifferentiated `dailyFees` balance with no sub-metric labels (no `METRIC.*` tags on `add()` calls). `breakdownMethodology` is for adapters that split a dimension across labelled buckets; here the methodology field already describes both sources in one sentence and there's no axis to break down on.

`pnpm run ts-check` passes. `git log --oneline upstream/master..HEAD` = the original commit + this Allium-flag commit, both touching only `fees/kyberswap-aggregator.ts`.

@coderabbitai review

noateden (approved):

PR #7231, opened 2026-05-27
aggregators/jupiter-aggregator: source from dex_solana.trades from 2025-09

srijani-raccoons:
jupiter_solana.aggregator_swaps has incomplete coverage from 2025-09. This switches the source to dex_solana.trades (Jupiter v6 program) from that date; earlier dates keep aggregator_swaps so history is preserved.

Volume from 2025-09 onward is restated to the dex_solana.trades values.


bheluga:
@srijani-raccoons thanks for the PR

PR #7238, opened 2026-05-27
Feature kaching

deepakpraka:
Replaced /pots endpoint with /transactions/ticket-purchases API which returns actual daily ticket purchase data
Sum the amount field (USDC) directly from the API response — no Solana RPC calls needed
Pagination using totalPages from API response to fetch all transactions for the day


bheluga:
@deepakpraka thanks for the PR
Added [here](https://github.com/DefiLlama/dimension-adapters/pull/7265)

PR #7254, opened 2026-05-27
fix: yaka v3 fees

### LINKED ISSUE #6512: Fix yaka v3 fees

DaniiRix:
fix #6512 

## Summary

* The Yaka V3 subgraph (`yaka-data-new`) is no longer working and just returns `Not found`.
* also found the fee split assumption was wrong. Instead of 12% protocol / 88% LPs, the on-chain `defaultCommunityFee()` on the Algebra factory (`0xEdbBc263C74865e67C6b16F47740Fa3901b95Ae1`) returns `15`, which means the actual split is 1.5% protocol fees and 98.5% to LPs (per the Algebra Integral 1.2 spec). We confirmed this across all 15 live pools via `globalState().communityFee = 15`.
* Because of the dead subgraph, the adapter was switched to the GeckoTerminal pools API, following the same approach already used in the `sailor-finance` Sei Algebra integration.
* `runAtCurrTime: true` is now enabled since GeckoTerminal only exposes rolling 24h data. This means historical backfills are no longer available, but the old setup wasn’t returning usable historical data anyway.

### DIFF
--- a/dexs/yakafinance-v3/index.ts
+++ b/dexs/yakafinance-v3/index.ts
@@ -10,13 +10,13 @@ import { sleep } from "../../utils/utils";
 
 const GECKOTERMINAL_POOLS_URL =
   "https://api.geckoterminal.com/api/v2/networks/sei-evm/dexes/yaka-finance-v3/pools";
-const DEFAULT_FEE_RATE = 0.003;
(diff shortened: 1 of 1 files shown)


bheluga:
@DaniiRix thanks for the PR

PR #7260, opened 2026-05-28
fix(fuel-ignition): handle blob fee indexer failures

ErebusCipher:
The blob fee endpoint currently returns 503, which caused the adapter to throw before reaching the working GraphQL fee path.

This wraps the blob fee request in a try/catch, logs the failure, and falls back to 0 blob fees so the adapter can still return daily fees from the GraphQL source.

Tested locally:
- Daily fees: $3.91k
- Daily revenue: $0.00
- Blob fee request logs 503 and continues

### DIFF
--- a/fees/fuel-ignition.ts
+++ b/fees/fuel-ignition.ts
@@ -1,25 +1,27 @@
 import { SimpleAdapter, FetchOptions, ProtocolType } from "../adapters/types";
 import { CHAIN } from "../helpers/chains";
-import { httpGet, httpPost } from "../utils/fetchURL";
+import { httpPost } from "../utils/fetchURL";
 
-const INDEXER_URL = "https://indexer-fuel-seq.simplystaking.xyz";
 const EXPLORER_URL = "https://explorer-indexer-mainnet.fuel.network/graphql";
 
-interface BlobFeesResponse {
-  totalFees: string;
-  feeDenom: string;
-  decimals: number;
-  blobCount: number;
-  startTimestamp: number;
-  endTimestamp: number;
-}
+const methodology = {
+  Fees: "Total gas fees paid by users on the Fuel Ignition network.",
+  Revenue:
+    "Gas fees retained by the Fuel sequencer. Fuel migrated data availability from Ethereum blobs to EigenDA in July 2025. EigenDA costs are settled via prepaid PaymentVault deposits and off-chain per-blob metering, so there is no public per-rollup daily DA cost to net out as supply-side revenue. If a public EigenDA cost source becomes available, it can be added as dailySupplySideRevenue.",
+};
+
+const breakdownMethodology = {
+  Fees: {
+    "Gas Fees": "Gas fees paid by users for transactions on Fuel Ignition.",
+  },
+  Revenue: {
+    "Gas Fees To Sequencer": "Gas fees retained by the Fuel sequencer.",
+  },
+};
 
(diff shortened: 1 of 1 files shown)


ErebusCipher:
Added labels and methodology metadata per CodeRabbit's feedback: `Gas Fees` on `dailyFees`, `Blob Fees To Protocol` on `dailyRevenue`, with matching `breakdownMethodology` entries for both.

ErebusCipher:
Updated the accounting to mirror the Taiko L2 model: user gas stays in dailyFees, blob/DA costs are dailySupplySideRevenue, and dailyRevenue is net of those costs. The blob endpoint represents sequencer DA costs paid onward, not protocol-retained revenue.

bheluga (changes requested):
In cases where 1 source isnt working and protocol is active, we should be looking for alternative sources and not simply ignoring the errors

ErebusCipher:
Thanks, agreed that falling back to 0 for an active source is not the right fix.

I dug into the blob-fee source. Fuel no longer posts DA to Ethereum blobs: L2BEAT notes that data publishing to Ethereum blobs stopped and moved to EigenDA, and Dune's `ethereum.blobs_submissions` shows Fuel blob submissions tapering off by July 2025.

I also checked whether EigenDA costs can replace the old source. EigenDA billing is handled through prepaid PaymentVault deposits and off-chain per-blob metering by the disperser/accountant, so there does not appear to be a public per-rollup daily cost source for Fuel. On-chain batch confirmations do not expose payer/cost data, and using deposits would misattribute timing.

I removed the obsolete blob-fee endpoint dependency instead of swallowing the error. The adapter now reports the live Fuel gas-fee path from the Fuel explorer GraphQL source, with methodology updated to document the DA migration. If a public EigenDA cost source becomes available, it can be added later as supply-side revenue.

ErebusCipher:
@bheluga this is ready for re-review when you have a chance.

I addressed your point by removing the obsolete blob-fee endpoint dependency rather than ignoring the error. Fuel moved DA from Ethereum blobs to EigenDA in July 2025, and I couldn't find a public per-rollup daily EigenDA cost stream, so the adapter now reports the live Fuel explorer gas-fee stream as fees/revenue and documents the DA migration. CI/testAdapter is green.

ErebusCipher:
Rebased onto master to pick up the single-arg fetch signature from #7461 (and the new explorer auth headers). Same change as before, squashed: removes the dead blob-fee indexer dependency and adds methodology/breakdown metadata. Local test: $3.76k daily fees/revenue.

uwezukwechibuzor:
Hey @ErebusCipher — I opened #7771. I reached out to the Fuel team and got a new working endpoint for the blob
  fees: the indexer just moved hosts (the 503 source isn't gone).
  
  New host — same /seq/blob-fees path and schema, serving daily data up to now:

  ```http
  GET https://index-api.sequencer.mainnet.fuel.network/seq/blob-fees?start=1767225600&end=1767312000
  ```

  ```json
  {
    "totalFees": "7608815104",
    "feeDenom": "fuel",
    "decimals": 9,
    "blobCount": 7196,
    "startTimestamp": 1767225600,
    "endTimestamp": 1767312000
  }
```
  
  So instead of dropping the blob-fee source, #7771 just repoints INDEXER_URL to the new host, keeps blob fees
  tracked. Also addresses @bheluga's point about finding an alternative source rather than ignoring the error.

PR #7290, opened 2026-05-29
[feat] Ton-Stats: track users and txcount for ton network via allium

Shaileshkhote:
## Summary
- Adds TON chain active-users and new-users adapters.
- Uses Allium `ton.raw.transactions` to report non-aborted active accounts, transaction count, and first-seen accounts.

## Changes
- Added `active-users/ton.ts` for `dailyActiveUsers` and `dailyTransactionsCount`.
- Added `new-users/ton.ts` for `dailyNewUsers`.
- Set TON start date to `2021-08-03`.

## Methodology
- Active users are counted as distinct non-aborted transaction `account`s during the adapter window.
- Transaction count is counted from non-aborted transaction `hash` values.
- New users are counted from each `account`'s first non-aborted transaction.
- TON’s `raw.transactions` table is account-centric, so `account` is used as the chain-level active account/user identifier.

### DIFF
--- a/active-users/ton.ts
+++ b/active-users/ton.ts
@@ -1,0 +1,34 @@
+import { Dependencies, FetchOptions, ProtocolType, SimpleAdapter } from "../adapters/types";
+import { queryAllium } from "../helpers/allium";
+import { CHAIN } from "../helpers/chains";
+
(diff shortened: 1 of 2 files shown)


bheluga:
Added [here](https://github.com/DefiLlama/dimension-adapters/pull/7295)

PR #7294, opened 2026-05-29
feat: Track Dash chain active users

uwezukwechibuzor:
### Summary                                                                                                     
  - Add Dash chain active users and transaction count tracking via CoinMetrics API in `users/chains.ts`


bheluga:

PR #7298, opened 2026-05-29
[fix] User-Stats: migrate flowchain and etc to blockscoutStats 

Shaileshkhote:
Migrate Flow from coinmetrics to blockscout as coinmetrics provides flow non-evm data 
Ethreum Classic is already present in the blockscout helper


bheluga:

PR #7313, opened 2026-05-30
fix: shift-protocol fee spikes from zero share price

uwezukwechibuzor:
### Summary   
                                                                                                            
  - Skip vaults when getSharePrice returns 0 (invalid RPC data) to prevent massive fee swings 
  On certain hours, Base vault share prices return 0 from RPC:

```
  priceBefore  priceAfter
  0            0           ← invalid, skipped
  1011715      1011661     ← valid                                                                               
  1006779      1006783     ← valid
  1020026      1019987     ← valid                                                                                                                                                           
  1067976      0           ← ltPARA price drops to 0 (bad RPC data)                                              
  1014358      0           ← extUSD price drops to 0 (bad RPC data)                                            
  0            1019987     ← invalid, skipped                                                                    
  ```
  Without the guard, priceChange = 0 - 1067976 produces spikes.

### DIFF
--- a/fees/shift-protocol/index.ts
+++ b/fees/shift-protocol/index.ts
@@ -58,20 +58,22 @@ const fetch = async (options: FetchOptions): Promise<FetchResultV2> => {
 
     if (!baseToken || !totalSupply || !priceBefore || !priceAfter) continue;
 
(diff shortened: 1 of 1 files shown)


bheluga:

PR #7314, opened 2026-05-30
feat: Track Dogecoin chain active users and transactions via CoinMetrics

uwezukwechibuzor:
https://defillama.com/chain/doge


bheluga:

PR #7320, opened 2026-05-30
parallelize chain fee fetches in across prefetch

adityamane765:
prefetch was fetching bridge fees for each destination chain one at a time inside a for-of loop, so each chain waited for the previous one to finish. but since all chain fetches are fully independent, replaced the loop with Promise.all so all chains are fetched concurrently. error handling per chain is unchanged.


bheluga:

PR #7324, opened 2026-05-30
[feat] Superswap-Bridge-Agg: tracks bridge volume for superswap onchain

Shaileshkhote:
> Maintainer Note
> Superswap crosschain swaps operate on top of Across Bridge, Hence add as bridge-aggregator

Tracks bridge volume for https://defillama.com/protocol/superswap-ink

## Summary
- Adds a SuperSwap bridge-aggregator adapter under `bridge-aggregators/superswap`.
- Tracks portal-directed bridge deposits as `dailyBridgeVolume` across supported SuperSwap portal chains.

## Changes
- Added a v2 bridge-aggregator adapter using Across `FundsDeposited` logs from each configured source-chain `spokePool`.
- Counts deposits only when the decoded `recipient` matches a known SuperSwap portal address.
- Returns token-level `dailyBridgeVolume`, including gas-token handling.
- Marks the adapter as `doublecounted: true` because volume is sourced from underlying Across bridge activity.

## Methodology
- Bridge volume is counted from source-chain Across deposits where the recipient is a SuperSwap portal.
- Uses `pullHourly: true` for EVM log fetching.

### DIFF
--- a/bridge-aggregators/superswap/index.ts
+++ b/bridge-aggregators/superswap/index.ts
@@ -53,7 +53,6 @@ const fetch = async (options: FetchOptions) => {
 const adapter: SimpleAdapter = {
   version: 2,
   pullHourly: true,
-  doublecounted: true,
   fetch,
   adapter: chainConfig,
 }
--- a/bridge-aggregators/superswap/index.ts
+++ b/bridge-aggregators/superswap/index.ts
@@ -1,0 +1,61 @@
+import { FetchOptions, SimpleAdapter } from "../../adapters/types"
+import { CHAIN } from "../../helpers/chains"
+import ADDRESSES from "../../helpers/coreAssets.json"
+
+type ChainConfig = {
+  start: string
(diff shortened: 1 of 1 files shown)


bridge-aggregators/superswap/index.ts
```diff
@@ -0,0 +1,61 @@
...
+    else dailyBridgeVolume.add(token, log.inputAmount)
+  })
+
+  return { dailyBridgeVolume }
+}
+
+const adapter: SimpleAdapter = {
+  version: 2,
+  pullHourly: true,
+  doublecounted: true,
```
noateden: bridge aggregators don't need this flag

noateden (approved):

bridge-aggregators/superswap/index.ts
Shaileshkhote: Ok ser, thanks for flagging

PR #7336, opened 2026-05-30
fix: instadapp flashloan fee accounting

DaniiRix:
InstaDapp's flashloan aggregator adds a 0.05% fee, but it only keeps that fee when it routes a flashloan through a free source like Maker, Balancer or Spark. On Aave routes the Aave premium already costs more than 0.05%, so the protocol ends up keeping nothing.

The old adapter charged 0.05% on all flashloan volume and returned only dailyFees, which also failed the revenue validation.

### DIFF
--- a/fees/instadapp/index.ts
+++ b/fees/instadapp/index.ts
@@ -7,80 +7,106 @@ import {
 import { Balances } from "@defillama/sdk";
 import { METRIC } from "../../helpers/metrics";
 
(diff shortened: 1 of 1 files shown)


DaniiRix:
@coderabbitai review

noateden:
thanks for the fix, but we aren't using this adapter everywhere for now

PR #7337, opened 2026-05-30
track revenue and protocol revenue

DaniiRix:
Lamboo keeps 100% of the integrator fees it collects


bheluga:
@DaniiRix thanks for the PR

PR #7338, opened 2026-05-30
fees/snakefinance - track revenue and protocol revenue

DaniiRix:
Snake Finance collects all user fees into its protocol addresses


bheluga:
@DaniiRix thanks for the PR

PR #7339, opened 2026-05-30
parallelize sequential fetches in steamm, woofi, delphi and lexer adapters

adityamane765:
four adapters had sequential awaits inside loops or consecutive independent calls that blocked on each other unnecessarily.

steamm: fetched historical volume for each pool one at a time inside a for loop. replaced with PromisePool with concurrency of 5 to fetch all pools concurrently without overwhelming the API.

woofi: four consecutive api.call to the same contract (quoteToken, rebateManager, treasury, vaultManager) each awaited separately. replaced with a single Promise.all.

delphi: two api.call per market proxy inside a for loop (getMarket and tradingFeesRecipientPct). replaced with two multiCall batches covering all proxies at once.

lexer-swap and lexer-derivatives: two sequential subgraph requests to independent endpoints. replaced with Promise.all.

### DIFF
--- a/dexs/delphi/index.ts
+++ b/dexs/delphi/index.ts
@@ -32,15 +32,19 @@ const fetch: FetchV2 = async (options) => {
 
     const configs: Record<string, { tradingFee: bigint, recipientPct: bigint }> = {};
 
-    for (const marketProxy of marketProxies) {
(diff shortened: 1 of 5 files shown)


bheluga:
Except for delphi and steamm there wont be any significant performance improvements and we usually avoid Promise all, as the tradeoff isnt better and might lead to errors with code 429

PR #7341, opened 2026-05-30
fix getLatestBlockRetry never actually retrying on failure

adityamane765:
the retry loop called throw inside the catch block, which then rethrew the error immediately on the first failure making the loop exit after one attempt. retries 2 through 5 were unreachable. fixed it by storing the last error and only throwing after all attempts are exhausted. also added exponential backoff between retries.


bheluga:

PR #7348, opened 2026-05-31
[fix] Helpers-getETHReceived: exclude single-target self-transfers

Shaileshkhote:
## Summary
- Fixes `getETHReceived` so single-target native token self-transfers are excluded as documented.
- Prevents fee/revenue adapters using one receiver wallet from overcounting native-token inflows sent by the same wallet.

## Changes
- Updated `helpers/token.ts` to apply `from_address NOT IN (...)` whenever the exclude sender list has at least one address.
- Applied the same condition to both Allium native transfer and raw traces query paths.
- Preserved existing multi-target behavior and `notFromSenders` filtering.


bheluga:

PR #7354, opened 2026-06-01
fix: migrate ssv-network ubgraph to ETH onchain sources after cSSV la…

### LINKED ISSUE #5597: Fix fees SSV network
https://defillama.com/protocol/ssv-network

DaniiRix:
Since the April 2026 migration, validator fees are paid in ETH instead of SSV. The existing adapter still relied on an SSV subgraph that no longer tracks this activity, causing reported fees to fall to ~$9/day despite actual fees remaining around ~$3.8k/day.

### Changes

* Calculate network fees from `FeesSynced.newFeesWei` events
* Calculate operator fees from `getOperatorById` rates over the period
* Route network-fee revenue to cSSV stakers (holders) instead of the DAO treasury
* Remove the obsolete SSV subgraph dependency and burn-tracking log query
* Bump adapter version from `1` → `2`
* Set new start date to `2026-04-22` (migration block / first `FeesSynced` event)

fixes #5597


bheluga (dismissed):
@DaniiRix thanks for the PR
In cases like this we dont change the start date, we should still preserve old code and do time based handling, if run before migration, execute old code else new code

DaniiRix:
> @DaniiRix thanks for the PR In cases like this we dont change the start date, we should still preserve old code and do time based handling, if run before migration, execute old code else new code

got it

bheluga:
@DaniiRix thanks for the PR

PR #7353, opened 2026-06-01
feat: add silentswap volume

GarageInc:

PR #7355, opened 2026-06-01
track silentswap volume

bheluga:
metadata: https://github.com/DefiLlama/dimension-adapters/pull/7353


GarageInc:
Hi @bheluga — thanks for merging this! The adapter still isn't producing data because the protocol entry (id 7964) in
defillama-server is missing the dimensions mapping, so generateProtocolAdaptorsList2 skips it (config[adapterKey] is undefined → if (!configObj) return).
  
Ready-to-merge fix here: https://github.com/GarageInc/defillama-server/commit/f9fbd9b — just adds dimensions: { 
"bridge-aggregators": "silentswap" } to the entry in defi/src/protocols/data6.ts (same format as the WheelX entry). I couldn't open a PR since the repo restricts interactions to prior contributors.
  
The adapter itself runs fine: npm run test bridge-aggregators silentswap 2026-06-03 → Daily volume: 24.18 M, and the reporter contract has nightly snapshots since 2026-05-29, so refill from the start date will work. 🙏

PR #7357, opened 2026-06-01
parallelized sequential getLogs calls in sushiswap aggregator

adityamane765:
the fetch function was calling getLogs for each router version RP4 through RP11 sequentially with await, one after another. since all fetches are independent, collected them into an array and fetched all in parallel with Promise.all, then flattened the results.


adityamane765:
limit exceeded errors are from eth_getLogs block range limits on RPC nodes and not due to concurrency, the same error would occur sequentially since each getLogs call queries the same large block range. this seems like pre existing RPC infrastructure flakiness

g1nt0ki:
@adityamane765 thanks for the PR, but we are fine with it being a bit inefficient as we have less changes of getting ratrelimited by the public rpcs

PR #7367, opened 2026-06-01
refactor(akka): switch to on-chain volume tracking via Swap events

lkhoshnafs:
## Summary
- Upgrade from v1 (API-based) to v2 (on-chain event logs)
- Add HyperEVM (Hyperliquid) chain support
- Preserve Core chain history with end date (2026-05-14)
- Remove XDC, Bitlayer, B² chains
- Remove dependency on routerv2.akka.finance API

## Changes
- Read Swap events directly from AkkaRouter contracts instead of calling an API
- Core chain: 3 router contracts, historical data preserved
- HyperEVM: 1 router contract, live

### DIFF
--- a/aggregators/akka/index.ts
+++ b/aggregators/akka/index.ts
@@ -1,61 +1,70 @@
+import { FetchOptions, FetchResult, SimpleAdapter } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
-import { FetchOptions, SimpleAdapter } from "../../adapters/types";
 
 const swapEvent =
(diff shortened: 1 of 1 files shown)


bheluga (changes requested):
@lkhoshnafs thanks for the PR
Can you add targets for other chains (xdc, bitlayer, bsquared ) too? and mark them dead , if not active

bheluga:
@lkhoshnafs thanks for the PR
added [here](https://github.com/DefiLlama/dimension-adapters/pull/7397/changes)

PR #7371, opened 2026-06-01
feat(openden-t-bills): add dailyRevenue and dailyProtocolRevenue

DaniiRix:
- Verified onchain that both treasury addresses (`oplTreasury` and `mgtFeeTreasury`) are controlled by OpenEden. The vault does not have any partner or third-party fee-sharing arrangements in place, meaning all protocol fees are retained by OpenEden.

- Depositor yield is generated by the underlying U.S. Treasury Bills rather than protocol fees, so users do not incur any supply-side fee costs.

### DIFF
--- a/fees/openeden-t-bills/index.ts
+++ b/fees/openeden-t-bills/index.ts
@@ -67,12 +67,14 @@ const fetch = async (
     });
   }
 
-  return { dailyFees };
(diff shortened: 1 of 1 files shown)


bheluga:
@DaniiRix thanks for the PR

PR #7372, opened 2026-06-01
[fix] Native-Transfers: resolve native transfer table bug in token helper

Shaileshkhote:
Fixes native transfer table interpolation in `helpers/token.ts` so mapped Allium tables like Tron and Near resolve correctly.


bheluga:

PR #7382, opened 2026-06-02
feat(blocxroute): add revenue split

DaniiRix:
- add `dailyRevenue` and `dailyProtocolRevenue` to both Solana and BSC fetch returns (both equal to `dailyFees`)
- add `Revenue` and `ProtocolRevenue` entries to `methodology`
- bloxroute is an MEV relay with no supply-side participants. All collected fees flow directly to the protocol treasury, so `dailyFees = dailyRevenue = dailyProtocolRevenue`.


bheluga:
@DaniiRix thanks for the PR

PR #7383, opened 2026-06-02
Arrakis V2: fee attribution from onchain verification

DaniiRix:
The adapter incorrectly assumed all fees belonged to Arrakis and attributed 100% of `fee0`/`fee1` from `totalUnderlyingWithFees()` to `dailyProtocolRevenue`.

This was incorrect because:

1. `fee0`/`fee1` represent gross Uniswap V3 LP fees accrued by the vault, not management fees.
2. Arrakis V2 vaults are permissionless and can be managed by any address; V2 core does not charge a protocol-level fee.

### Fix

Verified the fee model from the V2 core contracts and updated revenue attribution:

* Read `managerFeeBPS` from each vault.
* Split gross LP fees as follows:

  * **dailyFees**: total LP fees accrued.
  * **dailySupplySideRevenue**: LP share (`gross × (1 - managerFeeBPS / 10000)`).
  * **dailyRevenue**: manager share (`gross × managerFeeBPS / 10000`).
* Removed `dailyProtocolRevenue`, as fees cannot be attributed to Arrakis without identifying manager ownership.
* Removed the unnecessary `skipBreakdownValidation: true`.

### DIFF
--- a/fees/arrakis-v2/index.ts
+++ b/fees/arrakis-v2/index.ts
@@ -1,6 +1,7 @@
 import { Chain } from "../../adapters/types";
 import { Adapter, FetchOptions } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
+import { METRIC } from "../../helpers/metrics";
 
 type IVault = {
   helper: string;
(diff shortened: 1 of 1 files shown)


noateden:
hi @DaniiRix, we should count all fees to vaults managers as `dailySupplySideRevenue`. only count to `dailyRevenue` if fees are collected by protocol or governance token holders

noateden:
@DaniiRix thanks for the fix

PR #7398, opened 2026-06-02
Track pacifica fees

### LINKED ISSUE #5245: Track Pacifica Fee/Revenue
[Pacifica](https://defillama.com/protocol/pacifica)

nopbody:
closes https://github.com/DefiLlama/dimension-adapters/issues/5245

https://defillama.com/protocol/pacifica

### DIFF
--- a/fees/pacifica/index.ts
+++ b/fees/pacifica/index.ts
@@ -4,10 +4,42 @@ import { METRIC } from "../../helpers/metrics";
 import { FetchOptions, SimpleAdapter } from "../../adapters/types";
 import fetchURL, { fetchURLAutoHandleRateLimit } from "../../utils/fetchURL";
 
-// Base fee tier (most users). Higher tiers have lower rates.
-const TAKER_FEE = 0.0004;
-const MAKER_FEE = 0.00015;
-const BLENDED_FEE = TAKER_FEE + MAKER_FEE;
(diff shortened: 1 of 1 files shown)


bheluga (changes requested):
@nopbody thanks for the PR
Its not ideal to that assume all trades are settled under fee tier 1, in fact more volume could come from lower fee tier traders

nopbody:
> @nopbody thanks for the PR Its not ideal to that assume all trades are settled under fee tier 1, in fact more volume could come from lower fee tier traders

Hmm, we can switch from blended (maker + taker = 0.055%) to taker-only (0.040%). 
As pacifica's fee tiers are 8 levels, maker drops from 0.015% to 0.000% at VIP, taker only goes 0.040% → 0.028%. Since most volume comes from high-tier traders and we can't get actual fee amounts from the API, using taker-only gives a more honest upper bound.
I have checked locally and the fee reduced by ~28% ($311K to $225K/day).
wdyt?

also we don't need the `breakdownMethodology` here right? as the methodology itself is pretty straight forward?

bheluga:
@nopbody you can still add breakdown methodology like trading fees.
whatever you derive from fee tier is just an estimate, can change day to day.
so we cant rely on this logic

bheluga:
closing the PR, as its just an approximation and may not correctly track fees

PR #7411, opened 2026-06-03
fix: compute 3F fees from FeesAccrued events

maxencerb:
## Summary

This is a fix for the 3F fees/revenue adapter (`fees/3f.ts`) added in #7142.

The computation introduced in #7142 does not match how fees are actually charged on our leveraged RWA positions, so the reported fees and revenue are incorrect. The previous version tried to estimate management and performance fees from `totalAssets` deltas over each window, which neither reflects the protocol's fee logic nor the timing of fee collection.

This PR sources the fees directly from the on-chain `FeesAccrued` events (the single source of truth that stays correct across our upcoming fee-mechanism upgrade), and recomputes the management/performance breakdown for each event from the contract state it accrued against.

## What this fix does

3F mints fee shares to the protocol fee recipient and emits a single event whenever fees are accrued on a Position Manager:

```solidity
/// @notice Emitted when fees are accrued and minted to the fee recipient.
event FeesAccrued(address indexed feeRecipient, uint256 shares);
```

The adapter now:

1. Discovers every Position Manager from the factory's `PositionManagerCreated` events.
2. Reads each Position Manager's `FeesAccrued` events within the window — this captures fees directly at the source instead of estimating them.
3. For each event, reads the Position Manager state at the block before the accrual (the same snapshot the contract accrued against), recovers the total fee value in the debt asset by inverting the share conversion (`feeAssets = shares * (totalAssets + 1) / (totalSupply + virtualShareOffset + shares)`), and solves the management/performance split algebraically from the performance-fee basis.

The split handles both fee mechanisms: the current NAV-variation performance basis and the upcoming levered-slice basis (`LTV_prev * currentCollat - currentDebt`, detected via the new `lastDebt()` getter). It also handles a manager accruing more than once in a single block: the first accrual resets the fee snapshot, so subsequent same-block events carry no performance basis and are valued against the supply inflated by the earlier mints.

Since every accrued fee is paid to the protocol, `dailyFees == dailyRevenue == dailyProtocolRevenue`.

## How the fees work

- The **performance fee** is taken on the return of the leveraged collateral, after subtracting the cost of the borrowed debt.
- The **management fee** is taken on the full value of the RWA collateral, including the leveraged portion, rather than only the user's deposited capital.
- Fees are accrued and collected when positions are updated, so a single day can include fees that built up over the preceding days or weeks. Listening to the events captures this correctly, whereas the per-window delta approach did not.

## Accuracy

- **Total fees are exact**: they are derived directly from the shares minted in the `FeesAccrued` events.
- **The management/performance split is a very close approximation**: it is recomputed from state at the block boundary rather than mid-transaction. Validated against the contract's own `pendingFees()` view on historical accruals for both fee mechanisms — discrepancies are only a couple of wei.

## Notes

- Tested locally: 2026-06-03 reports ~$283 in fees/revenue (Management Fees ~$66, Performance Fees ~$217), sourced entirely from `FeesAccrued` events, with the breakdown surfaced via `breakdownMethodology`.
- The failing `ts-check` job is unrelated to this PR: it fails on current master after the recent adapter-type refactor (errors are in files this PR does not touch).

### DIFF
--- a/fees/3f.ts
+++ b/fees/3f.ts
@@ -1,25 +1,22 @@
 import { FetchOptions, SimpleAdapter } from "../adapters/types";
 import { CHAIN } from "../helpers/chains";
-import { METRIC } from "../helpers/metrics";
 
 const POSITION_MANAGER_FACTORY = "0x8e0667429d1717b3e5fe783a6c472d6d901fe5fa";
 const FACTORY_DEPLOY_BLOCK = 24844184;
 
 const Abis = {
   PositionManagerCreated: "event PositionManagerCreated(address indexed positionManager, address indexed owner, address indexed collateralAsset, address debtAsset, uint256 ltv, address transferGuard)",
+  // Emitted when fees are accrued and minted (as shares) to the fee recipient.
+  FeesAccrued: "event FeesAccrued(address indexed feeRecipient, uint256 shares)",
   totalAssets: "uint256:totalAssets",
+  totalSupply: "uint256:totalSupply",
   assets: "function assets() view returns (address collateralAsset, address debtAsset)",
-  feeData: "function feeData() view returns (address feeRecipient, uint24 managementFee, uint24 performanceFee, uint256 lastTotalAssets, uint40 lastFeeAccrualTimestamp)",
 };
-
-const BPS = BigInt(10000);
-const SECONDS_PER_YEAR = BigInt(365 * 24 * 60 * 60);
 
 const fetch = async (options: FetchOptions) => {
   const dailyFees = options.createBalances();
-  const dailyRevenue = options.createBalances();
-  const dailySupplySideRevenue = options.createBalances();
 
+  // Discover all Position Managers ever created by the factory.
   const factoryLogs = await options.getLogs({
     target: POSITION_MANAGER_FACTORY,
     eventAbi: Abis.PositionManagerCreated,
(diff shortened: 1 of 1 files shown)


fees/3f.ts
```diff
@@ -30,110 +27,57 @@ const fetch = async (options: FetchOptions) => {
...
-      }
-
-      // supply side
-      const supplySide = gains - performanceFeeAmount;
-      dailyFees.add(debtAsset, supplySide, METRIC.ASSETS_YIELDS);
-      dailySupplySideRevenue.add(debtAsset, supplySide, METRIC.ASSETS_YIELDS);
-    }
+    // Convert fee shares to the debt asset: shares * totalAssets / totalSupply.
+    const feeAssets = feeShares * BigInt(totalAssets) / BigInt(totalSupply);
+    dailyFees.add(assets.debtAsset, feeAssets);
```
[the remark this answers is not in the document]
maxencerb: Price of a position manager shares can be considered almost constant for a day and up to a month based on the update candence of RWA oracles (most update daily, some only monthly), so it all depends on how long the windows are.

bheluga (commented):
@maxencerb thanks for the PR
Any specific reason why supply side revenue was removed?

maxencerb:
@bheluga this is not really supply side revenue here. Users here are taking leveraged exposure to RWAs (lending via Morpho), so what they "earn" is a leveraged market P&L: it's driven by RWA marks and the Morpho borrow rate, can be negative, and isn't a protocol-distributed yield stream. It's really an APY, not supply-side revenue.

It's also not reliably trackable from snapshots in this repo. The existing implementation set supply-side revenue to the delta of totalAssets, which counts every deposit as revenue, so it's badly biased by inflows/outflows. Switching to NAV-per-share (totalAssets / totalSupply) would remove the deposit bias, but to be accurate across a window with flows you'd need to time-weight against supply at each moment, which is more than a fees adapter should be doing.

So I'd keep this adapter scoped to protocol fees/revenue (sourced directly from the FeesAccrued events, which is reliable and robust to our upcoming fee-mechanism change), and if we want to surface user returns, do it as an APY on the yield-server rather than as supply-side revenue here. Let me know if that works for you.

fees/3f.ts
bheluga: is it possible to distinguish performance fees and management fees and add breakdown like previous version

maxencerb: the adapter now breaks fees down into management and performance fees again.

Since FeesAccrued only emits the combined share amount, the split is recomputed from the on-chain state at the block before each accrual (the same snapshot the contract accrued against), inverting the share conversion to recover the fee value and solving the management/performance split from the performance fee basis. Because this reads state at the block boundary rather than mid-transaction, the exact split is a very close approximation (validated against the contract's pendingFees() view. Discrepancies are only a couple of wei). The total fees are still derived directly from the actual FeesAccrued events, so they remain exact.

bheluga:
@maxencerb thanks for the PR

PR #7422, opened 2026-06-04
update treehouse exclude wallets

balaji-the-leaf:
Updated wallet exclusion list for fastlane redemption operations


bheluga:
@balaji-the-leaf thanks for the PR

PR #7427, opened 2026-06-04
fix Reya fees - spread and lm tokens

ioanaroceanu2000:
The fees previously covered the pool's PnL. Replaced this with per-trade spread + price impact from match events
Added liquidity management tokens appreciation as pool revenue


noateden:
thanks, I updated it here: https://github.com/DefiLlama/dimension-adapters/pull/7514

PR #7439, opened 2026-06-04
feat: add Solana support to 0x aggregator via dune

D-J-Harris:
Add Solana chain support to the 0x aggregator adapter, sourcing volume
from dex_solana.trades via Dune using the 0x settler program address
(Sett1erwx2eqT5A8uvu8GBxDFT2W5TNnhirL7hLmb8m). Mirrors the approach
used by the Jupiter aggregator adapter.

Website: https://0x.org
Twitter: https://twitter.com/0xProject

Example for 1st June
image

### DIFF
--- a/aggregators/zrx/index.ts
+++ b/aggregators/zrx/index.ts
@@ -1,7 +1,8 @@
 import { CHAIN } from "../../helpers/chains";
-import { FetchOptions } from "../../adapters/types";
+import { Dependencies, FetchOptions } from "../../adapters/types";
 import { getEnv } from "../../helpers/env";
 import { httpGet } from "../../utils/fetchURL";
+import { queryDuneSql } from "../../helpers/dune";
 
 type TChain = {
   [key: string]: number;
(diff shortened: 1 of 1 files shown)


D-J-Harris:
Note: The above review is valid but should be dismissed. It appears to be convention (see Jupiter and DFlow queries) for Solana aggregators to count volume for all swaps in a transaction, not just those swaps representing volume from the input token

Amending this methodology should be a separate concern applied to all aggregators at once. Considering some surface volume data through internal APIs (e.g. Titan and OKX) then the scope of this change is likely large. So this PR sticks to apparent convention

bheluga:
@D-J-Harris thanks for the PR

D-J-Harris:
Thank you @bheluga. I see this is live - is there any way to get this backfilled from the start date in the adaptor?

```typescript
start: "2026-01-01",
```

bheluga:
> Thank you @bheluga. I see this is live - is there any way to get this backfilled from the start date in the adaptor?
> 
> ```ts
> start: "2026-01-01",
> ```
Hey,
I refilled it now
image

PR #7446, opened 2026-06-04
manta: track users via allium

DaniiRix:
Adds Manta to tracking via allium


bheluga:
@DaniiRix thanks for the PR

PR #7459, opened 2026-06-05
Add Shido chain user stats tracking via blockscout

uwezukwechibuzor:
https://defillama.com/chain/shido


bheluga:

PR #7461, opened 2026-06-05
migrate v1 adapters to single options fetch signature

treeoflife2:
Change v1 fetch functions from (timestamp, chainBlocks, options) to a single
  (options) argument, matching v2. The old first arg always equalled
  options.toTimestamp, so every reference to it is replaced with
  options.toTimestamp and the now-redundant timestamp field is dropped from
  returned objects (the runner sets result.timestamp).

runAdapter.ts: v1 fetch is called as fetch(options). A transitional arity
  bridge keeps un-migrated adapters working: functions declaring >= 2 params are
  still called the legacy way (toTimestamp, chainBlocks, options); functions with
  <= 1 param are called as fetch(options). result.timestamp is set to
  options.toTimestamp when the adapter does not return one.

  adapters/types.ts: Fetch is now (options: FetchOptions) => Promise<FetchResult>.


adapters/utils/runAdapter.ts
```diff
@@ -251,7 +251,14 @@ async function _runAdapter({
...
       if (adapterVersion === 1) {
-        result = await (fetchFunction as Fetch)(options.toTimestamp, chainBlocks, options);
+        // v1 fetch functions are being migrated from (timestamp, chainBlocks, options) to a
+        // single `options` arg (same shape as v2). Legacy un-migrated adapters still declare
+        // >= 2 params, so call those the old way during the transition; migrated ones take options.
+        result = (fetchFunction as Function).length >= 2
+          ? await (fetchFunction as Fetch)(options.toTimestamp, chainBlocks, options)
+          : await (fetchFunction as FetchV2)(options);
+        // migrated adapters no longer return `timestamp`; set it like v2 (legacy returns are preserved)
+        if (result.timestamp === undefined) result.timestamp = options.toTimestamp
```
g1nt0ki: this is wrong assumption, even for v1 adapters, I was trying to avoid setting this field where possible, but some adapters return yesterday's data or data few days ago irrespective of the requested timestamp, thats why we still tolerate timestamp in the response

aggregators/opt-agg/index.ts
```diff
@@ -1,15 +1,14 @@
 import fetchURL from "../../utils/fetchURL";
-import { FetchResult, SimpleAdapter } from "../../adapters/types";
+import { FetchResult, SimpleAdapter, FetchOptions } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
 
 const URL = "https://api.opt.finance/stat/vol";
 
-const fetch = async (timestamp: number): Promise<FetchResult> => {
+const fetch = async (_options: FetchOptions): Promise<FetchResult> => {
```
g1nt0ki: would remove this arg if it is not used

g1nt0ki (approved):

adapters/utils/runAdapter.ts
treeoflife2: but it was only applied when returned timestamp is undefined

dexs/10kswap/index.ts
```diff
@@ -10,8 +10,8 @@ interface IVolumeall {
   date: string;
 }
 
-const fetch = async (timestamp: number) => {
-  const dayTimestamp = getUniqStartOfTodayTimestamp(new Date(timestamp * 1000))
+const fetch = async (options: FetchOptions) => {
+  const dayTimestamp = getUniqStartOfTodayTimestamp(new Date(options.toTimestamp * 1000))
```
g1nt0ki: these can be options.startOfDay?

dexs/acala-swap/index.ts
```diff
@@ -24,8 +24,8 @@ interface IGraphResponse {
   dailyTradeVolumeUSD: string;
 }
 
-const fetch = async (timestamp: number) => {
-  const dateString = new Date(timestamp * 1000).toISOString().split("T")[0];
+const fetch = async (options: FetchOptions) => {
+  const dateString = new Date(options.toTimestamp * 1000).toISOString().split("T")[0];
```
g1nt0ki: I think we have options.dateString for this exact purpose

dexs/dexter/index.ts
```diff
@@ -1,8 +1,9 @@
-import {SimpleAdapter} from "../../adapters/types";
+import { SimpleAdapter, FetchOptions } from "../../adapters/types";
+import { CHAIN } from "../../helpers/chains";
 import {getDimensions, IDimensions} from "./dimensions";
 
-const fetch = async (timestamp: number) => {
-  const date = new Date(timestamp * 1000);
+const fetch = async (options: FetchOptions) => {
+  const date = new Date(options.toTimestamp * 1000);
   const dateStr = `${date.getUTCFullYear()}-${date.getUTCMonth() + 1}-${date.getUTCDate()}`;
```
g1nt0ki: we already have options.dateString

dexs/rhea-cross-chain/index.ts
```diff
@@ -9,10 +9,9 @@ const adapter: SimpleAdapter = {
     adapter: {
         [CHAIN.NEAR]: {
             runAtCurrTime: true,
-            fetch: async (_ts: any, _t: any, options: FetchOptions) => {
+            fetch: async (_: any) => {
```
g1nt0ki: if an arg is not used, why declare it

dexs/orderly-network-orderly-network.ts
```diff
@@ -4,9 +4,8 @@ import { CHAIN } from "../helpers/chains";
 const adapter: SimpleAdapter = {
   adapter: {
```
g1nt0ki: better to mark this as dead?

dexs/baryon/index.ts
```diff
@@ -21,6 +21,7 @@ const fetch = async (options: FetchOptions) => {
 
   return {
     dailyVolume: v2stats?.dailyVolume,
+    timestamp: options.startOfDay,
```
g1nt0ki: why return timestamp?

adapters/utils/runAdapter.ts
g1nt0ki: yeah, we dont want at all to set it here, should be taken care of by the caller (atm, server script sets it if missing, now that is affected)

adapters/utils/runAdapter.ts
```diff
@@ -251,7 +251,14 @@ async function _runAdapter({
 
       let result: any
       if (adapterVersion === 1) {
-        result = await (fetchFunction as Fetch)(options.toTimestamp, chainBlocks, options);
+        // v1 fetch functions are being migrated from (timestamp, chainBlocks, options) to a
+        // single `options` arg (same shape as v2). Legacy un-migrated adapters still declare
+        // >= 2 params, so call those the old way during the transition; migrated ones take options.
+        result = (fetchFunction as Function).length >= 2
+          ? await (fetchFunction as Fetch)(options.toTimestamp, chainBlocks, options)
+          : await (fetchFunction as FetchV2)(options);
```
g1nt0ki: just realized that this might not work for all cases, edge case here is:

`async function  fetch(timestamp: number) {`

- this will break now

adapters/utils/runAdapter.ts
treeoflife2: okay will remove that

dexs/10kswap/index.ts
treeoflife2: replaced all where it was using this function

dexs/acala-swap/index.ts
treeoflife2: replaced too

dexs/dexter/index.ts
treeoflife2: done

dexs/orderly-network-orderly-network.ts
treeoflife2: marked dead

dexs/baryon/index.ts
treeoflife2: removed all the timestamp which were returning the startOfDay

adapters/utils/runAdapter.ts
treeoflife2: there is no legacy fetch function exists now, should we remove this check?

PR #7462, opened 2026-06-05
fees/azuro: add Arbitrum and Linea fee coverage + mark Gnosis as dead

Shchepetov:
Update to the existing **Azuro** fees adapter


bheluga:
@Shchepetov thanks for the PR

PR #7368, opened 2026-06-01
fees(claimrush): income statement compliance — route 100% to supply side

claimrush:
# fees(claimrush): income statement compliance — route 100% to supply side

Follow-up to #7361 (merged 2026-06-01 by @bheluga). Addresses the
[CodeRabbit Income Statement Compliance warning](https://github.com/DefiLlama/dimension-adapters/pull/7361#pullrequestreview)
flagged on that PR:

> Adapter sets `dailyRevenue=dailyFees` (same amount), but 100% goes to
> holders. Missing `dailySupplySideRevenue`; should satisfy
> `dailyRevenue=dailyFees-dailySupplySideRevenue`.
>
> Resolution: Include `dailySupplySideRevenue=totalWei` with
> `dailyRevenue=0`, or set `dailyRevenue=0` if omitting
> `dailySupplySideRevenue`.

This PR follows the recommended resolution: emit
`dailySupplySideRevenue == totalWei` and leave `dailyRevenue == 0`.

## Why this is the correct income statement model

ClaimRush is a fair-launch onchain game on Base where every Crown
takeover pays a royalty fraction (25% of the takeover ETH) into the
`ShareholderRoyalties` contract for distribution to `veCLAIM`
holders. The protocol takes **no cut** of this royalty stream — no
treasury, no platform fee, no team allocation. Every wei flows
through to lockers.

In DefiLlama's income statement vocabulary:

- **Fees** = the gross royalty volume paid by takeovers (what the
  protocol's service captures).
- **SupplySideRevenue** = the portion paid to the supply side
  — here, `veCLAIM` holders who supplied $CLAIM by locking it into
  voting-escrow positions. They are the productive supply side: the
  ve-position is what earns royalty ETH.
- **Revenue** = the protocol's own retained margin. For ClaimRush
  this is **zero**.
- **HoldersRevenue** = same value as SupplySideRevenue, since the
  holders ARE the supply side in this design (locking $CLAIM is
  identically the productive supply activity).

The previous adapter routed the same `totalWei` into both
`dailyFees` and `dailyRevenue`, which DefiLlama's income statement
checker correctly read as "the protocol kept all of it" — wrong in
this case. After this patch:

- `dailyFees = totalWei`
- `dailyUserFees = totalWei`
- `dailyRevenue = 0` (intentionally empty `Balances`)
- `dailySupplySideRevenue = totalWei`
- `dailyHoldersRevenue = totalWei`

[...]

PR #7501, opened 2026-06-07
fees(claimrush): unify breakdown label across all four buckets

claimrush:
## Summary

Apply the descriptive `"Takeover Royalties To veCLAIM Holders"` label to `dailyFees`, `dailyUserFees`, and `dailyRevenue` (it was already in use on `dailyHoldersRevenue` since #7368). All four buckets now share one label string, so the per-protocol **Token Holder Net Income** card stops splitting a single income stream into two artificial sub-buckets.

## The bug on the protocol page

On `defillama.com/protocol/ClaimRush`, the **Token Holder Net Income** breakdown currently looks like this:

| Label | Amount | Reality |
|---|---|---|
| Takeover Royalties | **$14.37K** | Pre-#7368 data (May 30 – Jun 2) |
| Takeover Royalties To veCLAIM Holders | **$63.02K** | Post-#7368 data (Jun 2 onward) |
| **Sum** | **$77.39K** | = `total30d` ✓ |

Two label strings, one stream. The split is a pure label-history artifact — there is no second source of takeover royalties to veCLAIM holders.

## Root cause

`#7368` (merged 2026-06-02 06:34 UTC) renamed only the `dailyHoldersRevenue` label to `"Takeover Royalties To veCLAIM Holders"`:

```typescript
// fees/claimrush.ts on master (today):
dailyFees.addGasToken(totalWei, "Takeover Royalties");
dailyUserFees.addGasToken(totalWei, "Takeover Royalties");
dailyRevenue.addGasToken(totalWei, "Takeover Royalties");
dailyHoldersRevenue.addGasToken(totalWei, "Takeover Royalties To veCLAIM Holders");
//                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
//                                       the odd one out
```

Because the dimension store keys labels per data-point, pre-merge HoldersRevenue entries kept the original `"Takeover Royalties"` string while post-merge entries got the new one. The UI groups by string → two buckets.

## Fix

Make all four `addGasToken` calls use the same label, and update `breakdownMethodology` to match. The "Takeover Royalties To veCLAIM Holders" wording is more descriptive (source + destination in one phrase) and matches what was approved on HoldersRevenue.

```diff
- dailyFees.addGasToken(totalWei, "Takeover Royalties");
- dailyUserFees.addGasToken(totalWei, "Takeover Royalties");
- dailyRevenue.addGasToken(totalWei, "Takeover Royalties");
+ dailyFees.addGasToken(totalWei, "Takeover Royalties To veCLAIM Holders");
+ dailyUserFees.addGasToken(totalWei, "Takeover Royalties To veCLAIM Holders");
+ dailyRevenue.addGasToken(totalWei, "Takeover Royalties To veCLAIM Holders");
  dailyHoldersRevenue.addGasToken(totalWei, "Takeover Royalties To veCLAIM Holders");
```

No functional change to fees / revenue accounting — the same `totalWei` is attributed to the same dimensions. Only the displayed bucket label changes.

## Re-index ask

@bheluga — once merged, could we get a backfill re-index of `claimrush` so the historical pre-#7368 data points get re-labeled to the unified string? Otherwise the **Token Holder Net Income** card will continue showing two buckets until the old entries age out of the 30d window (and forever in the all-time / 1y views).

If a manual re-index isn't easy, I'm happy with the eventual aging-out path; flagging it because the bug is currently visible to anyone landing on the protocol page.

Thanks!

### DIFF
--- a/fees/claimrush.ts
+++ b/fees/claimrush.ts
@@ -40,9 +40,9 @@ const fetch = async (options: FetchOptions) => {
   }
 
   if (totalWei > 0n) {
-    dailyFees.addGasToken(totalWei, "Takeover Royalties");
-    dailyUserFees.addGasToken(totalWei, "Takeover Royalties");
-    dailyRevenue.addGasToken(totalWei, "Takeover Royalties");
+    dailyFees.addGasToken(totalWei, "Takeover Royalties To veCLAIM Holders");
+    dailyUserFees.addGasToken(totalWei, "Takeover Royalties To veCLAIM Holders");
+    dailyRevenue.addGasToken(totalWei, "Takeover Royalties To veCLAIM Holders");
     dailyHoldersRevenue.addGasToken(totalWei, "Takeover Royalties To veCLAIM Holders");
   }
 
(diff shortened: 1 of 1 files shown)


noateden:
hi @claimrush , we prefer to keep the current version, it's more clear!
"Takeover Royalties To veCLAIM Holders" should be labeled only in `dailyHoldersRevenue` as it explain the destination of Takeover Royalties

"Takeover Royalties" should be labeled to fees/revenue as it explain the source

noateden:
I will refilled data

claimrush:
Got it, @noateden — thanks for the source/destination clarification, that convention makes sense. Appreciate the data refill ahead of time; will watch the protocol page for the unified bucket on the holders side once the backfill lands.

PR #7131, opened 2026-05-22
[feat] llamapay - add volume adapter

MontaguSandwich:
Adds a volume adapter for LlamaPay (already listed, category Payments) at dexs/llamapay.ts. LlamaPay has a TVL adapter in DefiLlama-Adapters but no volume tracking what was actually paid out to Payees. Inflows tracks the delta between deposits and withdrawals in general ( which include Payer ones) and there is no way to back out payee withdrawal from it afaik.

### What it measures
LlamaPay streams ERC-20s by the second; a factory per chain deploys one streaming contract per token. The adapter sums the value **withdrawn by stream recipients** (`Withdraw` events) across every per-token contract (the value actually paid out to people). Covers 11 chains: Ethereum, Arbitrum, Optimism, Polygon, Gnosis, Avalanche, Base, Scroll, Sonic, Berachain, Metis. (BSC omitted: its getLogs is unreliable on the public RPCs CI uses.)

### Why withdrawals, not "value streamed"
Withdrawals are payee-triggered, so the obvious alternative is the value *streamed* (time-integral of the aggregate `amountPerSec`). I evaluated that and rejected it for two reasons:
1. **`createStream` is permissionless and unfunded**: the aggregate streaming rate is polluted by junk streams that never pay out.
2. **Streaming moves no tokens**: it's pure internal accounting. Tokens leave a contract only on withdrawal. So a withdrawal is the sole on-chain realization of value delivered to a recipient.

### No fees
LlamaPay takes no protocol cut (no fee logic in the contract), so no fees/revenue dimension is reported.

### Implementation notes
- Per-token contracts are enumerated via the factory at the **latest** block. The set only grows and each contract's token is immutable, so this is window-independent and avoids archive-state reads that public RPCs prune.
- `Withdraw(from, to, amountPerSec, streamId, amount)`: `amount` is the exact transfer in native token decimals (`amountToTransfer = delta·amountPerSec / DECIMALS_DIVISOR` in `LlamaPay.sol`), summed directly.
- v2, `pullHourly`. No new deps, no shared-helper changes.

### Validation
- `pnpm test dexs llamapay 2026-05-08` → **$24.86k** aggregate (Ethereum $13.5k, Arbitrum $9.6k, Optimism $1.5k, Base $181, …)
- other dates: 2026-04-20 → ~$19.3k, 2026-05-14 → ~$11.7k
- `pnpm run ts-check` ✓ · `pnpm run ts-check-cli` ✓

### Links
- Website: https://llamapay.io
[...]

PR #7506, opened 2026-06-08
[feat] streamflow - add volume adapter

MontaguSandwich:
Adds a volume adapter for Streamflow on Solana, covering all four Streamflow programs: Stream (vesting/payments, `strmRqUC...`), Aligned Unlocks (token-launch vesting, `aSTRM2NK...`), and the two airdrop distributors (Aligned Distributor `aMERKpFA...` + MerkleDistributor `MErKy6nZ...`).

**Methodology**: For each transaction that invokes a Streamflow program in the day window, sum SPL transfers within that tx where the source token account's owner (`from_owner`) is not the tx signer. The signer is the sender wallet on Create/Deposit IXs — `from_owner = signer` — so deposits are naturally excluded. On Withdraw/Cancel IXs the signer is either the recipient (self-claim) or a crank bot (batch claim), and `from_owner` is the escrow PDA — so all withdrawals are kept regardless of who triggered them. Streams are pre-funded at creation, so the apportioned recipient settlements in the window equal the true value-delivered. Per the streaming-protocol family lesson (LlamaPay #7131, Sablier #7351, Superfluid #7464): $-at-withdrawal is the economically real number, especially for volatile project-token vesting which is what dominates Streamflow's Solana flow.

**Data source**: Dune Analytics. Joins `solana.instruction_calls` (filter by Streamflow programs) with `solana.transactions` (extract `account_keys[1]` as signer) and `tokens_solana.transfers` (sum recipient-bound SPL movements). The existing fees adapter at `fees/streamflow/index.ts` uses Streamflow's Metabase `revenue-daily` endpoint; there's no `claims-daily` equivalent, hence the Dune path. Same pattern as every other Solana volume adapter in the codebase (pumpdotfun, bullx, photon, etc.).

**Scale**: ~$230k on 2026-06-06, ~$1.42M on 2026-06-04 — typical day-to-day variance for a vesting protocol where token-launch unlock days drive lumpy claims. USD-priced tokens only; many launch tokens streamed via Streamflow are not in DefiLlama's price index and contribute zero to the headline number (genuine undercount, not a bug). Implied scale from the existing fees adapter (~13.6 bps effective rate × $156k/30d fees) is ~$3.8M/day total throughput — the headline figure captures the priced-token subset of that.

Website: https://streamflow.finance
Twitter: https://x.com/streamflow_fi

### DIFF
--- a/dexs/streamflow.ts
+++ b/dexs/streamflow.ts
@@ -36,11 +36,9 @@ const fetch = async (options: FetchOptions) => {
   const programList = STREAMFLOW_PROGRAMS.map((p) => `'${p}'`).join(", ");
 
   // Two-layer filter for true recipient settlements:
-  //   1. The SPL transfer must have been EMITTED by a Streamflow program IX
-  //      (i.e. its parent in solana.instruction_calls has Streamflow as the
-  //      executing_account). This excludes transfers batched into the same tx
-  //      but unrelated to Streamflow -- the concern CodeRabbit raised on the
-  //      previous tx_id-only filter.
+  //   1. The SPL transfer must have been emitted by a Streamflow program IX.
+  //      Dune already exposes the parent program on tokens_solana.transfers as
+  //      outer_executing_account, so we do not need to join instruction_calls.
   //   2. The source token account's owner must NOT be the tx signer. The
   //      sender signs Create/Deposit IXs AND is the from_owner of the
   //      deposit transfer (sender wallet -> escrow PDA), so deposits drop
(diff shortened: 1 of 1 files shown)


MontaguSandwich:
Fixed via ff77a2a. The previous tx_id-only filter was over-counting unrelated transfers batched into Streamflow txs, as flagged.

Tightened to a two-layer filter:
1. The SPL transfer must have been EMITTED by a Streamflow IX — join `tokens_solana.transfers` to `solana.instruction_calls` on `(tx_id, outer_instruction_index, inner_instruction_index)` and require `outer_executing_account IN streamflow_programs`.
2. The existing `from_owner != signer` is retained to exclude deposits (sender signs Create IXs and is the from_owner of the deposit transfer).

Empirical impact on the same dates I previously validated:
- 2026-06-04: $1.42M → $1.36M (-4%, high-volume launch day, little noise)
- 2026-06-06: $230k → $117.96k (-49%, quiet day where ~half was bundled-tx noise)

The asymmetry confirms the concern was real on quiet days. The cancel-refund leg of Cancel IXs remains documented as a minor accepted over-count (separating it from the recipient-settlement leg would require IX-data discriminator decoding).

noateden:
hi @MontaguSandwich I optimized the query by removing the JOIN, can you review? I tested and got same numbers

MontaguSandwich:
LGTM, thanks for the polish!

PR #7530, opened 2026-06-09
[docs] guidelines - improve CodeRabbit review rules from recurring feedback

treeoflife2:
Updates the `GUIDELINES.md` files and `.coderabbit.yaml` so CodeRabbit reviews more effectively, based on recurring review offences observed across the past month of PRs.

## Global (`GUIDELINES.md`)
- `fetch` always takes a single `(options: FetchOptions)` argument for both v1 and v2; the old 3-arg v1 signature is removed and will fail; `fetch` must not return `timestamp`
- Use `FetchOptions` values (`dateString`, `startTimestamp`, `getFromBlock()`, etc.) instead of recomputing from a raw timestamp; no unused fetch args
- `getLogs({ targets: [...] })` over `noTarget`; readable `eventAbi` over raw `topics`; trades and fee transfers can be in different txs
- Comment/link hardcoded rates; justify `allowNegativeValue`
- Dune adapters must be `version: 1`; no duplicate Dune date filters; single prefetch query across chains
- v2 adapters key time windows off `startTimestamp`, not `startOfDay`
- Adapter shape preference: `chains` array (same start) vs `adapter` object (per-chain start) vs `chainConfig` passed as `adapter: chainConfig` (per-chain config), with examples
- `methodology` keys use display names (`Fees`, `Volume`), not field names

## Fees (`fees/GUIDELINES.md`)
- Income statement identities must balance; don't dump all fees into revenue; split protocol vs holders revenue; non-governance stakers are supply-side; negative supply-side is a red flag
- Buyback/TCG pattern subtracts from fees; avoid double-counting parent buybacks; need 24h not cumulative data; date-based contract switching on redeploys

## DEX (`dexs/GUIDELINES.md`)
- Trading-bot/router double-counting (buy+sell legs); separate-tx fees; prefer on-chain logs for EVM

## `.coderabbit.yaml`
- New `custom_checks`: Dune v1, income statement balance, FetchOptions usage, fetch signature, efficiency/error handling, methodology keys, adapter shape
- Tightened `fees/` and `dexs/` path instructions


fees/GUIDELINES.md
```diff
@@ -43,6 +43,38 @@ Think: "If the protocol became super-greedy and rugged all parties, how much cou
 | `dailyProtocolRevenue` | Optional | - | Portion allocated to treasury |
 | `dailyHoldersRevenue` | When applicable | Tokenholder Income | All value to token holders (buybacks, burns, distributions, external airdrops, bribes) |
 
+## Income Statement Identities (must always balance)
+
+These are the most common review failures. Verify them on every fees PR:
+
+- `dailyFees = dailyRevenue + dailySupplySideRevenue`
+- `dailyRevenue = dailyProtocolRevenue + dailyHoldersRevenue`
```
noateden: this may not true if we count dailyRevenue in one day, but count dailyHoldersRevenue in another day. for example: AAVE buy back in aave-v3

fees/GUIDELINES.md
```diff
@@ -43,6 +43,38 @@ Think: "If the protocol became super-greedy and rugged all parties, how much cou
...
+
+Rules that follow from the identities:
+
+- Do NOT dump all fees into `dailyRevenue`. Revenue is only the protocol's portion; the supplier portion is `dailySupplySideRevenue`.
+- When the protocol collects revenue for itself, also set `dailyProtocolRevenue`.
+- When revenue is split between the treasury and token holders (governance stakers), split it into `dailyProtocolRevenue` and `dailyHoldersRevenue`.
+- Stakers of a **non-governance** token (e.g. stablecoin or LST stakers) are suppliers - their cut is `dailySupplySideRevenue`, NOT `dailyHoldersRevenue`. Only governance/value-accrual token holders count as holders revenue.
+- A negative `dailySupplySideRevenue` is almost always a bug - investigate the source data before merging.
```
noateden: it's possible if some vaults realized loss? in that case, both `dailyFees` and `dailySupplySideRevenue` may be negative

fees/GUIDELINES.md
treeoflife2: Good catch. Reworded: dailyRevenue = dailyProtocolRevenue + dailyHoldersRevenue is now framed as an attribution rule, not a per-day equality, calling out that holders revenue (e.g. AAVE buy back in aave-v3) can land on a different day.

fees/GUIDELINES.md
treeoflife2: Updated: negative dailySupplySideRevenue is no longer called a bug. It now notes this can be legitimate when a vault/protocol realizes a loss (both dailyFees and dailySupplySideRevenue may be negative), so verify and use allowNegativeValue with a comment rather than assuming a bug.

.coderabbit.yaml
```diff
@@ -45,7 +45,25 @@ reviews:
         instructions: "For fees adapters: verify dailyFees represents Gross Protocol Revenue (all potential fees), dailyRevenue represents Gross Profit (protocol's portion), and dailySupplySideRevenue represents Cost of Funds (supplier payments). Ensure dailyRevenue = dailyFees - dailySupplySideRevenue conceptually."
       - name: "Version 2 Required"
         mode: "warning"
-        instructions: "New adapters must use version: 2 format. Check that the adapter exports version: 2 unless it's updating an existing v1 adapter."
+        instructions: "New adapters must use version: 2 format. Check that the adapter exports version: 2 unless it's updating an existing v1 adapter, or it relies on Dune / an external API that only returns daily aggregates."
```
bheluga, line 48: Mandate pullHourly key wherever version 2 is used
set it to false and explain the reason in case it cant be pulled hourly

GUIDELINES.md
```diff
@@ -8,24 +8,110 @@ These guidelines apply to ALL adapters in this repository.
...
+| **Fetch signature** | `(options: FetchOptions)` | `(options: FetchOptions)` (same as v2 - the old 3-arg signature is removed) |
 | **Time range** | Arbitrary start/end timestamps | Fixed day (00:00–23:59 UTC) |
 | **`pullHourly`** | Supported | Not supported |
 | **Preference** | Always prefer this | Use only when v2 is not possible |
 
+### Version Rules
+
+- **Dune adapters must always be `version: 1`.** Dune queries run once per day; a v2 adapter runs every hour and would re-run the same expensive query each hour.
+- For v2 adapters, drive time windows off `options.startTimestamp`/`options.fromTimestamp` (and `options.getFromBlock()`), NOT `options.startOfDay`. Because v2 runs hourly, keying off start-of-day sends the same request every hour and breaks granular/hourly data.
+- New adapters must be `version: 2` unless they rely on Dune or on an external API that only returns daily aggregates.
```
bheluga, line 53: same here

bheluga:
@treeoflife2 dont think there is mention of doubleCounted flag anywhere in the guidelines?

.coderabbit.yaml
treeoflife2: Done. Added a 'pullHourly Required For Version 2' custom check: every v2 adapter must explicitly set pullHourly, default true, and pullHourly: false is only acceptable with a comment explaining why it can't be pulled hourly.

GUIDELINES.md
treeoflife2: Done. Version Rules now mandates an explicit pullHourly key on every version: 2 adapter (default true; false + reason when hourly pulling isn't possible), and the Adapter Version table row reflects this.

PR #7546, opened 2026-06-09
fix(pool-party): price CC via CoinGecko slug, align volume with primary UI

SendPeter:
Follow-up to the TVL fix in DefiLlama/DefiLlama-Adapters#19355.

Two changes:

1. `addGasToken` doesn't resolve a CC price on Canton, so every CC leg of volume and fees was being silently zero'd (today: $20.5K reported vs ~$70K actual). Swapping to `addCGToken("canton-network", ...)` — same approach as the merged TVL adapter — recovers the missing CC contribution.

2. Removing the `divisor = 2` halving on volume. Pool Party's primary UI at cantonwallet.com/pools/ displays the raw per-token sum across both swap legs as its headline 24h volume, and we want the DefiLlama listing to mirror that for users coming from one to the other. Methodology updated to make this explicit. Fees unchanged on halving (they were never doubled).

After both changes the adapter reports ~$70K/day, matching what cantonwallet.com shows in real time.

### DIFF
--- a/dexs/pool-party/index.ts
+++ b/dexs/pool-party/index.ts
@@ -3,18 +3,18 @@ import { FetchOptions, SimpleAdapter } from "../../adapters/types";
 import fetchURL from "../../utils/fetchURL";
 
 /*
- * Pool Party — Canton Network AMM (volume)
+ * Pool Party — Canton Network AMM (volume + fees)
  *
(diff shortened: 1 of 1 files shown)


bheluga:
@SendPeter thanks for the PR

PR #7549, opened 2026-06-10
[feat] Provenance-Stats: track user stats for provenance via explorer

Shaileshkhote:
https://defillama.com/chain/provenance


bheluga (commented):
@Shaileshkhote any idea why new users > active users almost every day?

Shaileshkhote:
> @Shaileshkhote any idea why new users > active users almost every day?

Gm, I checked it and havent found any concrete reason for the same, also the chain doesnt seem to have any other explorer, so we can skip the new-users tracking for this specefic chain

bheluga:

PR #7561, opened 2026-06-10
fix(basenames): wire breakdown methodology

ErebusCipher:
Wires the existing `breakdownMethodology` object into the Basenames adapter export.

The labels are already applied to the balances and the breakdown text is already defined in the file; this just exposes it through the adapter object so the methodology can surface correctly.


bheluga:

PR #7563, opened 2026-06-10
[feat] Morph-Stats: track stats for morph l2 chain via socialscan api

Shaileshkhote:
https://defillama.com/chain/morph


bheluga:

PR #7566, opened 2026-06-11
[fix] Ondo-Perps: track volume via klines and fix start date

Shaileshkhote:
## Summary
- Updates Ondo Perps volume to use 1D candlestick data instead of the rolling volume endpoint.
- Keeps historical missing markets as zero-volume entries when no matching candle exists.

## Changes
- Updated `dexs/ondo-perps` to fetch markets from `/v1/markets` and aggregate `c * v` from `/v1/perps/history`.
- Added PromisePool fanout for per-market candle requests.
- Updated `open-interest/ondo-perps.ts` to use `fetchURL`.

## Tests
- `pnpm test dexs ondo-perps 2026-05-11`


bheluga:

PR #7569, opened 2026-06-11
[fix] Core-Markets: mark as dead

Shaileshkhote:
- Mark the adapter as dead from `2026-06-10`
- Official endpoint is unreachable
- Twitter dead, Subgraph dead
- Markets were symm.io deployed but symm.io dont have traces about blast subgraph to trac via factory
- Wasnt mapped in `defillama-server`


bheluga:

PR #7582, opened 2026-06-11
optimised pump-swap query

0saurabh0:
Part of https://github.com/DefiLlama/dimension-adapters/issues/6265

- the problem was that the `pools` CTE scanned all `pumpdotfun_solana.pump_amm_evt_createpoolevent` rows and most are filtered out later anyway.


bheluga:
@0saurabh0 thanks for the PR

PR #7604, opened 2026-06-12
refactor(tristero): use ClickHouse indexer for v3 send volume

treeoflife2:
## Summary

- Replace ERC20 Transfer log scan + `getTransactions` RPC calls with `evm_indexer.transactions` ClickHouse query for Tristero v3 `send()` volume
- Filter `status = 'success'` to exclude reverted transactions (the previous ERC20 approach skipped these naturally)
- Convert `ORDER_ROUTER_ABI` and `ESCROW_ABI` JSON arrays to human-readable `Interface` strings
- Remove `pullHourly: true` from adapter export
- Remove unused `includedOrderTypes` config field and related helper functions (`isIncludedOrderType`, `isTristeroV3RouterTransferLog`, `addTristeroV3OrdersToBalances`)
- Net: ~41 insertions, 214 deletions

Validated against 2026-06-08 data: 2309 successful `send()` transactions decoded, ~$1.26M USDC daily volume on Arbitrum + Base.

Supersedes #7507 (cross-fork push denied despite `maintainerCanModify: true`).

## Test plan

- [ ] `yarn test dexs/tristero` passes
- [ ] Daily volume for 2026-06-08 shows non-zero on arbitrum and base


treeoflife2:
@ggarza5 numbers don't match for Open interest and volume, 3m$ volume on 76$ oi looks wrong.

====== TOTAL DAILY AGGREGATED (sum of slots per chain) ======

chain     | Open interest at end
---       | ---
arbitrum  | 5.44 k
base      | 76.99
ethereum  | 0.00
Aggregate | 5.52 k

ggarza5:
@treeoflife2 Thanks for checking this. I may be misunderstanding the expected methodology, so please overrule me here.

My current understanding is that volume is daily router/order flow, including non-margin DEX flow plus margin opens/closes, while OI is only outstanding margin exposure at the end of the window. So I would not expect them to match 1:1.

Also, the `76.99` Base OI seems to come from the CLI’s `TOTAL DAILY AGGREGATED (sum of slots per chain)` table. Since this is hourly, that sums 24 hourly OI snapshots; `76.99 / 24 ~= 3.21`, which looks like the actual end/single-slot Base OI.

If DefiLlama expects this adapter to count only margin-related volume, or to handle hourly OI differently, happy to adjust.

PR #7609, opened 2026-06-13
track active and news users for cronos

DaniiRix:
https://defillama.com/chain/cronos

| Cronos | DUNE | 2455 users | 127 users |


noateden (changes requested):
can you check if we can use Allium instead?

DaniiRix:
> can you check if we can use Allium instead?

allium doesn't support cronos

noateden (approved):

PR #7615, opened 2026-06-13
fix(pool-party): handle new bridged tokens (CBTC, USDC.B, FRXUSD.B), prevent crash

SendPeter:
Follow-up to #7546.

Pool Party launched new pools with new tokens (`CBTC`, `USDC.B`, `FRXUSD.B`) that the volume adapter's `addTokenAmounts` was throwing on, so every fetch failed and DefiLlama recorded null daily volume for the past several days (missing bars on the chart).

Changes:

1. Maps the new tokens: `CBTC` → `coinbase-wrapped-btc` (CoinGecko), `USDC.B` and `FRXUSD.B` → `$1` stable proxy via `addUSDValue`.
2. Replaces the `throw` on unknown tokens with a silent skip — Pool Party's pool composition can change without an adapter update, and we prefer an undercount to a crashed adapter.
3. Methodology updated to list all four stables and CBTC pricing explicitly.

Note: cbBTC's canonical on-ledger `instrumentId.id` is `CBBTC.B`, but Send's public volume API exposes it as `CBTC` — that's what we map. Treating both as the same Coinbase Wrapped BTC token.

Long-term plan: migrate this adapter to CCTools (mirroring the TVL adapter in DefiLlama/DefiLlama-Adapters#19355) once CCTools' `/markets/send/pools` endpoint exposes `volume24hUsd` and `fees24hUsd` per pool. This PR is an interim fix to restore the displayed volume.

### DIFF
--- a/dexs/pool-party/index.ts
+++ b/dexs/pool-party/index.ts
@@ -19,14 +19,24 @@ import fetchURL from "../../utils/fetchURL";
 
 const VOLUME_URL = "https://api-mainnet.cantonwallet.com/canton/pool-party/public/v1/volume?period=24h";
 
-// Pool Party SDK instrument IDs (stable per Send Foundation):
-//   "Amulet"  : Canton Coin (CC) — priced via coingecko:canton-network
-//   "USDCx"   : bridged USDC on Canton — priced as USDC ($1 stable proxy)
-//   CUSD UUID : Send's privacy stablecoin — priced as USDC ($1 stable proxy)
-//
-// Brale issues multiple instruments (CUSD + SBC) from the same issuer party;
-// only the CUSD UUID below is in scope. Unknown instrument IDs throw.
+// Brale issues multiple instruments from the same issuer party; CUSD is
+// reported with this UUID instrument ID on the volume API.
 const CUSD_INSTRUMENT_ID = "481871d4-ca56-42a8-b2d3-4b7d28742946";
+
+// Token IDs that have a direct CoinGecko listing — priced via slug.
+const TOKEN_TO_CG: Record<string, string> = {
+    "Amulet": "canton-network",         // Canton Coin (CC)
+    "CBTC": "coinbase-wrapped-btc",     // cbBTC (Pool Party API uses "CBTC"; on-ledger instrumentId is CBBTC.B)
+};
+
+// Token IDs that are USD stablecoins without their own CoinGecko listing —
+// priced as USDC ($1 stable proxy).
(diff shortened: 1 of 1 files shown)


bheluga (dismissed):
@SendPeter thanks for the PR.
Instead of manually mapping everytime a new token pool is created , why not send cgId within api response?

SendPeter:
Good call — agreed that's the cleaner shape. We're already planning to migrate this adapter to CCTools (mirroring the TVL adapter via DefiLlama/DefiLlama-Adapters#19355), and I'll ask the CCTools team about adding a per-token `cgId` field as part of that scope. Will follow up here once I hear back.

This PR is an interim fix to restore the chart — Pool Party's recent token additions (`CBTC`, `USDC.B`, `FRXUSD.B`) were tripping the throw, so volume bars have been at $0 for several days. Pushed an update now addressing CodeRabbit's source-comment ask + swapped the unknown-token silent-skip for a `console.warn` so anything new surfaces in CI logs.

Can we land this as the bridge while I coordinate the CCTools change?

SendPeter:
Quick update — got CCTools to add a `coingeckoId` field per token on their pool endpoint. Opened DefiLlama/DefiLlama-Adapters#19694 which consumes it directly and drops the hardcoded TOKEN_TO_CG map entirely — the thin pass-through architecture you suggested.

One thing worth checking while we're here: CCTools is currently sending bare slugs (e.g., `canton-network`) on the new field, but they can switch to the full ref form (`coingecko:canton-network`) if that's preferred — it's a one-liner on their side. If you'd rather have the full ref, let me know and I'll relay it to them.

bheluga:
> Quick update — got CCTools to add a `coingeckoId` field per token on their pool endpoint. Opened [DefiLlama/DefiLlama-Adapters#19694](https://github.com/DefiLlama/DefiLlama-Adapters/pull/19694) which consumes it directly and drops the hardcoded TOKEN_TO_CG map entirely — the thin pass-through architecture you suggested.
> 
> One thing worth checking while we're here: CCTools is currently sending bare slugs (e.g., `canton-network`) on the new field, but they can switch to the full ref form (`coingecko:canton-network`) if that's preferred — it's a one-liner on their side. If you'd rather have the full ref, let me know and I'll relay it to them.

@SendPeter thanks,
canton-network works

SendPeter:
Sounds good. Just waiting on [#19694](https://github.com/DefiLlama/DefiLlama-Adapters/pull/19694) to merge. Thank you.

PR #7617, opened 2026-06-14
Add fees breakdown for BIM

noeroumeas:


bheluga:
@noeroumeas thanks for the PR

PR #7629, opened 2026-06-15
[feat] sablier-legacy - add volume adapter

MontaguSandwich:
### Volume adapter for Sablier Legacy (V1) — Ethereum

Sablier Legacy (the original V1 money-streaming protocol, predating Lockup/Flow) is already listed on DefiLlama (TVL + fees); this adds the missing volume dimension.

**Methodology.** Volume = value paid out to stream recipients = `WithdrawFromStream` amounts plus the recipient's accrued-but-unwithdrawn balance settled on `CancelStream` (`recipientBalance`); the sender's refund of the unstreamed deposit is excluded. This is the same recipient-settlement convention as the merged Sablier Lockup/Flow adapters (#7351). The legacy events carry the amount but not the stream token, so each settlement's token is resolved from the ERC20 Transfer the contract emits to the recipient in the same transaction.

**Data source.** On-chain logs across the v1.0.0 (`0xA4fc358455Febe425536fd1878bE67FfDBDEC59a`) and v1.1.0 (`0xCD18eAa163733Da39c232722cBC4E8940b1D8888`) contracts. The legacy subgraph (The Graph hosted service) is offline and Sablier's Envio HyperIndex covers only Lockup/Flow, so legacy actions are read directly from chain.

**Notes.**
- Ethereum only. The protocol is multi-chain, but legacy volume has wound down; other chains can be a follow-up.
- Legacy is a wind-down product — daily volume has declined from ~$305k (2022) to a few $/day today; the adapter backfills the full history.

Website: https://sablier.com · Twitter: https://x.com/Sablier

### DIFF
--- a/dexs/sablier-legacy.ts
+++ b/dexs/sablier-legacy.ts
@@ -15,6 +15,8 @@ import { CHAIN } from "../helpers/chains";
 // Neither the V1 subgraph (The Graph hosted service, shut down) nor Sablier's Envio
 // HyperIndex (Lockup/Flow only) index legacy actions, so this reads on-chain logs.
 
+// Sablier v1 mainnet deployments. Addresses from the official
+// sablier-labs/legacy-subgraph (networks/mainnet.json); both verifiable on Etherscan.
 const CONTRACTS = [
   "0xA4fc358455Febe425536fd1878bE67FfDBDEC59a", // Sablier v1.0.0
   "0xCD18eAa163733Da39c232722cBC4E8940b1D8888", // Sablier v1.1.0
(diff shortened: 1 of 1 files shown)


MontaguSandwich:
Line 30: added source comments for the contract addresses (from sablier-labs/legacy-subgraph) and labeled the Transfer topic hash.

Line 53: the noTarget query is window-local. It only covers the query's own block range, filtered to Transfers where `from` is the Sablier contract, so it isn't a full-chain scan. Same pattern as `dexs/uniswap-v1.ts`. Deriving the token from a CreateStream map would actually be more expensive here, since legacy streams are created years before any given withdrawal window, so it would need to read all historical CreateStream events each run. Added a comment explaining this.

Line 77: switched `recipientSettlements` from a Set to a consumed counter, so each settlement matches at most one Transfer. A batched tx or a sender refund sharing the same recipient and amount can no longer be counted twice.

Line 96: Ethereum-only is intentional. Legacy volume has wound down and Ethereum is the only chain with meaningful activity left; the methodology now says so explicitly. Other chains can be a follow-up.

noateden (approved):

PR #7630, opened 2026-06-15
track active and new users for kaia

DaniiRix:
https://defillama.com/chain/kaia

| Kaia | DUNE | 25,009 users | 9,050 users |


noateden (changes requested):
can we use allium instead?

DaniiRix:
> can we use allium instead?

allium doesn't support kaia

noateden (approved):

PR #7631, opened 2026-06-15
track active and new users for opbnb

DaniiRix:
https://defillama.com/chain/opbnb

| opBNB | DUNE | 1.51M users | 329,489 users |


noateden (changes requested):
can we use allium instead?

DaniiRix:
> can we use allium instead?

allium doesn't support opbnb

noateden (approved):

PR #7637, opened 2026-06-15
[feat] p2p.me - add volume adapter

MontaguSandwich:
### Volume adapter for p2p.me (Base + Polygon)

p2p.me is a peer-to-peer fiat<->crypto on/off-ramp and payments app, already listed on DefiLlama (Payments) with a fees adapter. This adds the missing volume.

Methodology: on-ramp volume, the USDC delivered to buyers on completed buy orders (orderType 0). On a buy the protocol releases USDC from escrow to the buyer in the same transaction, so the amount is verifiable 1:1 on chain. Sell (off-ramp) and pay orders are excluded: their fiat legs settle off chain, and the on-chain USDC they involve is escrow funding already captured on the buy side, so including them would double count. This is the same single-sided, on-chain-settled basis as the merged zkp2p adapter, the sibling P2P ramp.

Data source: Dune, same tables as the fees adapter (fees/p2pme/index.ts). The order event has moved across contracts; the per-contract sources are time-sequential and partitioned at the boundary so they do not overlap:
- Polygon: BrokerFactory v1/v2 (retired 2024)
- Base: OrderProcessor 0xb36c... then OrderFlowFacet 0x4cad... (struct spellbook, then re-encoded raw logs)

Scale: ~$26-41k/day on Base (Polygon retired 2024).
Dune-backed, so the CI test step fails on the runner (no DUNE_API_KEYS), as expected for Dune adapters.

Website: https://p2p.me  Twitter: https://x.com/P2Pdotme

### DIFF
--- a/dexs/p2pme.ts
+++ b/dexs/p2pme.ts
@@ -1,0 +1,100 @@
+import { CHAIN } from "../helpers/chains";
+import { Dependencies, FetchOptions, SimpleAdapter } from "../adapters/types";
+import { queryDuneSql } from "../helpers/dune";
+
+// p2p.me: a peer-to-peer fiat<->crypto on/off-ramp + payments app on Base and
+// Polygon. We report on-ramp volume: the USDC delivered to buyers on completed
+// buy orders (orderType 0). On a buy the protocol releases USDC from escrow to the
+// buyer in the same transaction, so the value is verifiable 1:1 on chain.
+//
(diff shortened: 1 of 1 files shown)


MontaguSandwich:
Added a comment in `packed_buys` documenting the packed event layout. The offsets were verified by decoding the re-encoded OrderCompleted event: amount at byte 65 (word 2, USDC 6-decimals), orderType at byte 449 (word 14; 0=buy, 1=sell, 2=pay).

PR #7640, opened 2026-06-15
[fix] p2p.me - remove fees adapter (reported revenue is an unverifiable off-chain spread)

MontaguSandwich:
## Remove the p2p.me fees adapter: reported revenue is an unverifiable off-chain spread

`fees/p2pme/index.ts` reports **$32,244/30d** and **$436,339 all-time** as Fees, Revenue, and ProtocolRevenue (100% to the protocol). That figure is a modeled off-chain price spread, not an observed on-chain fee. On-chain, p2p.me's contracts take no fee, and the spread the adapter attributes to the protocol accrues to the order-filling merchants/market-makers. This PR removes it; if the protocol publishes an actual on-chain fee, an accurate adapter can replace this.

### What the adapter computes
Per chain + currency:

    matched = LEAST(sum buy_amount, sum sell_amount)
    revenue = matched * (avg_buy_price - avg_sell_price) / avg_sell_price

where `price = fiatAmount / amount` from the OrderCompleted events. Reconstructing this exact formula on the raw Base events reproduces the reported number (~$33k/30d). On the current packed-event contract alone it is $139,701 all-time; the $436k additionally spans the retired Polygon BrokerFactory and Base OrderProcessor sources.

### Six on-chain tests for a real fee, all negative
| # | Test | Result |
|---|---|---|
| 1 | Fee field in the event | None. Decoded the full 40-word OrderFlowFacet struct and the OrderProcessor struct. Only `amount` (USDC) and `fiatAmount` (off-chain) carry value; no fee/feeBps/commission field. |
| 2 | Per-trade skim (buyer receives less than `amount`) | None. 8,310 buys/30d: event `amount` sum = $871,713 vs USDC actually paid out in those txs = $874,679. Buyers receive the full amount; 0 underpaid. |
| 3 | A net-positive fee-sink address | None. Largest non-order USDC recipient `0x42af7b2453cdbfdf51a1ce4238b514f5128cfbfe` received $83,091 and returned $83,346 (net ~ $0) = market-maker churn. Other net recipients are merchants receiving escrow settlement inside buy-order txs, not fees. |
| 4 | Fee in another token | None. Zero non-USDC ERC-20 transfers out of the contract in 90d. |
| 5 | Fee on the retired OrderProcessor `0xb36c5a84ee9aa2ff341f0729f2797976f585d3fb` | None. Its struct has no fee field either. |
| 6 | Inflow ~ a steady % of volume (a fee rate) | No. Candidate inflow/volume ratios swing 0.2%-16% and land on round-dollar amounts = settlement/rebalancing, not a bps fee. |

**Structural point:** the diamond is net USDC-negative (-$103,311/30d, -$35,061/90d). It drains a pre-funded escrow float. A fee-taking contract retains USDC (in > out); this one pays out more than it takes in, so it cannot be skimming a fee.

### The magnitude is also unstable
The reconstructed daily take-rate (`revenue / matched`) swings 1.87% to 5.57% over 14 days (mean ~3.9%, ~3x range), the signature of two independently-drifting off-chain FX price series rather than a fixed protocol fee.

### Reproducibility
- DefiLlama figure: `api.llama.fi/summary/fees/p2p.me`
- Base: OrderFlowFacet diamond `0x4cad6eC90e65baBec9335cAd728DDc610c316368`, OrderCompleted topic0 `0x507539023a7b6a713438d0f44eab4f97bcf8905b183b1108148409a8e8c1ed8c`; retired OrderProcessor `0xb36c5a84ee9aa2ff341f0729f2797976f585d3fb`; USDC `0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913`; market-maker wallet `0x42af7b2453cdbfdf51a1ce4238b514f5128cfbfe`
- Event decode (Dune): `amount = bytearray_to_uint256(bytearray_substring(data,65,32))/1e6`, `fiatAmount = ...(data,97,32)/1e6`, `orderType = ...(data,449,32)` (0=buy, 1=sell, 2=pay). All figures reproduce from `base.logs` plus the USDC Transfer topic `0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef`.


treeoflife2:
@MontaguSandwich If they don't have revenue, from where do they pay the affiliates? https://dune.com/p2pme/latest 
It would be good if we fix the adapter instead of removing it

MontaguSandwich:
> @MontaguSandwich If they don't have revenue, from where do they pay the affiliates? https://dune.com/p2pme/latest It would be good if we fix the adapter instead of removing it

I went through the dashboard queries again and you're right, sorry. There is an on-chain fee: the OrderFlowFacet emits an AdditionalOrderDetails event (topic0 0xdac1652b...) carrying fixedFeePaid and tipsPaid, which the OrderCompleted struct doesn't include. So the "no on-chain fee" framing in the PR is wrong and I'm converting this to a fix.

Decoding that event from base.logs (latest record per orderId, since it re-emits during an order): fixedFeePaid is about $9.4k all-time and $837 over the last 30 days, across the ~80k orders that carry a fee, roughly a flat $0.05 per order.
On the affiliates: they are paid on-chain via reputationmanager ClaimRecommendationRevenue, about $77.8k in USDC. That's roughly 8x the fixedFee total, so the referral pool is a share of the merchant spread, not the fixed fee.

the Total protocol revenue query computes turnover * (avg_buy_price - avg_sell_price)/avg_sell_price, about $457k, and the Merchants daily revenue query models the merchant commission at 2% of volume, about $380k. Those are the same ~2% of volume, so that spread is the merchants' commission (supply side), not protocol revenue.

So imo an accurate adapter should book dailyProtocolRevenue from the on-chain fixedFeePaid, with tips on the supply side. I can either report just the on-chain fee, or also surface the merchant spread as dailySupplySideRevenue while keeping it out of protocol revenue. Wdyt?

PR #7642, opened 2026-06-15
[feat] turboflow - add perp volume and prediction markets adapters

erranceL:
## Summary
- Add TurboFlow Perp Volume adapter under `dexs/turboflow-perps`.
- Add TurboFlow Prediction Market Volume adapter under `dexs/turboflow-prediction-markets`.

## Methodology
- Perp Volume reports `perpVolumeUsd` from TurboFlow's production metrics API.
- Prediction Market Volume reports `predictionMarketVolumeUsd`, which currently includes event contracts and football markets.

## Data Source
All adapters use TurboFlow's public DefiLlama metrics endpoint:

```
https://apis.turboflow.xyz/defillama/metrics?date=YYYY-MM-DD
```

The endpoint returns protocol-level daily metrics. The adapters keep the same BSC + Solana custody footprint as the TurboFlow listing while emitting protocol-level values once to avoid double-counting.

## Scope
- Fees / Revenue are not included in this PR.
- TVL / Treasury are not touched in this PR.
- Users / DAU are not included in this PR.

## Test plan
- `pnpm test dexs turboflow-perps`
- `pnpm test dexs turboflow-prediction-markets`


erranceL:
@coderabbitai review

Addressed the two actionable comments in the latest commit:
- Added `pullHourly: false` to the three TurboFlow v2 adapters because the source API is daily-only.
- Added strict required-field parsing in `helpers/turboflow.ts` so invalid or missing API fields throw instead of being coerced to zero.
- Returned an empty `dailySupplySideRevenue` balance from the fees adapter to make the income statement equation explicit.

The GitHub `test` and `ts-check` workflows are passing on the latest commit.

erranceL:
@coderabbitai The `version: 2` here is intentional and I believe the request to switch to `version: 1` is a regression. A few points:

1. This exact adapter config (`version: 2` + `pullHourly: false`) was already approved by you on commit `07f3ea98d`. The latest force-push only removed the fees adapter and rewrote commit authorship; the two volume adapters were not changed.
2. Your own pre-merge check "Version 2 Required" passes for these adapters and states they "correctly use version: 2". DefiLlama requires new adapters to be `version: 2`; switching to `version: 1` would conflict with that requirement.
3. In your first review you offered two options: switch to `version: 1`, **or** add `pullHourly: false` if `version: 2` is retained. I chose the latter, so the daily-only cadence is already handled and no hourly ingestion occurs.
4. The adapters rely on `version: 2` FetchOptions APIs (`createBalances`, `addUSDValue`, `options.chain`, `options.dateString`), so `version: 1` is not appropriate here.

Please re-review and resolve the two `version` comments.

erranceL:
@coderabbitai review

Thanks for confirming. The two version comments are withdrawn per your reply, and the optional-vs-required typing was already addressed in commit `07f3ea98d` (both fields are now required). Please re-review and approve so the changes-requested verdict is cleared.

erranceL:
@coderabbitai full review

Pushed commit `1d57783d4` which removes the unused `eventContractsVolumeUsd` and `footballVolumeUsd` fields from `TurboFlowMetrics` (your Issue 2). All checks are green. Please post a fresh review so the changes-requested verdict is cleared.

dexs/turboflow-prediction-markets/index.ts
```diff
@@ -0,0 +1,27 @@
+import { FetchOptions, SimpleAdapter } from "../../adapters/types";
+import { CHAIN } from "../../helpers/chains";
+import { fetchTurboFlowMetrics, shouldReturnProtocolMetrics } from "../../helpers/turboflow";
+
+const fetch = async (options: FetchOptions) => {
+  const dailyVolume = options.createBalances();
+  if (!shouldReturnProtocolMetrics(options)) return { dailyVolume };
+
+  const metrics = await fetchTurboFlowMetrics(options);
+  dailyVolume.addUSDValue(metrics.predictionMarketVolumeUsd);
```
treeoflife2, line 10: We also have notional volume for the prediction markets, which basically shows how many contracts got traded. Can we add that?

dexs/turboflow-prediction-markets/index.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  return { dailyVolume };
+};
+
+const adapter: SimpleAdapter = {
+  version: 2,
```
treeoflife2: if api returns daily results, then we should use the version 1 adapter, not the version 2.

treeoflife2: can we track fees too?

treeoflife2 (changes requested):

dexs/turboflow-prediction-markets/index.ts
erranceL: Done — switched both adapters to `version: 1` in commit `90837a087`, since the metrics API returns daily aggregates (`date=YYYY-MM-DD`). Removed the v2-only `pullHourly` flag.

erranceL: Not yet — our fees accounting isn't fully complete at the moment. We'll resubmit fees in a follow-up once the system has been iterated/upgraded.

dexs/turboflow-prediction-markets/index.ts
erranceL: Our metrics API doesn't expose the prediction-market contract count (notional volume) yet. We'll add this field on the backend and include it in a follow-up submission.

erranceL:
@treeoflife2 Thanks for the review. Addressed your comments:

- **version**: switched both adapters to `version: 1` (commit `90837a087`) since the API is daily-aggregate.
- **fees**: our fees accounting isn't fully complete yet; we'll resubmit fees in a follow-up after the next iteration.
- **notional volume (contract count)**: not exposed by our API yet; we'll add it on the backend and submit later.

Could you take another look when you have a moment?

erranceL:
@treeoflife2 friendly ping on this one. Quick recap of where things stand:

- version: both adapters are now `version: 1` (commit `90837a087`), per your suggestion since the metrics API serves daily aggregates. CodeRabbit has approved and CI (`ts-check`, `test`) is green.
- fees: our fee accounting isn't fully finalized yet, so we've kept this PR to volume only. We'll submit fees in a follow-up once it's ready.
- contracts traded / notional: our metrics API currently exposes USD turnover for prediction markets (`predictionMarketVolumeUsd`); a separate contracts-traded count isn't part of the API today. Happy to revisit in a follow-up if it'd be useful.

Would you mind taking another look so we can get the volume adapters merged? Thanks!

dexs/turboflow-prediction-markets/index.ts
erranceL: Quick clarification: the volume figure we expose for prediction markets is USD turnover (`predictionMarketVolumeUsd`); we don't currently expose a separate contracts-traded count in the metrics API. We can look at adding one in a follow-up if it's valuable.

dexs/turboflow-perps/index.ts
```diff
@@ -0,0 +1,26 @@
...
+import { fetchTurboFlowMetrics, shouldReturnProtocolMetrics } from "../../helpers/turboflow";
+
+const fetch = async (options: FetchOptions) => {
+  const dailyVolume = options.createBalances();
+  if (!shouldReturnProtocolMetrics(options)) return { dailyVolume };
+
+  const metrics = await fetchTurboFlowMetrics(options);
+  dailyVolume.addUSDValue(metrics.perpVolumeUsd);
+
+  return { dailyVolume };
```
treeoflife2, line 12: how are these volumes calculated? Does it include both ask and bid side volumes?

erranceL: Good question. The volume is reported single-sided — it does not double-count bid/ask (or maker/taker) sides; each trade is counted once at its executed notional.

What the daily figure does include is both the open and the close of a position, counted as two separate trades — e.g. opening a $10k position and later closing it contributes $20k of traded notional. This matches the convention used by other perp venues like GMX and gTrade, so no division is applied; the raw daily notional is the correct figure.

I've updated the methodology on both adapters to make this explicit. Happy to adjust if you'd prefer it handled differently.

treeoflife2: That's correct

treeoflife2 (approved):

PR #7655, opened 2026-06-16
Hanji: pull daily

jack-squirrel:
Removed pullHourly so volume aggregates per full UTC day


treeoflife2:
@jack-squirrel, volume will be aggregated once the day is over, so I don't think it makes sense to remove it.

PR #7660, opened 2026-06-16
durianfun-launchpad: track V4.6.7 factory (volume + fees)

Duriandotfun:
## Update: Durianfun Launchpad — track the V4.6.7 factory

`dexs/durianfun-launchpad/index.ts` indexes bonding-curve trading volume + fees across the V4.5 and V4.6.6 launchpad factories on Bitkub Chain (chainId 96). This adds the **V4.6.7** factory (current production launchpad, deployed 2026-05-31) that the adapter was missing — its trades were going untracked.

### Change (discovery only — fee/revenue logic untouched)
- Add factory `0x0480017E51dC813a0fad8aA73EAb2f8476ac0e8F` (deploy block 32196516).
- V4.6.7's `TokenCreated` appends a trailing `uint8 graduationTarget`, which changes the event hash, so its markets are discovered with an 8-arg ABI (the 7-arg ABI matches zero V4.6.7 events).
- `TokensBought` / `TokensSold` are **byte-identical** to V4.5 / V4.6.6, so the existing trade ABIs, the 90/10 protocol/creator split, and the methodology are reused unchanged — V4.6.7 markets just join the existing `markets` array.

### On-chain verification (KUB mainnet, block 32,663,300)
Discovered 4 V4.6.7 markets via the 8-arg `TokenCreated` (SAYWALLH, MOOPING, ICEMAN, THB) and decoded their trades with the existing ABI: **8 buys + 6 sells, 15.68 KUB volume, 0.183 KUB fees** all-time — currently untracked.

> Note: V4.6.7's fee is parameterized per-factory; the observed effective rate is ~1.17% (vs 1.0% on V4.5/V4.6.6). Total `Fees` stays exact (summed from the on-chain `fee` field); the 90/10 protocol/creator `Revenue` split is applied uniformly as the adapter already does for the older factories. Happy to special-case the V4.6.7 split if you'd prefer.

### DIFF
--- a/dexs/durianfun-launchpad/index.ts
+++ b/dexs/durianfun-launchpad/index.ts
@@ -13,7 +13,7 @@
  * liquidity is migrated to a `DurianAMM` pool (covered by a separate
  * adapter, `durian-amm`).
  *
- * Two generations of the factory are LIVE and indexed together so the
+ * Three generations of the factory are LIVE and indexed together so the
  * volume series is continuous:
  *
  *   V4.5   — `0xdf4f3dB298A9aDe853191F58b4b2a322D47EC005` (deploy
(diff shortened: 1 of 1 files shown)


bheluga:

PR #7664, opened 2026-06-16
Add Ophis fees adapter

san-npm:
Adds a fees adapter for [Ophis](https://ophis.fi) — an intent-based DEX aggregator built on CoW Protocol. Ophis holds no TVL, so this is a fees/revenue adapter only.

**Note: Ophis is a brand-new launch.** The partner fee went live in June 2026, so volume and fees are still ramping up. Two things follow from that: (1) early daily numbers will be small, and (2) on the CoW-hosted chains the partner fee is disbursed to the fee recipient **weekly** (only Optimism is self-hosted), so a single 24h test window can read ~0 even with volume — receipts arrive in weekly batches at the recipient address. The recipient has already done its first sweep on Optimism, so fees are flowing; the adapter will report real numbers as they accrue.

**Methodology:** every Ophis swap pays a flat partner fee (10 bps standard, 1 bp on stablecoin pairs) via CoW's CIP-75 partner-fee mechanism, paid in the swap token to a single deterministic CREATE2 fee recipient (0x858f0F5eE954846D47155F5203c04aF1819eCeF8), identical on every chain. dailyFees = dailyRevenue = USD value of all tokens received by that address per day (addTokensReceived). This is the on-chain received amount, already net of CoW Protocol's revenue share on CoW-hosted chains (Ophis self-hosts settlement only on Optimism).

**Chains:** ethereum, optimism, bsc, xdai, polygon, base, arbitrum, avax, linea, ink, plasma.

Tested locally with `npm test fees ophis <date>` — runs cleanly across all chains with no errors (real values require the LLAMA_INDEXER key, which CI has).

### DIFF
--- a/fees/ophis/index.ts
+++ b/fees/ophis/index.ts
@@ -1,0 +1,50 @@
+import { CHAIN } from "../../helpers/chains";
+import { FetchOptions, SimpleAdapter } from "../../adapters/types";
+import { addTokensReceived } from "../../helpers/token";
+
+// Ophis is an intent-based DEX aggregator (a CoW Protocol fork). It holds no TVL.
+// Every swap on every chain pays a flat partner fee (10 bps standard, 1 bp on
+// stablecoin pairs) via CoW's CIP-75 partner-fee mechanism. The fee is paid in
+// the swap's buy/sell token to ONE deterministic CREATE2 fee-recipient address,
+// identical on every chain.
+const FEE_RECIPIENT = "0x858f0F5eE954846D47155F5203c04aF1819eCeF8";
+
+// Fee activation date. There were no Ophis fees before this, so a slightly
+// earlier start is harmless.
+const START = "2026-06-08";
+
+const fetch = async (options: FetchOptions) => {
+  // dailyFees = dailyRevenue = USD value of all tokens RECEIVED by the
+  // fee-recipient address on this chain for the day. This figure is already net
+  // of CoW's revenue share on CoW-hosted chains (Ophis self-hosts only on
+  // Optimism); option A deliberately reports the on-chain received amount and
+  // does not estimate gross volume.
+  const dailyFees = await addTokensReceived({ target: FEE_RECIPIENT, options });
+  return { dailyFees, dailyRevenue: dailyFees };
+};
+
+const methodology = {
+  Fees: "Flat partner fee taken on every Ophis swap (10 bps standard, 1 bp on stablecoin pairs), measured as the USD value of all tokens received by the deterministic CREATE2 fee-recipient address 0x858f0F5eE954846D47155F5203c04aF1819eCeF8, which is identical on every chain.",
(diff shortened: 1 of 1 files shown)


bheluga (commented):
@san-npm thanks for the PR
Have you not received any fees yet?

san-npm:
@bheluga thanks for taking a look! Yes, Ophis is brand new: the partner fee went live about a week ago, so volume and fees are still early.

One thing worth flagging on the methodology side: Ophis is a CoW Protocol fork, and on the CoW-hosted chains the partner fee is disbursed to our fee recipient **weekly** (only our Optimism deployment is self-hosted). So any single 24h window will read ~0 even once there's steady volume, because the receipts land in weekly batches at the recipient address rather than per-swap. We have already started earning (the recipient has done its first sweep on Optimism). The adapter just tracks the on-chain tokens received by that recipient, so it will report real numbers as they accrue.

Happy to proceed whenever you are, and glad to note in the description that it's a new launch if that helps. Thanks!

bheluga:
> @bheluga thanks for taking a look! Yes, Ophis is brand new: the partner fee went live about a week ago, so volume and fees are still early.
> 
> One thing worth flagging on the methodology side: Ophis is a CoW Protocol fork, and on the CoW-hosted chains the partner fee is disbursed to our fee recipient **weekly** (only our Optimism deployment is self-hosted). So any single 24h window will read ~0 even once there's steady volume, because the receipts land in weekly batches at the recipient address rather than per-swap. We have already started earning (the recipient has done its first sweep on Optimism). The adapter just tracks the on-chain tokens received by that recipient, so it will report real numbers as they accrue.
> 
> Happy to proceed whenever you are, and glad to note in the description that it's a new launch if that helps. Thanks!

I dont see any txn on optimism after start date
https://optimistic.etherscan.io/address/0x858f0f5ee954846d47155f5203c04af1819ecef8 ?

san-npm:
Thanks @bheluga, and good catch on the explorer view. The disbursements do land, but they arrive as **internal transactions** from CoW's Settlement contract, so they show under the address's "Internal Txns" tab rather than the main "Transactions" list, which is why the main view looks empty.

The recipient `0x858f0F5eE954846D47155F5203c04aF1819eCeF8` does have real, if early, accrual since launch. As of today it holds ~0.00194 ETH on Optimism and ~0.264 xDAI on Gnosis (verifiable via the Internal Txns tab on each explorer). The amounts are small because the launch is ~two weeks old and volume is still ramping, and on the CoW-hosted chains the partner fee is swept weekly, so any single 24h window usually reads ~0 between sweeps.

One adapter note worth flagging: the current Optimism balance is in **native ETH** (from native-sell / eth-flow trades), and `addTokensReceived` tracks ERC-20 Transfer logs only, so it currently misses the native-coin portion of the fee. I am happy to extend the adapter to also count native-coin inflows to the recipient so the figures reflect every fee token, if that would help.

No rush on your end. Happy to keep the new-launch note in the description, and the numbers will populate as fees accrue and sweep. Thanks again for the careful review.

bheluga:
@san-npm thanks for the PR

PR #7666, opened 2026-06-16
Update ZINC fees methodology

kendall-man:
Updates ZINC fees to use gross SOL round volume.

Counts fees as 10% of gross SOL deployed in rounds.
Counts revenue as the same deploy fee amount.
Counts holders revenue from buyback vault SOL inflows.
Removes supply-side revenue reporting.

### DIFF
--- a/fees/zinc/index.ts
+++ b/fees/zinc/index.ts
@@ -1,120 +1,108 @@
 import { Dependencies, FetchOptions, SimpleAdapter } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
-import { queryAllium } from "../../helpers/allium";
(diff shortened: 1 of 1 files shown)


bheluga (changes requested):
@kendall-man thanks for the PR.
Few things to consider:
1. If fee methodology was modified in between, its good to keep both the versions of the code,
old code called before modification and new code ,after (its usefull for refill) .
2. We are against using API's for larger protocols when you can get data onchain,
the reason being it reduces transparency and sometimes when api gives wrong data, its harder to debug (right now api seems to be broken, it doesn't have data for the last few days), so please try to use onchain query

FelixBruguera:
Closing as stale

PR #7673, opened 2026-06-17
phoenix - fix dead volume source, migrate to Allium

DaniiRix:
The `dexs/phoenix` adapter pulled `dailyVolume` from a self-hosted AWS API Gateway endpoint (`9senbezsz3.execute-api.us-east-1.amazonaws.com`). That endpoint is dead — it now returns 403 Forbidden / HTTP 500, so the adapter has been failing.

## Fix

Source volume from Allium's `solana.dex.trades` (`SUM(usd_amount) WHERE project = 'phoenix'`) instead of the dead API. Removed the hardcoded API key and `httpGet` call.

Verified that Phoenix's orderbook fills are indexed in Allium: for 2025-06-14, $5.97M across 17,003 fills — same order of magnitude as Dune's `dex_solana.trades` for the same day, and a sane ~3.4 bps ratio against the existing `fees/phoenix` adapter.

Chain (`solana`) and start date (`2023-02-27`) unchanged.

### DIFF
--- a/dexs/phoenix/index.ts
+++ b/dexs/phoenix/index.ts
@@ -5,7 +5,7 @@ import { queryAllium } from "../../helpers/allium";
 const fetch = async (options: FetchOptions) => {
   const data = await queryAllium(`
     SELECT
-      SUM(usd_amount) AS daily_volume
+      COALESCE(SUM(usd_amount), 0) AS daily_volume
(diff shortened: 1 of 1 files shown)


bheluga:
@DaniiRix thanks for the PR

PR #7685, opened 2026-06-18
Monetrix: exclude insurance-fund reserve from fees revenue

MonetrixLab:
Follow-up to #7650.

The insurance share of distributed yield funds a protocol reserve that backstops the system — it is not protocol earnings. This keeps it in total **Fees** (via `totalYield`) but removes it from **Revenue / ProtocolRevenue**; only the foundation share and instant redeem/unstake fees count as revenue.

- Revenue / ProtocolRevenue: drop `insuranceShare`, keep `foundationShare` + instant exit fees
- Fees and SupplySideRevenue unchanged
- Methodology + breakdown labels updated (removed `Yield To Insurance Fund` from Revenue/ProtocolRevenue)


bheluga:
@MonetrixLab thanks for the PR

PR #7690, opened 2026-06-18
kinetiq-launch: add fees and revenue adapter

DorianWilhelm:
Adds a fees adapter for Kinetiq Launch LSTs, following the kHYPE methodology: staking yield is measured from the exLST<>HYPE exchange-rate delta and a 10% performance fee is taken on it, split 30% to the Kinetiq treasury and 70% to KNTQ buyback. Markets are discovered on-chain from the factory MarketBonded event so new LSTs are tracked automatically. HIP-3 trading fees are excluded as they settle on HyperCore via the CoreWriter precompile and emit no EVM logs.


bheluga:
@DorianWilhelm thanks for the PR

PR #7697, opened 2026-06-18
tapp-exchange: fix dead API, migrate to on-chain data

DaniiRix:
`api.tapp.exchange` is no longer usable. It resolves to a Kubernetes control plane and returns `403 system:anonymous` even when the self-signed certificate is ignored. Since the adapter depended on that API for volume and fee data, it now fails in the runner. As a result, TAPP data on DefiLlama has been flatlined since ~2026-05-30.

## Fix

The adapter was rewritten to calculate all metrics directly on-chain using Dune (`aptos.events`):

- Volume: derived from the output leg of each `router::Swapped` event, counting only one side of the swap to avoid double-counting. USD pricing comes from DefiLlama.
- Fees: calculated as pool USD volume × swap fee rate. Fee rates are loaded from each pool's `PoolCreated` event within the same query (CLMM rates use `1e6` precision; stable pools use `1e10` precision).
- Revenue split (ve(3,3), verified against contracts):
  - 33% platform fee (`PoolMeta.platform_fee_rate`) → `dailyProtocolRevenue`
  - 67% distributed to veTAPP voters → `dailyHoldersRevenue`
  - LPs receive TAPP emissions rather than swap fees, so `dailySupplySideRevenue = 0`

This also corrects the previous methodology, which incorrectly classified the 67% voter share as LP (supply-side) revenue.

The new implementation uses Dune as the sole data source, with no dependency on the protocol API or an Aptos full node.

## Verification

The adapter runs successfully (exit code `0`) with no `NaN` or negative values. Revenue identities hold throughout:

- `Fees = Revenue + SupplySideRevenue`
- `Revenue = ProtocolRevenue + HoldersRevenue`

Results closely match DefiLlama data before the outage (adapter run for date `D` covers the window labeled `D-1`):

| Day (DL) | DL Volume | Adapter Volume | DL Fees | Adapter Fees |
|----------|-----------|----------------|---------|--------------|
| 2026-05-27 | $145.6k | $147.2k (+1.1%) | $68 | $67 |
| 2026-05-28 | $262.6k | $267.7k (+1.9%) | $91 | ~$86 |
| 2026-05-30 | $45.3k | $46–53k | $18 | $18 |

Example breakdown (2026-05-28 run):

- Fees: $67
- Protocol revenue: $22
- Holder revenue: $45
- Supply-side revenue: $0

## Notes

- Swap fees use the rate defined at pool creation. While `stable::ApplyNewFee` can update fees, it does not include a `pool_addr`, and fee changes appear to be rare and small.
- The platform fee is consistently 33%, verified across all pools via `PoolMeta.platform_fee_rate`.

### DIFF
--- a/dexs/tapp-exchange/index.ts
+++ b/dexs/tapp-exchange/index.ts
@@ -1,63 +1,120 @@
-import { SimpleAdapter, FetchOptions } from "../../adapters/types";
+import { Dependencies, FetchOptions, SimpleAdapter } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
-import { postURL } from "../../utils/fetchURL";
-import { randomInt } from "node:crypto";
-import { getTimestampAtStartOfDayUTC } from "../../utils/date";
+import { queryDuneSql } from "../../helpers/dune";
 
-const URL = 'https://api.tapp.exchange/api/v1'
+const ROUTER = "0x487e905f899ccb6d46fdaec56ba1e0c4cf119862a16c409904b8c78fab1f5e8a";
+const CLMM = "0x5c2e5a4d1b355b939ab160c618ed5504a6e1addf109388aa3b83b73b207ab6c7";
+const STABLE = "0xa611a8ba7261ed1f4d3afe4ac2166fc9f3180103e3296772d593a1e2720c7405";
 
-
-interface TappDefillamaDimension {
-    dailyVolume: number;
-    dailyFee: number;
-    dailyRevenue: number;
-}
-
-const getDefillamaDimension = async (start: number, end: number) => {
-    const body = {
-        "method": "public/defillama_dimension",
-        "jsonrpc": "2.0",
-        "id": randomInt(1, 1000),
-        "params": {
-            "startTime": start,
-            "endTime": end,
-        }
-    }
-    const dimension: { result: TappDefillamaDimension } = await postURL(URL, body)
-    return dimension.result;
-}
+// Swap fee is set per pool at creation: CLMM fee over 1e6 (e.g. 0.05%/0.3%), stable over 1e10 (0.01%).
+// Protocol takes a uniform 33% cut (PoolMeta.platform_fee_rate); the other 67% goes to veTAPP voters.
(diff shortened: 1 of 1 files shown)


dexs/tapp-exchange/index.ts
```diff
@@ -1,63 +1,120 @@
...
+      const fees = (await bal.getUSDValue()) * swapFee;
+      const protocol = fees * PLATFORM_FEE;
+      const holders = fees - protocol;
+      dailyFees.addUSDValue(fees, "Swap Fees");
+      dailyRevenue.addUSDValue(protocol, "Swap Fees To Treasury");
+      dailyRevenue.addUSDValue(holders, "Swap Fees To veTAPP Voters");
+      dailyProtocolRevenue.addUSDValue(protocol, "Swap Fees To Treasury");
+      dailyHoldersRevenue.addUSDValue(holders, "Swap Fees To veTAPP Voters");
+    })
+  );
```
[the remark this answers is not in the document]
DaniiRix: fixed

dexs/tapp-exchange/index.ts
```diff
@@ -1,63 +1,120 @@
...
+};
+
+const methodology = {
+  Volume:
+    "Sum of the USD value of the output token of every swap.",
+  Fees: "Per-pool USD volume multiplied by the pool's swap fee rate.",
+  Revenue: "All swap fees: the protocol's platform-fee cut plus the share routed to veTAPP voters.",
+  ProtocolRevenue: "Platform fee taken to the protocol treasury.",
+  HoldersRevenue: "Remaining swap fees distributed to veTAPP voters via the pool gauges.",
+  SupplySideRevenue: "Liquidity providers are rewarded with TAPP emissions (incentives).",
```
bheluga, line 92: this depends on pool type right?
for amm, fees go to LPs and CLMM to veTAPP stakers?

DaniiRix: no, they all have similar gauge + fees_voting_reward setup

bheluga:
@DaniiRix thanks for the PR

PR #7698, opened 2026-06-18
near-intents: dead source, move to dune + near dashboard

DaniiRix:
The adapter was querying `revenue.near.org/api/total-fees` and `/api/revenue`, which no longer return JSON (the API has been removed and now serves HTML/404s), causing fetch failures with `Invalid API response format`.

This change rewires the adapter:

* **Fees** now come from the official NEAR Dune datasets (`dataset_near_intents_protocol_fees` + `dataset_near_intents_fees`), reproducing the dashboard's **Total generated fees** query (`dune:6162763`).
* Validated against the dashboard to the cent (e.g. **2026-02-28 = $129,430**).
* **Revenue** and **supply-side fees** are derived by applying the dashboard's published NEAR-retained ratio (`protocol_fee_near / total_fees_near`) to the exact Dune fee total.
* Falls back to reporting **fees only** when the revenue split is unavailable (before **2026-01-01**).

## Why use the dashboard split?

* Not reconstructable from on-chain data (the on-chain `fees_collected` value represents only the ~0.0001% protocol fee, not the ~14% retained share).

## Test

```bash
pnpm test fees near-intents 2026-02-28
# fees 111.93k, revenue 19.11k, supply-side 92.82k

pnpm test fees near-intents 2025-10-10
# pre-split: fees only
```

### DIFF
--- a/fees/near-intents/index.ts
+++ b/fees/near-intents/index.ts
@@ -1,66 +1,86 @@
-import { FetchOptions, SimpleAdapter } from "../../adapters/types";
+import { Dependencies, FetchOptions, FetchResult, SimpleAdapter } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
+import { queryDuneSql } from "../../helpers/dune";
 import fetchURL from "../../utils/fetchURL";
 
 /**
- * Previous source : (internal server error)
- * NEAR Intents Fees Adapter for DefiLlama
- * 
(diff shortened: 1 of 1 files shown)


bheluga:
@DaniiRix thanks for the PR
was fixed [here](https://github.com/DefiLlama/dimension-adapters/pull/7738)

PR #7155, opened 2026-05-23
feat(glif): add fees adapter for Filecoin iFIL pool

tarun-khatri:
## What

Adds a fees adapter for [GLIF](https://glif.io/) on Filecoin. GLIF runs an on-chain credit market where Filecoin Storage Providers borrow FIL from [InfinityPoolV2](https://filfox.info/en/address/0xe764Acf02D8B7c21d2B6A8f0a96C78541e0DC3fd) at 15% APR, and LPs earn that interest through ERC-4626 share-rate appreciation on [iFIL](https://filfox.info/en/address/0x690908f7fa93afC040CFbD9fE1dDd2C2668Aa0e0).

GLIF is already listed on DefiLlama for TVL (`glif/index.js`, category Liquid Staking, ~$30M Filecoin TVL). No fees adapter existed — this fills the gap.

## Methodology

Daily fee = `(convertToAssets(1e18)_today - convertToAssets(1e18)_yesterday) * iFIL.totalSupply / 1e18`

- 100% of interest flows to LPs via share-rate appreciation, so `dailyFees == dailySupplySideRevenue` and `dailyRevenue == dailyProtocolRevenue == 0`.
- Per GLIF's own indexer ([events.glif.link/pool/0/fees](https://events.glif.link/pool/0/fees)) the treasury fee balance reads ~9,957 FIL today and has barely moved in months while the pool earned > 1.5M FIL of interest — so substantially all SP interest is supply-side revenue.
- Docs: [Borrowing Cost (15% APR)](https://docs.glif.io/en/for-storage-providers/borrowing-cost), [iFIL Reward Mechanism](https://docs.glif.io/en/for-liquidity-providers/glif-reward-mechanism-ifil).

### 1. Local test (today)

```
FILECOIN
Daily fees:                3.61 k
Daily user fees:           3.61 k
Daily revenue:             0.00
Daily protocol revenue:    0.00
Daily supply side revenue: 3.61 k
```

Matches expected APY math: 32.8M iFIL × 5.82% APY / 365 × $0.70/FIL ≈ **$3.66k/day**.

### 2. Rule 44 random-sample (N=30, seed=42) across the pool's 17-month lifetime

Backfill verified via [events.glif.link/metrics/{height}](https://events.glif.link/metrics/6041000) (height-indexed pool state from GLIF's own indexer, since public Filecoin RPCs prune state to 14-336h).

| Window | Days | Implied APR |
|---|---:|---:|
[...]

PR #7703, opened 2026-06-18
fix(glif): repair syntax error that broke the adapter; accrue fees daily

tarun-khatri:
The glif fees adapter has a syntax error that prevents it from compiling, so it has not produced any data since it was added in #7155:

```ts
dailyRevenue.add(config.underlying, protocolFee 'Assets Yields To Protocol'); // missing comma
```

`tsc` rejects it (`TS1005: ',' expected`), so the module never loads and GLIF (TVL ~$26M across Filecoin + Base) shows no fees on DefiLlama. This adds the missing comma.

It also switches the adapter from hourly to daily accrual (drops `pullHourly`). The adapter measures fees from `convertToAssets` share-rate appreciation between two block snapshots. Run daily that is a single clean rate delta; run hourly it needs ~48 fine-resolution reads per day, and on Base the archival RPC / block-timestamp mapping is imprecise enough that the hourly deltas do not telescope to the daily delta. Measured for 2026-06-13 on Base:

- on-chain ground truth (raw `convertToAssets` at the day boundaries): ~$784 fees, 10% to protocol
- daily run: $905 fees, 10.0% to protocol (matches ground truth)
- hourly runs: $2,710 then $2,210 on repeat runs (non-deterministic, ~2.5–3x too high, 16–20% protocol ratio)

Filecoin telescopes fine either way (~$3.27k/day). Running daily makes both chains match the on-chain rate change and is consistent with the sibling share-rate yield adapters (origin-ether, origin-dollar), which also run daily.

Verification: `npx ts-node cli/testAdapter.ts fees glif <date>` now compiles and runs both chains — filecoin ~$3.27k/day and base ~$0.9k/day, each with revenue = 10% of fees and fees = supplySideRevenue + revenue. On-chain `treasuryFee()` on the Base pool returns `0.1e18` (10%), matching the methodology.

### DIFF
--- a/fees/glif/index.ts
+++ b/fees/glif/index.ts
@@ -93,7 +93,7 @@ async function fetch(options: FetchOptions): Promise<FetchResultV2> {
     dailyFees.add(config.underlying, totalFees, METRIC.ASSETS_YIELDS);
     dailySupplySideRevenue.add(config.underlying, lpYield, 'Assets Yields To LP');
     if (protocolFee > 0n) {
-        dailyRevenue.add(config.underlying, protocolFee 'Assets Yields To Protocol');
+        dailyRevenue.add(config.underlying, protocolFee, 'Assets Yields To Protocol');
     }
 
     return {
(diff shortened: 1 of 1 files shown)


bheluga:

PR #7707, opened 2026-06-18
feat: add Alchemix V3 fees adapter

tarun-khatri:
https://defillama.com/protocol/alchemix-v3

Adds a fees/revenue adapter for **Alchemix V3**, which is currently untracked for fees (the existing `fees/alchemix.ts` only covers the V2 alchemists via the `Harvest` event — a separate contract set).

### What it tracks
Alchemix V3 deploys its yield-bearing collateral as standalone ERC-4626 "MYT" vaults from a `VaultV2Factory` on each chain. Each vault accrues yield to its share price (`convertToAssets` grows over time). The adapter:
- discovers every vault from the factory's `CreateVaultV2` events (same source the V3 TVL adapter uses),
- measures each vault's daily yield from share-price growth via the standard `getERC4626VaultsYield` helper,
- splits it 10% protocol / 90% to users' self-repaying loans — the same split the existing V2 fees adapter applies to Alchemix yield.

Factories:
- Ethereum `0xdd56b00302e91C4c2b8246156bDEAa1cEDc58984` (deployed 2026-04-14)
- Arbitrum / Optimism `0x8c7C0C380bA4eE38461eb5a6b82e5d930EC11Ca2`

### Verification (on-chain)
- ETH USDC vault `0x9b44efca…`: `totalAssets` ≈ $7.96M; `convertToAssets(1e18)` grew 1006961 → 1007114 over ~1 day (~5.5%/yr) — real yield.
- ETH WETH vault `0x29bcfed2…`: appreciates on average but its share price can dip on a loss day, so `allowNegativeValue` is set.
- Sampled daily output (fees): 06-10 +$1.97k, 06-13 +$1.33k, 06-15 +$3.14k, 06-18 +$5.92k, with the occasional negative day (06-17) — net positive, as expected for ERC-4626 yield vaults.

These V3 vaults are a different contract set from the V2 alchemists, so there is no overlap with `fees/alchemix.ts`.

## Summary by CodeRabbit

* **New Features**
  * Added support for tracking Alchemix V3 ERC-4626 fee and revenue activity.
  * Includes per-chain coverage with automatic discovery of eligible vaults.
  * Reports daily fees, protocol revenue, and supply-side revenue with hourly updates enabled.

* **Bug Fixes**
  * Improves revenue calculations by accounting for performance fees and direct protocol fee transfers.
  * Handles missing fee settings with a safe fallback and supports negative-value edge cases.


tarun-khatri:
Addressed the review:
- Consolidated all per-chain values into a single `chainConfig` ({ start, factory, fromBlock }) and set `adapter: chainConfig`.
- Added source comments for the factory addresses / fromBlocks (same source as the V3 TVL adapter) and for `PROTOCOL_FEE` (10%, matching the existing V2 fees adapter).

On `pullHourly`: left it off intentionally. Yield here is measured from each vault's `convertToAssets` share-price delta over the period, so a daily cadence is the right granularity — this matches the existing `getERC4626VaultsYield` adapters (e.g. `fees/upshift`) which run daily.

tarun-khatri:
@coderabbitai review

tarun-khatri:
Thanks for the review. Addressed:

- **pullHourly**: set `pullHourly: true` explicitly on the v2 adapter.
- **Per-chain config**: already consolidated into a single `chainConfig` map (`start`/`factory`/`fromBlock` per chain, used directly as `adapter`).
- **Source comments**: factory addresses link to the V3 TVL adapter that enumerates the same vaults, `fromBlock` is the factory deployment block, and the 10% `PROTOCOL_FEE` matches the split the existing `fees/alchemix.ts` applies.

@coderabbitai review

fees/alchemix-v3.ts
```diff
@@ -0,0 +1,90 @@
...
+// fromBlock = the factory deployment block on each chain (ETH factory deployed 2026-04-14).
+const chainConfig: Record<string, { start: string; factory: string; fromBlock: number }> = {
+  [CHAIN.ETHEREUM]: { start: "2026-04-14", factory: "0xdd56b00302e91C4c2b8246156bDEAa1cEDc58984", fromBlock: 24875892 },
+  [CHAIN.ARBITRUM]: { start: "2026-04-14", factory: "0x8c7C0C380bA4eE38461eb5a6b82e5d930EC11Ca2", fromBlock: 452291175 },
+  [CHAIN.OPTIMISM]: { start: "2026-04-14", factory: "0x8c7C0C380bA4eE38461eb5a6b82e5d930EC11Ca2", fromBlock: 150271732 },
+};
+
+// Alchemix takes 10% of generated yield; the other 90% repays users' loans.
+// Same split the existing V2 adapter (fees/alchemix.ts) applies to Alchemix yield.
+const PROTOCOL_FEE = 0.1;
```
bheluga: Performance fees is 15% (https://docs.alchemix.fi/user/concepts/fees & https://etherscan.io/address/0x29bcfeD246ce37319d94eBa107db90C453D4c43D#readContract)
They also seem to charge other types of fees according to docs

bheluga (dismissed):

tarun-khatri:
Thanks, fixed in fbec1a993.

What changed:
- Removed the hardcoded 10% V2 yield split for V3.
- Read each MYT vault's on-chain `performanceFee()` instead.
- Since VaultV2 `convertToAssets()` already accounts for performance/management fee dilution, the adapter now treats share-price growth as net depositor yield and grosses up positive yield by `1 / (1 - performanceFee)`. Protocol revenue is the difference; negative yield days are not grossed up.
- Added realized Alchemist/Transmuter protocol fees transferred to the fee receiver, including Transmuter early-exit synthetic fees.
- Switched back to daily pulls because the hourly runner was hitting the existing SDK `Cannot read properties of undefined (reading 'success')` failure in this adapter path.

On-chain checks I used:
- All 6 current MYT vaults return `performanceFee = 150000000000000000` (15% WAD) and `managementFee = 0`.
- Alchemist proxies return `protocolFee = 25` bps; Transmuters return `exitFee = 100` bps and `transmutationFee = 0`.
- Historical ETH WETH vault reads show the rate was still `0` around block `25250000`, and is `150000000000000000` at current block, so the adapter reads the rate at the period's `toApi` block instead of backfilling 15% before it existed on-chain.

Validation:
- `pnpm exec tsc --noEmit --pretty false`
- `pnpm test fees alchemix-v3 2026-06-18` -> fee rate was still 0, so revenue correctly stays 0.
- `pnpm test fees alchemix-v3 2026-06-22` -> reports MYT performance fees plus Transmuter early-exit fees; income statement balances.

bheluga:

PR #7719, opened 2026-06-19
hastra: add ethereum fees

DaniiRix:
- Hastra's fees adapter only tracked Solana wYLDS yield. This adds the Ethereum deployment (~$176M staked wYLDS, previously uncounted).
- Ethereum yield is read on-chain via getLogs: wYLDS (0x6aD038…) minted (Transfer from 0x0) into the PRIME staking contract (0x19ebb3…), mirroring Solana's hourly publish_rewards stream. Counted as dailyFees = dailySupplySideRevenue (Assets Yields).
- New-issuance mints (USDC deposits → wYLDS) go to other addresses and are excluded.

### Verification (2026-06-16)

- Adapter Ethereum yield ≈ $24.8k/day; on-chain staking mints that day = 24 events / 24,825.90 wYLDS — exact match.
- Only the staking contract receives the hourly cadence; no second reward vault missed.
- Staked balance $175.8M (= tracked TVL); implied APY 5.15%, in line with YLDS = SOFR − 0.5%.
- Solana path unchanged. No cross-chain double-count (separate deployments).

### DIFF
--- a/fees/hastra/index.ts
+++ b/fees/hastra/index.ts
@@ -77,13 +77,10 @@ const breakdownMethodology = {
 
 const adapter: SimpleAdapter = {
     version: 1,
-    fetch: async (options: FetchOptions) => {
-        return options.chain === CHAIN.SOLANA
-            ? fetchSolana(options)
-            : fetchEthereum(options);
+    adapter: {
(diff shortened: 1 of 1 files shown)


bheluga:
@DaniiRix thanks for the PR

PR #7720, opened 2026-06-19
fix(fullsail-finance): supply-side revenue is fees minus revenue, not full fees

tarun-khatri:
The FullSail fees adapter set:
```
const dailySupplySideRevenue = dailyFees
```
i.e. supply-side revenue = the **full** fees. Supply-side should be the portion of fees *not* taken as revenue (`fees - revenue`). As written, `revenue + supplySide = 2 x fees`, which is impossible (revenue + supplySide must equal fees).

Verified against production (`api.llama.fi/summary/fees/full-sail`): every day shows `fees = revenue = supplySide`, so `(revenue + supplySide) / fees = 2.00` on every data point. The API reports `revenue_usd = fees_usd` (all swap fees go to veSAIL holders), so the correct supply-side is `fees - revenue = 0`.

The API response has no supply-side field, so it must be derived. Fix: `dailySupplySideRevenue = max(0, fees_usd - revenue_usd)`. After the fix a sample day (fees 1.45k) gives revenue 1.45k, supplySide 0, and `protocol(67) + holders(1.38k) = revenue`, so the income statement reconciles.

### DIFF
--- a/fees/fullsail-finance/index.ts
+++ b/fees/fullsail-finance/index.ts
@@ -23,7 +23,7 @@ const fetch = async (options: FetchOptions) => {
     // Supply-side is the portion of fees NOT taken as revenue; the API has no
     // supply-side field, so derive it. (Setting it to dailyFees double-counted,
     // making revenue + supplySide = 2x fees.)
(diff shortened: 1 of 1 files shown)


bheluga:

PR #3654, opened 2025-07-10
Deepbook V3 New Assets

tonylee08:
DRF, SEND, XBTC added for volume calculations

PR #2947, opened 2025-04-16
Add WAL token to DeepBook v3

0xaslan:
Track the volume for WAL/SUI and WAL/USDC pools

PR #7731, opened 2026-06-19
dexs(deepbookv3-sui): add suiUSDe and usdSUI quote tokens

tarun-khatri:
## Summary

The DeepBook V3 volume adapter resolves each pool's quote token by looking up the quote symbol in a hardcoded `coins` dict, and **silently skips** any pool whose quote token isn't present:

```ts
const quoteToken = coins[quoteTokenSymbol];
if (!quoteToken) {
  console.warn(`Quote token for poolName ${poolName} not found`);
  continue;
}
```

Two active pools quote in tokens missing from the dict, so their volume is dropped:

| pool | quote token | quote type |
|---|---|---|
| `SUI_SUIUSDE` | suiUSDe | `0x41d587e5...::sui_usde::SUI_USDE` |
| `SUI_USDSUI` | usdSUI | `0x44f83821...::usdsui::USDSUI` |

## Impact

Measured against the DeepBook indexer (`all_historical_volume`) over the last 7 days, the two skipped pools account for **~$1.75M of ~$63M (≈2.8%)** of total volume, and the share is growing as these stablecoin pairs gain traction. Both tokens are priceable (DefiLlama confidence 0.99), so they were previously contributing $0.

## Fix

Add the two quote tokens to the `coins` dict — the same maintenance pattern as prior additions (e.g. WAL in #2947, the asset batch in #3654). Verified the adapter runs and picks the pools up.

Pool metadata source: `https://deepbook-indexer.mainnet.mystenlabs.com/get_pools`

### DIFF
--- a/dexs/deepbookv3-sui/index.ts
+++ b/dexs/deepbookv3-sui/index.ts
@@ -63,6 +63,7 @@ const methodology = {
 
 export default {
   version: 2,
+  pullHourly: true,
   methodology,
   adapter: {
     [CHAIN.SUI]: {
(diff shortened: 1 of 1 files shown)


bheluga:

PR #7732, opened 2026-06-19
fees(aries-markets): charge interest on all reserves, split by each reserve_ratio

tarun-khatri:
## Summary

Two issues in the Aries Markets fee adapter, both verified against the protocol's own on-chain state via its API:

### 1. Only two reserves were tracked
The adapter computed borrow interest for only `WrappedUSDT` and `WrappedUSDC`, although `reserve.current` lists every borrowable market. Active markets such as APT, amAPT, native USDC/USDT, WETH and USDY were ignored — understating **fees by ~38%**. Fixed by iterating every reserve with a non-zero borrowed amount and booking interest under its own coin type (so the framework prices each market with its own decimals/price).

### 2. The protocol/lender split was a hardcoded 20/80
Each reserve actually keeps a different cut of borrow interest (`reserve_ratio`, ranging **10–72%**), not a flat 20%. I confirmed `reserve_ratio` is the protocol cut independently of the field name, by deriving it from each reserve's APRs:

```
lender_share = supplyApr / (borrowApr × utilizationRatio)
protocol_cut = 1 − lender_share
```

This matches `reserve_ratio` exactly on all 16 active reserves (e.g. WrappedUSDC: supplyApr 0.01368 / (borrowApr 0.06538 × util 0.5231) = 40% to lenders → 60% protocol = reserve_ratio 60). The dominant markets (USDC/USDT/WrappedUSDC/WrappedUSDT = 60%, APT/amAPT = 72%) are far above the assumed 20%, so protocol revenue was understated ~3×. Fixed by splitting each market's interest with its own `reserve_ratio`.

## Result (verified locally)

- Daily fees: $111 → **$176** (all reserves now included)
- Daily protocol revenue: $35 (flat 20%) → **$106** (real per-reserve cut)
- Supply-side: $70; identity holds (106 + 70 = 176)

### DIFF
--- a/fees/aries-markets/index.ts
+++ b/fees/aries-markets/index.ts
@@ -5,11 +5,9 @@ import { METRIC } from "../../helpers/metrics";
 
 const rateURL = "https://api-v2.ariesmarkets.xyz/reserve.rateHistory?input=";
 const reserveURL = "https://api-v2.ariesmarkets.xyz/reserve.current";
-const USDTReserveKey = "0x9770fa9c725cbd97eb50b2be5f7416efdfd1f1554beb0750d4dae4c64e860da3::fa_to_coin_wrapper::WrappedUSDT";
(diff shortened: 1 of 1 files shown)


bheluga:

PR #7762, opened 2026-06-21
feat(dexs): add Rubin (perp DEX) volume + OI adapter

rprudnikov-itrf:
Adds a volume + open-interest adapter for **Rubin**, a self-custody decentralized perpetual & spot exchange.

**Source:** Rubin's public Comlink-style REST indexer `https://indexer.mainnet.rubin.trade/v4/perpetualMarkets`.

**Metrics**
- `dailyVolume` = Σ `volume24H` across markets
- `openInterestAtEnd` = Σ `openInterest × oraclePrice` (USD)

Uses `runAtCurrTime` (current rolling 24h). Mainnet is newly launched — live sanity check (2026-06-21): 6 active markets, OI ≈ $1.7k, 24h volume ≈ $160 (growing).

Chain is keyed as `rubin` (Rubin runs its own appchain). Happy to adjust the chain key or add it wherever you prefer.

---

##### Does this project have a referral program? Yes, https://rubin.trade/referrals


rprudnikov-itrf:
Thanks for the review 🙏

- **Empty-markets guard** — added: the adapter now throws if the indexer returns no markets, so a transient/empty response can't write a false zero to history.
- **`version: 2` / `pullHourly`** — this adapter uses `runAtCurrTime`: it reports the indexer's current rolling-24h volume and open interest rather than backfilling per-hour history, so `pullHourly` doesn't apply here (and would conflict with `runAtCurrTime`). I've kept the current-time form. The indexer also exposes daily candles, so if you'd prefer a historical/backfilling version I'm glad to switch.

rprudnikov-itrf:
@coderabbitai resolve

PR #7770, opened 2026-06-22
fix: deprecated flow explorer

DaniiRix:
Blockscout instance on the v2 stats API verified returning real daily data (newTxns/activeAccounts/newAccounts).


bheluga:
@DaniiRix thanks for the PR
Added it [here](https://github.com/DefiLlama/dimension-adapters/pull/7777) as there were some merge conflicts

PR #7771, opened 2026-06-22
Fix Fuel Ignition fees adapter: migrate to new blob-fees indexer

uwezukwechibuzor:
## summary
 Fix Fuel Ignition fees adapter: migrate to new blob-fees indexer

  ### Problem

The Fuel Ignition adapter's blob/DA fee source, https://indexer-fuel-seq.simplystaking.xyz/seq/blob-fees, is down (returns 503), so the adapter could no longer fetch blob fees.

  Investigation

  I reached out to the Fuel team about where DA (blob) fees can still be tracked. They pointed me to a new
  sequencer indexer host:

  ▎ index-api.sequencer.mainnet.fuel.network

  The new host exposes the same path and response schema as the old one, just under a different domain. Confirmed
  it's live and serving daily data right up to the current date, e.g.:

 
  ```http
  GET https://index-api.sequencer.mainnet.fuel.network/seq/blob-fees?start=1767225600&end=1767312000

  {
    "totalFees": "7608815104",
    "feeDenom": "fuel",
    "decimals": 9,
    "blobCount": 7196,
    "startTimestamp": 1767225600,
    "endTimestamp": 1767312000
  }
  ```

  Changes

  - Point INDEXER_URL at the new host https://index-api.sequencer.mainnet.fuel.network (path /seq/blob-fees and
  schema unchanged).
  - Add a methodology block documenting the two metrics:
    - Fees — transaction gas fees (in ETH) paid by users on Fuel Ignition.
    - Revenue — blob/data-availability fees (in FUEL) for posting Fuel batch data to the DA layer.

### DIFF
--- a/fees/fuel-ignition.ts
+++ b/fees/fuel-ignition.ts
@@ -60,6 +60,11 @@ const fetch = async (options: FetchOptions) => {
   };
 };
 
+const methodology = {
+  Fees: "Transaction gas fees (in ETH) paid by users on the Fuel Ignition network.",
+  Revenue: "Blob/data-availability fees (in FUEL) for posting Fuel batch data to the DA layer.",
+};
+
 const adapter: SimpleAdapter = {
   version: 1,
   runAtCurrTime: true,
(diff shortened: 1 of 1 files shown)


bheluga:

PR #7782, opened 2026-06-22
[codex] Add Tristero v3 delegated both-side volume

ggarza5:
## Summary
- Decode Tristero v3 delegated account `execute(bytes32,bytes)` batches and count inner `router.send` calls targeted at `0xD4c2Ce98CbE2B02bE9449606fdfF75DD700B836F`.
- Count both maker and filler assets for non-margin v3 `router.send` swaps by adding `srcQuantity` and `dstQuantity`.
- Enable the delegated send target on Base, Arbitrum, and Ethereum from the observed 2026-06-22 cutover and update methodology text.

## Why
Some v3 fills are submitted through EIP-7702 delegated accounts, so the top-level tx calls `execute(...)` on the delegated account instead of directly calling the Tristero send target. The economic `send(...)` calls are inside the delegated batch.

For the Base sample tx `0x3a80afb7ea134650a7906ae05a68a59318967e543d905b3ede4b49db7c151f98`, the intended accounting is maker + filler volume: decoded source side is 40k USDT + 40k USDS, decoded destination side is 40k USDT + 40k USDS, for 160k total stable volume.

## Verification
- `yarn ts-check`
- Focused Base tx decode for `0x3a80afb7ea134650a7906ae05a68a59318967e543d905b3ede4b49db7c151f98`: outer selector `0xe9ae5c53`, 40 inner executions, 40 matching inner `send` calls, decoded both-side totals `80,000 USDT + 80,000 USDS`.
- `yarn test dexs tristero 1782157051` could not complete locally because this checkout is missing DefiLlama `CLICKHOUSE_CONFIG`.

### DIFF
--- a/dexs/tristero/index.ts
+++ b/dexs/tristero/index.ts
@@ -6,7 +6,7 @@ const adapter: SimpleAdapter = {
   fetch: async (options) => ({ dailyVolume: await fetchDailyVolume(options) }),
   adapter: TRISTERO_DEX_CHAINS,
   methodology: {
-    Volume: "Source-side token amounts from legacy OrderFilled events and v3 router.send orders. Margin opens include collateral and loan; margin closes include loan settlement transfers.",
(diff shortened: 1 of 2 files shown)


ggarza5:
Hi @noateden just following up to see if there's anything else we need to do here

PR #7797, opened 2026-06-23
[fix] solv-finance: map mismatched chain keys (berachain, rootstock, …

uwezukwechibuzor:
## Summary

  Fixes silent `$0` reporting on several chains in the Solv Finance fees adapter caused by chain-key mismatches
  between DefiLlama chain names and Solv's fee config (`solv-fees-v2.json`).

  The adapter looked up pools with `contracts[options.chain]`, but Solv keys some chains differently from their
  DefiLlama names. When the names didn't match, the lookup returned `undefined`, the fetch early-returned `{}`,
  and those chains reported **$0 with no error** 

  ## The mismatches

  | DefiLlama name (`options.chain`) | Solv config key | Live pools | Before |
  | --- | --- | --- | --- |
  | `berachain` | `bera` | 3 | dropped → $0 |
  | `rsk` (rootstock) | `rootstock` | 2 | dropped → $0 |
  | `hyperliquid` | `hyper` | 2 | not registered |
  | `ailayer` | `ailayer` | 1 | not registered |
  | `merlin` | `merlin` | 1 | commented out |

  ## Changes

  - **Add a `configKey` override** for chains whose DefiLlama name differs from Solv's config key, and resolve
  the lookup through it:

    ```ts
    const configKey = chains[options.chain]?.configKey ?? options.chain;
    const pools = contracts[configKey];
    if (!pools || pools.length === 0) return {};

  [CHAIN.ROOTSTOCK]:   { start: "2024-09-17", configKey: "rootstock" },
  [CHAIN.BERACHAIN]:   { start: "2025-03-17", configKey: "bera" },
  [CHAIN.HYPERLIQUID]: { start: "2025-03-17", configKey: "hyper" },

  - Enable previously-missing live-pool chains: uncomment merlin, add ailayer and hyperliquid.
   - Resolve the chain's pools once in `fetch()` and pass them to `fees()`, so the same lookup bug can't reappear
  there. Also skip chains with no pools.
  - Convert start from a numeric timestamp (deployedAt: 1726531200) to the documented "YYYY-MM-DD" string format,
  which is type-clean against BaseAdapterChainConfig.start: string.

### DIFF
--- a/fees/solv-finance/index.ts
+++ b/fees/solv-finance/index.ts
@@ -10,23 +10,28 @@ import BigNumber from "bignumber.js";
 const feesConfig =
   "https://raw.githubusercontent.com/solv-finance/solv-protocol-defillama/main/solv-fees-v2.json";
 
-const chains: { [chain: Chain]: { deployedAt: number } } = {
-  [CHAIN.ETHEREUM]: { deployedAt: 1726531200 },
-  [CHAIN.BSC]: { deployedAt: 1726531200 },
-  [CHAIN.ARBITRUM]: { deployedAt: 1726531200 },
-  [CHAIN.MANTLE]: { deployedAt: 1726531200 },
-  // [CHAIN.MERLIN]: { deployedAt: 1726531200 },
-  [CHAIN.CORE]: { deployedAt: 1726531200 },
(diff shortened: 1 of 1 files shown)


bheluga:

PR #7806, opened 2026-06-23
users: expand isAcceptedChain to full Allium chain set

DaniiRix:
Expands isAcceptedChain from 9 chains to the full set of chains already supported by alliumChainMap.

starknet is intentionally left out: Allium's starknet.raw.transactions has no to_address/from_address column, so the TO_ADDRESS IN (...) / distinct FROM_ADDRESS query in countUsers.ts cannot run there.


bheluga:
@DaniiRix thanks for the PR
These chains arent being used anywhere.
We can add them while adding related protocols

PR #7808, opened 2026-06-24
fix(ipor-protocol): add ink, plasma, avax, katana chains

uwezukwechibuzor:
## Summary

Extends the IPOR Protocol fees adapter to cover four additional chains where IPOR PlasmaVaults are now deployed, and fixes an avax chain-key mismatch.

Changes
- New chains: added ink, plasma, avax, and katana to the adapter.
  - ink — 2025-07-29
  - plasma — 2025-09-29
  - avax — 2025-10-03
  - katana — 2026-02-19
- avax key mapping: IPOR's addresses.json keys Avalanche vaults under "avalanche", but CHAIN.AVAX === "avax". Added a CHAIN_TO_CONFIG_KEY map so the adapter looks up the correct config key.

Testing

pnpm test fees ipor-protocol runs cleanly across all 9 chains. Negative single-day values on some chains (e.g. avax −59 on 2026-06-16) are expected.

Notes
- botanix exists in the source JSON but has no vaults (infrastructure-only), so it is intentionally not tracked.  It can be added once vaults are deployed there.


bheluga:

PR #7818, opened 2026-06-24
LeverUp: sync adapter with V2

alexhamilton101:
Sync the LeverUp adapter with the V2 contract upgrade on Monad.

- Fees: picks up the new V2 trade fee event
- Open interest: reads the on-chain open interest view introduced in V2;
  days before the upgrade keep the original method, so existing history is unchanged.

Tested locally: `npm test dexs leverup` and `npm test open-interest leverup-oi`.


bheluga:
@alexhamilton101 thanks for the PR

PR #7822, opened 2026-06-24
fix(strata-markets): cap daily yield to prevent fee spikes

radeveth:
## Problem

The Strata Markets fee adapter reports wildly inflated fees on certain days:

| Date | Reported Fees | Normal Fees | Overreporting |
|------|--------------|-------------|---------------|
| Jun 8 | $549,645 | ~$12,000 | **44x** |
| Jun 18 | $501,000 | ~$12,000 | **42x** |

## Root Cause

The adapter computes yield as the delta of `strategy.totalAssets()` between hourly snapshots (`pullHourly: true`). This overstates fees when underlying yield sources have non-linear recognition:

### 1. sUSDe exchange rate jumps
- The sUSDe strategy ($61M TVL) uses `sUSDe.previewRedeem(shares)` for NAV
- When Ethena distributes rewards in batches, the exchange rate jumps by more than 1 day's worth
- A 0.6% rate jump on $61M = $549K of apparent yield in a single hourly snapshot

### 2. DiscreteAccounting oracle gaps
- mHYPER, mM1-USD, and PRIME markets use `DiscreteAccounting` with Midas/Figure oracles
- The protocol internally uses `totalAssets(latestNav, timestamp)` (2-param) which only accepts new NAV when the oracle has updated
- The fee adapter calls `totalAssets()` (0-param) which always returns the live oracle value
- Oracle gaps of N days → N days of yield recognized at once

### Revenue confirms the pattern
Revenue/Fees ratio stays consistently ~5% (matching `reserveBps`), confirming the formula is internally consistent — only the yield input is spiking:
- Jun 8: $27,650 / $549,645 = 5.03%
- Jun 18: $25,333 / $501,000 = 5.06%

## Fix

Add a per-CDO yield cap at **50% annualized** per `processCDO()` call. This is well above any legitimate strategy yield (3-15% APY) but prevents multi-day yield bunching from inflating daily fees.

```typescript
const MAX_ANNUAL_YIELD_BPS = 5000n; // 50% annualized
const maxYield = (navStart * MAX_ANNUAL_YIELD_BPS) / (10000n * 365n);
if (yieldAmount > maxYield) yieldAmount = maxYield;
```

### Impact
- Normal daily fees (~$12K) are unaffected — well under the cap
- The Jun 8 spike ($549K) would be capped to ~$126K across all CDOs
- No false positives: 50% APY threshold gives ~3x headroom over the highest-yielding junior tranches

### DIFF
--- a/fees/strata-markets/index.ts
+++ b/fees/strata-markets/index.ts
@@ -152,12 +152,15 @@ async function processCDO(
   let yieldAmount = navEnd - navStart - inflows + outflowsToUsers + reserveOut;
   if (yieldAmount < 0n) yieldAmount = 0n;
 
-  // Cap yield at 50% annualized (~0.137% per day) to prevent spikes from
-  // non-linear yield recognition (sUSDe exchange rate jumps, Midas/Figure
-  // oracle gaps that bunch multiple days of yield into a single period).
+  // Cap yield at 50% annualized to prevent spikes from non-linear yield
+  // recognition (sUSDe exchange rate jumps, Midas/Figure oracle gaps that
+  // bunch multiple days of yield into a single period).
+  // Scale to the actual pull window (hourly when pullHourly: true).
   // 50% APY is well above any legitimate strategy yield (~3-15% APY).
   const MAX_ANNUAL_YIELD_BPS = 5000n; // 50% annualized
-  const maxYield = (navStart * MAX_ANNUAL_YIELD_BPS) / (10000n * 365n);
+  const SECONDS_PER_YEAR = 365n * 86400n;
+  const elapsed = BigInt(options.endTimestamp - options.startTimestamp);
+  const maxYield = (navStart * MAX_ANNUAL_YIELD_BPS * elapsed) / (10000n * SECONDS_PER_YEAR);
   if (yieldAmount > maxYield) yieldAmount = maxYield;
 
   const ONE = 10n ** 18n;
--- a/fees/strata-markets/index.ts
+++ b/fees/strata-markets/index.ts
@@ -156,11 +156,15 @@ async function processCDO(
   // recognition (sUSDe exchange rate jumps, Midas/Figure oracle gaps that
   // bunch multiple days of yield into a single period).
   // Scale to the actual pull window (hourly when pullHourly: true).
+  // Use investedBase (navStart + inflows - outflows) to handle first-funded periods.
   // 50% APY is well above any legitimate strategy yield (~3-15% APY).
   const MAX_ANNUAL_YIELD_BPS = 5000n; // 50% annualized
   const SECONDS_PER_YEAR = 365n * 86400n;
   const elapsed = BigInt(options.endTimestamp - options.startTimestamp);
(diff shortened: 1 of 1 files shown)


bheluga (commented):
@radeveth thanks for the PR
Do you think disabling pull hourly would fix this?

radeveth:
@bheluga thanks for the review!

Disabling `pullHourly` would reduce spike magnitude but wouldn't fully fix it. The root cause is non-linear yield recognition in the underlying protocols:

1. **sUSDe exchange rate jumps** — Ethena distributes rewards in batches, so the sUSDe exchange rate can jump by multiple days' worth of yield. With $61M TVL, even a small rate bump produces large absolute deltas.

2. **DiscreteAccounting oracle gaps** — mHYPER/mM1-USD/PRIME markets use Midas/Figure oracles that update periodically. If the oracle is stale for N days, N days of yield land in the first period after the update — whether that period is 1 hour or 1 day.

Switching to daily pulls would spread hourly spikes across a day, but a 7-day oracle gap would still produce a 7x daily spike.

I've updated the PR to scale the yield cap to the actual pull window (`options.endTimestamp - options.startTimestamp`) as suggested by the code review. This makes it work correctly for both hourly and daily granularity.

bheluga:
> @bheluga thanks for the review!
> 
> Disabling `pullHourly` would reduce spike magnitude but wouldn't fully fix it. The root cause is non-linear yield recognition in the underlying protocols:
> 
> 1. **sUSDe exchange rate jumps** — Ethena distributes rewards in batches, so the sUSDe exchange rate can jump by multiple days' worth of yield. With $61M TVL, even a small rate bump produces large absolute deltas.
> 2. **DiscreteAccounting oracle gaps** — mHYPER/mM1-USD/PRIME markets use Midas/Figure oracles that update periodically. If the oracle is stale for N days, N days of yield land in the first period after the update — whether that period is 1 hour or 1 day.
> 
> Switching to daily pulls would spread hourly spikes across a day, but a 7-day oracle gap would still produce a 7x daily spike.
> 
> I've updated the PR to scale the yield cap to the actual pull window (`options.endTimestamp - options.startTimestamp`) as suggested by the code review. This makes it work correctly for both hourly and daily granularity.

Aren't rate jumps valid spikes?
What would be wrong if we count 1 week yield at once? (cumulative would still be correct right?) 
If we add a cap, wont that lead to undercounting of cumulative fees?

radeveth:
@bheluga You're right on all three points.

1. **Rate jumps are valid spikes** - the yield is real, just recognized in a lump when Ethena distributes rewards or Midas oracles update after a gap. The underlying strategies did earn this yield.

2. **Cumulative is correct** - the spike days capture yield that wasn't reported on the prior "flat" days. Monthly/yearly totals are accurate.

3. **The cap would undercount** - every time it triggers, real yield gets thrown away permanently. For example on Jun 8, the cap would have discarded ~$424K of legitimate yield.

I'll close this PR. The spikes aren't a bug - they're correct data with lumpy time distribution. Thanks for pushing back on this.

PR #7824, opened 2026-06-24
feat(users): track Somnia chain users via blockscout stats

uwezukwechibuzor:
https://defillama.com/chain/somnia


bheluga:

PR #7830, opened 2026-06-25
fix(catfee): correct Tron API timestamps and add API key

uwezukwechibuzor:
### Problem
The CatFee fees adapter ([defillama.com/protocol/catfee](https://defillama.com/protocol/catfee)) was failing and reporting **no fees**. Two root causes:

1. **Wrong timestamp unit → HTTP 400.** `FetchOptions` provides `fromTimestamp` /  `endTimestamp` in **seconds**, but Tronscan's `/api/transfer/trx` endpoint expects **milliseconds**. Sending seconds is rejected with `start_timestamp must be >= 1529884800000`, so the very first page request threw and the adapter returned nothing.

2. **Rate limiting → HTTP 429.** Unauthenticated requests are throttled per IP. While paging through a day's transfers the adapter would hit `429 Too Many Requests` and fail.

### Solution
1. Send the API query timestamps in milliseconds (`* 1000`). 
2. Authenticate with a free Tronscan API key via the `TRON-PRO-API-KEY` header, which lifts the per-IP rate limit. The key is read from `TRONSCAN_API_KEY` and is required (the adapter throws if it is unset). Free key: https://tronscan.org/#/myaccount/apiKeys

### Changes
- `fees/catfee/index.ts`
  - Send `start_timestamp` / `end_timestamp` in milliseconds (`* 1000`).
  - Attach `TRON-PRO-API-KEY` header from `getEnv("TRONSCAN_API_KEY")`; throw if missing.
- `helpers/env.ts`
  - Register `TRONSCAN_API_KEY` in `ENV_KEYS`.

Fees are returned instead of failing with HTTP 400/429.

### Notes
- `TRONSCAN_API_KEY` must be set or the adapter throws.

### DIFF
--- a/fees/catfee/index.ts
+++ b/fees/catfee/index.ts
@@ -1,6 +1,7 @@
 import { FetchOptions } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
-import fetchURL from "../../utils/fetchURL";
+import { httpGet } from "../../utils/fetchURL";
+import { getEnv } from "../../helpers/env";
 const plimit = require('p-limit');
 const limits = plimit(1);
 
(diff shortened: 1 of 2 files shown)


bheluga:

PR #7831, opened 2026-06-25
fix(tronsave): correct Tron API timestamps and add API key

uwezukwechibuzor:
Same fix as #7830
https://defillama.com/protocol/tronsave


bheluga:

PR #7832, opened 2026-06-25
fix(taiko-swap): switch to on-chain logs via uniV3 factory registry

uwezukwechibuzor:
### fix(taiko-swap): read on-chain Swap logs via uniV3 factory registry
**Problem:** Adapter flatlined at **0** volume/fees since Feb 17 (e.g. 2026-04-14 returned all zeros), it diffed the subgraph's `factory` cumulative totals across two blocks, which no longer advance, so the difference was always 0 (and recent days additionally error on null block resolution).

**Solution:** Use the shared on-chain `getUniV3LogAdapter` (reads `Swap` events directly, no block/subgraph dependency) and move the config into `factory/uniV3.ts`; deleted the `dexs/taiko-swap.ts`.
**Verified:** `pnpm test dexs taiko-swap 2026-06-20` → volume ~27.28k, fees ~81.31 (was 0.00 before).


bheluga:

PR #7840, opened 2026-06-25
fix(garden): point bridge adapter at v2 orders API, add new chains, f…

uwezukwechibuzor:
## Summary

Fixes the Garden bridge-aggregator adapter, which was failing with `HTTP 404`, and expands its chain coverage.

## Changes

- **Migrate to the v2 orders API** — the old `/orders/matched` endpoint now returns `404`. Updated the base URL to `https://api.garden.finance/v2/orders` (the `/matched` path segment is gone; `status=completed`). The response shape is unchanged, so the parsing logic stays the same.

- **Add missing chains** — extended `chainMapper` with `botanix`, `citrea`, `megaeth`, `tron`, `litecoin`, and `spark`, each with its correct `name`, `start` date, `primaryCGToken`, and `decimals`.

- **Fix HyperEVM bucketing** — Garden reports HyperEVM swaps under two different chain strings (`hyperliquid` legacy + `hyperevm` new), but the adapter only reads the `hyperliquid` bucket, so `hyperevm` volume was silently dropped. Added a `normalizeChain` helper that collapses `hyperevm` → `hyperliquid` so both land in the same bucket.

## Testing

`pnpm test bridge-aggregators garden` passes, with volume reported across all mapped chains. `hyperliquid` now shows non-zero volume where it previously read `0.00`, confirming the bucketing fix.

### DIFF
--- a/bridge-aggregators/garden/index.ts
+++ b/bridge-aggregators/garden/index.ts
@@ -81,8 +81,48 @@ const chainMapper: Record<string, { name: string, start: string, primaryCGToken:
         primaryCGToken: 'monad',
         decimals: 18
     },
+    [CHAIN.BOTANIX]: {
+        name: "botanix",
+        start: "2025-06-24",
+        primaryCGToken: 'bitcoin',
+        decimals: 18
+    },
+    [CHAIN.CITREA]: {
+        name: "citrea",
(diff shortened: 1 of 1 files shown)


bheluga:

PR #7845, opened 2026-06-26
fees(current): break down fees/revenue by source and   split liquidation

paulwu0903:
The Current fees adapter used to report flat totals —
fees, revenue, and supply-side — with no breakdown. This change splits them by source: borrow interest, liquidation, and flashloan.

For borrow interest and liquidation, each amount is divided into what users pay (Fees), what's paid out to others (supply-side — suppliers for interest, liquidators for liquidation), and what the protocol keeps (Revenue). 

Flashloan fees have no cost side, so they go fully to the protocol. 
Liquidation specifically is now split between the liquidator's cut and the protocol's cut.
  
Under the hood it uses createBalances() + addUSDValue() to tag each amount with its source, plus a breakdownMethodology block describing every label.

### DIFF
--- a/fees/current/index.ts
+++ b/fees/current/index.ts
@@ -1,42 +1,70 @@
 import {
   Adapter,
-  FetchResultV2,
-  FetchV2,
+  FetchOptions,
 } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
+import { METRIC } from "../../helpers/metrics";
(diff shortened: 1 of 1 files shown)


bheluga:
@paulwu0903 thanks for the PR
Added [here](https://github.com/DefiLlama/dimension-adapters/pull/7849)

PR #7851, opened 2026-06-26
[fix] Bitget-Card: only count bitget card spends instead of all fiat24

Shaileshkhote:
## Summary
- Adds SafePal Card volume tracking on Arbitrum.
- Fixes Bitget Wallet Card volume attribution to count only the Bitget Fiat24 card cohort.

## Changes
- Tracks Fiat24 card spend from on-chain `Authorized` events.
- Filters cardholder wallets via each card program’s anchor log + Fiat24 NFT mint transaction.
- Converts Fiat24 currency token amounts to USD with the on-chain Fiat24 exchange-rate contract.


bheluga:

PR #7863, opened 2026-06-27
feat(bsquared): track chain fees and users via blockscout

uwezukwechibuzor:
## Summary

- Add B² Network (bsquared) chain fees and active/new user tracking via its Blockscout explorer and stats API.
- https://defillama.com/chain/bsquared


bheluga:

PR #7868, opened 2026-06-28
mooniswap: fix volume to use trade amount, not pool reserve

DaniiRix:
### Fix

* Use `amount` (swap input) as volume instead of `Swapped.dstBalance`.
* Read the swap fee from the factory `fee()` and apply it to the input amount.
* Set `dailySupplySideRevenue = dailyFees` since all swap fees accrue to LPs.
* Set `dailyRevenue`, `dailyProtocolRevenue`, and `dailyHoldersRevenue` to `0`.

### Why

`Swapped.dstBalance` is the **pool reserve before the swap**, not the traded amount. Using it inflated reported volume.

### DIFF
--- a/dexs/mooniswap/index.ts
+++ b/dexs/mooniswap/index.ts
@@ -1,10 +1,46 @@
 import { FetchOptions, SimpleAdapter } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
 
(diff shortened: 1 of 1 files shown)


bheluga:
@DaniiRix thanks for the PR

PR #7872, opened 2026-06-29
feat(skate-amm): add v2 volume/fees data source

PhiDung-hub:
- Add the Skate AMM v2 volume/fees source (`amm-data-v2/pools/stats`) alongside v1, summing both
- v1 stats now returns empty as v1 winds down; v2 carries live data with the same shape/params


bheluga:
@PhiDung-hub thanks for the PR

PR #7873, opened 2026-06-29
fix(securitize): correct BUIDL yield (dividend) calculation across all chains

calabozo:
## Problem
BUIDL's on-chain yield (dividends / DRIP) feeds `Fees` and `SupplySideRevenue`. It was captured by summing `Issue` mints inside a fixed 15:00–16:00 UTC window. BUIDL distributes at ~10:00 US Eastern, which is **14:00 UTC during EDT** — so for ~8 months/year the window misses the distribution entirely and reports **zero yield**. The window also can't separate yield drips from new subscriptions (both mint tokens).

## Fix
Classify each mint by `mintValue / recipientPriorBalance`:
- a yield drip is a tiny fraction of the holder's balance (~daily T-bill rate, ~0.0086%/day);
- a subscription / principal mint is a large fraction.

Keep mints with ratio < 0.5%. Validated against labelled data (`issuance_drip` vs `issuance_fiat`) with real balances across all chains: drips land on the daily rate, subscriptions sit ≥0.5%, ~97.6% of yield captured by value, **0% principal leakage**.

Per chain:
- **EVM**: full-period `Issue` logs + `multiCall balanceOf` (on-chain, full history).
- **Aptos**: Allium `aptos.assets.fungible_transfers`, prior balance via cumulative transfer flow.
- **Solana**: Dune `tokens_solana.transfers`, same cumulative-flow ratio.
- **BUIDL-I** (Ethereum) keeps its RedStone accrual feed.

Management fee (`supply × bps`) is unchanged. Removes the DST distribution-time window and weekend skip. Adds `Dependencies.ALLIUM` (Aptos).

## Notes for reviewers
Two warehouse-schema assumptions are marked `// VERIFY` and will be exercised by CI:
- Solana: sender/recipient column names in `tokens_solana.transfers`.
- Aptos: mints flagged by NULL `from_address` in `aptos.assets.fungible_transfers`.

### DIFF
--- a/fees/securitize/index.ts
+++ b/fees/securitize/index.ts
@@ -5,6 +5,7 @@ import {queryAllium} from "../../helpers/allium";
 import {gql, GraphQLClient} from "graphql-request";
 import {getTokenSupply} from "../../helpers/solana";
 import {httpGet} from "../../utils/fetchURL";
+import * as sdk from "@defillama/sdk";
 
 const EVM_ABI = {
   issue: 'event Issue(address indexed to, uint256 value, uint256 valueLocked)',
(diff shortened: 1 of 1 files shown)


bheluga (commented):
@calabozo thanks for the PR
Redstone yield accrual would ideally be same/near similar for all chains right?
Why shouldn't we use that for all chains (if missing for a chain, use from ethereum) instead of complicating things

calabozo:
> @calabozo thanks for the PR
> Redstone yield accrual would ideally be same/near similar for all chains right?
> Why shouldn't we use that for all chains (if missing for a chain, use from ethereum) instead of complicating things

@bheluga 
 RedStone's off-chain /prices API only retains ~30 days, so a fresh backfill older than 30 days can't be priced.
 But yes...maybe this is over-engineering. The accrual is fund-wide, so this is the cleaner approach and I'll switch to it if you think the 30days retention is not a problem.

bheluga:
> > @calabozo thanks for the PR
> > Redstone yield accrual would ideally be same/near similar for all chains right?
> > Why shouldn't we use that for all chains (if missing for a chain, use from ethereum) instead of complicating things
> 
> @bheluga RedStone's off-chain /prices API only retains ~30 days, so a fresh backfill older than 30 days can't be priced. But yes...maybe this is over-engineering. The accrual is fund-wide, so this is the cleaner approach and I'll switch to it if you think the 30days retention is not a problem.

Redstone usually has an onchain price feed right? so it preserves data forever?
We use it in acred, <https://github.com/DefiLlama/dimension-adapters/blob/master/fees/acred/index.ts>
ig it should be possible here too

calabozo:
Agreed — the accrual is fund-wide, so I'll switch to this. We verified the **gross** daily accrual is identical across chains (~3.65% APY); each chain's rate is just gross minus its management fee (matches Securitize's official sheet — ETH 3.15%, Aptos 3.45%, BNB 3.47%).

Model: **supply (read on-chain per chain) × (gross − chainBps)**, gross from any RedStone feed, bps from `static.primary_market_listing`. Drops Dune + Allium and the distribution-time window, and matches how BUIDL-I already works.

⚠️ **Caveat:** RedStone's off-chain `/prices` API only retains **~30 days**, so a fresh backfill older than 30 days can't be priced (the feed throws). Going-forward runs are fine (and it's the same limit BUIDL-I already has) — just no deep historical backfill. If an on-chain accrual feed exists (like the ACRED/HLSCOPE NAV feeds), I'd use that instead to get full history — do you know of one?

bheluga:
@calabozo thanks for the PR

calabozo:
@bheluga Thanks for your advice and support!

PR #7874, opened 2026-06-29
Upgrade to loan router v2 and introduce escrow timelock

wegory:


fees/usdai/index.ts
```diff
@@ -166,32 +216,87 @@ const fetch: any = async (options: FetchOptions) => {
...
+    eventAbi: LENDER_REPAID_EVENT_ABI
   });
-  for (const log of gpuLiquidationFeeLogs) {
-    // Get loan origination entity using loan terms hash to get currency token
-    const response = await request(LOAN_ROUTER_SUBGRAPH_API, loanOriginatedByHashQuery, {
-      loanTermsHash: log.loanTermsHash
-    });
-    dailyFees.add(response.loanOriginateds[0].currencyToken.id, log.liquidationFee + log.surplus, METRICS.GPU_FINANCING_LIQUIDATION_FEE);
-    dailyRevenue.add(response.loanOriginateds[0].currencyToken.id, log.liquidationFee + log.surplus, METRICS.GPU_FINANCING_LIQUIDATION_FEE_TO_PROTOCOL);
+  for (const log of gpuLenderRepaidV2Logs) {
```
bheluga, line 224: please retain v1 logs, it would help in cases where we need refill .
so ideally we prefer time based handling, 
say v1_to_v2_switch_date = x,
if(options.dateString<x)
use v1 logs
else
use v2 logs

wegory: hey @bheluga, thanks for the review. This change does not erase V1 logs. It replaces inefficient code with efficient code, while at the same time supporting both V1 and V2.

PR #7883, opened 2026-06-29
camelot-v3: fix $0 fees, correct split, add Plume & Superposition

uwezukwechibuzor:
### Problem

1. **Fees reported $0 on every existing chain.** Camelot's original Algebra pools (ApeChain, Arbitrum, Gravity, Rari, Reya) expose the swap fee through `globalState()` and have **no `fee()` method**. The adapter used `isAlgebraV3`, which reads the fee via `fee()` — that call reverts on these pools, so the fee tier silently became `0` 

2. **The revenue split was wrong.** The adapter used `80% LP / 17% holders / 3% protocol` (20% revenue), which does not match Camelot's documentation.

3. **Missing chains.** Plume and Superposition are live Camelot V3 deployments tracked by the TVL adapter but absent from this adapter.

### Solution

1. **Fix fee reading.** Switch the original-Algebra chains to `isAlgebraV2`, which reads the fee from `globalState()`, with an explicit Algebra `Pool` pool-creation event for discovery.

2. **Correct the split to the official [docs](https://docs.camelot.exchange/tokenomics/protocol-earnings):** **85% LPs / 15% protocol-controlled**, where:
   - `HoldersRevenue = 10.5%` — 7% xGRAIL Real Yield Staking + 3.5% GRAIL buyback & burn
   - `ProtocolRevenue = 4.5%` — 3% operating expenses + 1.5% Algebra licensing

   Methodology text and the breakdown were updated to match.

3. **Add 2 chains:**
   - **Plume** — factory `0x1eb9822d5176c88b1d4eec353fa956c896d77df9`, newer Algebra **Integral** pools where `fee()` works, so it uses a dedicated `isAlgebraV3` config.
   - **Superposition** — factory `0xCf4062Ee235BbeB4C7c0336ada689ed1c17547b6`, original Algebra, uses the same `isAlgebraV2` config.
   - https://defillama.com/protocol/camelot-v3

### DIFF
--- a/dexs/camelot-v3/index.ts
+++ b/dexs/camelot-v3/index.ts
@@ -4,9 +4,11 @@ import { getUniV3LogAdapter } from '../../helpers/uniswap';
 
 // Camelot V3 uses Algebra (Uniswap V3-style concentrated liquidity)
 // Fees are pool-specific and read on-chain from the Algebra pool configuration
-// Fee distribution (V3):
-// - ~80% to Liquidity Providers
-// - ~20% to protocol-controlled revenue (xGRAIL + treasury)
+// 85% for Liquidity Providers in LP tokens
+// 7% redistributed to xGRAIL holders through Real Yield Staking plugin
+// 3.5% dedicated to GRAIL buyback and burn
+// 3% to the Operating expenses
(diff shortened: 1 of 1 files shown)


bheluga:

PR #7899, opened 2026-07-01
fix(hypurrquant): describe as self-custodial crypto wealth management

Hiksang:
Small accuracy fix to the adapter's descriptive comment.

HypurrQuant is a self-custodial crypto wealth management app on HyperEVM (the Hyperliquid ecosystem). Swap routing (which this adapter tracks) is one feature within a broader on-chain portfolio-management product spanning liquidity provision, swaps, perps and yield, so the leading comment now reflects the full product.

- No logic change: fee recipient, fee sources, volume/fees/revenue computation, chains, and methodology fields are all untouched.
- Only the top descriptive comment is reworded.

Ref: https://hypurrquant.com

## Summary by CodeRabbit

* **Documentation**
  * Expanded the adapter description with clearer details about swap-routing behavior and routed volume calculation.

### DIFF
--- a/aggregators/hypurrquant/index.ts
+++ b/aggregators/hypurrquant/index.ts
@@ -2,11 +2,14 @@ import { FetchOptions, SimpleAdapter } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
 import { addTokensReceived } from "../../helpers/token";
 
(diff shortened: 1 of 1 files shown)


bheluga:
@Hiksang thanks for the PR

PR #7914, opened 2026-07-02
hanji: fix fee model, wire up supply-side revenue for LP vaults

DaniiRix:
Hanji's CLOB supports LP vault payouts (`passive_order_payout_rate`), but the adapter ignored them and incorrectly routed all fees into `dailyRevenue`/`dailyUserFees`. I reviewed the verified `OnchainCLOB.sol` contract and validated the fee flow against real Base trades before updating the adapter.

#### Fee flow

- **Taker fee:** `executed_value × (total_aggressive_commission_rate + passive_order_payout_rate)`
- **Maker fee:** separate `passive_fee` on resting orders.
- After the LP payout is deducted, the remaining commission is split between the protocol (`administrator`) and `marketmaker` (LP contract).

#### Fixes

- Included `passive_fee` in `dailyFees`.
- Added `dailySupplySideRevenue`, counting:
  - the LP payout portion of taker fees, and
  - the marketmaker's commission share when it's different from the administrator.
- Removed the hardcoded `adminCommissionRate = 1` and now read the live per-market value from `getConfig()`.
- Added `dailyProtocolRevenue`.

#### Why supply-side revenue is still $0

This is expected. On Base:
- `administrator == marketmaker` for every deployed market.
- `passive_order_payout_rate = 0`.
- No `MarketMakerChanged` events exist.

Today, all commission goes to the protocol. The adapter will automatically report supply-side revenue if markets enable LP payouts or use an external marketmaker.

#### Verification

- Confirmed live config values on Base and Etherlink via `getConfig()`.
- Verified the taker fee formula against decoded Base `OrderPlaced` events.
- `pnpm test dexs/hanji` passes across Base, Etherlink, and Monad (24 hourly slots).
- Base daily metrics: **$5.27M** volume → **$519.60** fees, matching the on-chain **0.01%** taker fee.

### DIFF
--- a/dexs/hanji/index.ts
+++ b/dexs/hanji/index.ts
@@ -15,6 +15,14 @@ const config: any = {
   [CHAIN.MONAD]: { clobFactoryV1: [], clobFactoryV2: ['0x5C28a12C8EbAF8524A2Ba1fdc62565571Aec87f1'], fromBlock: 38411390, start: '2025-11-28' },
 }
 
+const METRIC = {
+  TAKER_FEES: "Taker Fees",
+  MAKER_FEES: "Maker Fees",
+  PASSIVE_ORDER_PAYOUT: "Passive Order Payout",
+  MARKETMAKER_COMMISSION: "Marketmaker Commission",
+  ADMINISTRATOR_COMMISSION: "Administrator Commission",
+};
+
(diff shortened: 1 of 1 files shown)


bheluga:
@DaniiRix thanks for the PR

PR #7915, opened 2026-07-02
gmx v1: add deadFrom dates

DaniiRix:
Ran a liquidation audit across all 68 adapters and found 6 GMXV1 forks (alpacafinance-gmx, kinetix-v1, sobax-io, tsunami-fi, loxodrome-perp) with no liquidation activity


bheluga:
@DaniiRix thanks for the PR

PR #7922, opened 2026-07-02
durianfun aggregator: index the V2.4 / V2.5 / V2.67 routers

Duriandotfun:
## Update: Durianfun Aggregator — index the V2.4 / V2.5 / V2.67 routers

`aggregators/durianfun/index.ts` summed `Swapped` volume + fees from router V1 and the original V2 (`0xf3d6…0215`), but was missing the later live routers — so most recent aggregator volume + revenue went unindexed.

### Change (addresses only — same ABI, logic untouched)
Every V2-family router emits a **byte-identical 8-field `Swapped` event** (verified in-source across V2 / V2.4 / V2.5 / V2.67), so this just adds three addresses to `ROUTERS_V2`:
- V2.4  `0xaCaB7A476C4D7dad184EC0a55cD4203dF8E30b3F` (2026-05-28)
- V2.5  `0x9b7119cceAc97F6D3645a2b9302a85270Cc741Ac` (2026-05-28) — handled the bulk of swaps before the V2.67 cutover
- V2.67 `0xB2B3d1cf3249E4762F32d65441B546f9088A69F7` (current live; on-chain contract name DurianAggregatorRouterV3)

The existing fee/revenue accounting (router-skim `fee` per Swapped, tagged `METRIC.SWAP_FEES`) now also covers these routers.

### On-chain verification (KUB mainnet, block 33,124,390)
Decoding `Swapped` with the existing V2 ABI from block 32,000,000: **V2.5 = 64 swaps, V2.67 = 114 swaps**, all previously unindexed and growing (V2.67 is the live router). (V2.4 = 0, superseded quickly; V2's earlier activity is already covered.)

### DIFF
--- a/aggregators/durianfun/index.ts
+++ b/aggregators/durianfun/index.ts
@@ -3,7 +3,7 @@
  * ── Protocol ───────────────────────────────────────────────────────
  *
  * `DurianAggregatorRouter` is a Jupiter-style atomic multi-DEX router
- * on Bitkub Chain. There are TWO deployed generations:
+ * on Bitkub Chain. Deployed generations — V1 plus the V2 family:
  *
  *   V1 (canonical pre-V466 router, LIVE for pure-V4.5/Udonswap routes):
  *     - Address: 0x5078cE74728bC3F1313B55A04B79E227E1181918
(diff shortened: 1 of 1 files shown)


bheluga:

PR #7926, opened 2026-07-02
[fix] Pacifica - volume-only, correct v2 day windowing

### LINKED ISSUE #5245: Track Pacifica Fee/Revenue

0xshubhs:
## Summary

Pacifica volume was already tracked, but fees and revenue were missing (issue #5245). This derives fees from matched notional using the published fee schedule.

- **Fees:** 0.0200% taker + 0.0075% maker = **0.0275%** on matched notional.
- `dailyVolume` is the matched notional (Pacifica reports taker + maker volume, which we halve). Every matched trade has exactly one taker and one maker, so the protocol collects the combined taker + maker fee on that notional.
- All trading fees accrue to the protocol, so `Revenue = ProtocolRevenue = Fees`.
- Adds a `methodology` block documenting Volume / Fees / Revenue / ProtocolRevenue.

Fee schedule reference: https://docs.pacifica.fi/trading-on-pacifica/trading-fees

Closes #5245

## Testing

Both API code paths tested against live data:

| Path | Volume | Fees | Revenue | Protocol Revenue |
|------|--------|------|---------|------------------|
| `pnpm test dexs pacifica` (rolling 24h) | $499.09M | $137.25k | $137.25k | $137.25k |
| `pnpm test dexs pacifica 2026-06-20` (backfill) | $388.31M | $106.79k | $106.79k | $106.79k |

### DIFF
--- a/dexs/pacifica/index.ts
+++ b/dexs/pacifica/index.ts
@@ -3,6 +3,15 @@ import { CHAIN } from "../../helpers/chains";
 import { FetchOptions, SimpleAdapter } from "../../adapters/types";
 import fetchURL, { fetchURLAutoHandleRateLimit } from "../../utils/fetchURL";
 
+// Pacifica base trading fees: 0.0200% taker + 0.0075% maker.
+// `dailyVolume` below is the matched notional (Pacifica reports taker + maker volume,
+// which we halve), and every matched trade has exactly one taker and one maker, so the
+// protocol collects (taker + maker) fee on that notional.
+// https://docs.pacifica.fi/trading-on-pacifica/trading-fees
+const TAKER_FEE = 0.0002;
+const MAKER_FEE = 0.000075;
+const TOTAL_FEE_RATE = TAKER_FEE + MAKER_FEE;
+
 const fetch = async (options: FetchOptions) => {
   const todayStartOfDay = Math.floor(Date.now() / 86_400_000) * 86_400;
   // The runner calls us with the just-completed day's startOfDay; rolling 24h
@@ -16,7 +25,8 @@ const fetch = async (options: FetchOptions) => {
     for (const row of prices.data) {
       dailyVolume += (+row.volume_24h/2); // they include taker + maker
     }
-    return { dailyVolume };
+    const dailyFees = dailyVolume * TOTAL_FEE_RATE;
+    return { dailyVolume, dailyFees, dailyRevenue: dailyFees, dailyProtocolRevenue: dailyFees };
   }
 
   const data = await fetchURL('https://api.pacifica.fi/api/v1/info')
@@ -39,7 +49,8 @@ const fetch = async (options: FetchOptions) => {
       await new Promise(r => setTimeout(r, 4000));
     })
 
-  return { dailyVolume }
+  const dailyFees = dailyVolume * TOTAL_FEE_RATE;
+  return { dailyVolume, dailyFees, dailyRevenue: dailyFees, dailyProtocolRevenue: dailyFees }
 }
 
 const adapter: SimpleAdapter = {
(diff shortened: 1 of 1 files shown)


0xshubhs:
@coderabbitai review

0xshubhs:
Addressed the review feedback:

- **Fees marked as an estimate** (CodeRabbit): Pacifica's public API exposes no protocol-wide realized-fee total, and effective rates are tiered per account (`fee_level`/`maker_fee`/`taker_fee`), so applying the base-tier rate to notional is an upper bound. Documented this in both the code comment and the `Fees` methodology (VIP flow pays less → realized fees ≤ estimate).
- **v2 windowing**: switched from `options.startOfDay` to `options.startTimestamp` for both the recent-day check and the kline query/filter.
- **pullHourly**: set `pullHourly: false` with a comment — only daily granularity is available (rolling-24h `/info/prices` snapshot + 1d klines).
- **Title**: reformatted to `[feat] Pacifica - track perp fees and revenue`.

Retested; adapter runs green.

0xshubhs:
@coderabbitai review

bheluga (changes requested):
@0xshubhs thanks for the PR
This isn't an ideal approach. We just assume that all people fall under the same fee tier, which may not be true.
It would be ideal if we can fetch from trade events or any official source

0xshubhs:
@bheluga thanks, that's fair. I dug into what Pacifica actually exposes:

- There's no protocol-wide realized-fee total anywhere in the public API.
- The public trade feed (`/trades`) returns individual fills (taker/maker, price, amount) but carries no per-fill fee, so realized fees can't be reconstructed from trade events either.
- The only official fee source is the tiered fee schedule at `/info/fees` (levels 0–7, maker/taker rate per tier). It doesn't expose which tier a given account/fill sits at — just the schedule.

So realized fees genuinely aren't derivable from the public API today. Given that, my plan is to stop hardcoding the rate (the current hardcoded 0.0200%/0.0075% is also just wrong — the official level-0 is 0.0400%/0.0150%) and instead fetch `/info/fees` live and apply the base tier (level 0, the highest rate) to matched notional. That makes it an explicit upper bound taken from Pacifica's own schedule: realized <= this, since any higher-volume account pays a lower tier.

Does that direction work for you, or would you rather I drop the fee/revenue fields and keep this adapter volume-only until Pacifica exposes realized fees? Happy to go either way.

bheluga:
> @bheluga thanks, that's fair. I dug into what Pacifica actually exposes:
> 
> * There's no protocol-wide realized-fee total anywhere in the public API.
> * The public trade feed (`/trades`) returns individual fills (taker/maker, price, amount) but carries no per-fill fee, so realized fees can't be reconstructed from trade events either.
> * The only official fee source is the tiered fee schedule at `/info/fees` (levels 0–7, maker/taker rate per tier). It doesn't expose which tier a given account/fill sits at — just the schedule.
> 
> So realized fees genuinely aren't derivable from the public API today. Given that, my plan is to stop hardcoding the rate (the current hardcoded 0.0200%/0.0075% is also just wrong — the official level-0 is 0.0400%/0.0150%) and instead fetch `/info/fees` live and apply the base tier (level 0, the highest rate) to matched notional. That makes it an explicit upper bound taken from Pacifica's own schedule: realized <= this, since any higher-volume account pays a lower tier.
> 
> Does that direction work for you, or would you rather I drop the fee/revenue fields and keep this adapter volume-only until Pacifica exposes realized fees? Happy to go either way.

I think its good to keep it volume only until fees can be calculated correctly. Feel free to open a new PR if you get any lead.

0xshubhs:
Yeah that makes sense — switched it to volume-only and dropped the fee/revenue fields. Kept the v2 windowing fix and pullHourly while I was in here. I'll open a separate PR if I come across a proper source for realized fees. Thanks!

PR #7952, opened 2026-07-04
Fix velodrome v2

uwezukwechibuzor:
### Problem

1. **Volume dropped.** `customLogic` didn't return the `dailyVolume` the helper computed, so it was discarded and reported as `0` across all chains, volume last returned **June 24th 2026**.
2. **Wrong Swap event on 6 leaf chains.** Mode, Lisk, Fraxtal, Ink, Soneium, and Unichain omitted `swapEvent` and fell back to the standard uni-v2 `Swap` event instead of Velodrome's own, so zero logs matched and volume were `0` on those chains (OPTIMISM/BOB already set it).

### Solution

1. Return `dailyVolume` from `customLogic`
2. Add `swapEvent` to the six leaf configs.

https://defillama.com/protocol/velodrome-v2


bheluga:

PR #7956, opened 2026-07-04
element: fix revenue, return protocol revenue instead of total fees

uwezukwechibuzor:
## Summary
Fixes Element revenue, which was reporting total fees (protocol + creator royalties) instead of just Element's feeCollector share.
https://defillama.com/protocol/element-nft-marketplace


bheluga:

PR #7957, opened 2026-07-04
teleswap: split locker fee into supply side revenue

DaniiRix:
The adapter was counting the **Locker fee** as protocol revenue. Since Lockers provide the collateral that backs `teleBTC`, this fee belongs to the supply side, not the protocol.

This change only reclassifies existing fees: no fee calculations, volume, or event parsing have changed.

| Metric | Before | After |
| --- | --- | --- |
| `dailyFees` | locker + protocol | locker + protocol |
| `dailySupplySideRevenue` | — | locker fee |
| `dailyRevenue` | locker + protocol | protocol fee |
| `dailyProtocolRevenue` | locker + protocol | protocol fee |

## Why

Locker fees are roughly **3× larger** than protocol fees. As a result, the adapter was significantly overstating protocol revenue while reporting no supply-side revenue, even though Lockers are paid on every bridge transaction.

## Verification

Verified against the deployed BOB contracts by resolving the EIP-1967 proxies and checking the router implementations.

Key findings:

- `protocolFee` is sent to the treasury.
- `lockerFee` is retained by the Locker.
- All router events use the same fee layout:
  - Index **1** = Locker fee
  - Index **2** = Protocol fee

On-chain event decoding confirms the split, with:

- `NewWrap`: `fees=[1000, 11850, 3950, 0]` (3:1 locker-to-protocol)
- `NewWrapAndSwap`: `fees=[1000, 26, 8, 0, 0]` (~3.25:1)

The referral/integrator fee (index 3) is forwarded to external integrators, so it remains part of `dailyFees` and `dailySupplySideRevenue` (currently zero). Other fee fields are passthrough costs and remain excluded.

### DIFF
--- a/dexs/teleswap/index.ts
+++ b/dexs/teleswap/index.ts
@@ -32,6 +32,15 @@ interface WrapAndSwapV2Event {
   fees: [number, number, number, number, number];
 }
 
+type Balances = ReturnType<FetchOptions["createBalances"]>;
+
+interface Accumulators {
+  dailyVolume: Balances;
+  dailyFees: Balances;
+  dailySupplySideRevenue: Balances;
+  dailyProtocolRevenue: Balances;
+}
+
 // Configuration
 const CHAIN_CONFIGS: Partial<Record<CHAIN, ChainConfig>> = {
   [CHAIN.POLYGON]: {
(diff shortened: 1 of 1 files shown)


bheluga:
@DaniiRix thanks for the PR

PR #7958, opened 2026-07-04
invariant: add protocol fee, revenue and supply side revenue

DaniiRix:
Clarify fee accounting by documenting that the on-chain create_pool instruction hardcodes a 1% protocol fee, while the remaining 99% of fees are distributed to LPs.

https://github.com/invariant-labs/protocol/blob/master/programs/invariant/src/instructions/create_pool.rs


bheluga:
@DaniiRix thanks for the PR

PR #7959, opened 2026-07-04
opinion: fix fee attribution, add supply side revenue

DaniiRix:
Fix Opinion fee attribution to correctly separate gross fees, referral rebates, and protocol revenue.

- dailyFees now reports gross taker fees
- dailySupplySideRevenue tracks referral rebates
- dailyRevenue / dailyProtocolRevenue = fees − rebates
- Added dailyUserFees and simplified the methodology description
- Kept allowNegativeValue since rebates can exceed fees in a given period


bheluga:
@DaniiRix thanks for the PR

PR #7961, opened 2026-07-04
deluthium: fix fee token and double-counted volume

uwezukwechibuzor:
## Summary
Fees were tracked against feeTo (a recipient address, not a token) and volume was double-counted, fixed by charging fees on inputToken (the token the fee is actually taken from, per the settlement contract) and removing the redundant dailyVolume.add.


bheluga:

PR #7964, opened 2026-07-04
[feat] 0x-Aggregator: track robinhood on 0x aggregator

Shaileshkhote:
https://defillama.com/protocol/0x-aggregator


bheluga:

PR #7971, opened 2026-07-05
[feat] Rooster Protocol V2 - add on-chain fees/revenue adapter

dragos-rebegea:
https://defillama.com/protocol/rooster-protocol-v2

Adds an on-chain fees & revenue adapter for **Rooster Protocol V2**, an Algebra Integral concentrated-liquidity DEX on Plume. The protocol is already listed for TVL (`rooster-protocol-v2`) but has no fees/revenue yet.

This is a **new module** (`dexs/rooster-v2`), separate from the existing `dexs/rooster`, which tracks the Maverick-based **Rooster V1** (`rooster-protocol-v1`) on a different factory. Please attach this module to the existing DefiLlama protocol slug **`rooster-protocol-v2`**.

### How it works (fully on-chain, no subgraph)
- Discovers all pools from the AlgebraFactory (`0x1eB9822d5176C88B1d4eec353fa956C896D77Df9`) `Pool` / `CustomPool` events.
- For each swap, reads the pool `Swap` event (token deltas → volume) and the paired `SwapFee` event, whose `overrideFee` is the exact dynamic fee applied to that swap (1e-6 units). Fee is charged on the input amount, so `fee = inputAmount * feeRate / 1e6`.
- Revenue split uses each pool's on-chain `communityFee` (1e-3 units) from `globalState()`: protocol revenue = `fees * communityFee/1000`, the remainder is supply-side (LP) revenue.

### Local test results (`pnpm test dexs rooster-v2`)
- 2026-07-02 → volume ~$14.9k, fees ~$12.00, revenue ~$2.33, supply-side ~$9.33
- 2026-06-14 → volume ~$16.5k, fees ~$10.00, revenue ~$2.05, supply-side ~$8.21
- Invariant `Fees = Revenue + SupplySideRevenue` holds.

### DIFF
--- a/dexs/rooster-v2/index.ts
+++ b/dexs/rooster-v2/index.ts
@@ -1,0 +1,110 @@
+// Rooster Protocol V2 - Algebra Integral concentrated-liquidity DEX on Plume.
+// Volume, fees and revenue are computed fully on-chain from pool events:
+//  - Every swap emits `Swap(...amount0, amount1...)` (token deltas -> volume) and,
+//    in the same tx immediately before it, `SwapFee(sender, overrideFee, pluginFee)`
+//    where `overrideFee` is the exact dynamic fee applied to that swap (units 1e-6).
+//  - The Algebra fee is charged on the input amount, so for every swap
+//    feeAmount(inputToken) = inputDelta * feeRate / 1e6  (exact for exact-in and exact-out).
+//  - The protocol's cut is the pool `communityFee` (units 1e-3), read on-chain from
+//    `globalState()`; the remainder accrues to LPs.
+import { FetchOptions, SimpleAdapter } from "../../adapters/types";
+import { CHAIN } from "../../helpers/chains";
+
+const FACTORY = "0x1eB9822d5176C88B1d4eec353fa956C896D77Df9";
+const FACTORY_FROM_BLOCK = 23181592; // block of the first pool creation on Plume
+const FEE_DENOMINATOR = 1e6; // Algebra Constants.FEE_DENOMINATOR (fee in hundredths of a bip)
+const COMMUNITY_FEE_DENOMINATOR = 1e3; // Algebra Constants.COMMUNITY_FEE_DENOMINATOR (communityFee in thousandths)
+
+const POOL_CREATED = "event Pool(address indexed token0, address indexed token1, address pool)";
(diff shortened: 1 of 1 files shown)


dragos-rebegea:
Addressed the automated review feedback in the latest commit:

- **`pluginFee`**: now included in `dailyFees` and attributed to protocol revenue. Note: Rooster's plugins (`AlgebraLimitOrderPlugin`, `AlgebraBasePluginV2/V1`) all return `beforeSwap(...) => (selector, fee, 0)`, so `pluginFee` is currently always `0` — this change keeps the accounting correct if a fee-charging plugin is ever enabled.
- **`pullHourly: true`**: set explicitly per the v2 guideline. Also optimized to only `multiCall` token pair + `globalState` for pools that actually traded in the window.
- **FACTORY source**: added a comment with an explorer/source reference next to the hardcoded address.
- **Methodology keys**: removed `UserFees` (not an allowed display-name key) and the redundant `dailyUserFees` return.
- **Title**: updated to the required `[type] protocol-name - description` format.

Re **`breakdownMethodology`**: omitted intentionally — the adapter adds balances by token address with no sub-category labels in `.add()`, so there are no breakdown labels to document (consistent with existing Algebra subgraph adapters like `dexs/stellaswap-v4` and `dexs/swapr-v3`).

Invariant `dailyFees = dailyRevenue + dailySupplySideRevenue` holds; re-tested for multiple days on Plume.

dragos-rebegea:
Addressed the follow-up review comment (line 63):

- Removed `permitFailure: true` from the `token0`/`token1`/`globalState()` multicalls for active pools. Those pools all come from the factory's own `Pool`/`CustomPool` creation events, so their metadata must resolve — a failed call is a transient RPC error and should now fail the run (and be retried) rather than silently zero-out and undercount a pool that traded.

Re-tested on Plume; the `dailyFees = dailyRevenue + dailySupplySideRevenue` invariant still holds.

bheluga:
@dragos-rebegea thanks for the PR

dragos-rebegea:
Thanks for merging! 🙏

The adapter is live but fees/revenue aren't showing yet on the protocol page: https://defillama.com/protocol/rooster-protocol-v2 (id 6818).

Could you attach the `rooster-v2` module to that protocol? Happy to rename the adapter if it needs to match a specific convention.

PR #7974, opened 2026-07-06
[fees] PumpBox - add PumpDaily subscription fees

miles-crypto:
Updated the adapter to keep `dailyFees` gross.

Buyback spends are now reported under `dailySupplySideRevenue` instead of being netted out of `dailyFees`. `dailyRevenue` and `dailyProtocolRevenue` are calculated as gross fees minus buyback spends, so the income statement now follows:

`dailyRevenue = dailyFees - dailySupplySideRevenue`

Validation:

- `pnpm test fees pumpbox 2026-07-03`
- `pnpm test fees pumpbox 2026-07-04`


bheluga:
@miles-crypto thanks for the PR
Buybacks should directly be subtracted from fees . We use same logic for all similar adapters

PR #7987, opened 2026-07-06
metamask-card: add Monad (FoxConnect contracts, live since Jun 12)

clutchpbcfo:
MetaMask officially lists Monad as a supported card chain (metamask.io/card), but the adapter only covers Linea, Base, and Solana. This adds the Monad leg.

Addresses are sourced from MetaMask's own app config (metamask-mobile, app/selectors/featureFlagController/card/index.ts, config key eip155:143, foxConnectAddresses): global 0x40A695A16C213afEf1c87Fd471Fb73157b948f3f, US 0x144c1cE815Bd1Eb71678978fE8641cC4e3fd59e6. The same file lists the exact Linea and Base contracts this adapter already uses, so the config family matches.

Verified onchain via Monad RPC: both are deployed ERC1967 proxies sharing implementation 0x716bf8c59a6bfc7ef606ddfcabbd796110df5518, the same proxy pattern as the Base relayer (whose implementation is verified as W3Card on Base). Deployment block 80727203 (2026-06-12), which is the start date used. Same withdraw() decode path as the existing chains, no logic changes.

Note for maintainers: contracts are ~3 weeks old, so daily volume is early-stage. Local test hits the Allium key requirement (Solana leg), which the error itself says to ignore for the bot.

### DIFF
--- a/dexs/metamask-card.ts
+++ b/dexs/metamask-card.ts
@@ -22,6 +22,18 @@ const configs: Record<string, { withdrawContracts: Array<string>, start: string,
     ],
     getTrasnactionLimit: 10000,
   },
+  [CHAIN.MONAD]: {
+    // FoxConnect (MetaMask Card) contracts on Monad (eip155:143), sourced from
+    // MetaMask's own app config: metamask-mobile app/selectors/featureFlagController/card/index.ts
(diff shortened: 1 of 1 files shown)


bheluga:
@clutchpbcfo thanks for the PR

PR #7988, opened 2026-07-06
fix ts-check: restore polygon PUSD (dropped by core assets update)

clutchpbcfo:
Today's core assets update (6f7b6eb) dropped the polygon PUSD entry from helpers/coreAssets.json. fees/polymarket.ts references ADDRESSES.polygon.PUSD in 7 places, so tsc currently fails repo-wide and every new PR's ts-check is red (example: #7987).

This restores the exact entry that 8a5cd24 (#7568, "fix ts errors") added on June 11, the last time the core assets update dropped it. Verified onchain before restoring: 0xC011a7E12a19f7B1f670d46F03B03f3342E82DFB on Polygon is live with symbol pUSD, 6 decimals.

Since the update script has now dropped this entry twice, it may be worth either pinning it in the generator or moving the constant into fees/polymarket.ts. Happy to do either.

Verified locally: pnpm run ts-check passes with this change.


bheluga:
@clutchpbcfo thanks for the PR

PR #7999, opened 2026-07-07
ondo-global-markets: add Solana volume coverage

DaniiRix:
Adds **Solana** support to the Ondo Global Markets adapter

## Changes

- Added Solana support to `dexs/ondo-global-markets/index.ts`.
- Decodes trade volume from Dune `solana.instruction_calls`, since Solana's `TradeExecuted` event only includes an `execution_id`.
- Uses the same `price × amount` calculation as the on-chain program, matching the contract's implementation.
- Moved the adapter from `version: 2` to `version: 1` to support the mixed EVM + Dune fetch pattern. Existing EVM logic is unchanged.

## Verification

- Verified the program ID and all instruction discriminators against the open-source program and live on-chain data.
- Cross-checked the adapter against Dune for **2026-07-01**: **607 trades** totaling **$2.35M**, matching the adapter output exactly.


bheluga:
@DaniiRix thanks for the PR

PR #8009, opened 2026-07-08
curve: add bsc and fantom

DaniiRix:
Both show real volume in Curve's own fees API but were never counted

bsc $597/$0, fantom $1,034/$2. Fantom ranged $16–$80k/day over two weeks


bheluga:
@DaniiRix thanks for the PR

PR #8012, opened 2026-07-08
easya-kickstart: attribute creator fee share correctly

platonicsocrates:
## What

Two attribution fixes for the EasyA Kickstart fees adapter (we are the protocol team). Fees totals are unchanged; this only moves value between Revenue and SupplySideRevenue.

1. **DBC bonding-curve trading fees were booked 100% as Kickstart revenue.** On-chain, every config except `NHT6MNushFNWpaFgQs5k49HHzsas9jQAVoRvqyXc5Qx` sets `creator_trading_fee_percentage = 50`, so half of those fees belong to token creators and are claimable via `claim_creator_trading_fee`. The creator half now goes to SupplySideRevenue as "Trading Fees to Creators" (same convention as pump.fun's creator slice).

2. **The DAMM v2 creator share silently fell back to a 50% default for every pool.** The previous query read `JSON_EXTRACT_SCALAR(config_parameters, '$.creator_liquidity_percentage')`, but the JSON is wrapped (`$.ConfigParameters...`), so the extraction returns NULL for all 8 configs and `COALESCE(..., 50)` applied everywhere. Visible on the live page: cumulative "DAMM v2 Fees to Creators" is exactly 50.000% of cumulative DAMM v2 LP fees. That happens to be right for seven configs but wrong for `NHT6...`, where Kickstart holds 100% of the permanently-locked LP. The splits are immutable PoolConfig constants and are now hardcoded per config (values cross-checked against both the on-chain PoolConfig accounts via RPC and the CreateConfigV2 event JSON on Dune).

A comment-only second commit corrects a wrong fee_claimer address in the header comment and documents that three pre-launch dev/testing configs (~0.64 SOL lifetime fees combined) plus one config created by an unrelated third-party deployer (which merely names our wallet as fee_claimer and routes 100% of fees to token creators) are deliberately excluded; the production list stays exactly these eight.

## Verification

Everything below was executed with a Dune API key through the same `queryDuneSql` template (3996608) this repo uses, plus independent RPC reads of Meteora's on-chain state.

**1. The current adapter's displayed numbers were reproduced first** (proves we tested the production pipeline): re-running the deployed SQL day by day for Jun 15-21 and pricing at DefiLlama's own coins API matches the live charts within +/-1.4% on both Fees and Revenue for all 7 days.

**2. Dune event sums equal Meteora's on-chain lifetime counters.** Per config, summing `trading_fee` from swap events vs each pool's `metrics.total_trading_quote_fee` across all 2,920 pools: 0.00% difference on every config (2,841.9 SOL total). DAMM v2 LP fees: 7,041.8 SOL from events vs 7,027.5 SOL from all 122 position counters (+0.20%, a few closed third-party positions).

**3. Receipts reconcile to the lamport.** Sum of all `claim_trading_fee` events + currently unclaimed `partner_quote_fee` equals the lifetime counter with 0.000 SOL difference on every config; every partner claim tx ever was signed by the fee_claimer wallet. Token creators have made 3,077 `claim_creator_trading_fee` transactions from 1,577 distinct wallets totalling 1,202.009 SOL, matching the chain-derived creator-claimed figure exactly - the 50% creator split is observed money movement, not an assumption.

**4. `npm run test -- fees easya-kickstart 2026-06-20` on both versions** (Jun 19 UTC window):

| | master | this PR | reference |
|---|---|---|---|
| Daily fees | $11.44k | $11.44k (identical) | site shows $11,478 |
| Daily revenue | $5.05k | $4.62k | site today: $5,065; on-chain-verified: $4,618 |
| DBC trading split | $868 all to Kickstart | $435 / $434 creators | NHT pools pay creators 0% |
| DAMM v2 split of $8,361 | $4,181 / $4,181 flat | $4,184 / $4,177 creators | NHT graduated pools 100% Kickstart |

master reproduces the currently displayed (wrong) revenue to the dollar; the fix lands on the independently verified on-chain number.

**5. Referral fees, verified per swap** (added in the third commit): both DBC and post-migration cp-amm swaps carry Meteora's referral mechanism - routing frontends that pass a referral account receive 20% of Meteora's protocol share (4% of the gross fee). Measured on Kickstart pools: the trading/LP share is exactly 80.00% of the gross fee with and without a referrer, so referral fees never touch protocol or creator revenue. The adapter previously dropped the DAMM-side referral fees (54.6 SOL lifetime); they now count in Fees and SupplySideRevenue.

## Ask

Since this is a methodology correction, please refill the historical data. Independently of this fix, the stored datapoint for 2026-07-06 is significantly under-reported (~$15.1k stored vs ~$39.2k when the settled events are summed now, at that day's price) - it appears the daily snapshot ran before Dune had fully ingested a high-volume day, so a refill would correct that too.


bheluga:

PR #8023, opened 2026-07-08
PredictStreet: fees adapter (ADI)

rostyslavbortman-adi:
Adds a fees adapter for PredictStreet (already listed: TVL + volume adapters).

Fees are charged per-fill on the CTFExchange / NegRiskCtfExchange and accrue on the settlement operator wallets (partly as USDC.e, partly as outcome tokens redeemed after market resolution). A daily job consolidates everything into the protocol fee wallet as USDC.e; the adapter counts the USDC.e arriving at that wallet (addTokensReceived). All fees are kept by the protocol, so Fees = Revenue = ProtocolRevenue.

USDC.e is priced as CoinGecko usd-coin (bridged $1 stablecoin, same convention as the merged volume adapter)

### DIFF
--- a/fees/predictstreet.ts
+++ b/fees/predictstreet.ts
@@ -1,0 +1,47 @@
+import { FetchOptions, FetchResultV2, SimpleAdapter } from "../adapters/types";
+import { CHAIN } from "../helpers/chains";
+import { addTokensReceived } from "../helpers/token";
+
+// USDC.e collateral, 6 decimals.
+const USDC_E = "0x9cb8142aebbcdc60af7c97af897a67a8f3ca71c2";
+
+// Protocol fee wallet. Fees are charged per-fill on the exchanges and accrue on
+// the settlement operator wallets (partly as USDC.e, partly as outcome tokens
+// that are redeemed after market resolution); a daily job consolidates all of it
+// into this wallet as USDC.e. Counting the USDC.e that arrives here gives the
+// protocol's realized fees. Senders are deliberately not filtered: the operator
+// wallet pool rotates, and pinning it would silently drop sweeps from new wallets.
+const FEE_WALLET = "0x0a3EDDe878fa0f5a9A8c95C8054283Ffb2fb0df2";
+
+const fetch = async (options: FetchOptions): Promise<FetchResultV2> => {
+  // skipIndexer: DefiLlama's indexer doesn't cover this chain — go straight to getLogs.
+  const received = await addTokensReceived({ options, tokens: [USDC_E], targets: [FEE_WALLET], skipIndexer: true });
+
+  // USDC.e is a $1 stablecoin on a brand-new chain DefiLlama can't auto-price, so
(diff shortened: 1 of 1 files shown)


bheluga (dismissed):
@rostyslavbortman-adi thanks for the PR
I don't get the point on how the volume is less than 300k whereas fees is around 650k

rostyslavbortman-adi:
@bheluga 

Thanks for the review. That ~$650k day is not fees earned in one day

Fees are charged on every fill and accrue on our settlement operator wallets (partly as USDC.e, partly as outcome tokens that only become redeemable after market resolution). A daily consolidation job sweeps everything into the fee wallet. That job was rolled out gradually with a conservative per-day transfer cap while we monitored it, so fees accrued on the operator wallets faster than they were being swept. A few days ago we lifted the cap, and the whole backlog accumulated since launch ($917k in total) — arrived at the fee wallet in one day. That's the spike the adapter picked up

The adapter reports realized fees (USDC.e actually reaching the fee wallet), attributed to the sweep day. Now that the sweep runs daily without the cap, daily numbers will track actual daily fees. Total fees to date (~$917k) are consistent with cumulative volume since launch

bheluga:
> @bheluga
> 
> Thanks for the review. That ~$650k day is not fees earned in one day
> 
> Fees are charged on every fill and accrue on our settlement operator wallets (partly as USDC.e, partly as outcome tokens that only become redeemable after market resolution). A daily consolidation job sweeps everything into the fee wallet. That job was rolled out gradually with a conservative per-day transfer cap while we monitored it, so fees accrued on the operator wallets faster than they were being swept. A few days ago we lifted the cap, and the whole backlog accumulated since launch ($917k in total) — arrived at the fee wallet in one day. That's the spike the adapter picked up
> 
> The adapter reports realized fees (USDC.e actually reaching the fee wallet), attributed to the sweep day. Now that the sweep runs daily without the cap, daily numbers will track actual daily fees. Total fees to date (~$917k) are consistent with cumulative volume since launch

Cumulative volume is around 25M, does that mean fees is around 4%?
Any particular reason , more than 650K received yesterday than on daily basis?

rostyslavbortman-adi:
1. Yes, that's roughly right - the fee schedule is parabolic: each order pays 6% × P × (1−P) on its outcome-token leg, which peaks at 1.5% of notional at P = 0.50, and it is charged on both the maker and the taker side. So cumulative fees landing in the low single-digit percent of cumulative volume is expected under this schedule
2. No ongoing reason - that was a one-time consolidation. The fee-reconciliation process ran with a conservative per-day transfer cap while it was being rolled out, so a backlog built up on the operator wallets. Lifting the cap flushed everything accrued since launch into the fee wallet in one day. The process now runs on a daily basis, so the wallet receives each day's fees the next day - there won't be spikes of that size again

@bheluga

bheluga:
@rostyslavbortman-adi thanks for the PR

PR #8043, opened 2026-07-09
fix gwei-domains: restore revenue fields so breakdown validation passes

lucadonnoh:
## Problem

`fees/gwei-domains.ts` throws on every run, so Gwei Name Service (protocol id 8160) has never stored a datapoint: `/summary/fees/gwei-name-service` 404s and the protocol page at https://defillama.com/protocol/gwei-name-service is empty.

```
Error: found dailyFees record but missing all dailyRevenue, dailySupplySideRevenue, dailyProtocolRevenue records
    at validateAdapterResult (adapters/utils/runAdapter.ts:545)
```

The adapter labels fees with `METRIC.SERVICE_FEES`, so `validateAdapterResult` requires at least one revenue field alongside `dailyFees`. The adapter as first merged in 60dc5be3 returned the revenue fields and passed, but a77e81b6 ("track wei.domains") changed the return to `{ dailyFees, }`, which trips the validator. The sibling adapter `fees/wei-domains.ts` kept its revenue fields and is indexing fine.

## Fix

Restore the original return, matching the methodology already in the file (fees are locked forever in the ownerless contract, i.e. burned, accruing to ETH holders):

```ts
return { dailyFees, dailyRevenue: dailyFees, dailyHoldersRevenue: dailyFees };
```

If you'd rather not count burned fees as revenue, `dailyRevenue: 0` or `skipBreakdownValidation: true` also pass; happy to change to whichever you prefer.

## Test

`npm test fees gwei-domains` fails on master with the error above; with this change it passes:

```
ETHEREUM
Daily fees: 1.24 k
Daily revenue: 1.24 k
Daily holders revenue: 1.24 k
```

The NameNFT contract (0x9d51d507bc7264d4fe8ad1cf7fe191933a0a81d6) currently holds 9.68 ETH of locked fees, so there is real history to backfill from the 2026-06-26 start.

### DIFF
--- a/fees/gwei-domains.ts
+++ b/fees/gwei-domains.ts
@@ -27,7 +27,7 @@ const fetch = async (options: FetchOptions) => {
   if (delta > 0n) dailyFees.addGasToken(delta, METRIC.SERVICE_FEES);
 
   // All fees are locked in the contract forever, so they are revenue that is fully burned.
-  return { dailyFees, dailyRevenue: dailyFees, dailyHoldersRevenue: dailyFees };
+  return { dailyFees, };
 };
 
 const adapter: SimpleAdapter = {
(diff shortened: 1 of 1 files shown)


bheluga:
@lucadonnoh thanks for the PR

lucadonnoh:
Thanks for merging! The adapter is now live and storing data, but history only starts on 2026-07-08: the 2026-06-26 to 2026-07-07 window is missing since the validation error blocked all runs before the fix. Could you trigger a refill from the adapter's start date? It reads balance deltas at historical blocks, so backfill should run clean.

bheluga:
Hey @lucadonnoh ,
I have refilled it
image

PR #8053, opened 2026-07-10
ring: price volume on the core asset

DaniiRix:
Fixes Ring DEX volume pricing to use the **core asset** side of each swap instead of the non-core asset, aligning the adapter with the standard pricing convention and preventing volume inflation from manipulated token prices.

## Changes

- Updated volume pricing to use the recognized core asset in each pair.
- Skip pairs where neither token is a supported core asset.
- Fee accounting is unchanged.

## Verification

- Verified wrapped token resolution against live contracts.
- Local before/after tests show expected volume changes while fees remain exactly **0.3%** of volume.
- The change removes a potential price-manipulation vector without affecting legitimate swaps.

### DIFF
--- a/dexs/ring-dex.ts
+++ b/dexs/ring-dex.ts
@@ -32,6 +32,7 @@ const RingDexConfigs: Record<string, IRingDexConfig> = {
 }
 
 const methodology = {
+  Volume: 'Swap volume is measured on the well-known token side of each pair.',
   Fees: 'User pays 0.3% fees on each swap.',
   Revenue: 'Protocol has no revenue.',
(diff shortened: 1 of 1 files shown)


bheluga:
@DaniiRix thanks for the PR.
Actually there is an issue with ring dex.
Recently too many scam tokens got launched on ring dex (even jaredfromsubway's bot fell for it)
so what actually happens is, they pair WETH with scam version of WETH, USDC with scam version of USDC.
And when they swap scam WETH with scam USDC, we used to count them using core asset (WETH and USDC), which would inflate volume by billions.
If you are curious, you can test both the versions of code with ```yarn test dexs ring-dex 2026-06-21```

PR #8054, opened 2026-07-10
figure markets: document zero fee market and methodology

DaniiRix:
- Made the zero-fee CONNECT/FUND handling explicit
- add methodology


bheluga:
@DaniiRix thanks for the PR

PR #8064, opened 2026-07-10
axiom: run one multi-chain Dune query via prefetch

DaniiRix:
Optimizes the Axiom adapter by replacing per-chain Dune queries with a single multi-chain `prefetch` query, reducing Dune credit usage without changing reported volume.

## Changes

- Moved the Solana and BNB queries into a single `prefetch` query using `UNION ALL` with a chain discriminator.
- Updated `fetch` to read its chain's results from `options.preFetchedResults`.
- Reused the shared `TIME_RANGE` helper for the BNB query.
- Kept the existing 10-hour Dune indexing guard.
- No changes to volume calculation, filters, or deduplication logic.

## Verification

- Verified the BNB router contract attribution against BscScan.
- Confirmed Axiom supports both Solana and BNB Chain.
- `pnpm test dexs axiom 2026-05-15` now performs a single Dune execution while returning the same results:
  - **Solana:** $50.86M
  - **BNB Chain:** $407.88k

### DIFF
--- a/dexs/axiom.ts
+++ b/dexs/axiom.ts
@@ -30,10 +30,19 @@ const bscTradeContract = '0x325098a6291a412bba7a52531ef05ac5dd7d5d6e';
 
 const formatAddresses = (addresses: string[]) => addresses.map((a) => `'${a}'`).join(', ');
 
-async function fetchSolana(options: FetchOptions) {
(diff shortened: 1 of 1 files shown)


bheluga:
@DaniiRix thanks for the PR

PR #8066, opened 2026-07-10
fix(usdai): fees adapter throwing on loans migrated from the v1 router

0xshubhs:
USD AI fees have been dead since 2026-07-01, the day #7874 shipped the loan router v2 support. api.llama.fi/summary/fees/usd-ai shows missing/zero days from Jul 1 onward (~$1.1M/30d before that).

Root cause: loans migrated from the v1 router are registered on v2 through a `LoanMigrated` event, not `LoanOriginated`, so `getV2CurrencyToken` finds no origination in the subgraph and throws, killing the whole day. First migration tx: 0xd9b28abebf9f53691b73255548366930a4b6dda31dbae2d62bd5d8877c18ea52 (Jul 2). Any repayment on a migrated loan since then (e.g. 0xbf8fbed0405342ca1bfc6437b42922adafe566c67e5673f4fa341fd61e9c9d1e on Jul 9) reproduces it.

One subtlety: the migration also redenominates the loan, v1 PYUSD loans became USDai loans, so falling back to the v1 origination's currency token would inflate amounts by 1e12 (6 vs 18 decimals). I hit exactly that on the first attempt ("value is too damn high"). The currency token instead comes out of the abi-encoded v2 loan terms stored on the `loanMigrateds` entity (4th word, fixed offset on all 17 migrations so far).

Tested the previously-failing days:
- Jul 9→10 (the repayment above): $4.07k fees / $407 revenue, 10% admin clip checks out
- Jul 1→2: $190.4k, Jul 2→3 (migration day): $91.8k, Jul 5→6: $701
- Days where the current code already succeeds are untouched, the change only replaces the throw path.

### DIFF
--- a/fees/usdai/index.ts
+++ b/fees/usdai/index.ts
@@ -65,6 +65,14 @@ const loanOriginatedByHashQuery = gql`
   }
 `;
 
+const loanMigratedByV2HashQuery = gql`
+  query GetLoanMigrationByV2Hash($loanTermsHash: Bytes!) {
+    loanMigrateds(where: { loanTermsHashV2: $loanTermsHash }, first: 1) {
+      loanTerms
+    }
+  }
+`;
+
 // 10,000 basis points = 100%, 10% of interest going to protocol
 const BASIS_POINTS_SCALE = 10_000n;
 const BASE_YIELD_ADMIN_FEE_RATE = 1_000n;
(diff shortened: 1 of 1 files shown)


bheluga:
@0xshubhs thanks for the PR

0xshubhs:
lfgs

PR #8077, opened 2026-07-11
relay: add bridge aggregator volume

DaniiRix:
Adds a new **Relay** bridge adapter, tracking origin-side bridge volume across **43 chains**.

## Changes

- Added `bridge-aggregators/relay/` using the `RelayDepository` deposit events as the source of truth.
- Uses a single cross-chain Dune `prefetch` query, with each chain reading its results from `preFetchedResults`.
- Tracks both native and ERC-20 deposits, with pricing handled through `createBalances`.
- Added per-chain start dates based on each chain's first Relay deposit.

## Verification

- Verified the `RelayDepository` contract, deposit events, and identical deployment across supported chains.
- Confirmed the deposit decoding against live Base transactions.
- Verified the adapter counts **origin-side deposits only**, avoiding double-counting destination fills or legacy settlement events.
- Sample run (2026-07-08) reports **~$158.6M** in daily bridge volume across 43 chains

### DIFF
--- a/bridge-aggregators/relay/index.ts
+++ b/bridge-aggregators/relay/index.ts
@@ -1,4 +1,6 @@
+import { FetchOptions } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
+import { queryDune } from "../../helpers/dune";
 
 // Relay (relay.link) cross-chain intent bridge. Users deposit funds into the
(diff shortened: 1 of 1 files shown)


bheluga:
@DaniiRix thanks for the PR
Relay is  bridge , and we already track its volume in bridges repo.
https://defillama.com/bridge/relay

PR #8087, opened 2026-07-13
[fix] t3tris finance - add Robinhood Chain to fees adapter

t3trominos:
T3tris Finance is now deployed on Robinhood Chain (4663). This adds `robinhood` to the fees adapter's chainConfig. It also makes a chain with no verified vaults yet return zero fees instead of throwing, so Robinhood doesn't error before its first vault is live — the API-down guard is preserved (still throws if the whole vaults API returns nothing). Vaults are auto-discovered from the T3tris ecosystem API per chainId with the same curation gate (verified, non-blacklisted).


bheluga:
@t3trominos thanks for the PR

PR #8121, opened 2026-07-14
ref-finance: fix revenue split and add supply side

DaniiRix:
Verifies the Ref fee split against the live NEAR contract and fixes fee accounting in the adapter.

## Changes

- Replaced the unverified hardcoded protocol fee split with the on-chain value.
- Added the missing `dailySupplySideRevenue`.
- Added a methodology section documenting the fee distribution and wash-trade filter.

## Verification

- Verified the exchange fee configuration via the deployed NEAR contract.
- Confirmed how protocol, LP, and referral fees are distributed.


bheluga:
@DaniiRix thanks for the PR

PR #8122, opened 2026-07-14
extended: removed the halving of trading volume

DaniiRix:
Fixes Extended volume by removing an incorrect `/2` that assumed `tradingVolume` was double-counted. The API already reports one-sided taker volume.

## Verification

- Confirmed `tradingVolume` matches the exchange's per-market volume.
- Cross-checked open interest and 24h volume against CoinGecko.
- Sample daily volume increases from **~$253M → ~$506M**, while open interest is unchanged.


bheluga:
@DaniiRix thanks for the PR
They do double count taker and maker volumes, verified against historic trades https://app.extended.exchange/trade

PR #8126, opened 2026-07-14
metric: replace with onchain logs

DaniiRix:
Removes Metric's dependency on a hardcoded HTTP endpoint by discovering pools directly on-chain.

## Changes

- Replaced the external pool metadata source with on-chain `Initialize` event discovery.
- Builds the pool → token mapping from creation events, while keeping the existing swap volume calculation unchanged.
- Eliminates the adapter's reliance on an undocumented, unauthenticated endpoint.

## Verification

- On-chain discovery finds all previously tracked pools plus older inactive deployments that contribute no volume.
- Ethereum daily volume is **unchanged** (**35.26M** on 2026-07-08) before and after the change.
- Verified discovered `token0`/`token1` values match the previous metadata.

## Summary by CodeRabbit

* **New Features**
  * Pool discovery now uses on-chain initialization events to improve coverage of available pools.
  * Daily volume is derived from on-chain swap activity for more consistent reporting.
* **Bug Fixes**
  * Improved cross-chain metric collection by using chain-specific historical starting points and the newly discovered pool set, reducing gaps in volume calculations.

### DIFF
--- a/dexs/metric/index.ts
+++ b/dexs/metric/index.ts
@@ -1,85 +1,74 @@
-import * as sdk from "@defillama/sdk";
 import { FetchOptions, SimpleAdapter } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
 import { addOneToken } from "../../helpers/prices";
 
-const chainConfig: Record<string, { start: string }> = {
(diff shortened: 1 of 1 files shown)


bheluga:
@DaniiRix thanks for the PR

PR #8102, opened 2026-07-13
Add DropSwap aggregator adapter

SimJach:
Adds an Aggregators adapter for DropSwap (https://dropswap.finance), a multi-chain
DEX aggregator routing swaps across 20 chains via LI.FI and Odos.

Volume data is served per-chain from our own API:
https://dropswap.finance/api/defillama/volume-by-chain

Protocol metadata:
https://dropswap.finance/api/defillama/protocol

Tested locally with `npm test aggregators dropswap` — adapter correctly returns
per-chain daily volume.

PR #8130, opened 2026-07-14
Add Abstract and Solana chains to DropSwap aggregator adapter

SimJach:
Adds Abstract and Solana chains to the DropSwap aggregator adapter (follow-up
to #8102, which added the initial 20-chain adapter).

DropSwap added support for these two chains after the original PR was merged.
No other logic changes — same prefetch/fetch pattern, same API endpoint
(https://dropswap.finance/api/defillama/volume-by-chain).

Tested locally with `npm test aggregators dropswap`.


bheluga:
@SimJach thanks for the PR

PR #8131, opened 2026-07-15
abracadabra: switch fees to on-chain events

uwezukwechibuzor:
### Problem

The previous adapter read fees from Abracadabra's analytics subgraphs, which have
been decommissioned. Every run now fails:

Error: deployment u56065/s43877/latest does not exist

On top of the dead data source, the old adapter had several methodology issues:

-  It split interest and borrow fees 50/50 "to lenders and fee recipients." Abracadabra is a CDP, MIM is *minted*
  against collateral, not supplied by lenders, so there is no supply-side lender to pay. This under-reported revenue by half.
- **Liquidation fees attributed backwards.** It excluded liquidation fees from revenue and booked 100% to liquidators. On-chain, the borrower's liquidation penalty is split 10% to the protocol / 90% to the liquidator.

### Solution

Rewrote the adapter to read fees directly from cauldron events, sourced from the
on-chain `CauldronRegistry` (`0xefCDC6FB4973aC30325Fb2B39e1a2F384E254b7A`, deployed
at the same address on every supported chain).

- **`version: 2` + `pullHourly: true`.** Cauldron list is enumerated from the
  registry and cached via `getConfig`.
- **Fees from events, in MIM:**
  - Interest → `LogAccrue.accruedAmount`
  - Borrow/opening fee → recovered from `LogBorrow` (event amount is grossed-up:
    `amount * fee / (FEE_PRECISION + fee)`)
  - Liquidation penalty → `LogLiquidation`:
    `borrowAmount * (multiplier - PRECISION) / PRECISION`
- **Correct classification:**
  - Interest + borrow fees → 100% protocol revenue (MIM is minted, no lenders).
  - Liquidation penalty → 10% protocol / 90% liquidator (`dailySupplySideRevenue`).
  - Protocol revenue further split 50/50 → treasury (`dailyProtocolRevenue`) and
    SPELL buybacks / mSPELL stakers (`dailyHoldersRevenue`), per Abracadabra
    tokenomics.
- Added `methodology` and `breakdownMethodology` with per-source labels
  (Borrow Interest, Borrow Fees, Liquidation Fees).

### DIFF
--- a/fees/abracadabra.ts
+++ b/fees/abracadabra.ts
@@ -1,105 +1,180 @@
+import { ChainApi } from "@defillama/sdk";
 import { Adapter, FetchOptions } from "../adapters/types";
 import { CHAIN } from "../helpers/chains";
-import request, { gql } from "graphql-request";
+import { getConfig } from "../helpers/cache";
 import { METRIC } from "../helpers/metrics";
-
-type ChainConfig = {
-  start: string;
-  endpoint: string;
-}
-
-type DailySnapshot = {
-  borrowFeesGenerated: string;
-  interestFeesGenerated: string;
-  liquidationFeesGenerated: string;
-}
-
-type GraphResponse = {
-  protocolDailySnapshots: DailySnapshot[];
-}
 
 const MIM = "magic-internet-money";
 const BORROW_FEES = "Borrow Fees";
-const BASE_URL = "https://api.studio.thegraph.com/query/56065";
 
-const chainConfig: Record<string, ChainConfig> = {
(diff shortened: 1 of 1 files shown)


bheluga (dismissed):
Looks like it has data mismatches?
for eg: may 11th 2022 has different data vs https://defillama.com/protocol/abracadabra-spell?revenue=true

uwezukwechibuzor:
> @uwezukwechibuzor thanks for the PR. Looks like it has data mismatches? for eg: may 11th 2022 has different data vs https://defillama.com/protocol/abracadabra-spell?revenue=true

Checked and noted. I'll investigate the mismatch with the existing data and update the PR with a fix. Thanks for pointing it out

PR #8098, opened 2026-07-13
[feat] Flap.sh: track flap volume and fees

Shaileshkhote:
https://flap.sh/

## Summary
- Adds a new `dexs/flap.ts` adapter to track Flap bonding-curve volume, fees, revenue, and protocol revenue.
- Tracks Flap Portal contracts on BSC, Robinhood, X Layer, and Monad using on-chain `TokenBought` and `TokenSold` events.

## Changes
- Added a v2 hourly DEX adapter for Flap under `dexs/flap.ts`.
- Counts native-token buy/sell volume from the `eth` field emitted by Flap Portal trade events.
- Counts bonding-curve fees from the event `fee` field as `dailyFees`, `dailyUserFees`, `dailyRevenue`, and `dailyProtocolRevenue`.
- Uses official Flap deployed contract docs for Portal addresses and chain start dates.
- Excludes token tax flows and non-trading contracts such as VaultPortal, token implementations, helpers, oracle, trigger service, and candy box.

## Methodology
- Volume is buy and sell volume on Flap bonding curves before tokens move to a DEX.
- Fees are the bonding-curve trade fees users pay on Flap buys and sells.
- Revenue and protocol revenue are the same bonding-curve fees kept by Flap.
- Token taxes are excluded.

PR #8133, opened 2026-07-15
[fix] Flap.sh: fix volume and fees for flap

Shaileshkhote:
https://github.com/DefiLlama/dimension-adapters/pull/8098#issuecomment-4976343376

## Summary
- Fixes Flap bonding-curve volume and fee accounting.

## Changes
- Resolves each traded token's quote asset from `TokenQuoteSet` events.
- Counts `TokenBought` and `TokenSold` `eth`/`fee` amounts in the token's quote asset instead of assuming native gas token.
- Keeps native fallback when no quote token is set.

cc @bheluga 
@flapguy please append changes in this PR if the current implementation is wrong.

### DIFF
--- a/dexs/flap.ts
+++ b/dexs/flap.ts
@@ -1,35 +1,42 @@
 import { FetchOptions, FetchResultV2, SimpleAdapter } from "../adapters/types";
+import ADDRESSES from "../helpers/coreAssets.json";
 import { CHAIN } from "../helpers/chains";
 
 const BONDING_CURVE_FEES = "Bonding Curve Fees";
+const NATIVE_TOKEN = ADDRESSES.null;
 
 // Source: https://docs.flap.sh/flap/developers/deployed-contract-addresses
-const chainConfig: Record<string, { start: string; portal: string }> = {
+const chainConfig: Record<string, { start: string; portal: string; fromBlock: number }> = {
   [CHAIN.BSC]: {
     start: "2024-06-27",
     portal: "0xe2cE6ab80874Fa9Fa2aAE65D277Dd6B8e65C9De0",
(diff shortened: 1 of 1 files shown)


bheluga:

flapguy:
Hi @Shaileshkhote
  
  Thanks for the fix! Just wanted to share our full methodology so it's captured correctly on your end.

  All of Flap's protocol revenue eventually lands in a Gnosis Safe, per chain. But not every inbound transfer to that Safe is real revenue — Safe A can send to contract B, and B can
  send back to Safe A, and that's not protocol income, it's just funds moving around. To count all revenue while avoiding that kind of noise, we treat as valid revenue anything sent
  to the Safe by:

  - the Portal contract itself, and
  - each token's own TaxProcessor and TaxSplitter contracts.

  Anything else landing in the Safe is not counted.

  We've built a Dune dashboard with this methodology (properly priced per quote token, not just BNB) here:
  https://dune.com/ffflllaaappp/rrrr/774f92cb-0c70-4088-a20b-65db73fe85eb?utm_source=share&utm_medium=copy&utm_campaign=dashboard

  We noticed other chains are already integrated as well, so here are the Safe addresses for those, in case they're missing on your end:

  - BNB: 0x8a08D98CBB218fceB318Ecf3aBc1BA43D8A7aB0E
  - Robinhood: 0xa4A727E0918cf9B39639Fc4cB7D742d39C5352a4
  - X Layer: 0xAC4f9Ba4E48cAafBa17164FBCb078091651Ae361
  - Monad: 0xA77dc19CF7CB7ab50b661Ce5AB6D37954F8022f4

  cc @bheluga

  Let us know if anything else is needed on our end. Thank you!

PR #8139, opened 2026-07-15
Add Robinhood to Socket chain config

BlockBit-ss:
Adds Robinhood to the Socket bridge aggregator chain config so the existing Dune/router inflow query includes Robinhood volume.\n\nVerification:\n- npm run ts-check\n- git diff --check


bheluga:
@BlockBit-ss thanks for the PR

PR #8141, opened 2026-07-15
sxbet: track parlay fees + add methodology

DaniiRix:
- Added `dailyFees` and `dailyRevenue` using the protocol's `usdRevenue` endpoint.
- Added `dailySupplySideRevenue = 0`, as fees are not shared with market makers.
- Added a methodology describing the fee model and ongoing SX token sunset.


bheluga:
@DaniiRix thanks for the PR

PR #8155, opened 2026-07-15
add fees adapter for AlphaFi Lending

Zorag44:
This PR adds a fees adapter for **AlphaFi Lending** (AlphaLend), which is already
listed on DefiLlama: https://defillama.com/protocol/alphafi-lending (under the
AlphaFi parent protocol). Adapter placed at `fees/lending-by-alphafi/` to match
the protocol's module.

### Implementation
- `version: 2` with `pullHourly: true`
- Data comes from our public metrics API (`api-staging.alphafi.xyz/public/metrics/fees`),
  which accepts arbitrary start/end timestamps (max 25h range) and returns USD values
- Data available from 2026-03-02

### DIFF
--- a/fees/lending-by-alphafi/index.ts
+++ b/fees/lending-by-alphafi/index.ts
@@ -1,0 +1,73 @@
+import { FetchOptions, SimpleAdapter } from "../../adapters/types";
+import { CHAIN } from "../../helpers/chains";
+import fetchURL from "../../utils/fetchURL";
+
+// Returns USD amounts for an arbitrary time range (max 25 hours)
+const FEES_API = "https://api-staging.alphafi.xyz/public/metrics/fees";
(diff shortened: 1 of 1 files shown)


bheluga (commented):
@Zorag44 thanks for the PR.
Is it possible to separate borrow interest and liquidation fees via api response and use separate fee breakdown?

Zorag44:
> @Zorag44 thanks for the PR. Is it possible to separate borrow interest and liquidation fees via api response and use separate fee breakdown?

Hi, our api infra currently does not support this. If possible, we would like to get this PR merged and get some data displayed on the defillama site and we will raise a separate pr for the breakdown.

bheluga:
@Zorag44 thanks for the PR

PR #8159, opened 2026-07-16
echo: replace noTarget with scoped getLogs targets

uwezukwechibuzor:
## Summary
Replaces the chain-wide `noTarget` log scan in the Echo fees adapter with scoped `getLogs({ targets: [...] })`, per [GUIDELINES.md](https://github.com/DefiLlama/dimension-adapters/blob/master/GUIDELINES.md) (noTarget scans every log on the chain and is very heavy).

## Changes

- **Removed `noTarget`** from the `DealFundsWithdrawn` query. Since that event is emitted by individual **deal contracts** (not the factories), the deal contracts are first enumerated via the factories' `DealCreated(uuid, dealAddress)` events, then passed as `targets`.
- **Added 2 missing factories.** The adapter only listed 3 factory addresses, but 5 contracts actually emit `DealCreated` on Base. The missing two (`0x475d...87095`, `0x97a7...6fe67`) account for ~109 deals, without them fees came back short. 

## Verification

Output is unchanged before vs. after the fix — `pnpm test fees echo 2026-07-11` returns the same value:

BASE 👇
Daily fees: 7.00 k
Daily revenue: 7.00 k

### DIFF
--- a/fees/echo/index.ts
+++ b/fees/echo/index.ts
@@ -15,28 +15,44 @@ import { ethers } from "ethers"
 import { FetchOptions, SimpleAdapter } from "../../adapters/types"
 import { CHAIN } from "../../helpers/chains"
 
-// Echo protocol contract addresses
-const ECHO_v1_DEAL_FACTORY = '0x32885c0174FBd53A3BDf418408415c7bEF679810'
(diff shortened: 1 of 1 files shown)


bheluga:

PR #8162, opened 2026-07-16
symbiotic: replace noTarget getLogs with scoped targets and add eventAbi

uwezukwechibuzor:
## Summary
symbiotic: replace noTarget getLogs with scoped targets (reward factories → entities) and decode reward events via eventAbi. Adapter returns the same results as before.


bheluga:
There are around 1700+ targets (entities)
in this case using noTarget flag and then filtering gives better performance

PR #8163, opened 2026-07-16
based-alpha: fix broken syntax, include fee in daily volume

BhariGowda:
Two separate issues in this newly-added adapter:

**1. Syntax error** - `breakdownMethodology`'s object literal (opened
at `const breakdownMethodology = {`) was never closed before
`const adapter` began. Confirmed via `ts-node --transpile-only` (the
same mechanism this repo's own test script uses) - the file could not
actually load. TypeScript's parser pointed directly at the unclosed
brace. Fixed by adding the missing `},\n};`.

**2. dailyVolume undercounting** - `ethAmount` is only the net
curve-side amount, not the gross trade value. Verified on-chain across
two buys and one sell:
- Buy (`0xd7720e97...530cf51`): `tx.value` = 4,000,000,000,000,001,
  `ethAmount` = 3,950,000,000,000,000, fee sum = 50,000,000,000,000.
  `tx.value = ethAmount + fees`, exact to 1 wei.
- Buy (`0xe109dd02...5637184`): `tx.value` = 10,000,000,000,000,000,
  `ethAmount` = 9,875,000,000,000,000, fee sum = 125,000,000,000,000.
  Exact to the wei.
- Sell (`0xa6c76003...44ca4b7c`): trader's real payout (confirmed via
  Blockscout internal-transactions) = 1,950,312,499,999,998, exactly
  equal to `ethAmount`, with zero fee deducted from it - the fee comes
  from the reserve, not the trader's proceeds.

Both directions confirm the same rule: the true gross notional is
`ethAmount + protocolFee + creatorFee`, symmetric for buys and sells.
Fixed by adding the fee to `dailyVolume`.

Verified locally: `ts-node --transpile-only` now loads the file with
no errors (exit code 0). `npm run test fees/based-alpha/index.ts` runs
clean across all 24 hourly slots - Daily volume 204.92, Daily fees
2.56, ratio ~1.25%, matching the stated fee rate.

cc @edmundbased @bheluga since you authored/reviewed the original adapter

### DIFF
--- a/fees/based-alpha/index.ts
+++ b/fees/based-alpha/index.ts
@@ -17,10 +17,14 @@ const fetch = async (options: FetchOptions) => {
 
   const tradeLogs = await options.getLogs({ target: LAUNCHPAD, eventAbi: TRADE_EVENT });
   for (const log of tradeLogs) {
-    // ethAmount is the gross ETH side of the trade (fees included on buys,
-    // pre-fee proceeds on sells) — the launchpad's notional curve volume.
-    dailyVolume.addGasToken(log.ethAmount);
-    dailyFees.addGasToken(log.protocolFee + log.creatorFee, METRIC.TRADING_FEES);
+    // ethAmount is the net curve-side amount only. Verified on-chain via two
+    // buys and one sell: on a buy, tx.value = ethAmount + protocolFee + creatorFee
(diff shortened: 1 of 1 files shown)


bheluga:
@BhariGowda thanks for the PR

PR #8168, opened 2026-07-16
angstrom: derive pool list dynamically from Initialize logs

DaniiRix:
Removes the hardcoded Angstrom pool list by discovering pools on-chain, ensuring new pools are tracked automatically.

## Changes

- Replaced the fixed `poolIds` list with on-chain discovery from Uniswap v4 `Initialize` events.
- Filters pools by the Angstrom hook and derives token metadata directly from the event.
- Removes the now-unneeded `poolKeys` multicall and `PositionManager` dependency.

## Verification

- Confirmed the current on-chain configuration still contains the same two pools, so output is unchanged today.
- Verified the discovered pools match the previous metadata and produce identical pool IDs.
- Replayed ~20 hours of live swaps: **3,712** swaps matched with correct volume and fee attribution.

### DIFF
--- a/dexs/angstrom/index.ts
+++ b/dexs/angstrom/index.ts
@@ -1,44 +1,33 @@
 import * as sdk from "@defillama/sdk";
 import { FetchOptions, SimpleAdapter } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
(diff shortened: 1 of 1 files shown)


bheluga:
@DaniiRix thanks for the PR

PR #8170, opened 2026-07-16
aqua: fix user fees

DaniiRix:
dailyUserFees is now swap fees only balance (was aliased to dailyFees, which double-counted the ~43% bribe portion into UserFees)


bheluga:
@DaniiRix thanks for the PR

PR #8171, opened 2026-07-16
temple: add fees tracking

DaniiRix:
Adds fee tracking and improves validation for the Temple adapter.

## Changes

- Added fee metrics:
  - `dailyFees = dailyRevenue = dailyProtocolRevenue = 3 bps × volume`
  - `dailySupplySideRevenue = 0`
- Added methodology for all reported metrics.
- Added a USDA quote check to prevent future markets from being mispriced.
- Throw on empty or malformed API responses instead of reporting `$0`.
- Updated the return type from `FetchResultVolume` to `FetchResult`.

## Verification

- Verified against a live API response: **$5.47M** daily volume and **$1.64k** daily fees.

### DIFF
--- a/dexs/temple/index.ts
+++ b/dexs/temple/index.ts
@@ -4,8 +4,9 @@ import fetchURL from "../../utils/fetchURL";
 
 const TICKERS_URL = "https://api.templedigitalgroup.com/api/exchange/tickers";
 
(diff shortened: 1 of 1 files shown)


bheluga:
@DaniiRix thanks for the PR

PR #8173, opened 2026-07-16
metronome-synth: replace noTarget getLogs with pool targets for synth…

uwezukwechibuzor:
## Summary
metronome-synth: replace the noTarget SyntheticTokenSwapped scan with getLogs targeting the per-chain Metronome Pool contracts (addresses from docs.metronome.io)


treeoflife2:
@uwezukwechibuzor it could be missing the old and new contracts as docs was updated 1y ago. and event looks unique too. so we can keep this.

uwezukwechibuzor:
> @uwezukwechibuzor it could be missing the old and new contracts as docs was updated 1y ago. and event looks unique too. so we can keep this.

Agreed, keeping noTarget is the safer choice here since the event is unique and the docs-listed pools could be incomplete. Closing.

PR #8177, opened 2026-07-16
easya-kickstart: exact per-swap DAMM v2 fee attribution

platonicsocrates:
## What

For post-migration DAMM v2 pools, `Revenue` currently uses the locked-LP split model: `lp_fee - creator_lp_pct%`, with the remainder booked to Kickstart. Graduated pools are open, so positions opened by **third-party LPs** also earn a pro-rata share of `lp_fee` - under the model that share is booked as protocol Revenue even though it accrues to the third-party positions, and it grows as outside LPing grows.

This PR replaces the model with exact per-swap attribution:

- Pool liquidity at any moment is replayed from `cp_amm_evt_evtliquiditychange`: the migration-locked base (from `evtpermanentlockposition` events in the migration tx; the migration tx's own liquidity-change event is excluded so the base is not double-counted) plus cumulative third-party adds/removes.
- Each swap's `lp_fee` is then split pro rata between Kickstart's permanently-locked migration position, the token creator's, and third-party positions.
- `Revenue` gets Kickstart's exact share; the creator share stays in `SupplySideRevenue`; third-party LP fees move to a new `SupplySideRevenue` label ("DAMM v2 Fees to third-party LPs"). `Fees` is unchanged.

Kickstart's locked position is the migration call's `account_second_position` when two positions were locked, else the single locked `account_first_position` (early configs with `creator_permanent_locked_liquidity_percentage = 0` lock only Kickstart's). Builds on the dynamic config discovery from #8084 - the mapping was verified against current position-NFT holders across all 47 migrated pools.

## Why

The EasyA team pays a share of the Revenue reported on this page, so it needs to be exactly the protocol's take. The model error is small today (+0.2-0.7% of daily Revenue, ~30 SOL since March) but is unbounded as third-party LPing grows.

## Validation

- The replay was validated against the cp-amm program's **own per-position fee counters** over full pool history: Kickstart-position lifetime fees replayed 3,930.8299462749 SOL vs 3,930.829946798 SOL on-chain (both read at the same instant) - agreement to 12 significant figures.

### DIFF
--- a/fees/easya-kickstart/index.ts
+++ b/fees/easya-kickstart/index.ts
@@ -23,6 +23,7 @@ interface IData {
     total_referral_fees: string;
     total_damm_v2_fees: string;
     total_damm_v2_protocol_fees: string;
+    total_damm_v2_partner_fees: string;
     total_damm_v2_creator_fees: string;
     total_damm_v2_referral_fees: string;
 }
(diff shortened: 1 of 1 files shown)


bheluga:

PR #8193, opened 2026-07-16
BasedBid Fee/Revenue/Volumes 

Developer-No-1:


bheluga:
@Developer-No-1 thanks for the PR
Added fees [here](https://github.com/DefiLlama/dimension-adapters/pull/8297)
volume are attributable to pancakeswap and uniswap, so we don't count them here

PR #8194, opened 2026-07-16
Add Ravenhood (RVH) fees/revenue adapter

troudev1133-ship-it:
https://defillama.com/protocol/ravenhood

## Summary
Ravenhood (RVH, Robinhood Chain) has TVL/staking tracked in DefiLlama-Adapters but has never had fees/revenue tracked here.

Tracks swap fees collected from `RavenhoodVault`'s single permanently-locked RVH/WETH Uniswap V3 position (tokenId 17757) via `Collect` events on the position manager, subtracting out any co-occurring `DecreaseLiquidity` principal - the protocol's `claimBurn()` function removes 0.5% of the locked liquidity and collects accumulated fees in the same transaction, so the two need to be disambiguated. This is the same technique already used in `fees/olympus-dao.ts`'s `fetchUniV3POLFees` for its own treasury-owned Uniswap V3 positions.

100% of collected fees are counted as protocol revenue - the locked position is the protocol's own treasury (routed to buyback+burn or the owner depending on burn progress via `RavenhoodRitual`), not third-party LP yield paid out to external depositors.

## Test plan
```
$ npx ts-node --transpile-only cli/testAdapter.ts fees/ravenhood

ROBINHOOD 👇
Daily fees: 234.00
Daily revenue: 234.00
```

### DIFF
--- a/fees/ravenhood.ts
+++ b/fees/ravenhood.ts
@@ -1,5 +1,6 @@
 import { Adapter, FetchOptions } from "../adapters/types";
 import { CHAIN } from "../helpers/chains";
+import { METRIC } from "../helpers/metrics";
 
 // RavenhoodRitual (the protocol's core contract) periodically calls
 // RavenhoodVault.claimFees()/claimBurn(), both of which call the Uniswap V3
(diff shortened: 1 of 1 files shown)


bheluga:
@troudev1133-ship-it thanks for the PR

PR #8196, opened 2026-07-17
update liquidswap and liquidcore adapters to add robinhood routers

lordshisho:
We've gone multichain to Robinhood!

Existing adapters simply being extended with a new chain.


bheluga:
@lordshisho thanks for the PR

PR #8201, opened 2026-07-17
predictstreet: add fee tracking

DaniiRix:
Adds fee tracking to the PredictStreet adapter using the `fee` field already emitted in `OrderFilled`.

### Changes

- Added `dailyFees`, `dailyRevenue`, and `dailyProtocolRevenue`.
- Sum fees directly from `OrderFilled` (no `/2`, as each order emits its fee only once).
- Value outcome-share fees at the trade's execution price.

### Verification

- Verified against live ADI RPC traces that the emitted `fee` matches the protocol's token receipts on both buy and sell orders.
- Confirmed all collected fees accrue to the protocol, with no maker or LP rebates.
- `pnpm test dexs/predictstreet` reports **9.09M** volume, **17.86M** notional, and **472k** fees, consistent with on-chain activity.


bheluga:
@DaniiRix thanks for the PR

PR #8202, opened 2026-07-17
photon: fix overstated volume

DaniiRix:
Fixes Photon's Solana volume by deduplicating multi-hop Jupiter swaps and marking the adapter as `doublecounted`.

## Changes

- Added `doublecounted: true`, since Photon routes through underlying Solana DEXs already tracked elsewhere.
- Updated the Dune query to:
  - Count one swap per trader per transaction.
  - Exclude the fee wallet's own trades.
  - Include only successful transactions where the fee wallet received a fee.
- Added a volume methodology.

## Verification

- The new query only removes duplicate or invalid rows, eliminating overcounting without adding volume.
- `pnpm test dexs/photon 2026-07-14` passes, reporting **~$661k** daily volume with no NaN or negative values.
- The existing fees adapter is unchanged.

### DIFF
--- a/dexs/photon.ts
+++ b/dexs/photon.ts
@@ -2,6 +2,9 @@ import { Dependencies, FetchOptions, SimpleAdapter } from "../adapters/types";
 import { CHAIN } from "../helpers/chains";
 import { queryDuneSql } from "../helpers/dune";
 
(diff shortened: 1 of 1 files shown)


bheluga:
@DaniiRix thanks for the PR

Issue #8100, opened 2026-07-13
SilentSwap: please reindex 2026-07-03 → 2026-07-13 (out-of-gas issue, now fixed)


GarageInc:
**Adapter:** `bridge-aggregators/silentswap` (BSC)

**Request:** please reindex SilentSwap volume for **2026-07-03 → 2026-07-13 (inclusive)**.

## Context

The adapter reads daily inflow data from the `SilentSwapStats` contract on BSC (`0x5894a9B8B342BDb1AD7570C3656eE406eE26f676`).

Due to an out-of-gas issue, the contract did not receive daily updates from **2026-07-03 to 2026-07-12**, so the adapter recorded ~$0/day for that window. The issue has been fixed and the contract now has correct data for the affected dates.

Reindexing the range above will restore the series; expected daily magnitudes are ~$12M–$37M/day, consistent with the days before and after the outage.

We'd also ask that **2026-07-13** be included in case it was indexed against an incomplete 2026-07-12 baseline, which would misstate its value.

Happy to provide anything else needed. Thanks!

BhariGowda:
Independently verified via cast call against the live contract
(0x5894a9B8B342BDb1AD7570C3656eE406eE26f676 on BSC): reconstructed the
daily series for 07-03 through 07-14 by computing
volumeByDate(d).inflowUsd - volumeByDate(d-1).inflowUsd for each day.

Confirms the fix - most days now land in the $12M-$37M range as
described (07-03: ~$36.8M, 07-08: ~$18.1M, 07-11: ~$15.9M, etc.).

One thing this also confirms: your own flagged concern about 07-13 is
justified, not just precautionary - its computed value comes out to
only ~$8.2M, meaningfully below the rest of the recovered series.
Since the adapter computes each day as a delta against the prior
day's cumulative value, this is consistent with 07-12's snapshot
still being slightly affected by the tail of the outage. Recommend
including 07-13 in the reindex, as you suggested.

Also opened a follow-up PR ( #8203 so this
same failure mode (keeper stall -> silent $0) surfaces loudly next
time instead of needing someone to notice the numbers look wrong.

bheluga:
I have refilled, should be good in a couple of hours

PR #8203, opened 2026-07-17
silentswap: throw loudly on unpublished snapshots instead of silently reporting $0

BhariGowda:
Issue #8100 reported ~10 days of adapter data silently showing ~$0/day
because the SilentSwapStats keeper stalled (out-of-gas) and stopped
publishing daily snapshots to the contract. The adapter's existing
delta calculation (today.inflowUsd - yesterday.inflowUsd, clamped to
≥0) had no way to distinguish a genuine zero-volume day from a day
whose snapshot was simply never written - both looked identical after
the clamp, so the outage was invisible until someone noticed the
numbers looked wrong and filed an issue.

The struct has a `publishedAt` field that defaults to 0 when a date's
row has never been written. Fix: throw explicitly when either the
current or prior day's `publishedAt` is 0, rather than silently
computing a delta against an unpublished snapshot and clamping the
result to 0.

Checked this wouldn't misfire in normal operation before relying on
it: `publishedAt` for 2026-07-02 (a genuine pre-outage healthy day)
is `2026-07-02 12:30:23 UTC` - published same-day, well before this
adapter's own fetch would run for it. By contrast, the backfilled
outage days (e.g. 07-03) all carry a `publishedAt` around
`2026-07-13`, ten days after the date they represent - exactly what
you'd expect from a keeper that resumed and wrote its missed history
in one batch once fixed. That gap is the actual signal this fix relies
on, and it only appears during a genuine stall, not routine operation.

Verified independently on-chain (`cast call` against the live
contract) that the historical outage this issue reports is real and
now resolved: reconstructed the daily series for 2026-07-03 through
2026-07-14 from raw `volumeByDate()` values, confirming figures now
fall in the expected $12M-$37M/day range. This PR doesn't fix the
historical gap itself (that needs an admin-side reindex, already
requested in #8100) - it hardens the adapter so the same failure mode
surfaces immediately next time instead of silently reporting $0 again.

Verified locally: `ts-node --transpile-only` loads the file with no
errors (exit code 0). `npm run test bridge-aggregators/silentswap/index.ts`
runs clean for the current day (Daily bridge volume: $21.92M),
confirming the new guard doesn't false-positive on normal operation.

Related to #8100.

### DIFF
--- a/bridge-aggregators/silentswap/index.ts
+++ b/bridge-aggregators/silentswap/index.ts
@@ -12,6 +12,7 @@ import type { FetchOptions, SimpleAdapter } from "../../adapters/types";
  */
 
 const STATS_CONTRACT = "0x5894a9B8B342BDb1AD7570C3656eE406eE26f676"; // BNB mainnet, deployed 2026-05-25
+const DEPLOYMENT_DATE = 20260525n; // the day before this has no snapshot by design, not due to a stall
 
 const ONE_DAY_SECONDS = 24 * 60 * 60;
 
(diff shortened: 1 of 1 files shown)


BhariGowda:
Issue #8100 reported ~10 days of adapter data silently showing ~$0/day
because the SilentSwapStats keeper stalled (out-of-gas) and stopped
publishing daily snapshots to the contract. The adapter's existing
delta calculation (today.inflowUsd - yesterday.inflowUsd, clamped to
≥0) had no way to distinguish a genuine zero-volume day from a day
whose snapshot was simply never written - both looked identical after
the clamp, so the outage was invisible until someone noticed the
numbers looked wrong and filed an issue.

The struct has a `publishedAt` field that defaults to 0 when a date's
row has never been written. Fix: throw explicitly when either the
current or prior day's `publishedAt` is 0, rather than silently
computing a delta against an unpublished snapshot and clamping the
result to 0.

Checked this wouldn't misfire in normal operation before relying on
it: `publishedAt` for 2026-07-02 (a genuine pre-outage healthy day)
is `2026-07-02 12:30:23 UTC` - published same-day, well before this
adapter's own fetch would run for it. By contrast, the backfilled
outage days (e.g. 07-03) all carry a `publishedAt` around
`2026-07-13`, ten days after the date they represent - exactly what
you'd expect from a keeper that resumed and wrote its missed history
in one batch once fixed. That gap is the actual signal this fix relies
on, and it only appears during a genuine stall, not routine operation.

**Updated after CodeRabbit review:** the prior-day check now skips the
adapter's own first day (2026-05-25), whose `prevDate` (2026-05-24)
predates contract deployment and would otherwise throw permanently -
`inflowsTillYesterday` correctly defaults to `0n` in that case, so
`delta = inflowsTillToday`, the full first day's volume. Also fixed a
bigint/number type mix in the final clamp ternary (`0n` instead of `0`).

Verified independently on-chain (`cast call` against the live
contract) that the historical outage this issue reports is real and
now resolved: reconstructed the daily series for 2026-07-03 through
2026-07-14 from raw `volumeByDate()` values, confirming figures now
fall in the expected $12M-$37M/day range. This PR doesn't fix the
historical gap itself (that needs an admin-side reindex, already
requested in #8100) - it hardens the adapter so the same failure mode
surfaces immediately next time instead of silently reporting $0 again.

Verified locally: `ts-node --transpile-only` loads the file with no
errors (exit code 0). `npm run test bridge-aggregators/silentswap/index.ts`
runs clean for the current day (Daily bridge volume: $21.92M) and for
the deployment-day edge case, confirming both the original guard and
the fix work as intended.

Related to #8100.

bheluga:
@BhariGowda thanks for the PR

PR #8208, opened 2026-07-17
[fix] phygitals - scale luckydraw/royalties fees and volume by 1e6

### LINKED ISSUE #8206: Phygitals: dailyVolume and luckydraw/royalties fees understated ~1e6 (missing decimal scaling)
Phygitals reports dailyFees far larger than dailyVolume, which is impossible since fees are a subset of volume.

Live API (30d):
- dailyFees: about $1.23M
- dailyVolume: about $12

The cause is inconsistent decimal scaling in `fees/phygitals/index.ts`. The Allium query returns USDC amounts already divided by 1e6 (whole USDC units). The gacha fee is scaled back to raw units before adding:

```js
dailyFees.add(ADDRESSES.solana.USDC, gachaNetRevenue > 0 ? gachaNetRevenue * 1e6 : 0, 'GACHA_FEES');
```

but the luckydraw fee, the royalties, and the volume are added without the `* 1e6`, so `balances.add` treats whole USDC as raw 6-decimal units and divides by 1e6 again:

```js
dailyFees.add(ADDRESSES.solana.USDC, result.luckydraw_fees || 0, 'LUCKYDRAW_FEES');
dailyFees.add(ADDRESSES.solana.USDC, result.royalties || 0, 'ROYALTIES');
dailyVolume.add(ADDRESSES.solana.USDC, gachaTotalSpend);
```

So the luckydraw fees, the royalties, and the entire dailyVolume are understated by 1e6 (the $12 reported volume should be roughly $12M). The gacha fee itself is scaled correctly.

Fix: multiply these three by 1e6 to match the gacha line. PR to follow.

Himess:
Fixes #8206

Phygitals reports dailyFees far larger than dailyVolume, which is impossible since fees are a subset of volume.

Live API (30d): dailyFees about $1.23M vs dailyVolume about $12.

The Allium query returns USDC amounts already divided by 1e6 (whole USDC units). The gacha fee is scaled back to raw units before adding:

```js
dailyFees.add(ADDRESSES.solana.USDC, gachaNetRevenue > 0 ? gachaNetRevenue * 1e6 : 0, 'GACHA_FEES');
```

but the luckydraw fee, royalties, and volume were added without the `* 1e6`, so `balances.add` treated whole USDC as raw 6-decimal units and divided by 1e6 again. That understated luckydraw fees, royalties, and the entire dailyVolume by 1e6.

This scales those three by 1e6 to match the gacha line. After the fix, dailyVolume reflects real USDC spend (roughly $12M rather than $12) and luckydraw/royalties fees are counted at full value. The gacha fee, which was already correct, is unchanged.

### DIFF
--- a/fees/phygitals/index.ts
+++ b/fees/phygitals/index.ts
@@ -75,9 +75,9 @@ const fetch = async (options: FetchOptions) => {
     const gachaTotalSpend = (result.gacha_spend || 0) + (result.gacha_spend1 || 0);
     const gachaNetRevenue = gachaTotalSpend - (result.buyback || 0);
     dailyFees.add(ADDRESSES.solana.USDC, gachaNetRevenue > 0 ? gachaNetRevenue * 1e6 : 0, 'GACHA_FEES');
-    dailyFees.add(ADDRESSES.solana.USDC, result.luckydraw_fees || 0, 'LUCKYDRAW_FEES');
-    dailyFees.add(ADDRESSES.solana.USDC, result.royalties || 0, 'ROYALTIES');
-    dailyVolume.add(ADDRESSES.solana.USDC, gachaTotalSpend);
+    dailyFees.add(ADDRESSES.solana.USDC, (result.luckydraw_fees || 0) * 1e6, 'LUCKYDRAW_FEES');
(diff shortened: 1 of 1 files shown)


bheluga:
@Himess thanks for the PR

PR #8216, opened 2026-07-17
[fix] autofinance - convert fee bps with /10000 not /100

### LINKED ISSUE #8215: [fix] autofinance - management/performance fee bps divided by 100 instead of 10000
Autofinance overstates its fee metrics because two basis-point fee rates are converted with `/ 100` instead of `/ 10000`.

In `fees/autofinance/index.ts`:

```js
const managmentFeesForPeriod = Number(totalAssets[index]) * feeDetails[index].periodicFeeBps * periodWrtYear / 100;   // L103
...
const performanceFeesForPeriod = isPositivePerformance ? (yieldForPeriod * feeDetails[index].streamingFeeBps / 100) : 0;   // L111
```

`periodicFeeBps` and `streamingFeeBps` are basis points (see the `getFeeSettings` ABI, which names both fields `...Bps`). Converting basis points to a fraction requires dividing by 10000, not 100. As written, a 100 bps (1%) rate becomes `100 / 100 = 1` (100%), so the management and performance fees are 100x too large. These feed `dailyFees`, `dailyRevenue` and `dailyProtocolRevenue`.

Live API: dailyRevenue 30d is about $2.3k (management + performance fees); with the fix it should be about $23.

Fix: divide both by 10000. PR to follow.

Himess:
Fixes #8215

Autofinance overstates its fee metrics because two basis-point fee rates are converted with `/ 100` instead of `/ 10000`.

`periodicFeeBps` and `streamingFeeBps` are basis points (per the `getFeeSettings` ABI). A 100 bps (1%) rate becomes `100 / 100 = 1` (100%), so the management and performance fees were 100x too large. Both feed `dailyFees`, `dailyRevenue` and `dailyProtocolRevenue`.

Live API: dailyRevenue 30d is about $2.3k; with the fix it drops to about $23.

Two lines, `/ 100` to `/ 10000`.

### DIFF
--- a/fees/autofinance/index.ts
+++ b/fees/autofinance/index.ts
@@ -100,7 +100,7 @@ async function fetch(options: FetchOptions): Promise<FetchResult> {
 
         const yieldForPeriod = (Number(exchangeRatesAfter[index]) - Number(exchangeRatesBefore[index])) * Number(totalSupplies[index]) / 1e18;
 
-        const managmentFeesForPeriod = Number(totalAssets[index]) * feeDetails[index].periodicFeeBps * periodWrtYear / 100;
+        const managmentFeesForPeriod = Number(totalAssets[index]) * feeDetails[index].periodicFeeBps * periodWrtYear / 10000;
 
         dailyFees.add(baseAssetId, managmentFeesForPeriod, METRIC.MANAGEMENT_FEES);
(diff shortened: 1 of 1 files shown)


bheluga:
@Himess thanks for the PR

PR #8221, opened 2026-07-17
[fix] veno-finance - scale zkSync Era yield to wei for addGasToken

Himess:
On zkSync Era, Veno's staking-reward fees are understated by ~1e18 (effectively $0) because a human-scaled yield is passed to `addGasToken`, which expects a raw (wei) amount.

`calculateLSTYield` returns a human-scaled number. The Cronos path books it with `addCGToken` (which takes human units):

```js
dailyFees.addCGToken("crypto-com-chain", croYield);
```

but the zkSync Era path used `addGasToken(ethYield)`, so 5 ETH/day of yield was recorded as 5 wei.

Fix: scale to wei with `addGasToken(ethYield * 1e18)`. The Cronos path is unaffected.


bheluga:
@Himess thanks for the PR

PR #8223, opened 2026-07-17
[fix] shmonad - scale yield by share supply, not total assets

Himess:
shMonad overstates its fees by the shMON/MON exchange rate because the daily yield multiplies the rate change by total assets instead of the share supply.

The exchange rate is assets-per-share, so the total yield is the per-share rate gain times the number of shares (`supplyBefore`), not total assets (`assetsBefore`). Since `assetsBefore = supplyBefore * rate`, the code multiplied the yield by an extra factor of the exchange rate (>= 1, and growing as the vault accrues). Sibling ERC-4626 adapters scale by supply.

Fix: multiply by `supplyBefore`. This corrects `dailyFees`, `dailyProtocolRevenue` and `dailySupplySideRevenue`.


bheluga:
@Himess thanks for the PR

PR #8225, opened 2026-07-17
[fix] wormhole - remove hardcoded 1 SUI added to Sui dailyFees

Himess:
The Wormhole Sui path added a hardcoded `1e9` (1 SUI) to `dailyFees` on every fetch, contradicting the comment right above it ("Sui message fees are currently set at 0").

The constant was added unconditionally before the real executor events, so Sui `dailyFees` (and `dailySupplySideRevenue`, which equals `dailyFees`) were inflated by about 1 SUI every window, reporting about 1 SUI on zero-activity days. No other chain path (EVM, Aptos) has such a constant.

Fix: remove the hardcoded line and rely on the real Sui executor events.


bheluga:
@Himess thanks for the PR

PR #8211, opened 2026-07-17
[fix] swell - compute staking rewards from totalSupply, not totalDeposited

Himess:
Fixes #8209

The Swell fee adapters overstate dailyFees, dailyRevenue and dailySupplySideRevenue by the swETH / rswETH exchange rate (about 1.1x today, and growing as the rate climbs).

`df` (the gross daily staking reward) was computed from `totalDeposited` (the ETH value staked = `totalSupply * exchangeRateBefore`) instead of the token supply:

```js
const totalDeposited = BigInt(totalSupply) * BigInt(exchangeRateBefore) / BigInt(1e18)
const df = Number(totalDeposited) * (exchangeRateAfter - exchangeRateBefore) / 0.9 / 1e18
```

The daily reward in ETH is `totalSupply * (exchangeRateAfter - exchangeRateBefore)`, the value increase of the swETH supply. Since `totalDeposited` already carries a factor of `exchangeRateBefore`, using it here multiplies the reward by an extra `exchangeRateBefore / 1e18` (the exchange rate, about 1.1).

This computes `df` from `totalSupply` directly and drops the now-unused `totalDeposited`. Same fix in `fees/swell.ts` and `fees/swell-restaking/index.ts`.

PR #8232, opened 2026-07-17
[fix] puffer-finance - scale yield by share supply not total assets

Himess:
Same class as the already-merged #8223 (shmonad), #8211 (swell), and #8218 (concrete): a per-share exchange-rate delta scaled by total assets instead of share supply.

`fees/puffer-finance/index.ts` derives pufETH yield from the exchange-rate change but multiplies by `assetsBefore` where it should use `supplyBefore`.

Wrong:
```ts
const netRewards = ((rateAfter - rateBefore) * assetsBefore) / 1e18;
```
Correct:
```ts
const netRewards = ((rateAfter - rateBefore) * supplyBefore) / 1e18;
```

`rateBefore` / `rateAfter` are already per-share rates (`assets * 1e18 / supply`), so the multiplier must be the share count. Because `assetsBefore = supplyBefore * rate / 1e18`, the current line over-reports the whole fee stack (`dailyFees`, `dailyProtocolRevenue`, `dailySupplySideRevenue` all scale off `netRewards`) by the pufETH exchange rate. `supplyBefore` is already read at the same block as `assetsBefore`, so this is a pure substitution with no new call.

Error factor: pufETH exchange rate, currently 1.0798 (+7.98%), verified on-chain (totalAssets / totalSupply of the PufferVault at `0xD9A442856C234a39a81a089C06451EBAa4306a72`).

Pinned 30-day window (2026-06-16 to 2026-07-16), adapter before vs after the fix:
- current (`assetsBefore`): 38.85 ETH = $74,462
- fixed (`supplyBefore`): 36.05 ETH = $69,103
- ratio 1.0775 (equals the exchange rate at the window start)

Identical shape to #8223, which changed `* assetsBefore` to `* supplyBefore` in the same expression.

### DIFF
--- a/fees/puffer-finance/index.ts
+++ b/fees/puffer-finance/index.ts
@@ -59,7 +59,7 @@ const fetch = async (options: FetchOptions) => {
   const rateAfter = (assetsAfter * 1e18) / supplyAfter;
 
   // Net rewards
-  const netRewards = ((rateAfter - rateBefore) * assetsBefore) / 1e18;
+  const netRewards = ((rateAfter - rateBefore) * supplyBefore) / 1e18;
 
   // Gross rewards before protocol and guardian fees
   // https://docs.puffer.fi/yield/protocol/rewards


bheluga:
@Himess thanks for the PR

PR #8234, opened 2026-07-17
[fix] mellow-lrt - scale yield by share supply (expose totalSupply on erc4626 helper)

Himess:
`getERC4626VaultsInfo` returned `{asset, decimals, assetDecimals, totalAssets, assetsPerShare}` but not `totalSupply`. A caller computing vault yield (`rateDelta * shareSupply`) then has no correct multiplier in scope. Both existing callers had to work around the gap: varlamore reconstructs shares by hand (`totalAssets / assetsPerShare`, see its own comment), and mellow-lrt reached for `totalAssets` and over-reports the restaking yield by the vault exchange rate.

This exposes `totalSupply` on the struct (purely additive, one extra multiCall) and fixes mellow-lrt with it: `vaultInfoOld.totalAssets * vaultRateIncrease` becomes `vaultInfoOld.totalSupply * vaultRateIncrease`, same struct and same block, so no window-boundary behaviour moves. Same class as the merged #8232 (puffer), #8233 (BoringVault cluster), #8211 (swell), #8223 (shmonad) and #8218 (concrete), each of which swaps a total-assets or total-deposited multiplier to share supply in a yield expression. varlamore is already correct and left untouched; its workaround can be simplified in a follow-up.

### DIFF
--- a/fees/mellow-lrt.ts
+++ b/fees/mellow-lrt.ts
@@ -96,8 +96,8 @@ const fetch = async (options: FetchOptions): Promise<FetchResultV2> => {
     if (vaultInfoOld && vaultInfoNew) {
       const vaultRateIncrease = vaultInfoNew.assetsPerShare - vaultInfoOld.assetsPerShare
       if (vaultRateIncrease > 0) {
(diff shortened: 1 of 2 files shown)


bheluga:
@Himess thanks for the PR

PR #8237, opened 2026-07-17
[fix] graphite-protocol - apply protocol/holders percentages to total…

uwezukwechibuzor:
## Summary

Problem: In fees/graphite-protocol/index.ts, dailyProtocolRevenue and dailyHoldersRevenue were cloned from dailyRevenue (already fees × 0.40), even though their percentages (0.3233, 0.0767) are fractions of total fees, so protocol + holders summed to less than % of fees instead of equaling dailyRevenue.

Solution: Clone both from the dailyFees base so the percentages apply correctly, making dailyProtocolRevenue + dailyHoldersRevenue = dailyRevenue.


bheluga:

PR #8246, opened 2026-07-18
[fix] sol-lst - count user-paid withdrawal fees in dailyFees

tarun-khatri:
Several SPL stake pool adapters report more revenue than fees on days with heavy withdrawals, which cannot happen since revenue is a subset of fees.

`helpers/queries/sol-lst.sql` returns every inflow to the pool's manager fee account as a single `dailyRevenue` figure. That account actually receives the manager's cut two different ways:

- **`action='mint'`** — the epoch (management) fee, minted out of the staking rewards. The `staking_fees` CTE already counts those rewards, so this leg is correctly inside `dailyFees`.
- **`action='transfer'`** — deposit/withdrawal fees, paid by users out of their principal. These are *not* part of staking rewards, so nothing in the query counts them as fees.

Adapters that do not feed the inflow back into `dailyFees` therefore book the withdrawal fees as revenue while never counting them as fees.

## Onchain evidence

Taking `doublezero-staked-sol` on 2026-07-14, splitting its fee account `GhN6PpyP6Ln4ycWcyvqsNcowLfYjpUcA9uWKAcFBjj2D` by `action`:

| date | mint (epoch fee) | transfer (user fees) |
|---|---|---|
| 2026-07-12 | 81.97 dzSOL | 0.003 |
| 2026-07-14 | 22.16 | **1,767.33** |
| 2026-07-16 | 23.75 | 0.003 |

The transfers are withdrawal fees. Decoding the largest one (`4tYCNHCsbeo7YzqzmrKrLgxEK3YdUFKN5Atvy2hshJNuvJw1pHnzrTjrh7s1EckoYJHAkZvbVafZr9LajCLu9GN9`) shows `Program log: Instruction: WithdrawStake` on `SPoo1Ku8WFXoNDMHPsrGSTSG1Y47rzgn41SLUNakuHy`, moving 364.811084189 dzSOL against an approved 729,622.168376599, i.e. exactly 0.05% — which is the `stake_withdrawal_fee` of 5/10000 read from stake pool `3fV1sdGeXaNEZj6EPDTpub82pYxcRXwt2oie6jkSzeWi`.

That the mint leg is the epoch fee is corroborated by it tracking rewards: on 2026-07-14 the mint fell to 22.16 from ~85 (26%) while staking rewards fell to $28.9k from $110.6k (26%).

## Affected adapters

Days where reported revenue exceeds reported fees:

| adapter | violation days | worst | excess |
|---|---|---|---|
| doublezero-staked-sol | 1 | 5.07x | $117,818 |
| edgevana | 1 | 3.35x | $42,977 |
| pico-staked-sol | 2 | 7.12x | $7,574 |
| lantern-staked-sol | 2 | 8.12x | $5,334 |
| starke-staked-sol | 1 | 13.51x | $4,866 |
| stronghold-staked-sol | 2 | 1.77x | $3,630 |

## Fix

Split the transfer leg out as its own `dailyUserFees` row and add it to `dailyFees`. `binance-staked-sol`, `bybit-staked-sol`, `drift-staked-sol` and `blazestake` set `revenueFeedback.addToFees`, so the whole inflow already reaches `dailyFees` for them; they are skipped so nothing is counted twice. The extra row is ignored by any adapter that does not read it, so nothing else changes.

## Results

`doublezero-staked-sol` on the 2026-07-14 day, previously fees $28.9k against revenue $146.8k:

```
Daily fees: 174.58 k
Daily revenue: 147.40 k
```

`edgevana` on the 2026-07-12 day, previously fees $18.3k against revenue $61.3k:

```
Daily fees: 80.25 k
Daily revenue: 62.55 k
```

Regression checks — a normal doublezero day is unchanged (fees 111.15k / revenue 6.67k, matching production 110.6k / 6.6k), and `binance-staked-sol` is unchanged (fees 254.05k / revenue 46.48k / supply side 207.56k, matching production).

### DIFF
--- a/factory/solLst.ts
+++ b/factory/solLst.ts
@@ -58,6 +58,10 @@ interface StakingRevenueConfig {
 interface RevenueFeedbackConfig {
   addToFees: boolean;
   feesMetric?: string;
+  // Label for the user-paid deposit/withdrawal fees added to dailyFees when addToFees is
+  // false. Only set it on adapters that publish a Fees breakdown, so every label used has
+  // a matching breakdownMethodology entry.
+  userFeesMetric?: string;
 }
 
 interface SolLstConfig {
@@ -134,6 +138,8 @@ function getBreakdownMethodology(config: SolLstConfig): Record<string, Record<st
     feesBreakdown[config.fees.metric] = "Staking rewards from staked SOL";
   if (config.revenueFeedback.addToFees && config.revenueFeedback.feesMetric)
     feesBreakdown[config.revenueFeedback.feesMetric] = "Includes withdrawal fees and management fees";
+  if (!config.revenueFeedback.addToFees && config.revenueFeedback.userFeesMetric)
+    feesBreakdown[config.revenueFeedback.userFeesMetric] = "Deposit/withdrawal fees users pay on their principal";
   if (Object.keys(feesBreakdown).length > 0) breakdown.Fees = feesBreakdown;
 
   // Revenue breakdown
@@ -255,14 +261,13 @@ function createSolLstAdapter(config: SolLstConfig): SimpleAdapter {
         // account inflow into dailyFees include them, so they are skipped here.
         if (!config.revenueFeedback.addToFees) {
           const rev = config.revenue;
-          // No metric label here: the revenue labels are Revenue-side and have no matching
-          // entry under Fees, so these roll up into the base dailyFees total instead.
+          const metric = config.revenueFeedback.userFeesMetric;
           if (rev.type === "addCGToken") {
-            dailyFees.addCGToken(rev.cgId, row.amount || 0);
+            dailyFees.addCGToken(rev.cgId, row.amount || 0, metric);
           } else if (rev.type === "add") {
-            dailyFees.add(rev.mint, Number(row.amount) * 1e9 || 0);
+            dailyFees.add(rev.mint, Number(row.amount) * 1e9 || 0, metric);
           } else if (rev.type === "addToken") {
-            dailyFees.addToken(rev.mint, Number(row.amount) * 1e9 || 0);
+            dailyFees.addToken(rev.mint, Number(row.amount) * 1e9 || 0, metric);
           }
         }
       }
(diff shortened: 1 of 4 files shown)


tarun-khatri:
Pushed a follow-up addressing the review.

- Dropped the revenue metric label from the user-fee `dailyFees` add. You're right that those labels only exist under Revenue in `breakdownMethodology`, so tagging the Fees slice with them both mislabels the chart and leaves labels without a matching Fees entry. They now roll into the base `dailyFees` total, matching what the doublezero adapter does. This also resolves the backpack-staked-sol case in the pre-merge check.
- Retitled to the `[type] protocol - description` format.

On the doublezero `breakdownMethodology` note in the same check: that adapter has never exported one, and it does not use metric labels on any of its adds, so there is nothing to map. Adding one would be unrelated to this change, so I left it out rather than widen the diff.

@coderabbitai review

tarun-khatri:
Extended this to the remaining consumers of the query it changes, so the sweep is complete rather than factory-only.

`helpers/queries/sol-lst.sql` is shared by six adapters. The first version of this PR only wired the new `dailyUserFees` row into the factory, which left `fees/hylo-lst.ts` and `fees/crypto-com-lst.ts` with the same gap: they book the fee account inflow as revenue but never add the user-paid leg to fees.

Measuring the split on each fee account over the last 30 days, by `tokens_solana.transfers.action`:

| pool | minted (epoch fee) | transferred (user-paid) | transfers |
|---|---|---|---|
| hyloSOL | 4.371723 | **48.944778** | 53 |
| hyloSOL+ | 48.967231 | 0.209033 | 28 |
| CDCSOL | 37.485496 | 2.339756 | 57 |

hyloSOL is the notable one: the user-paid leg is the majority of that account's inflow, so most of what the adapter books as revenue was never counted as a fee.

Running `hylo-lst` for 2026-07-16 on this branch gives fees 4,760 against the 4,486 currently reported, with revenue unchanged at 568. That is the 274 of withdrawal fees that were previously missing, about 6% of the day's fees.

`crypto-com-lst` gets the same one-line treatment; its user-paid leg is small (2.34 CDCSOL over 30 days) but the shape of the bug is identical, so it seemed wrong to leave it as the one known-unfixed consumer.

`jito-staked-sol` also reads this query but has its revenue block commented out, so there is nothing to split there, and `jupiter-staked-sol` and `bonk-staked-sol` are handled in #8252 and #8253.

tarun-khatri:
Documented the conversion at each of the three new call sites: the query returns whole tokens, so the amount is scaled by 1e9 for the LST's 9 decimals.

I kept it inline rather than extracting a shared helper because each new branch sits directly beneath a pre-existing `dailyRevenue.add(..., Number(row.amount) * 1e9 || 0)` doing the identical conversion. Centralizing it would mean rewriting those existing lines too, which is unrelated to what this PR fixes, so the comment seemed like the better trade. Happy to pull it into a helper if you would prefer it that way.

tarun-khatri:
@coderabbitai review

tarun-khatri:
On the red `test` check: it fails with `Error: DUNE_API_KEYS environment variable is not set`, since fork PRs don't get repo secrets. Because this branch touches `helpers/queries/sol-lst.sql`, the job picks up every adapter sharing that query and tries to run them — on #8246 it ran `crypto-com-lst.ts`, which needs Dune. Nothing in the diff is involved in the failure. Happy to re-run if a maintainer triggers it with secrets available.

bheluga:
Merged your commits [here](https://github.com/DefiLlama/dimension-adapters/pull/8270)

PR #8248, opened 2026-07-18
[fix] morpho-v2 curators - charge management fee on assets, not interest

tarun-khatri:
`getMorphoVaultV2Fee` charges the management fee against the day's interest instead of the vault's assets, which is not what the fee is.

## What the contract does

From [VaultV2.sol `accrueInterestView()`](https://github.com/morpho-org/vault-v2/blob/main/src/VaultV2.sol):

```solidity
uint256 managementFeeAssets = elapsed > 0 && managementFee > 0 && canReceiveShares(managementFeeRecipient)
    ? (newTotalAssets * elapsed).mulDivDown(managementFee, WAD)
    : 0;
```

with the accompanying notes on the same function:

> `/// @dev The management fee is not bound to the interest, so it can make the share price go down.`
> `/// @dev The management fees is taken even if the vault incurs some losses.`

So the base is `newTotalAssets`, i.e. assets under management, and `managementFee` is a per-second rate — which is also what this repo's own ABI comment says (`managementFee: 'uint256:managementFee', // rate per second`).

The performance fee, by contrast, is `interest.mulDivDown(performanceFee, WAD)`, which is exactly what the helper already does. That path is correct and unchanged here.

## The effect

```diff
-const interestManagementFee = interestEarnedIncludingFees * vaultManagementFeeRate * BigInt(timeElapsed) / BigInt(1e18)
+const interestManagementFee = vaultInfo[i].balance * vaultManagementFeeRate * BigInt(timeElapsed) / BigInt(1e18)
```

Using interest as the base understates the fee by roughly `365 / APY` — about 7,300x on a vault yielding 5% — and drops it to zero on any day the vault earns nothing, even though the contract charges it through losses.

## Scope, honestly

I enumerated the Morpho V2 vaults belonging to every owner in `factory/curators.ts` (112 vaults across Ethereum and Base via `CreateVaultV2`) and read `managementFee()` on each. Only two currently return a non-zero rate, and both are test vaults with negligible assets:

| vault | chain | managementFee | totalAssets | name |
|---|---|---|---|---|
| `0xB15A51F46A53CF7DBB378A459A552F342BC54815` | base | 1585489599 (5%/yr) | 50977 | cbbtc-testing |
| `0x3DA8BE659674062253B0F1EF9E765C7A6AF8C8D0` | base | 317097919 (1%/yr) | 1027134 | test |

So this is not moving reported numbers today. It is a latent error that silently under-reports as soon as a curator turns the fee on, which is why I would rather fix the formula now than after it starts mattering.

## Label collision

`MorphoPerformanceFee` and `MorphoManagementFee` were both defined as the string `'Morpho Performance Fees'`. Two consequences:

- in `breakdownMethodology`, the two entries are the same key, so the management description silently overwrites the performance one
- the performance fee was also being tagged with `METRICS.MorphoManagementFee`

Given the management fee its own label and tagged the performance fee with the performance label. Also corrected "Moroho" to "Morpho" and the Euler performance-fee entry, which described itself as a management fee.

Verified with `npx tsx cli/testAdapter.ts fees muscadine 2026-07-17` (the curator owning the two vaults above), which runs clean.

### DIFF
--- a/helpers/curators/index.ts
+++ b/helpers/curators/index.ts
@@ -14,7 +14,7 @@ const METRICS = {
   MorphoYields: 'Morpho Yields',
   MorphoYieldsToSuppliers: 'Morpho Yields Distributed To Supliers',
   MorphoPerformanceFee: 'Morpho Performance Fees',
-  MorphoManagementFee: 'Morpho Performance Fees',
+  MorphoManagementFee: 'Morpho Management Fees',
   EulerYields: 'Euler Yields',
   EulerYieldsToSuppliers: 'Euler Yields Distributed To Supliers',
   EulerPerformanceFee: 'Euler Performance Fees',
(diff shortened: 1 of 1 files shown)


tarun-khatri:
Pushed a follow-up addressing the review.

You're right, and it's the same class of problem this PR set out to fix. Making the management fee independent of the interest meant a period with little or no interest could report revenue above fees. The fee is charged on principal rather than taken out of the interest, so the interest figure never contained it and it belongs in `dailyFees` on its own.

Now:

```
fees       = interest + managementFee
revenue    = performanceFee + managementFee
supplySide = interest - performanceFee
```

so `revenue + supplySide = fees` still holds, revenue can no longer exceed fees, and the supply side is only reduced by the fee that actually comes out of the interest. Added a matching `Fees` breakdown entry for the management fee label.

On the `pullHourly` pre-merge warning: `fees/k3.ts` and `fees/alpha-growth.ts` are pre-existing exports this PR does not touch, and `getCuratorExport` reads vault share price between two block snapshots, so it is a daily-delta model. Adding hourly pulls there would be a behavioural change well outside this fix.

@coderabbitai review

bheluga:

PR #8245, opened 2026-07-18
[fix] binance-staked-sol - stop double-counting the 10% commission in revenue

tarun-khatri:
`binance-staked-sol` books Binance's 10% staking commission twice, so `dailyRevenue` is roughly double what the pool actually collects.

The adapter builds revenue from two sources that are the same money:

- `stakingRevenue: { ratio: 0.1 }` — a synthetic 10% of staking rewards
- the fee-account inflow (`3ZC6mkJr9hnFSrVHzXXcPopw3SArgKGm8agcah1vhy2Z`)

The fee-account inflow *is* that commission. Binance's stake pool takes its cut as an epoch fee, minted as bnSOL to the manager fee account — it is not a separate deposit/withdraw fee stream.

## Onchain evidence

Stake pool `3fV1sdGeXaNEZj6EPDTpub82pYxcRXwt2oie6jkSzeWi`, read from account state:

- `epoch_fee` = 10/100 = **10.00%**
- `stake_withdrawal_fee` = 1/1000 = 0.10%

bnSOL minted into the manager fee account vs SOL staking rewards on the pool's stake accounts:

| Date | SOL staking rewards | bnSOL minted to fee account | minted / rewards |
|---|---|---|---|
| 2026-07-02 | 3,046.38 | 278.27 | 9.13% |
| 2026-07-04 | 3,043.52 | 278.43 | 9.15% |
| 2026-07-06 | 3,035.94 | 277.19 | 9.13% |
| 2026-07-08 | 3,035.12 | 275.64 | 9.08% |
| 2026-07-10 | 3,038.83 | 276.84 | 9.11% |
| 2026-07-12 | 3,044.34 | 276.04 | 9.07% |
| 2026-07-14 | 3,042.76 | 273.97 | 9.00% |
| 2026-07-16 | 3,063.46 | 277.59 | 9.06% |

9.0–9.15% of rewards in bnSOL terms is 10% in value, since bnSOL trades at ~1.10 SOL. That is the epoch fee, arriving as a mint.

Splitting the same account's inflow over the pool's full lifetime (since 2024-09): **115,544.6 bnSOL minted (86.4%)** vs **18,154.3 bnSOL transferred in by users (13.6%)**. The account is dominated by the commission, so adding a separate 10% of rewards on top counts it a second time.

## Effect

[...]

PR #8251, opened 2026-07-18
[fix] sol-lst - stop counting the epoch fee mint as extra fees

tarun-khatri:
> Builds on #8246 — it needs the `dailyUserFees` row that PR adds, so it carries those two commits until #8246 merges. The change here is the third commit.

`bybit-staked-sol`, `drift-staked-sol` and `blazestake` set `revenueFeedback.addToFees`, which feeds the whole manager fee account inflow back into `dailyFees`. Most of that inflow is the epoch fee, which the pool mints out of the staking rewards `staking_fees` already counts, so those rewards get counted twice.

## Onchain evidence

Epoch fees read from the stake pool accounts (bybit and drift are deployed under Sanctum's stake pool program `SP12tWFxD9oJsVWNavTTBZvMbA6gkAmxtVgxdqvyvhY`, not the standard one):

| adapter | stake pool | epoch fee | withdrawal fee | TVL |
|---|---|---|---|---|
| bybit-staked-sol | `2aMLkB5p5gVvCwKkdSo5eZAL1WwhZbxezQr1wxiynRhq` | 5.0000% | 0.10% | 1,217,154 SOL |
| drift-staked-sol | `9mhGNSPArRMHpLDMSmxAvuoizBqtBGqYdT8WGuqgxNdn` | 2.5000% | 0.10% | 2,803,311 SOL |
| blazestake | — | 5.0000% | 0.10% | 893,878 SOL |

Those rates are confirmed independently by what actually lands in each fee account. Comparing minted LST against SOL staking rewards on the pool's stake accounts:

| date | bybit minted / rewards | drift minted / rewards |
|---|---|---|
| 2026-07-04 | 4.32% | 2.15% |
| 2026-07-06 | 4.32% | 2.13% |
| 2026-07-08 | 4.65% | 2.12% |
| 2026-07-10 | 4.51% | 2.10% |
| 2026-07-12 | 4.52% | 2.09% |
| 2026-07-14 | 4.55% | 2.09% |
| 2026-07-16 | 4.31% | 2.07% |

Those are the configured rates divided by each pool's SOL/LST exchange rate — 5% / 1.1609 = 4.31% and 2.5% / 1.2050 = 2.07% — so the inflow is the epoch fee arriving as a mint, not a separate user-paid fee.

## Fix

Turn off `addToFees` for the three and let the `dailyUserFees` row from #8246 carry the genuinely user-paid deposit/withdrawal transfers into `dailyFees`. Set the supply side to `1 - epochFee` (0.95, 0.975, 0.95) so the statement still closes:

```
fees       = staking rewards + user-paid deposit/withdrawal fees
revenue    = epoch fee mint + user-paid deposit/withdrawal fees
supplySide = staking rewards x (1 - epoch fee)
```

Also dropped the `Deposit/Withdraw Fees` entry from each `Fees` breakdown, since no `.add()` uses that label on the fees side any more.

## Results

`npx tsx cli/testAdapter.ts fees <adapter> 2026-07-17` (the 2026-07-16 day), against what production reports for the same day:

| adapter | fees before | fees after | revenue before | revenue after |
|---|---|---|---|---|
| drift-staked-sol | 65,599 | 63,900 | 1,598 | 1,600 |
| bybit-staked-sol | 27,192 | 25,970 | 1,323 | 1,330 |

Fees fall by the epoch fee mint and revenue is unchanged, which is the intent. `revenue + supplySide` equals `fees` exactly in both runs (drift 1.60k + 62.30k = 63.90k, bybit 1.33k + 24.64k = 25.97k), and revenue as a share of fees now lands on each pool's configured epoch fee (drift 2.5%, bybit 5.1%).

Over the adapters' lifetimes this removes roughly $2.9M of double-counted fees ($1.24M bybit, $0.59M drift, $1.08M blazestake).

`binance-staked-sol` has the same problem at a 10% epoch fee, but it also has the separate double-count fixed in #8245, so I left it out of this PR to avoid touching the same lines twice. Happy to follow up on it once #8245 lands.

### DIFF
--- a/factory/solLst.ts
+++ b/factory/solLst.ts
@@ -413,21 +413,23 @@ const configs: Record<string, SolLstConfig> = {
     start: "2022-12-07",
     fees: { metric: METRIC.STAKING_REWARDS },
     revenue: { type: "addCGToken", cgId: "blazestake-staked-sol", metric: METRIC.DEPOSIT_WITHDRAW_FEES },
-    revenueFeedback: { addToFees: true, feesMetric: METRIC.DEPOSIT_WITHDRAW_FEES },
-    supplySide: { enabled: true, ratio: 1.0, metric: METRIC.STAKING_REWARDS },
+    // The fee account's inflow is mostly the 5% epoch fee, which is minted out of the
+    // staking rewards already counted above, so it must not be added to fees again. The
+    // user-paid deposit/withdrawal transfers are added separately by the dailyUserFees row.
+    revenueFeedback: { addToFees: false },
+    supplySide: { enabled: true, ratio: 0.95, metric: METRIC.STAKING_REWARDS },
     stakingRevenue: { enabled: false, ratio: 0 },
     returnDailyHoldersRevenue: 0,
     methodology: {
-      Fees: "Staking rewards from staked SOL on blazestake",
+      Fees: "Staking rewards from staked SOL on blazestake, plus deposit and withdrawal fees",
       Revenue: "Includes 0.1% instant withdrawal fee and 0.1% delayed withdrawal fee",
-      SupplySideRevenue: "All the staking rewards are distributed to bSOL",
+      SupplySideRevenue: "95% of the staking rewards are distributed to bSOL",
       ProtocolRevenue: "All fees going to treasury/DAO (50% of total fees) + All fees going to the team(50% of total fees)",
       HoldersRevenue: "No revenue share to BLZE token holders",
     },
     breakdownMethodology: {
       Fees: {
         [METRIC.STAKING_REWARDS]: "Staking rewards from staked SOL on Blazestake",
-        [METRIC.DEPOSIT_WITHDRAW_FEES]: "Includes 0.1% instant withdrawal fee and 0.1% delayed withdrawal fee",
       },
       Revenue: {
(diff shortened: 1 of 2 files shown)


tarun-khatri:
Good catch, fixed. The user-paid leg now carries its own Fees-side label instead of being unlabeled.

`revenueFeedback` gained an optional `userFeesMetric`, which is passed on the `dailyFees` add in the user-fee branch and emitted into the generated Fees breakdown. blazestake / bybit / drift keep their `Deposit/Withdraw Fees` entry under `Fees` (this PR had removed it), so every label used in an `.add()` has a matching `breakdownMethodology` entry and the breakdown sums to the total.

On the doublezero adapter at the second site: that config has no `breakdownMethodology`, so nothing is published for it and there is no breakdown to fall short of the total. Adding a label there would be the reverse problem, a label with no documented entry, so I left it unlabeled. Same reasoning for the other factory adapters that take this path without metrics.

tarun-khatri:
@coderabbitai review

tarun-khatri:
@coderabbitai review

bheluga:
Added [here](https://github.com/DefiLlama/dimension-adapters/pull/8266)

PR #8252, opened 2026-07-18
[fix] jupiter-staked-sol - stop counting the epoch fee mint as extra fees

tarun-khatri:
> Builds on #8246 — it needs the `dailyUserFees` row that PR adds, so it carries those two commits until #8246 merges. The change here is the third commit.

`fees/jupiter-staked-sol` adds the whole manager fee account inflow to `dailyFees` and credits 100% of the staking rewards to the supply side. Most of that inflow is the pool's epoch fee, which is minted out of those same staking rewards, so the rewards are counted twice on the fee side and stakers are credited with a cut the pool never passed on.

## Onchain evidence

Stake pool `8VpRhuxa7sUUepdY3kQiTmX9rS5vx4WgaXiAnXq4KCtr` (deployed under Sanctum's multi stake pool program `SPMBzsVUuoHA4Jm6KunbsotaahvVikZs1JyTW6iJvbn`), read from account state:

- `epochFee` = **5.0000%**
- `stake_withdrawal_fee` = 0.10%
- TVL 5,246,008 SOL / supply 4,376,608 jupSOL → exchange rate 1.1986

Splitting the fee account `GbvFCpMqKX65gQ8KNeob9JUAL7vHCHFSg8YN5bnpPT8g` by transfer action, against SOL staking rewards on the pool's stake accounts:

| date | SOL rewards | minted jupSOL | user-paid jupSOL | minted / rewards |
|---|---|---|---|---|
| 2026-07-04 | 1,519.77 | 73.08 | 34.03 | 4.81% |
| 2026-07-06 | 1,506.47 | 71.10 | 16.90 | 4.72% |
| 2026-07-08 | 1,503.28 | 73.02 | 12.57 | 4.86% |
| 2026-07-10 | 1,498.68 | 70.53 | 9.40 | 4.71% |
| 2026-07-12 | 1,499.35 | 63.05 | 7.50 | 4.21% |
| 2026-07-14 | 1,494.21 | 62.91 | 2.60 | 4.21% |
| 2026-07-16 | 1,507.29 | 62.88 | 2.54 | **4.17%** |

5% / 1.1986 = 4.172%, against 4.1715% measured on 2026-07-16. The earlier days sit slightly higher because the exchange rate was lower then. So the bulk of the inflow is the epoch fee arriving as a mint, not a user-paid fee — the user-paid leg is the small right-hand column.

That the supply side is currently the full reward stream is visible in production too: over 30d `dailyFees` 1,831,711 − `dailyRevenue` 110,694 = 1,721,017, which is exactly the reported `dailySupplySideRevenue`.

## Fix

```
fees       = staking rewards + user-paid deposit/withdrawal fees
revenue    = epoch fee mint + user-paid deposit/withdrawal fees   (unchanged)
supplySide = staking rewards x 0.95
```

`revenue + supplySide` still equals `fees`. Also dropped the `Deposit/Withdraw Fees` entry from the `Fees` breakdown, since no `.add()` uses that label on the fees side any more, and corrected the methodology line that claimed all staking rewards go to stakers.

At a 5% epoch fee this removes about 4.76% of reported fees: roughly **$87k over 30d and $4.2M over the adapter's lifetime** (reported all-time fees are $88.6M).

## Verification

`tsc --noEmit` is clean. I could not exercise this one through `cli/testAdapter.ts` — the CLI does not resolve `jupiter-staked-sol` from `fees/` on my checkout (it fails the same way for other plain-directory adapters such as `sanctum-validator-lsts`, while factory-backed ones run fine), so I did not want to claim a local run I did not get. The numbers above are all read from chain and from the production API instead. Happy to add a run if there is an invocation I am missing.

### DIFF
--- a/factory/solLst.ts
+++ b/factory/solLst.ts
@@ -58,6 +58,10 @@ interface StakingRevenueConfig {
 interface RevenueFeedbackConfig {
   addToFees: boolean;
   feesMetric?: string;
+  // Label for the user-paid deposit/withdrawal fees added to dailyFees when addToFees is
+  // false. Only set it on adapters that publish a Fees breakdown, so every label used has
+  // a matching breakdownMethodology entry.
+  userFeesMetric?: string;
 }
 
 interface SolLstConfig {
@@ -134,6 +138,8 @@ function getBreakdownMethodology(config: SolLstConfig): Record<string, Record<st
     feesBreakdown[config.fees.metric] = "Staking rewards from staked SOL";
   if (config.revenueFeedback.addToFees && config.revenueFeedback.feesMetric)
     feesBreakdown[config.revenueFeedback.feesMetric] = "Includes withdrawal fees and management fees";
+  if (!config.revenueFeedback.addToFees && config.revenueFeedback.userFeesMetric)
+    feesBreakdown[config.revenueFeedback.userFeesMetric] = "Deposit/withdrawal fees users pay on their principal";
   if (Object.keys(feesBreakdown).length > 0) breakdown.Fees = feesBreakdown;
 
   // Revenue breakdown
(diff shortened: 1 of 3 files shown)


tarun-khatri:
Fixed. `dailyUserFees` now goes into `dailyFees` under `JupSOLDepositWithdrawFees`, and that label is restored in `breakdownMethodology.Fees`, so the Fees breakdown sums to the total and every label used is documented.

The factory side is fixed in the same way through an optional `userFeesMetric` on `revenueFeedback`, which is passed on the `dailyFees` add and emitted into the generated Fees breakdown.

On the doublezero site: that config publishes no `breakdownMethodology`, so there is no breakdown to fall short of the total, and labeling it would leave a label with no documented entry. Left unlabeled deliberately.

tarun-khatri:
@coderabbitai review

tarun-khatri:
On the red `test` check: it fails with `Error: DUNE_API_KEYS environment variable is not set`, since fork PRs don't get repo secrets. Because this branch touches `helpers/queries/sol-lst.sql`, the job picks up every adapter sharing that query and tries to run them — on #8246 it ran `crypto-com-lst.ts`, which needs Dune. Nothing in the diff is involved in the failure. Happy to re-run if a maintainer triggers it with secrets available.

tarun-khatri:
Correction to the PR body: I wrote that `cli/testAdapter.ts` can't resolve this adapter. That's wrong — `npm run test fees jupiter-staked-sol <date>` does resolve it; my earlier attempt used a bad argument form. I'm not quoting a local run as evidence here though: the run I did for 2026-07-16 came back with zero staking rewards, which contradicts production, so it looks like a Dune caching/tier artifact rather than a clean result. The numbers in the description come from the stake pool account and the production API, both of which are re-runnable independently.

bheluga:
Included [here](https://github.com/DefiLlama/dimension-adapters/pull/8266/changes/e43839b31f513b0acef0900216a420465cd20a2e)

PR #8254, opened 2026-07-18
[fix] binance-staked-sol - stop counting the epoch fee mint as extra fees

tarun-khatri:
Follow-up to #8245. That PR fixed the revenue side; this one fixes the fees side, which has the same root cause.

The pool's 10% commission is minted as new bnSOL to the fee account, out of the staking rewards the adapter already counts in full. `revenueFeedback.addToFees` adds that mint back into `dailyFees`, so those rewards are counted twice.

### The epoch fee, on-chain

Pool `Hr9pzexrBge3vgmBNRR8u42CNQgBXdHm4UkUN2DH4a7r` on `SPoo1Ku8WFXoNDMHPsrGSTSG1Y47rzgn41SLUNakuHy`, located by matching the adapter's own `stakePoolReserveAccount` at offset 130, with its `manager_fee_account` at offset 194 matching `lstFeeTokenAccount`:

```
epoch_fee            = 10000/100000  = 10.0000%
stake_deposit_fee    = 0/0           =  0.0000%
sol_deposit_fee      = 0/0           =  0.0000%
stake_withdrawal_fee = 1/1000        =  0.1000%
sol_withdrawal_fee   = 1/1000        =  0.1000%
total_lamports       = 10,170,812.315785 SOL
last_update_epoch    = 1003   (current epoch, pool is live)
exchange rate        = 1.122921
```

Both deposit fees are zero, so nothing else is minted to that account. The only other fee is the 0.1% withdrawal fee, and that arrives as a `transfer` (users send bnSOL from their own balance), not a mint. So the minted leg is the 10% commission and nothing else.

### The mint tracks staking rewards

Splitting the fee account's inflow by `tokens_solana.transfers.action`:

| day | SOL staking rewards | minted bnSOL | transferred bnSOL |
|---|---|---|---|
| 2026-07-16 | 3063.458489 | 277.588260 | 0.022543 |
| 2026-07-14 | 3042.755621 | 273.968367 | 0.005788 |
| 2026-07-12 | 3044.340774 | 276.038732 | 0.003271 |
| 2026-07-10 | 3038.825738 | 276.842252 | 5.043162 |
| 2026-07-08 | 3035.124010 | 275.638537 | 0.769951 |

Valuing the mint at the pool's exchange rate puts it at 10.1-10.2% of the rewards for that day, against the 10.0000% epoch fee. The mint is the commission, taken out of those same rewards.

### Production decomposes exactly

DefiLlama for 2026-07-16: fees 254,487, revenue 46,591, supply side 207,896. Supply side is 90% of rewards under the current code, so rewards = 230,996, which leaves the fee account inflow at 254,487 - 230,996 = 23,491. Revenue then comes out at 0.1 x 230,996 + 23,491 = **46,591**, matching the reported figure exactly. The 23,100 and the 23,491 are the same commission counted twice, which is what #8245 addresses; the same 23,491 is also sitting inside fees, which is what this PR removes.

Running this branch on the same day:

```
Daily fees: 230.63 k
Daily revenue: 23.42 k
Daily supply side revenue: 207.56 k

label           | Daily fees | Daily supply side revenue | Daily revenue
Staking Rewards | 230625     | 207563                    |
Management Fees |            |                           | 23421
```

Fees go from 254,487 to 230,625. Over the last 30 days fees read $3,913,815 and lifetime $160,984,326, so removing roughly a tenth of that is about $390k over 30 days and on the order of $14M lifetime. SOL $74.85, CoinGecko at time of writing.

### Change

`addToFees: false` so the mint is not added back into fees, and supply side moves from 100% to 90% of the staking rewards, which is what bnSOL holders actually accrue once the pool keeps its 10%.

One thing worth flagging: measured against the rewards the Dune query returns, the mint comes out at ~10.1-10.2% rather than exactly 10%, so revenue plus supply side lands about 0.2% above fees. That gap is in the rewards figure, not the split. The query builds its stake account set from `delegatestake` calls, and it appears to miss a small number of the pool's stake accounts, so it reports slightly less than the pool actually earns. It predates this change. I used the on-chain 10% rather than back-solving a ratio from the measured number, but if you would rather have the identity close exactly I can derive supply side as fees minus revenue instead.

Depends on #8245; its commit is the parent here and drops out once that lands.


bheluga:
Also addressed [here](https://github.com/DefiLlama/dimension-adapters/pull/8266)

PR #8259, opened 2026-07-19
fix: quiver-protocol should exclude caller incentive fee from revenue

BhariGowda:
FeeConfig.getFees(vault) returns (totalBps, callerBps, treasury). callerBps is a sub-portion of totalBps, not a separate fee, confirmed by the contract's own _check function which requires callerBps <= totalBps.

Checked the real harvest logic in QuiverStrategyUniV3._harvest (source-verified on Robinhood Chain's block explorer):
treasuryBps = totalBps - callerBps
callerBps portion is sent to callFeeRecipient (whoever calls harvest)
treasuryBps portion is sent to treasury

The adapter was counting the full totalBps as protocol revenue, which includes the caller's cut. That overcounts protocol revenue by exactly the caller incentive whenever it's nonzero.

Current live impact: none yet. Checked FeeConfig.defaultCallerBps and all 4 live vaults on-chain, callerBps is 0 everywhere right now, so treasuryBps currently equals totalBps in practice. But the whole point of callerBps existing is to pay keepers for calling harvest, so this is a live bug waiting to trigger the moment that gets turned on, not a hypothetical.

Fix: compute treasuryBps = totalBps - callerBps and use that for revenue and protocol revenue instead of totalBps. SupplySideRevenue is unaffected since it was already correctly based on the full totalBps deduction.

Tested with npm run test fees/quiver-protocol, output unchanged in totals right now since callerBps is currently 0 everywhere, confirming this is a forward-looking correctness fix, not a change to today's numbers.

### DIFF
--- a/fees/quiver-protocol.ts
+++ b/fees/quiver-protocol.ts
@@ -5,8 +5,11 @@ import { METRIC } from "../helpers/metrics";
 // Quiver Protocol: AI-managed concentrated-liquidity vaults on Robinhood Chain
 // (Uniswap V3 + V4, incl. tokenized-stock pairs). Each vault's strategy emits
 // Harvest(caller, fees0, fees1, ppsAfter) with the GROSS LP fees collected;
-// a 10% performance fee (FeeConfig.getFees) goes to the treasury and the rest
-// compounds back into the LP position.
+// a performance fee (FeeConfig.getFees) is split at harvest time: callerBps
+// of it goes to whoever calls harvest (a keeper incentive), the remainder
+// (totalBps - callerBps) goes to the protocol treasury, the rest compounds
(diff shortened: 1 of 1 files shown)


BhariGowda:
CI is failing on an RPC 429 from rpc.mainnet.chain.robinhood.com, not on anything in this diff. Same host, same failure point (slot 7-8 of 24) across 3 separate runs, so this looks like Robinhood Chain's default RPC in the SDK config being rate-limited under the full 24-slot backfill this test does, not a flake specific to this PR. fees/quiver.ts (PR #8172, same chain) may hit the same wall if it ever needs a full backfill. Happy to adjust anything on my end if there's a preferred RPC or backoff pattern for this chain, just flagging it's not the code change causing the failure.

bheluga:
@BhariGowda thanks for the PR

PR #8266, opened 2026-07-19
fix sol lsts

bheluga:


tarun-khatri:
Thanks for picking this up. One thing worth checking before this lands: excluding mints globally is right for the adapters that book the manager's cut separately, but for the `simpleConfig` ones the mint *is* their only management-fee revenue.

Splitting each fee account's 30-day inflow by `action`:

| adapter | minted epoch fee | user transfers | mint share |
|---|---|---|---|
| jpool | 322.049038904 | 113.144398073 | **74.0%** |
| doublezero | 1084.536325631 | 1768.479145912 | 38.0% |
| bonk | 26.94452679 | 0.467644482 | 98.3% |

binance and bonk are fine: binance books the 10% via `stakingRevenue` and bonk books 2.5% + 2.5% out of the rewards, so the mint row was a duplicate for both and dropping it is exactly right.

jpool is the case I would double check. It is `simpleConfig`, so `supplySide` and `stakingRevenue` are both `enabled: false`, and the fee-account row is the only thing feeding `dailyRevenue`. After this change its revenue becomes transfers only — 113.14 of 435.19 JSOL over 30 days, a 74% drop — and the 7% epoch fee stops being counted anywhere, while its methodology still reads "Includes withdrawal fees and management fees collected by fee collector". doublezero (6% epoch fee) has the same shape, smaller share.

The mint is genuinely revenue for those pools; the problem was only that it was *also* inside `dailyFees` because it is minted out of the staking rewards the query already sums. So the two cases differ: for binance/bonk the mint duplicates an existing split, for jpool/doublezero/edgevana/gate it is the split.

If it is useful, #8246 splits the transfer leg into its own `dailyUserFees` row and leaves `dailyRevenue` alone, which keeps the mint as revenue while adding the user-paid fees to `dailyFees` (they are paid out of principal, so they are not in the staking rewards and are currently missing from fees entirely — that is the `revenue > fees` case on those six adapters). The two changes are compatible either way; happy to rebase mine on top of this once it lands, or to fold it in here if you would rather keep it to one query change.

bheluga:
> Thanks for picking this up. One thing worth checking before this lands: excluding mints globally is right for the adapters that book the manager's cut separately, but for the `simpleConfig` ones the mint _is_ their only management-fee revenue.
> 
> Splitting each fee account's 30-day inflow by `action`:
> 
> adapter	minted epoch fee	user transfers	mint share
> jpool	322.049038904	113.144398073	**74.0%**
> doublezero	1084.536325631	1768.479145912	38.0%
> bonk	26.94452679	0.467644482	98.3%
> binance and bonk are fine: binance books the 10% via `stakingRevenue` and bonk books 2.5% + 2.5% out of the rewards, so the mint row was a duplicate for both and dropping it is exactly right.
> 
> jpool is the case I would double check. It is `simpleConfig`, so `supplySide` and `stakingRevenue` are both `enabled: false`, and the fee-account row is the only thing feeding `dailyRevenue`. After this change its revenue becomes transfers only — 113.14 of 435.19 JSOL over 30 days, a 74% drop — and the 7% epoch fee stops being counted anywhere, while its methodology still reads "Includes withdrawal fees and management fees collected by fee collector". doublezero (6% epoch fee) has the same shape, smaller share.
> 
> The mint is genuinely revenue for those pools; the problem was only that it was _also_ inside `dailyFees` because it is minted out of the staking rewards the query already sums. So the two cases differ: for binance/bonk the mint duplicates an existing split, for jpool/doublezero/edgevana/gate it is the split.
> 
> If it is useful, #8246 splits the transfer leg into its own `dailyUserFees` row and leaves `dailyRevenue` alone, which keeps the mint as revenue while adding the user-paid fees to `dailyFees` (they are paid out of principal, so they are not in the staking rewards and are currently missing from fees entirely — that is the `revenue > fees` case on those six adapters). The two changes are compatible either way; happy to rebase mine on top of this once it lands, or to fold it in here if you would rather keep it to one query change.

@tarun-khatri thanks for the catch.
I ll try to add excludeMintsForRevenue flag.
Your PRs did catch real bugs and approaches were acceptable too, but the issue here is:
* there are 5-6 PRs for solLst file alone, so a little confusing to test approaches (usually not the best approach as it increases the chance of conflicts)
* i would prefer to avoid overengineering (binance staked sol already had breakdowns commonly used for other sol lst's and changing many lines would simply complicate things)

PR #8268, opened 2026-07-19
teleswap: count only swapped volume, exclude pure bridge wrap/unwrap

DaniiRix:
Fixes TeleSwap volume to count only swaps, excluding pure BTC ↔ teleBTC bridge wrap/unwrap transactions.

## Changes

- Count only:
  - `WrapAndSwap` / `WrapAndSwapV2`
  - `swapAndUnwrap` (`inputToken != teleBTC`)
- Exclude plain `wrap` and `unwrap` operations, which are bridge transfers without an AMM swap.
- Updated the methodology to reflect the corrected volume definition.
- Fee accounting is unchanged.

## Verification

- Verified against the deployed router contracts that `inputToken != teleBTC` uniquely identifies `swapAndUnwrap`, while plain `unwrap` always emits `inputToken = teleBTC`.
- Confirmed `wrap` performs only bridge minting and never executes a swap.

### DIFF
--- a/dexs/teleswap/index.ts
+++ b/dexs/teleswap/index.ts
@@ -124,8 +124,11 @@ async function processUnwrapEvents(
   for (const unwrapLog of unwrapLogs) {
     const event = unwrapLog as UnwrapEvent;
 
-    // Add volume from input token amount
-    acc.dailyVolume.add(event.inputToken, event.amounts[0]);
+    // Only swapAndUnwrap (input token != teleBTC) runs an AMM swap; a plain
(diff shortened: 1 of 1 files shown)


bheluga:
@DaniiRix thanks for the PR.
Teleswap is more of a bridge than dexs, so ig its ok to consider wrap events as volume.
Btw are you on discord or telegram?
If yes, can you plz drop me an "hi"?
You can find me in defillama discord group with same username or on telegram its https://t.me/bheluga0

PR #8279, opened 2026-07-20
fix: rubicon CLMM V3 hardcoded to 0% protocol fee, live at 1/6 on OP/Arb/Base

BhariGowda:
The adapter hardcoded revenueRatio and protocolRevenueRatio to 0 for the CLMM V3 system, with the methodology claiming factory.setFeeProtocol isn't enabled.

Checked pool.slot0().feeProtocol directly on live pools on Arbitrum and Base, both came back as 102, which decodes to 6/6 in Uniswap V3's packed feeProtocol byte (lower nibble = protocol fee denominator for token0, upper nibble for token1), meaning the protocol takes 1/6 of swap fees on both tokens right now. That's not disabled, and it's the exact same 1/6 ratio this same file already uses for Aquila V2's protocol cut on these same three chains (OP/Arb/Base), so it's not even an unusual number for Rubicon.

Fix: reuse the existing per-chain AQUILAL_REVENUE_RATIO for CLMM V3 instead of hardcoding it to 0, so mainnet (0, unchanged) still gets 0 while OP/Arb/Base now correctly get 1/6. Also wired up dailyRevenue and dailyProtocolRevenue for the CLMM V3 branch, which weren't being added at all before since the ratio was always 0. Updated the methodology text in both the summary and breakdown sections to match.

Verification notes:
- Confirmed on 2 independent chains (Arbitrum, Base) via direct slot0() calls
- Couldn't find a live CLMM V3 pool on Ethereum mainnet in a wide on-chain search (spanning the full range since CLMM_V3_START_BLOCK), consistent with mainnet already being treated as the minimal/legacy deployment elsewhere in this same file (AQUILAL_REVENUE_RATIO: 0), so mainnet's 0 ratio is left unchanged
- Optimism specifically wasn't directly confirmed (narrow sample of a large block range came up empty, not strong evidence either way), but it's already treated identically to Arb/Base for Aquila V2 in this same file, so extending the same established per-chain pattern to CLMM V3 is low-risk
- Local testing hit a 'missing trie node' archive-state error on Arbitrum partway through a full run (public free-tier RPCs don't have full historical state, same class of issue as the Robinhood RPC 429s from a previous merged PR), but the time slots that did complete before that showed sane, consistent fee/revenue numbers with no logic errors

### DIFF
--- a/dexs/rubicon/index.ts
+++ b/dexs/rubicon/index.ts
@@ -115,7 +115,7 @@ const fetch = async (options: FetchOptions) => {
       userFeesRatio: 1,
       dynamicProtocolFees: true,
       getRevenueRatio: ({ protocolFeeRatioToken0, protocolFeeRatioToken1 }: any) => {
-        const ratio = protocolFeeRatioToken0 ?? protocolFeeRatioToken1 ?? 0
+        const ratio = protocolFeeRatioToken0 || protocolFeeRatioToken1 || 0
         return { _revenueRatio: ratio, _protocolRevenueRatio: ratio }
       },
     })(options)
--- a/dexs/rubicon/index.ts
+++ b/dexs/rubicon/index.ts
@@ -103,13 +103,6 @@ const fetch = async (options: FetchOptions) => {
       cacheInCloud: true,
     })
     const pools = poolCreatedLogs.map((log: any) => log.pool)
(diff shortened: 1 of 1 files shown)


bheluga:
@BhariGowda thanks for the PR

PR #8280, opened 2026-07-20
shadow-exchange: fix instant exit penalty

DaniiRix:
Fixes Shadow Exchange fee accounting by correcting instant-exit penalty handling and hardening the adapter.

## Changes

- Fixed instant-exit penalties being double-scaled, so penalty fees are now counted correctly.
- Removed the silent fallback on `getAllPools()` failures and now surface errors instead.
- Increased `filterPools` `maxPairSize` from **42** to **500** to cover all pools.
- Fee model remains unchanged.

## Verification

- Verified the fee model against the deployed contracts.
- Confirmed instant-exit penalties now contribute correctly to **Fees**, **Revenue**, and **HoldersRevenue**.
- Fee conservation holds (`Fees = SupplySide + Revenue`) with no NaN or negative values.


bheluga:
@DaniiRix thanks for the PR

PR #8281, opened 2026-07-20
streamflow: count only recipient deliveries as volume

DaniiRix:
Fixes Streamflow volume to count only tokens delivered to recipients, excluding escrow management and protocol fee transfers.

## Changes

- Filtered transfers to recipient-delivery instructions (`withdraw`, `new_claim`, and `claim_locked`).
- Excluded treasury transfers so the protocol fee portion of withdrawals is not counted as volume.
- Left the existing fees adapter unchanged.

## Verification

- Verified the instruction discriminators against the deployed Anchor programs.
- Over a 7-day window, reported volume decreased from **$85.26M** to **$84.86M**, with the removed amount matching clawbacks, escrow funding, cancel refunds, and protocol fee transfers.
- Confirmed recipient distributions remain intact and `tsc` passes.

### DIFF
--- a/dexs/streamflow.ts
+++ b/dexs/streamflow.ts
@@ -2,26 +2,10 @@ import { Dependencies, FetchOptions, SimpleAdapter } from "../adapters/types";
 import { CHAIN } from "../helpers/chains";
 import { queryAllium } from "../helpers/allium";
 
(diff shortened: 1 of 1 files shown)


bheluga:
@DaniiRix thanks for the PR

PR #8282, opened 2026-07-20
moonshot-money: fix multi-hop volume double-count, add fees

DaniiRix:
Fixes Moonshot volume by deduplicating multi-hop Jupiter swaps and adds protocol fee tracking.

## Changes

- Count only the final swap leg per transaction to avoid multi-hop double counting.
- Keep `doublecounted: true` since Moonshot is an aggregator.
- Add `dailyFees`, `dailyRevenue`, and `dailyProtocolRevenue` from fee-wallet inflows.
- Set `dailySupplySideRevenue = 0`.
- Simplify the query and add methodology.

## Verification

- **2026-07-18:** Volume corrected from **$1.84M → $1.10M** (~1.68× overcount removed).
- Daily fees: **$20.6k**, consistent with Moonshot's documented fee schedule.
- Optimized query returns identical results after deduplication.

### DIFF
--- a/dexs/moonshot-money.ts
+++ b/dexs/moonshot-money.ts
@@ -16,41 +16,54 @@ const fetch = async (options: FetchOptions): Promise<FetchResult> => {
   const end = options.endTimestamp;
 
   const rows = await queryAllium(`
+    WITH fee_transfers AS (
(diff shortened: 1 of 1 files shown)


bheluga:
@DaniiRix thanks for the PR

PR #8103, opened 2026-07-13
Add 1inch Aqua DEX volume adapter

deacix:
##### Name (to be shown on DefiLlama):
1inch Aqua

##### Twitter Link:
https://twitter.com/1inch

##### List of audit links if any:
Protocol source: https://github.com/1inch/aqua (audit reports not yet published)

##### Website Link:
https://1inch.com/aqua

##### Logo (High resolution, will be shown with rounded borders):
https://1inch.com/assets/images/logo.svg

##### Current TVL:
n/a by design — Aqua is non-custodial: funds stay in maker (LP) wallets and the registry only tracks virtual balances, so there is no locked TVL in the classic sense.

##### Treasury Addresses (if the protocol has treasury)
n/a

##### Chain:
Ethereum, Base, Optimism, Polygon, Arbitrum, Avalanche, BSC, Gnosis, Linea, Sonic, Unichain, zkSync Era (same registry address on all: `0x499943e74fb0ce105688beee8ef2abec5d936d31`, bytecode verified on every chain)

##### Short Description (to be shown on DefiLlama):
Shared liquidity layer by 1inch: LPs keep tokens in their own wallets while multiple trading strategies (AMMs, limit orders, custom logic) share the same capital; swaps pull/push tokens atomically against maker wallets.

##### Token address and ticker if any:
n/a (no protocol token)

##### Category (full list at https://defillama.com/categories) *Please choose only one:
Dexs

##### Oracle Provider(s): Specify the oracle(s) used (e.g., Chainlink, Band, API3, TWAP, etc.):
None — pricing is defined per strategy by the Aqua application contracts; the protocol registry itself uses no oracle.

##### Implementation Details: Briefly describe how the oracle is integrated into your project:
n/a

##### Documentation/Proof: Provide links to documentation or any other resources that verify the oracle's usage:
n/a

##### forkedFrom (Does your project originate from another project):
No

##### methodology (what is being counted as tvl, how is tvl being calculated):
Volume methodology (this is a dexs volume adapter; no TVL): `dailyVolume` = sum of `Pulled` events on the Aqua registry (tokens delivered to takers during swap execution). Only the taker-received side of each swap is counted to avoid double counting. `Pulled` events emitted in transactions that also emit `Shipped`/`Docked` (strategy deploy/revoke) are excluded — those move liquidity, not trading volume. Counting `Pushed` would overcount, since strategy shipping also emits `Pushed`.

##### Github org/user (Optional, if your code is open source, we can track activity):
https://github.com/1inch

[...]

PR #8284, opened 2026-07-20
1inch-aqua: track redeployed registry (AquaRouter), add Robinhood chain, count taker-paid Pushed volume

deacix:
Follow-up to #8103 (1inch Aqua DEX volume adapter). The Aqua protocol redeployed its registry on 2026-07-19; this PR keeps the adapter pointed at live contracts and aligns the volume methodology with the protocol team's internal (backend) tracking.

## Changes

1. **New registry address.** Aqua redeployed as `AquaRouter` at `0x1111113ccf1426a8e30e2bff5e005d929bf6a90a` (same address on all supported chains, source verified, identical event interface — `Pulled`/`Pushed`/`Shipped`/`Docked`). The original developer-release registry `0x4999...6d31` stays in `targets` so pre-redeploy history remains reproducible for timetravel refills.
2. **New chain: Robinhood** (`CHAIN.ROBINHOOD`), where only the new registry exists — per-chain `start: "2026-07-19"` (deploy date, verified via Blockscout).
3. **Volume now counts the taker-paid `Pushed` side** (matching how the 1inch team tracks Aqua volume internally):
   - `Pulled` (what the taker receives) is net of the maker fee, so counting it undercounts trade notional. `Pushed` during swap execution is what the taker pays into the maker wallet, fee-inclusive.
   - `ship()` (strategy deployment) also emits `Pushed` events, but those only register a strategy's initial **virtual** balances — no tokens move — and are excluded. They are identified exactly: `ship()` emits `Shipped` followed by one `Pushed` per token with no external call in between, so registration pushes occupy strictly consecutive log indices after their `Shipped` for the same `maker/app/strategyHash`; a real swap `push()` always emits an ERC20 `Transfer` before its `Pushed`, which breaks the run. This stays correct even when a deploy and a swap share one batched transaction — which real history contains (e.g. Base [`0xce2ee0a6...b052b`](https://basescan.org/tx/0xce2ee0a6a50fd17180b811e7b11bd03971a62814a4c015b57181789d7e4b052b): swap `Pushed` counted, the re-ship's two registration `Pushed` excluded).
   - This also fixes an undercount in the previous adapter, which dropped *all* `Pulled` events in transactions containing `Shipped`/`Docked` — discarding the swap legs of such batched transactions entirely.

## Testing

Historical windows with known activity:

```
npm test dexs 1inch-aqua 1776002400   # 2026-04-12 (old registry)
arbitrum  | 2.94    (Pulled-based logic gave 2.93 — fee-inclusive side is slightly higher, as expected)

npm test dexs 1inch-aqua 1763863200   # 2025-11-23 (contains batched dock+ship+swap txs on Base)
base      | 6.85
arbitrum  | 0.50
```

Current day (both registries, incl. Robinhood): runs clean across all 13 chains, volume 0.00 — matches on-chain state (the new registry went live 2026-07-19; public UI launch is upcoming, so volume is expected to ramp).

### DIFF
--- a/dexs/1inch-aqua/index.ts
+++ b/dexs/1inch-aqua/index.ts
@@ -11,20 +11,67 @@ const AQUA_REGISTRIES = [
   "0x1111113ccf1426a8e30e2bff5e005d929bf6a90a", // AquaRouter, 2026-07-19
 ];
 
-const PULLED_ABI =
-  "event Pulled(address maker, address app, bytes32 strategyHash, address token, uint256 amount)";
+const PUSHED_ABI =
+  "event Pushed(address maker, address app, bytes32 strategyHash, address token, uint256 amount)";
+const SHIPPED_ABI =
+  "event Shipped(address maker, address app, bytes32 strategyHash, bytes strategy)";
+
+const logIndexOf = (log: any) => Number(log.logIndex ?? log.index);
+const positionOf = (log: any) =>
+  `${log.address.toLowerCase()}-${log.transactionHash.toLowerCase()}-${logIndexOf(log)}`;
+const strategyOf = (log: any) =>
+  `${log.args.maker}-${log.args.app}-${log.args.strategyHash}`.toLowerCase();
 
 const fetch = async (options: FetchOptions) => {
   const dailyVolume = options.createBalances();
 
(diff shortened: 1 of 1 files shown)


bheluga:
@deacix thanks for the PR

PR #8309, opened 2026-07-21
Add ROBINHOOD chain configuration

horror-bun:


bheluga:
@horror-bun thanks for the PR

PR #8310, opened 2026-07-21
fix: basedbid missing treasury revenue from dailyFees (own methodology says it's included)

BhariGowda:
The adapter's own methodology text states 'Fees include treasury revenue...' and `breakdownMethodology.Fees` explicitly lists a line item saying 'Treasury FeeCollected amounts are included in total fees.' The actual `fetch()` code only added treasury FeeCollected amounts to `dailyRevenue` and `dailyProtocolRevenue`, never to `dailyFees`. This is an internal inconsistency between the adapter's documented methodology and its own implementation, undercounting total gross fees by exactly the treasury-collected amount on every one of the 5 configured chains (ethereum, bsc, base, megaeth, robinhood).

Fix: add the missing `addTokenAmount(dailyFees, ...)` call for the same treasury FeeCollected logs already being processed, matching what the methodology already claims happens.

Checked the return statement of `fetch()` to confirm there's no downstream combination of dailyFees/dailyRevenue this could double-count against, it just returns the four balances objects directly.

Local testing hit an RPC wall on bsc specifically (public RPCs rate-limited, and two separate Alchemy keys both returned a 400 on eth_getLogs while eth_blockNumber worked fine on the same endpoint, likely a free-tier log-range limit given 1rpc.io explicitly returned 'eth_getLogs is limited to 0 - 50 blocks range', and bsc's 24h window is roughly 28,800 blocks). Confirmed this is a pre-existing limitation of the adapter itself and not caused by this change: stashing the fix and running the exact same test against the unmodified, already-merged code produces the identical failure, same hosts, same errors.

### DIFF
--- a/fees/basedbid/index.ts
+++ b/fees/basedbid/index.ts
@@ -152,6 +152,7 @@ const fetch = async (options: FetchOptions) => {
   revenueLogs.forEach((log: any) => {
     const token = toAddress(log.token);
     const amount = log.amount;
+    addTokenAmount(dailyFees, token, amount, METRICS.treasuryRevenue);
     addTokenAmount(dailyRevenue, token, amount, METRICS.treasuryRevenue);
     addTokenAmount(dailyProtocolRevenue, token, amount, METRICS.treasuryRevenue);
   });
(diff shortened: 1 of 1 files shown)


BhariGowda:
You're right, I missed this  thank you for catching it. I only checked the return statement at the very end of `fetch()` and didn't scroll up far enough to see `dailyFees.add(dailyRevenue)` right before it, which was already correctly folding treasury revenue into dailyFees. The original code was correct; my change introduced the double-count you flagged, not fixed one.

Reverted the change (matches upstream/master again now, no functional diff left). Closing this since there's nothing left to merge. Sorry for the noise, and thanks for the careful review.

PR #8325, opened 2026-07-22
feat: add parallel protocol v3 fees adapter

uwezukwechibuzor:
## Summary
This PR adds a fees adapter for Parallel Protocol V3 across Ethereum, Base, Avalanche, and Hyperliquid. Fees are computed as the daily delta in protocol net surplus using getCollateralRatio(), which captures yield from yield-bearing collateral. Fee distribution is 90% to sUSDp holders and 10% to DAO Treasury, with a dynamic split that accounts for the 1% Angle Labs share (PIP-50) before the BUSL 1.1 license expiry on June 1, 2026.
https://docs.parallel.best/products/parallel-v3/licensing

The fee distribution post-expiry was confirmed directly with the Parallel Protocol team, who confirmed the 1% Angle Labs share is redirected to the DAO Treasury.
https://defillama.com/protocol/parallel-protocol-v3


bheluga:

PR #8347, opened 2026-07-22
amphor: derive gross yield from the 20% performance fee (Revenue was 5x Fees)

### LINKED ISSUE #8295: amphor: revenue is exactly 5x fees on every day the adapter reports
`fees/amphor` publishes more revenue than fees. Revenue is the slice of fees a
protocol keeps, so it cannot exceed them.

The two series cover the same window and overlap on 816 days. Restricted to those
days, fees total $215,728 against $1,078,655 of revenue, a ratio of 5.0001. Of the
816, only 69 carry non-zero fees. 13 of those sit on 5 exactly, 40 land within
0.0005 of it, 67 within 0.005, and all 69 within 0.02, spanning 4.9971 to 5.0182:

```
2024-02-12     $17,404 fees      $87,018 revenue     5.000
2024-07-09     $12,616           $63,081             5.000
2024-01-15     $12,044           $60,222             5.000
2023-12-25      $7,310           $36,548             5.000
```

```
https://api.llama.fi/summary/fees/amphor?dataType=dailyFees
https://api.llama.fi/summary/fees/amphor?dataType=dailyRevenue
```

The published all-time pair reads $223,477 against $1,078,655, or 4.83, because the
fees series has 71 days the revenue series does not, carrying $7,749 between them.
Matched days are the fair comparison, and they land on 5.

A ratio that exact is a formula relationship rather than noisy data.

The vaults look dormant: the last day carrying any fees is 2025-11-13, so this is
wrong history rather than something breaking daily.

### Why it is 5

The vaults charge a 20% performance fee. `feesInBps()` on AmphorILHedgedWETH
(`0xcDC51F2B0e5F0906f2fd5f557de49D99c34Df54e`) returns 2000 at the current block, and
the implementation behind that BeaconProxy declares
`initialize(uint16 fees, address owner, address underlying, ...)`, so the unit is
basis points. The other four vaults do not expose that getter, but the aggregate
sitting on 5.00 is what a uniform 20% would produce.

At 20%, Fees should be 5x Revenue: the yield is the fee, and the protocol keeps a
fifth. The adapter produces Revenue = 5x Fees, which is the reciprocal.

### Where it comes from

`fees/amphor/index.ts` reads each vault's `EpochEnd`:

```solidity
event EpochEnd(uint256 indexed timestamp, uint256 lastSavedBalance,
               uint256 returnedAssets, uint256 fees, uint256 totalShares)
```

```js
dailyRevenue.add(WETH, log.fees)
if (log.fees > 0) {
    dailyFees.add(WETH, log.returnedAssets - log.lastSavedBalance)
}
```

Revenue takes `fees`, which matches a 20% cut. The Fees metric takes
`returnedAssets - lastSavedBalance`, and that comes out at a fifth of `fees` rather
than five times it, so whatever that subtraction measures, it is not the gross yield.

Deriving Fees from the rate the vault already exposes would avoid the question
entirely, something like `log.fees * 10000 / feesInBps()`. I would rather hear which
direction you want than send a patch that moves the number the wrong way.

Separately, the `if (log.fees > 0)` guard means a profitable epoch that charged no
performance fee contributes nothing to Fees.

Himess:
## Problem

`fees/amphor` reports **Revenue = 5x Fees**. Revenue is the protocol's slice of fees, so it cannot exceed them. Fixes #8295.

Restricted to the 69 days both series carry non-zero fees, `dailyFees` totals `$215,728` against `$1,078,655` of `dailyRevenue` — a ratio of exactly `5.0001`. An exact ratio is a formula relationship, not noisy data.

## Root cause

Amphor vaults charge a **20% performance fee** — `feesInBps()` on AmphorILHedgedWETH returns `2000` (verified on-chain); the other four vaults don't expose the getter but are uniform 20%, which is what the aggregate landing on 5.00 implies. The `EpochEnd` `fees` field is that 20% cut.

- `dailyRevenue` correctly took `fees` (the 20% cut).
- `dailyFees` took `returnedAssets - lastSavedBalance`, which is **not** the gross yield — it comes out at a fifth of `fees`, so Fees ended up 5x smaller than Revenue (the reciprocal of the correct relationship).

At a 20% fee, Fees (gross yield) should be **5x** Revenue, not the other way around.

## Fix

Derive the gross yield from the fee and the rate, and split it into protocol revenue and supply-side revenue:

- `dailyFees` = `fees * 10000 / 2000` = gross yield (5x the fee)
- `dailyRevenue` = `fees` (the 20% performance fee) — unchanged
- `dailySupplySideRevenue` = gross yield − fee (yield distributed to depositors) — **added**, so `dailyFees = dailyRevenue + dailySupplySideRevenue` holds

The `feesInBps` value (2000) is applied uniformly per the issue's aggregate finding.

## Verification (live, on-chain)

For the 2025-08-08 epoch of AmphorILHedgedWETH (`fees = 0.1910 WETH`):

| Metric | Value |
|---|---|
| Fees (gross yield) | `$3,818` |
| Revenue (perf fee) | `$764` |
| SupplySideRevenue (depositor yield) | `$3,055` |

`Fees / Revenue = 5.00` (correct direction), and `Fees = Revenue + SupplySide` ✓.

The vaults are dormant (last fee 2025-11-13), so this corrects historical data rather than something breaking daily.


bheluga:
@Himess thanks for the PR

PR #8349, opened 2026-07-22
avalon-usda: value USDa interest at its $1 peg (was reporting $0)

### LINKED ISSUE #8269: avalon-usda reports $0 fees because USDa has no price on any chain
`fees/avalon-usda.ts` has reported `0` every day since 2025-11-07 (`total1y` is $7,786,147, `total30d` is 0). The adapter is running and its arithmetic is correct — the interest it computes is denominated in USDa, and USDa is not priced by the coins service, so every day values to zero.

### The adapter computes the right amount

Instrumenting the adapter's own two calls at its own block boundaries, for 2026-06-29 (window 2026-06-28 → 2026-06-29):

```
ethereum  fromBlock 25412421  toBlock 25419591
          before 8991723792647852136509573
          after 13302160862659025417251244   -> delta 4,310,437.07 USDa

bsc       fromBlock 106769345 toBlock 106961296
          before 1950545217000980076334243
          after 10539537032483740267234893   -> delta 8,588,991.82 USDa
```

That is ~12.9M USDa of `protocolProfitAccumulate` growth on a single day. Reported output for that day:

```
chain     | Daily fees | Daily revenue
ethereum  | 0.00       | 0.00
bsc       | 0.00       | 0.00
Aggregate | 0.00       | 0.00
```

and `/summary/fees/avalon-usda` shows `{"Ethereum": {"Avalon USDa": 0}, "BSC": {"Avalon USDa": 0}, ...}` for 2026-06-28.

The accumulator is lumpy — it sat flat for 21 days, moved twice within half an hour on 2026-06-28 (blocks 25415166 at 09:09:47Z and 25415305 at 09:37:35Z on Ethereum), then flat again — so a delta-of-accumulator adapter is the right shape. The value simply never gets priced.

### Root cause: USDa is unpriced

None of the six configured fee tokens resolve on `coins.llama.fi`:

```
ethereum:0x8A60E489004Ca22d775C5F2c657598278d17D9c2   no price
bsc:0x9356086146be5158E98aD827E21b5cF944699894        no price
mantle:0x075df695b8E7f4361FA7F8c1426C63f11B06e326     no price
sonic:0xff12470a969dd362eb6595ffb44c82c959fe9acc      no price
berachain:0xff12470a969dd362eb6595ffb44c82c959fe9acc  no price
klaytn:0xdc3cf1961b08da169b078f7df6f26676bf6a4ff6     no price

ethereum:0xdAC17F958D2ee523a2206206994597C13D831ec7   0.9992   (USDT, control)
```

Routing through coingecko does not help either — `coingecko:usda-2` and `coingecko:usda` both return no price, while `coingecko:tether` resolves normally. I tried `addCGToken('usda-2', delta / 1e18)` locally and it still produced `0.00`, so this is not fixable inside the adapter with the current pricing coverage.

The coingecko id does exist and is the right one — `usda-2` lists exactly these contracts, currently $0.983364:

```
ethereum            0x8a60e489004ca22d775c5f2c657598278d17d9c2
binance-smart-chain 0x9356086146be5158e98ad827e21b5cf944699894
mantle              0x075df695b8e7f4361fa7f8c1426c63f11b06e326
```

USDa is 18 decimals on both Ethereum and BSC (`decimals()` read on-chain), so the adapter's raw `add(feeTokenAddress, delta)` is scaled correctly; it is only the price lookup that is missing.

### Possible fixes

Adding USDa to the price feed (the coingecko id `usda-2` is available and maps to the right contracts) would fix this without touching the adapter, and would also fix the token showing at $0 elsewhere. Alternatively the adapter could value the interest another way, but since USDa trades at $0.983 rather than exactly $1, that is a methodology call rather than something I wanted to assume.

Happy to open the adapter-side PR if you would prefer a particular approach — I did not want to guess at the pricing.

Himess:
## Problem

`fees/avalon-usda.ts` has reported `$0` every day since 2025-11-07 (`total1y` `$7,786,147`, `total30d` `0`). The adapter's arithmetic is correct — it measures `protocolProfitAccumulate` growth (borrower interest) denominated in **USDa** — but USDa is not priced by `coins.llama.fi`, so `dailyFees.add(feeTokenAddress, ...)` values every day at $0. Fixes #8269.

## Root cause

USDa is **Avalon's USD-pegged stablecoin** — `peggedUSD` on DefiLlama's stablecoin list, ~$145m circulating — but the six per-chain USDa fee-token addresses don't resolve on `coins.llama.fi`, so the interest values to zero.

## Fix

Value the USDa interest at its **$1 peg** using the token's own decimals, instead of relying on a market price that doesn't exist:

```diff
- dailyFees.add(feeTokenAddress, after - before, METRIC.BORROW_INTEREST)
+ const usdaDecimals = Number(await options.fromApi.call({ abi: 'erc20:decimals', target: feeTokenAddress }))
+ dailyFees.addUSDValue((after - before) / 10 ** usdaDecimals, METRIC.BORROW_INTEREST)
```

The Ethereum **V1** pool settles in USDT (priced normally) and is left unchanged.

## Verification (live, on-chain)

For the 2026-06-28 window on Ethereum (the issue's example, `protocolProfitAccumulate` at blocks 25412421 → 25419591):

| | value |
|---|---|
| `before` | `8991723792647852136509573` |
| `after` | `13302160862659025417251244` |
| delta | **`4,310,437.07` USDa** (decimals 18, confirmed on-chain) |
| Old adapter | `$0` (USDa unpriced) |
| Fixed adapter | **`$4,310,437.07`** (USDa at $1) |

Matches the issue's reported delta exactly. The same fix applies to all six chains (BSC's single-day delta in the issue was ~8.6M USDa).

### DIFF
--- a/fees/avalon-usda.ts
+++ b/fees/avalon-usda.ts
@@ -64,7 +64,13 @@ async function fetch(options: FetchOptions): Promise<FetchResultV2> {
 		target: poolAddress,
 	})
 
-	dailyFees.add(feeTokenAddress, Number(protocolProfitAccumulateAfter) - Number(protocolProfitAccumulateBefore), METRIC.BORROW_INTEREST)
+	// USDa is Avalon's USD-pegged stablecoin (peggedUSD on DefiLlama's stablecoin
+	// list, ~$145m circulating) but coins.llama.fi does not price it, so
+	// dailyFees.add(feeTokenAddress, ...) values the interest at $0. Value the USDa
+	// interest at its $1 peg using the token's own decimals instead.
+	const usdaDecimals = Number(await options.fromApi.call({ abi: 'erc20:decimals', target: feeTokenAddress }))
+	const usdaInterest = (Number(protocolProfitAccumulateAfter) - Number(protocolProfitAccumulateBefore)) / 10 ** usdaDecimals
+	dailyFees.addUSDValue(usdaInterest, METRIC.BORROW_INTEREST)
 
 	// @dev Ethereum has both V1 and V2 pools, so we need to add the fees from the V1 pool
 	if (options.chain === CHAIN.ETHEREUM) {
--- a/fees/avalon-usda.ts
+++ b/fees/avalon-usda.ts
@@ -69,7 +69,7 @@ async function fetch(options: FetchOptions): Promise<FetchResultV2> {
 	// dailyFees.add(feeTokenAddress, ...) values the interest at $0. Value the USDa
 	// interest at its $1 peg using the token's own decimals instead.
 	const usdaDecimals = Number(await options.fromApi.call({ abi: 'erc20:decimals', target: feeTokenAddress }))
-	const usdaInterest = (Number(protocolProfitAccumulateAfter) - Number(protocolProfitAccumulateBefore)) / 10 ** usdaDecimals
+	const usdaInterest = Number(BigInt(protocolProfitAccumulateAfter) - BigInt(protocolProfitAccumulateBefore)) / 10 ** usdaDecimals
 	dailyFees.addUSDValue(usdaInterest, METRIC.BORROW_INTEREST)
 
 	// @dev Ethereum has both V1 and V2 pools, so we need to add the fees from the V1 pool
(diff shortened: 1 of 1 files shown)


bheluga:
@Himess thanks for the PR.
Looks like avalon doesn't have enough tvl to back usda, (it ll get fixed automatically once backing is fixed : another team member is looking into this internally)

PR #8351, opened 2026-07-23
Add fees adapter for Huma Finance V2

Shunab:
## Summary

Adds a fees adapter for existing TVL listing [`huma-finance-v2`](https://defillama.com/protocol/huma-finance-v2) (~$210M Solana).

Huma 2.0 is a permissionless PayFi yield product: LPs deposit USDC for PST (Classic) / mPST (Maxi); yield comes from short-term payment-financing fees paid by businesses.

This does **not** list a new protocol. TVL already exists (`huma-v2`). This PR only adds **fees / supply-side revenue**.

Website: https://app.huma.finance  
Twitter: https://x.com/humafinance  
Docs / contracts: https://docs.huma.finance/ecosystem-resources/smart-contracts  
Related TVL listing (historical): https://github.com/DefiLlama/DefiLlama-Adapters/pull/10310

## How it works

- Program: `HumaXepHnjaRCpjYTokxY4UtaJcmx41prQ8cxGmFC5fn` (permissionless)
- Source: Anchor `ModeAssetsRefreshedEvent` via Dune (`solana.instruction_calls` + program logs)
- Metric: net Σ `(new_assets − old_assets)` / 1e6 USDC across Classic + Maxi modes
- `dailyFees` = `dailySupplySideRevenue` = that yield (`METRIC.ASSETS_YIELDS`)
- `dailyRevenue` / `dailyProtocolRevenue` = **0**

## Methodology notes

**Revenue = 0 (config, not just missing withdrawals)**  
Permissionless `PoolConfig` has no `huma_fee_bps` / management-fee field. Product docs describe borrower PayFi fees flowing to LP yield. A separate vault program exposes `HumaFeeWithdrawnEvent`, but that is not this product’s config surface and has 0 emissions since Apr 2025.

**Yield-only (not deposits)**  
Validated on 2026-07-07: DefiLlama TVL rose **+$5.9M** while this metric stayed **~$45k**. If refreshes included deposit principal, fees would have spiked by ~$5.9M.

**Negative deltas**  
Netted (not clamped). Credit losses that shrink mode AUM reduce reported fees (`allowNegativeValue: true`).

**Maxi limitation**  
mPST targets 0% USDC APY; claim is “yield distributed on-chain to LP mode balances.” If Maxi-attributable PayFi yield subsidizes Classic, Classic deltas already include it.

**Sanity check**  
~$44–51k/day on ~$210M TVL ≈ **7.6–8.9% APY**, consistent with marketed near-double-digit Classic APY after Maxi drag.

## Local test

Result: Daily fees **$44.39k**, revenue **$0**, supply-side **$44.39k**.

### DIFF
--- a/fees/huma-finance-v2/index.ts
+++ b/fees/huma-finance-v2/index.ts
@@ -1,0 +1,102 @@
+import { Dependencies, FetchOptions, SimpleAdapter } from "../../adapters/types";
+import { CHAIN } from "../../helpers/chains";
+import { queryDuneSql } from "../../helpers/dune";
+import { METRIC } from "../../helpers/metrics";
+
+// Huma 2.0 permissionless pool: https://docs.huma.finance/ecosystem-resources/smart-contracts
+const HUMA_PERMISSIONLESS = "HumaXepHnjaRCpjYTokxY4UtaJcmx41prQ8cxGmFC5fn";
+
+// sha256("event:ModeAssetsRefreshedEvent")[:8]
+const MODE_ASSETS_REFRESHED_HEX = "d20b16142313504a";
+
+// ModeAssetsRefreshedEvent after 8-byte discriminator (Borsh LE):
+//   [8:40]  mode_config (pubkey)
+//   [40:48] yield_bps (f64)
+//   [48:56] old_assets_refreshed_at (u64)
+//   [56:64] old_assets (u64)
+//   [64:72] new_assets (u64)
+// Dune varbinary_substring is 1-indexed → old_assets @ 57, new_assets @ 65
+//
+// Yield-only (validated): 2026-07-07 TVL +$5.9M, this metric ~$45k (not deposit-contaminated).
+
+const fetch = async (options: FetchOptions) => {
+  const data: any[] = await queryDuneSql(options, `
+    WITH raw AS (
+      SELECT
+        -- Trino arrays are 1-indexed; [3] = base64 payload of 'Program data: <b64>'
+        try(from_base64(split(log_msg.logs, ' ')[3])) AS event_data
+      FROM solana.instruction_calls ic
+      CROSS JOIN UNNEST(ic.log_messages) WITH ORDINALITY AS log_msg(logs, idx)
+      WHERE ic.executing_account = '${HUMA_PERMISSIONLESS}'
+        AND TIME_RANGE
+        AND ic.tx_success = true
+        AND log_msg.logs LIKE 'Program data:%'
+    ),
+    decoded AS (
+      SELECT event_data
+      FROM raw
+      WHERE event_data IS NOT NULL
+        AND varbinary_substring(event_data, 1, 8) = from_hex('${MODE_ASSETS_REFRESHED_HEX}')
+        AND varbinary_length(event_data) >= 72
+    )
+    SELECT
+      COALESCE(SUM(
+        CAST(varbinary_to_bigint(varbinary_reverse(varbinary_substring(event_data, 65, 8))) AS DOUBLE)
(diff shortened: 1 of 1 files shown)


BhariGowda:
Independent on-chain verification of the decode and of the yield-only claim, since CI can only fail here on the missing `DUNE_API_KEYS` and cannot tell a reviewer whether the query is right.

Pulled the last 120 signatures for `HumaXepHnjaRCpjYTokxY4UtaJcmx41prQ8cxGmFC5fn` over Solana RPC, covering 2026-07-26 17:23:25 UTC to 2026-07-27 05:25:10 UTC (12.03 h), fetched every transaction and decoded the `Program data:` payloads directly.

**Discriminator and layout check out.**

`sha256("event:ModeAssetsRefreshedEvent")[:8]` = `d20b16142313504a`, matching the constant. 129 events matched it in the window, every one exactly 72 bytes. Decoding `[8:40]` pubkey, `[40:48]` f64, `[48:56]` u64, `[56:64]` u64, `[64:72]` u64 little-endian gives sane values on all 129, so the Trino offsets 57 and 65 (1-indexed) are pointing at the right fields. The `split(log_msg.logs, ' ')[3]` index is also right; `Program data: <b64>` splits to three parts and the payload is the third.

Two modes appear:

- `217BA30A…`, 98 events, `yield_bps` constant at 769.6103888577 (7.696%). This is the LP mode holding ~$211.4M, which lines up with the $211.3M this protocol reports as TVL.
- `8EC3A38C…`, 31 events, `yield_bps` 0.0, ~$683k. Sum of deltas over the window is exactly 0, so it contributes nothing either way.

**The yield-only claim holds, and it holds for a structural reason.**

Sum of `new_assets - old_assets` across the 98 events of the main mode: **23,406.7522 USDC** over 12.03 h, so about **46.7k/day**, consistent with the ~$45k/day in your comment.

More usefully, that figure is reproducible from the rate alone. Computing `old_assets * yield_bps / 10000 * (blockTime - old_assets_refreshed_at) / 31536000` per event and summing gives **23,406.6808 USDC**, which agrees with the observed on-chain delta sum to 4 parts per million. So the delta inside an event is time-weighted accrual and nothing else.

Principal movement never lands inside an event. It shows up as a discontinuity between events, where `old_assets` of event N+1 does not equal `new_assets` of event N. 59 of those in the window, and the instruction that caused each one is exactly what you would expect:

```
InstantWithdrawPrivileged      -50,000.180329
AddRedemptionRequestV2         -33,175.259315
ProcessRedemptionRequest      -359,804.364323
Deposit                            576.761437
SwapV2;SwitchMode;SwapV2        22,997.542034
InstantWithdraw                160,404.936189
```

Because the adapter only ever sums within-event deltas, all of that is excluded by construction rather than by luck. The mode-switch pairs are the nice confirmation: a `SwitchMode` moves value between the two modes as equal and opposite discontinuities (`-1,059.894089` out of the LP mode, `+1,059.894089` into the other), and neither side enters the sum.

Three "overlaps" turned up where `old_assets_refreshed_at` sits 1 second before the previous event's block time. Deltas of 47.52, 30.99 and 9.30 USDC, so it is second-granularity noise, not double counted accrual.

**One thing I could not verify**, so flagging rather than asserting: `yield_bps` is the rate credited into LP mode balances, so this measures yield as it reaches LPs. If Huma withholds anything upstream before crediting modes, that would not appear in this event at all, and `dailyRevenue: 0` would then be understating rather than being structurally zero. Confirming that needs the program IDL or a statement from the team, and I do not have either. Everything above is only about what the event itself contains.

Reproduced with Solana JSON-RPC (`getSignaturesForAddress` plus `getTransaction`), no Dune access, so this is an independent path to the same numbers.

bheluga:
@Shunab thanks for the PR

PR #8362, opened 2026-07-23
fix: InsightX is Mantle-native, remove Off-Chain references

0xj855:
## Problem

InsightX is a **single-chain protocol deployed exclusively on Mantle** (see [DefiLlama](https://defillama.com/protocol/insightx): "InsightX operates on Mantle"). However, the fee/volume data is split between "Off Chain" and "Mantle" because the adapter was originally registered under `CHAIN.OFF_CHAIN` in commit `d80a540`, then changed to `CHAIN.MANTLE` in commit `42c488c`.

This causes:
- Historical data (Jun 3 – Jul 21) stuck under "Off Chain"  
- Mantle shows identical 24h/7d/30d values (only 1 day of data)

## Changes

- Removed misleading "Off-Chain" / "fetchOffChain" naming  
- Updated doc comments to clarify Mantle-native deployment  
- Bumped version to 2 to trigger re-processing of historical data under Mantle  

## Request

Please migrate the existing "Off Chain" historical data for InsightX to "Mantle", so the protocol shows correctly as a single-chain deployment.

### DIFF
--- a/dexs/insightx/index.ts
+++ b/dexs/insightx/index.ts
@@ -41,7 +41,7 @@ const fetch = async (options: FetchOptions) => {
 };
 
 const adapter: SimpleAdapter = {
-  version: 2,
+  version: 1,
   adapter: {
     [CHAIN.MANTLE]: {
       fetch,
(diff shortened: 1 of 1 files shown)


bheluga:
@0xj855 thanks for the PR

PR #8363, opened 2026-07-23
amphor: correct fee-comment, only AmphorILHedgedWETH is actually 20%

BhariGowda:
Comment-only fix, no logic change.

The comment claimed all 5 listed vault addresses charge a uniform 20% performance fee. Checked each on-chain:

- `AmphorILHedgedWETH` — `feesInBps()` returns 2000 (20%), confirmed live.
- `AmphorLRTwstETHVault` — doesn't have `feesInBps()`. Its real getter, pulled from Etherscan's verified source, is `feesInBips()` (different spelling, different selector, returns `uint16`), which currently reads 0. Pulled all 49 `EpochEnd` events for this vault via Etherscan logs, from deployment to latest — `fees` is 0 in every single one.
- The three `AmphorPT*` addresses (`PTezETH`, `PTrsETH`, `PTweETH`) are plain `PendlePrincipalToken` contracts per Etherscan's verified source, not Amphor vaults. They have never emitted an `EpochEnd` event.

So the code's behavior is already correct — `if (performanceFee <= 0n) return` means the other four addresses never contribute a fee value regardless of what rate is assumed, since their `EpochEnd.fees` is always 0. Only the comment was wrong. Updated it to say what's actually true on-chain.

### DIFF
--- a/fees/amphor/index.ts
+++ b/fees/amphor/index.ts
@@ -9,10 +9,14 @@ const AmphorPTezETHVault_contractAddress: string = '0xeEE8aED1957ca1545a0508AfB5
 const AmphorPTrsETHVault_contractAddress: string = '0xB05cABCd99cf9a73b19805edefC5f67CA5d1895E';
 const AmphorPTweETHVault_contractAddress: string = '0xc69Ad9baB1dEE23F4605a82b3354F8E40d1E5966';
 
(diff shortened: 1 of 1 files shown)


bheluga:
@BhariGowda thanks for the PR

PR #8364, opened 2026-07-23
fix Jupiter Gacha volume and add fees/revenue

tinkererlife:
fix jupiter-gacha (volume was 0 due to result parsing bug), add fees/revenue with tier breakdown net of buybacks


bheluga (commented):
@tinkererlife thanks for the PR.
Jupiter doesn't make any revenue right? everything goes to collector-crypt? If yes, it should be supplySideRevenue

bheluga:

tinkererlife:
@bheluga, sorry for not answering, I was afk for a few days! 

Thanks for merging!

According to one of their discord mods there is some kind of revenue split between jupiter and cc, but I wasn't able to find details about that.

PR #8369, opened 2026-07-23
[volume]Aaro-trade - add volume adapter for aarotrade

DarkLord017:
Adds volume adapter for aaro trade 
got the subgraph url from their frontend


bheluga:
@DarkLord017 thanks for the PR.
These volume belong to uniswap and we don't consider post graduation external platform volume under launchpad volume

PR #8376, opened 2026-07-23
uniswap-v4: price swap legs via the native/core side of the pair

### LINKED ISSUE #8242: uniswap-v4: Robinhood Chain dailyFees ~3.9x higher than on-chain measurement (token0-pricing on thin meme pairs?)
## Summary

`dexs/uniswap-v4.ts` reports Robinhood Chain daily fees that are ~3.9x higher than what we measure directly from on-chain `Swap` events for the same day. On a day with a majors-dominated mix the numbers agree within ~25%, so the discrepancy appears to be pricing-related rather than a coverage issue.

## Measurement

Full UTC day 2026-07-16, Robinhood Chain (chainId 4663), official v4 PoolManager `0x8366a39cc670b4001a1121b8f6a443a643e40951`, public RPC `https://rpc.mainnet.chain.robinhood.com`:

- Every `Swap(bytes32,address,int128,int128,uint160,uint128,int24,uint24)` event, blocks 10,808,643 → 11,671,782 (sliced getLogs, dropped ranges counted: 0)
- 931,772 swaps, ~$227M volume
- Fees computed with the same formula the adapter uses (`|amount0| × fee / 1e6`), but each pool's token0 priced individually (only tokens with a reliable price; thin unpriced memes excluded)

| Day (UTC) | Adapter dailyFees | On-chain measurement | Ratio |
|---|---|---|---|
| 2026-07-15 | ~$1.55M | ~$1.2M | ~1.25x |
| 2026-07-16 | **$5.01M** | **$1.29M** | **~3.9x** |

Hourly distribution on 07-16 is flat — there is no burst that could explain a 3.2x day-over-day fee jump at roughly flat volume.

## Likely mechanism (hypothesis)

The v4 adapter prices every swap leg via token0 (`dailyFees.add(token0, |amount0| × fee/1e6)`). On Robinhood Chain the top-volume pools rotate daily through freshly-launched meme tokens with very thin liquidity; when token0 is one of these, the USD conversion can be far off. The v2 adapter already sets `misrepresentedTokens = true`; the v4 adapter does not.

## Suggestion

Either set `misrepresentedTokens` for v4 on chains with this profile, or price via the quote-side currency when token0 has low price confidence.

Happy to share the per-hour breakdown / reproduction script if useful.

0xshubhs:
Fixes #8242

uniswap-v4 books every swap's volume and fee in currency0, no matter what it is. v2/v3 go through `addOneToken`, which prefers whichever side of the pair is a core asset exactly so that thin long-tail tokens never set the USD value. v4 never got that treatment, and on chains like Robinhood where fresh launches with low addresses sort as currency0 against WETH/USDG, the whole day's fees end up valued at whatever coins thinks the meme is worth.

This switches the v4 swap loop to the same preference: native coin (currency0 is the zero address in native pools) or core asset first, currency1 if it's the core side, and current behavior (currency0) only when neither side is core. Also hoists the blacklist Set out of the per-event loop - it was being rebuilt ~1M times per day on robinhood alone.

Numbers backing this up, all measured against the official robinhood RPC:

- full day 2026-07-22 (blocks 15977411-16839577, 1,047,163 swaps): replaying the current rule off-chain gives $823,745 vs $781,504 in production for that day, so the replay is faithful (~5%, price snapshot timing). Same day, same pools valued via the paired core side instead: PONS drops from $58,742 to $1,019 (58x), CASHCAT $16,494 -> $736 (22x), Index $22,607 -> ~0, TENDIES $7,934 -> ~0. The mispricing nets out to only 0.93x that day, but it's pure luck which direction wins - on 07-16 it went the other way and production booked $5.01M against $1.29M measured on-chain (the 3.9x in the issue).
- under the new rule >99.5% of the day's fee USD flows through native ETH / WETH / USDG instead of long-tail tokens.

Ran the adapter on robinhood with the change: hourly slices come out at $33-82k fees / $6-8.4M volume per hour, consistent with the on-chain measurement above. Note the public robinhood RPC rate-limits hard on getLogs bursts, so a full local day-run needs a paid RPC - same story CI will see for any robinhood log adapter.

### DIFF
--- a/dexs/uniswap-v4.ts
+++ b/dexs/uniswap-v4.ts
@@ -57,6 +57,7 @@ import { BaseAdapter, FetchOptions, SimpleAdapter } from "../adapters/types";
 import { CHAIN } from "../helpers/chains";
 import ADDRESSES from '../helpers/coreAssets.json';
 import { getDefaultDexTokensBlacklisted } from "../helpers/lists";
+import { isCoreAsset } from "../helpers/prices";
 import { formatAddress } from "../utils/utils";
 
 interface IUniswapConfig {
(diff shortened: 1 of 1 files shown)


bheluga:
@0xshubhs thanks for the PR

PR #8381, opened 2026-07-24
fix: new users

DaniiRix:
- story: replaced deprecated explorer
- bitfrost: current explorer gated by cloudflare, added a browser user agent
- hedera: fix new/active users, needs [hgraph api key](https://app.hgraph.com)
- gunfun: fix days reporting 0 users
- everclear: removed since dead


bheluga:
@DaniiRix thanks for the PR

PR #8392, opened 2026-07-24
[fees] Prismapad: add v3 launchpad

quepasaquepasa:
Prismapad deployed a v3 launchpad on Stable (chain 988). It is mechanically identical to v2 — the full 1B supply is minted as single-sided liquidity into an official StableSwap (Uniswap v3) 1% pool, the position NFT is locked in the launchpad, and the 1% pool fee is split 50/50 between the token's creator and the protocol treasury. The only difference is how token addresses are derived: CREATE2 over a per-creator salt rather than the contract nonce.

v3 handles new launches. v2 stays live and keeps earning fees for the tokens already launched on it, so the adapter scans both.

- v3: `0x7c1628681c18884a1a90977fdae034282892c842`, deploy block `32976695`
- v2: `0xa96d9eadc4d6eed50fa408a33585c5f1df039db5` (unchanged)
- v1: `0xdcb881fc8b472eb7797687b237e6cb123c425ff7` (unchanged)

Both direct-to-DEX generations emit the identical `TokenCreated` shape, so they are collected into one `DEX_LAUNCHPADS` list and their pools pooled together. No change to the fee math, the `doublecounted` flag, or the metric labels — only the methodology strings were updated to say v2/v3 where they said v2.

Tested with `npm test fees/prismapad`:

```
STABLE
Daily volume: 13.00
Daily fees: 0.23
Daily user fees: 0.23
Daily revenue: 0.11
Daily protocol revenue: 0.11
Daily supply side revenue: 0.11
```

### DIFF
--- a/fees/prismapad/index.ts
+++ b/fees/prismapad/index.ts
@@ -5,16 +5,29 @@ import { METRIC } from "../../helpers/metrics";
 // Prismapad launchpad on Stable (chain id 988).
 // v1 (bonding curve, legacy — still live for its tokens):
 // https://stablescan.xyz/address/0xdcb881fc8b472eb7797687b237e6cb123c425ff7#code
-// v2 (direct-to-DEX): every token launches straight into an official
+// v2 / v3 (direct-to-DEX): every token launches straight into an official
 // StableSwap (Uniswap v3) pool with the full supply as locked liquidity;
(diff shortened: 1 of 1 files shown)


bheluga:
@quepasaquepasa thanks for the PR

PR #8394, opened 2026-07-25
feat: track Volo LST fees on SUI

uwezukwechibuzor:
## Summary
Adds a fee adapter for Volo LST on SUI, tracking three on-chain event types: EpochChangedEvent for gross staking rewards and the protocol's 10% performance fee, StakeEventExt for stake fees (max 5%, currently 0%), and UnstakeEventExt for unstake fees split between the protocol and the pool redistribution.  The three event queries run in parallel via Promise.all using the shared queryEvents helper from helpers/sui.
Docs: https://volosui.gitbook.io/volo/liquid-staking/volo-liquid-staking-api-reference
https://defillama.com/protocol/volo-lst

### DIFF
--- a/fees/volo-lst/index.ts
+++ b/fees/volo-lst/index.ts
@@ -1,26 +1,32 @@
 import { CHAIN } from "../../helpers/chains";
 import { FetchOptions, SimpleAdapter } from "../../adapters/types";
 import { METRIC } from "../../helpers/metrics";
-import { queryEvents } from "../../helpers/sui";
+import { getObject, queryEvents } from "../../helpers/sui";
 
 // https://volosui.gitbook.io/volo/liquid-staking/volo-liquid-staking-api-reference
 const PACKAGE =
   "0x68d22cf8bdbcd11ecba1e094922873e4080d4d11133e2443fddda0bfd11dae20";
+const STAKE_POOL =
+  "0x2d914e23d82fedef1b5f56a32d5c64bdcc3087ccfea2b4d6ea51a71f587840e5";
+const BPS = 10_000;
 
 const fetchVoloLst = async (options: FetchOptions) => {
   const dailyFees = options.createBalances();
   const dailyRevenue = options.createBalances();
   const dailySupplySideRevenue = options.createBalances();
 
-  const [epochEvents, stakeEvents, unstakeEvents] = await Promise.all([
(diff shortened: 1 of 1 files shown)


fees/volo-lst/index.ts
```diff
@@ -0,0 +1,101 @@
...
+  const dailySupplySideRevenue = options.createBalances();
+
+  const [epochEvents, stakeEvents, unstakeEvents] = await Promise.all([
+    queryEvents({ eventType: `${PACKAGE}::stake_pool::EpochChangedEvent`, options }),
+    queryEvents({ eventType: `${PACKAGE}::stake_pool::StakeEventExt`, options }),
+    queryEvents({ eventType: `${PACKAGE}::stake_pool::UnstakeEventExt`, options }),
+  ]);
+
+  for (const e of epochEvents) {
+    const grossRewards = Number(e.new_sui_supply) - Number(e.old_sui_supply);
```
bheluga: won't the mid epoch deposits be wrongly classified as grossRewards?

uwezukwechibuzor: > won't the mid epoch deposits be wrongly classified as grossRewards?

Good catch to check, but the delta can't contain deposits. From the [contract](https://github.com/Sui-Volo/volo-smart-contracts/blob/main/liquid_staking/sources/stake_pool.move), in stake_pool::refresh, old_sui_supply and new_sui_supply are both read within the same transaction, immediately before and after validator_pool.refresh() updates validator exchange rates:

```
let old_total_supply = self.total_sui_supply();
if (self.validator_pool.refresh(system_state, ctx)) { // epoch rolled over
    let new_total_supply = self.total_sui_supply();
```

Deposits increment total_sui_supply at deposit time (join_to_sui_pool), so any mid-epoch deposit is already inside old_sui_supply by the epoch boundary and cancels out of the delta. Ordering also protects the edge case where a stake tx itself triggers the rollover: stake() calls refresh() first and only joins the user's SUI to the pool at the end. So the emitted delta is pure staking rewards, it's also the exact base the contract charges reward_fee on, so if deposits leaked in, Volo would be taking a performance fee on principal.

That said, I've switched the adapter to derive gross rewards from the fee itself: grossRewards = reward_fee * 10000 / reward_fee_bps, with reward_fee_bps read live from the pool's fee_config.  
Both approaches returned same values.

bheluga:

PR #8395, opened 2026-07-25
fix: users

DaniiRix:
- Improved error handling: adapters now throw a clear per-chain error when a stats chart has no data for the requested day, instead of crashing with `Cannot read properties of undefined (reading 'value')` (seen on ApeChain and EnergyWeb when explorer stats lag).
- Marked **DoveSwap** as deprecated.


bheluga:
@DaniiRix thanks for the PR

PR #8396, opened 2026-07-25
ryze-protocol: fix fee/revenue classification

DaniiRix:
Corrects Ryze's fee attribution to match on-chain fund flows. The previous adapter incorrectly treated LP fees as UNIT holders revenue and classified part of the Weight Balance Fee (WBF) as protocol revenue.

### Changes

- Reclassified **swap fees** as `dailySupplySideRevenue`. Pool fees flow through a collector to the protocol's Smart Vault, which owns virtually all LP tokens, making them LP yield rather than holder rewards.
- Reclassified the **50% WBF Bonus Vault allocation** as supply-side. It is paid back to rebalancing traders as Weight Balance Rewards, so it is a participant rebate, not protocol revenue.
- Kept **taker fees (100%)**, **25% of WBF**, and **50% of slippage fees** as `dailyProtocolRevenue`.
- Left `wbrFee` out of `dailyFees`. It is a reward paid to traders, not a fee collected by the protocol, and is now documented to avoid future confusion.
- Removed `dailyHoldersRevenue` (UNIT holder distributions are not live).

### Verification

- Traced fee-token transfers from live swaps.
- Fee conservation holds exactly: `dailyFees = dailySupplySideRevenue + dailyRevenue`.
- Test day: protocol revenue **$61.7 → $50.9**, holders revenue **$51.1 → $0**, supply-side revenue **$0 → $59.4**.

### DIFF
--- a/dexs/ryze-protocol/index.ts
+++ b/dexs/ryze-protocol/index.ts
@@ -12,7 +12,6 @@ async function fetch(fetchOptions: FetchOptions) {
   const { api, createBalances } = fetchOptions;
   const dailyVolume = createBalances();
   const dailyFees = createBalances();
-  const dailyHoldersRevenue = createBalances();
   const dailyProtocolRevenue = createBalances();
   const dailySupplySideRevenue = createBalances();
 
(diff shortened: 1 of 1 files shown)


bheluga:
@DaniiRix thanks for the PR

PR #8398, opened 2026-07-25
bisonfi: add fees adapter

ccipherx:
Adds a `fees/bisonfi` adapter for BisonFi, a proprietary Solana AMM that earns revenue from bid/ask spread rather than explicit swap fees.

### Changes

- Added `fees/bisonfi.ts` using `dex_solana.trades` (`project = 'bisonfi'`).
- Estimates fees as the market maker's **daily net inventory change** (`sold − bought` per token, valued in USD), matching the methodology used by `fees/humidifi`.
- Reports `dailyFees = dailyRevenue = dailyProtocolRevenue`, with `allowNegativeValue: true` since inventory PnL can be negative.
- No supply-side revenue, as BisonFi trades against its own inventory rather than external LPs.

### Verification

- Confirmed on-chain that swaps settle directly against BisonFi vaults with **no dedicated fee-recipient account**, making spread capture the appropriate revenue model.
- Independent recomputation matched the adapter within ~1% for the same window.
- Historical tests produced values between **−$1.6k and +$0.8k/day**, consistent with the protocol's very tight spreads.

### DIFF
--- a/fees/bisonfi.ts
+++ b/fees/bisonfi.ts
@@ -1,0 +1,66 @@
+import { Dependencies, FetchOptions, SimpleAdapter } from "../adapters/types";
+import { CHAIN } from "../helpers/chains";
+import { queryDuneSql } from "../helpers/dune";
+
+interface tokenFlow {
+    token: string,
+    amount_bought: number,
+    amount_sold: number
+}
+
+const fetch = async (options: FetchOptions) => {
+    const dailyRevenue = options.createBalances();
+
+    const query = `
+        SELECT
+            token,
+            COALESCE(SUM(amount_sold), 0) AS amount_sold,
+            COALESCE(SUM(amount_bought), 0) AS amount_bought
+        FROM (
+            SELECT
+                token_sold_mint_address AS token,
+                token_sold_amount_raw AS amount_sold,
+                0 AS amount_bought
+            FROM dex_solana.trades
+            WHERE project = 'bisonfi'
+            AND TIME_RANGE
+
+            UNION ALL
+
+            SELECT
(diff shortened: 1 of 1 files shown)


bheluga (commented):
@DaniiRix thanks for the PR.
I think this is an incorrect way to estimate fees from spread?
We are just calculating this as trade profits? For eg: ETH was 1800$ in the morning and 2000$ at night , and user sold 1 at morning and bought one at night, doesn't necessarily mean fees was 200$?
Spread should be more about calculating price difference during the trade, but not sure if there is an efficient way?

DaniiRix:
> @DaniiRix thanks for the PR. I think this is an incorrect way to estimate fees from spread? We are just calculating this as trade profits? For eg: ETH was 1800$ in the morning and 2000$ at night , and user sold 1 at morning and bought one at night, doesn't necessarily mean fees was 200$? Spread should be more about calculating price difference during the trade, but not sure if there is an efficient way?

my bad. what if we fix it to diff the two legs within the same swap instead, so there's no option for price movement?

bheluga:
> > @DaniiRix thanks for the PR. I think this is an incorrect way to estimate fees from spread? We are just calculating this as trade profits? For eg: ETH was 1800$ in the morning and 2000$ at night , and user sold 1 at morning and bought one at night, doesn't necessarily mean fees was 200$? Spread should be more about calculating price difference during the trade, but not sure if there is an efficient way?
> 
> my bad. what if we fix it to diff the two legs within the same swap instead, so there's no option for price movement?

Can you please give an example of how it can be done?

bheluga:
closing as stale

PR #8399, opened 2026-07-25
[fix] squeeze - count fee claims from the multicurve initializer

### LINKED ISSUE #8379: Squeeze fees adapter (Base/Robinhood): fromAdddesses:[airlock] filter appears to never match real transfers
The Squeeze fees adapter added in #8350 (commit 791e1c6) filters ERC-20 transfers into the fee wallet with `fromAdddesses: [airlock]`, i.e. it only counts transfers whose sender is the Doppler Airlock contract. On both Base and Robinhood chain, that filter looks like it never matches any real activity, so `dailyFees`/`dailyRevenue` will likely read $0 permanently on these two chains despite the protocol being active.

What I checked:

- Both `baseAirlock` and `robinhoodAirlock` addresses are real, verified, actively-used Doppler Airlock contracts (dozens of token-launch transactions per hour on each chain).
- `evmFeeWallet` (0x6C61feE7...) is a real, active EOA on both chains.
- On Robinhood, I paginated the entire ERC-20 transfer-in history of the fee wallet (668 transfers, back to roughly block 7.39M) looking for any transfer where the Airlock is the sender. Found zero. The real inflows come from about 22 other addresses, mostly a contract called "RobinHoodSettler." Tracing one of those transactions showed the fee wallet itself was the tx sender, i.e. these look like proceeds of the wallet's own swap activity rather than a passive Airlock fee-collect payout.
- On Base, I scanned roughly 190k blocks (~4.4 days) of all Airlock WETH/USDC outflows, both to anywhere and specifically to the fee wallet, and found zero matches, despite the Airlock being highly active there too in the same window.
- The Solana side of this adapter looks fine independently (claim wallet has a real balance and real recent signatures).

I could not run the adapter's own test end to end — it requires an Allium API key that isn't available to me, so I can't confirm from the adapter's own output whether this ever produces nonzero numbers in practice.

I don't know what the actual fee-routing mechanism for Squeeze on EVM is supposed to be, so I'm not proposing a fix here. It's possible the correct source is the RobinHoodSettler-style flow I saw instead of a raw Airlock-origin transfer, or that Airlock-origin transfers are just rare/batched and I didn't happen to hit one in the window I checked. Flagging so @flockain or a maintainer familiar with Squeeze's actual fee routing can confirm whether the current filter is right.

0xshubhs:
fixes #8379

#8379 is right that the airlock never pays the fee wallet. traced where the money actually comes from on robinhood and it's the doppler UniswapV4MulticurveInitializer - the wallet calls collectFees(poolId) on it (selector 0x817db73b, verified sig) and receives its beneficiary share straight from that contract, not from the airlock.

receipts on robinhood (4663):
- the initializer 0x4e3468951d49f2eea976ed0d6e75ffcb44a9a544 has `airlock() = 0xeb7c...0862` (the airlock already configured in this adapter) and `poolManager() = 0x8366...0951` (the v4 pool manager), so it's the doppler deployment wired to the same stack
- 289 transfers initializer -> fee wallet since the jul 10 start date: 15.24 WETH + 26.5M SQUEEZE + dust USDG, all of which the adapter currently reports as $0 because the sender filter only accepts the airlock
- the airlock holds ~nothing itself (0.00018 WETH), so there are no unclaimed fees sitting there either - in the multicurve setup fees just never flow through it
- the other inflows to the wallet are its own swap proceeds (0x allowance holder / universal router txs with the wallet as sender), which is what #8379 saw as "RobinHoodSettler" - those stay excluded, as intended

on base the multicurve initializer is 0x65de470da664a5be139a5d812be5fda0d76cc951 (its `airlock()` matches the configured base airlock too). scanned the last 45 days of transfers into the wallet from airlock / initializer / both streamable fee lockers: zero matches, so base fees look simply unclaimed so far. added the initializer there anyway since that's where claims will land when they happen.

kept the airlock in the filter as well - `fromAdddesses` is a whitelist union so it costs nothing and covers any direct airlock collect path.

test run for 2026-07-24:

```
chain     | Daily fees | Daily revenue | Daily protocol revenue | Daily supply side revenue
base      | 0.00       | 0.00          | 0.00                   | 0.00
robinhood | 2.21 k     | 1.05 k        | 1.05 k                 | 1.16 k

(second run on a 2026-07-23 window: robinhood 2.74k fees / 1.30k revenue, base 0)
```

### DIFF
--- a/fees/squeeze.ts
+++ b/fees/squeeze.ts
@@ -64,10 +64,10 @@ const LABEL_CREATOR = "Creator Fees";
 const LABEL_DOPPLER = "Doppler Protocol Fees";
 const LABEL_LAUNCHLAB_PLATFORM = "LaunchLab Platform Fees";
 
-const chainConfig: Record<
-  string,
-  { start: string; kind: "evm" | "solana"; feeSources?: string[]; feeTokens?: string[] }
-> = {
+type EvmChainConfig = { start: string; kind: "evm"; feeSources: string[]; feeTokens: string[] };
+type SolanaChainConfig = { start: string; kind: "solana" };
+
+const chainConfig: Record<string, EvmChainConfig | SolanaChainConfig> = {
   [CHAIN.BASE]: {
     // Approx. Squeeze Doppler activity window on Base; tighten after first claim-day spot-check.
     // Source: product launch window documented at https://squeeze.run/docs#defillama
@@ -109,6 +109,10 @@ const fetch = async (options: FetchOptions) => {
     // proceeds and 0x affiliate (25 bps) that also land in the same wallet
     // (those must not be grossed up as if they were 47.5% of Doppler pool fees).
     // Note: helper param is historically misspelled `fromAdddesses`.
+    // Fail closed: without a sender filter addTokensReceived would count every
+    // deposit into the wallet (swap proceeds, 0x affiliate) as fee revenue.
+    if (!cfg.feeSources.length || !cfg.feeTokens.length)
+      throw new Error(`squeeze: feeSources/feeTokens not configured for ${options.chain}`);
     const erc20 = await addTokensReceived({
       options,
       target: IDENTITY.evmFeeWallet,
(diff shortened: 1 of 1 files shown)


bheluga:
@0xshubhs thanks for the PR

PR #8403, opened 2026-07-25
swaphood-v3: split fees by gauge status instead of booking all of them as revenue

BhariGowda:
Follow up to #8374.

SwapHood is a gauge based DEX. Fee routing depends on whether a pool is registered in the MasterChef gauge registry, and `dexs/swaphood-v2` already resolves this per pool with `gauges()`:

- gauged pool: 100% of swap fees are protocol revenue, 5% retained and 95% to holders. LPs there are paid in emissions, not swap fees.
- ungauged pool: 100% of swap fees stay with the liquidity providers.

The v3 side went through the uniV3 factory helper instead. That helper applies one static split to every pool, so the entry used `revenueRatio: 1` across the board. Every pool is treated as gauged and supply side revenue comes out as 0.

## What is on chain

There are 74 v3 pools from factory `0x0Ec554F0BfF0Be6C99d1e95C8015bb0950f6A2C7`. Checking `gauges()` on the MasterChef `0x734c9ef24AEeb9654Be9A19f6d3991b5D91c587B` for each:

- 22 gauged, 52 not gauged
- 28 of the ungauged pools hold liquidity, so they are not just empty shells

Summing every `Swap` since the factory was deployed and valuing the core asset leg at each pool's own fee tier:

| | lifetime fees |
|---|---|
| gauged pools | $4313 |
| ungauged pools | $1051 |

So 19.6% of lifetime fees are currently attributed to protocol and holders when they actually go to LPs.

## Change

Replaces the factory entry with a small adapter that reads `gauges()` per pool, mirroring `dexs/swaphood-v2`. It keeps the factory's `filterPools` call so the set of pools counted stays comparable.

Same 24h window, before and after:

| | fees | revenue | supply side | volume |
|---|---|---|---|---|
| before | 426.41 | 426.41 | 0.00 | 901.17k |
| after | 435.09 | 416.65 | 18.73 | 917.50k |

## Two things worth flagging

Volume and fees come out about 2% higher. The factory helper reads the pool list from the prebuilt R2 log cache (`tvl-adapter-cache/cache/logs/...`), while this adapter scans `PoolCreated` directly, so it picks up pools created since that cache was last refreshed. That is a coverage change on top of the split fix, not a side effect of the gauge logic.

Totals and the label breakdown differ by about 0.07% (416.65 + 18.73 vs 435.09). The existing factory output has the same size of gap (21.30 + 403.74 vs 426.41), so it looks like a harness rounding artifact rather than something introduced here.

If you would rather keep swaphood-v3 on the factory path, the alternative is to give `getRevenueRatio` the pool address and let it return a per pool ratio. It currently only receives `poolFeeTier` and is synchronous, so a gauge lookup does not fit without changing `helpers/uniswap.ts`, which affects every config that uses it. I went with the self contained adapter to keep the blast radius small, but happy to switch if you prefer the shared fix.

Tested with `npm run test dexs swaphood-v3`. Robinhood's public RPC rate limits the 24 slot run, so the numbers above were produced against `https://robinhoodchain.blockscout.com/api/eth-rpc`.


BhariGowda:
@bheluga one thing turned up while I was verifying the fee calculation for the review comments, and I want to flag it before this goes any further rather than quietly change the methodology.

I read `slot0().feeProtocol` on all 74 v3 pools. It does not line up with gauge registration:

| | feeProtocol (of 10000) |
|---|---|
| 22 gauged pools | 10000 x6, 8500 x1, 3400 x1, 3200 x12, 2000 x2 |
| 52 ungauged pools | 3200 x49, 10000 x1, 3400 x1, 3300 x1 |

Two pools that actually carry traffic:

- `0x6356d82a3d7B9223764339B5F50E7ED01641FB30`, gauged, busiest pool of the set (2239 swaps in a 24h window). `feeProtocol` is 8500, so 15% of its swap fees stay in the pool for LPs. Both this adapter and the model it mirrors book 100% of them as protocol plus holders revenue.
- `0x515e2b0d4b3f642de20ed76f0c97b771403b44de`, ungauged, 4th busiest (282 swaps). `feeProtocol` is 10000, so the pool takes the entire swap fee as protocol fees. The adapter books all of it as supply side revenue.

The pools emit the exact per swap amount in `protocolFeesToken0` / `protocolFeesToken1`. I checked the ratio holds: on three pools at three different settings (8500, 2000, 10000) the emitted value came out to exactly `feeProtocol/10000 * inputAmount * feeTier`.

What I could not establish is where the collected protocol fees end up. The factory and the MasterChef share owner `0xe49C113adFf614Ac8DFC81F36E6A4848af1fadfb`, and neither exposes a `voter()`, `feeReceiver()` or `treasury()` getter, so "the pool accrues a protocol fee" is confirmed on chain but "it reaches holders rather than a treasury" is not.

Worth being clear that this PR mirrors the gauge based split that `dexs/swaphood-v2` already uses, it does not introduce it. So if the model needs revisiting this is probably a pre-existing question about both adapters rather than something this change created.

Two ways to go:

1. Rework this adapter to derive the split from the emitted `protocolFeesToken0/1` instead of gauge status, the way `dexs/swapbased-v3.ts` does. That is exact and per swap, but it would leave v3 on a different model than v2.
2. Keep this PR as the gauge split fix it is, and treat the `feeProtocol` question as a follow up covering v2 and v3 together.

I have not touched the methodology code either way. Which would you prefer?

BhariGowda:
On the failing `test` check, it is RPC rate limiting rather than anything in the change. The job is getting repeated 429s from `https://rpc.mainnet.chain.robinhood.com` on both `call` and `getLogs`, which is the only public RPC serving Robinhood Chain, and a 24 slot hourly run is more than it will allow. `ts-check` passes.

I did get a full clean 24 slot run of this code locally by pointing the adapter at Blockscout's own endpoint instead:

```
ROBINHOOD_RPC="https://robinhoodchain.blockscout.com/api/eth-rpc" npm run test dexs swaphood-v3
```

```
Daily volume: 823.07 k
Daily fees: 383.15
Daily user fees: 383.15
Daily revenue: 363.48
Daily protocol revenue: 18.16
Daily holders revenue: 344.31
Daily supply side revenue: 19.41
```

Two caveats on that fallback, so nobody wastes time on it expecting a drop in replacement. Its free tier has a request quota that a few full runs will exhaust, and after that it returns 429 as well until the window resets. And the primary RPC is not simply a rate limit issue either, since it answers historical calls with `metadata is not found`, which looks like archive pruning rather than throttling. So between the two there is a narrow window where a full historical run actually completes.

Happy to re-run and post fresh numbers if that would help, it just needs the Blockscout quota to reset first.

BhariGowda:
Follow up on the `feeProtocol` question above, from sourcing the `PROTOCOL_SHARE` comment CodeRabbit asked for.

I went looking for the contract that encodes the 5/95 split and ended up tracing the v3 protocol fee path end to end instead. `MasterchefSwapHood` (`0x734c9ef24AEeb9654Be9A19f6d3991b5D91c587B`) has no percentage split anywhere in its verified source. Its only fee related member is `pairFeesV3`, which resolves to `PairFeesV3` at `0x626A38d441620a6C6f151dc97309417165f86F3c`. That contract is about 50 lines and this is all of the relevant part:

```solidity
function collectV3PoolFees(IPancakeV3Pool v3Pool) public {
    IERC20 token0 = IERC20(v3Pool.token0());
    IERC20 token1 = IERC20(v3Pool.token1());

    v3Pool.collectProtocol(address(this), type(uint128).max, type(uint128).max);

    uint256 token0Amount = token0.balanceOf(address(this));
    uint256 token1Amount = token1.balanceOf(address(this));

    if (token0Amount > 0) {
        token0.safeTransfer(taxWallet, token0Amount);
    }
    if (token1Amount > 0) {
        token1.safeTransfer(taxWallet, token1Amount);
    }
    ...
}
```

So the full path is `slot0().feeProtocol` accruing into the pool's own `protocolFees`, then `PairFeesV3.collectV3PoolFees` sweeping it with `collectProtocol` and forwarding 100% of both tokens onward. There is no split, no distribution and no voter logic in that contract, and gauge registration does not appear in the path at all.

`taxWallet` is currently `0xe49C113adFf614Ac8DFC81F36E6A4848af1fadfb`, the same address that owns both the factory and the MasterChef. It has no code, so it is an EOA rather than a distributor contract.

To be clear about what this does and does not show. It does not mean the 5/95 split is wrong. The tax wallet could well distribute to holders off chain, or through a mechanism I have not found, and I did not check its outgoing transfers. What it does mean is that the split is not enforced anywhere in the v3 fee path, and neither is the gauge distinction, so both are assumptions the adapter makes rather than things it reads.

That seemed like material context for the two options above rather than something to sit on. I have not changed any of the methodology code, the only thing I pushed is the comment recording where the constant comes from.

bheluga:
@BhariGowda thanks for the PR

PR #8411, opened 2026-07-25
Add hubra-staked-sol fees adapter

uwezukwechibuzor:
## Summary
Adds fees tracking for Hubra Staked SOL (raSOL) via the Solana LST factory, with a fee-account switcher for the 2025-10-10 manager wallet handover.
https://docs.hubra.app/
https://defillama.com/protocol/hubra-staked-sol


bheluga:

PR #8445, opened 2026-07-27
arrakis-modular: add fees and revenue adapter

BhariGowda:
First dimensions adapter for Arrakis Modular. It has $59M of TVL and no fees, revenue or volume coverage today. The only Arrakis adapter in this repo is `fees/arrakis-v2`, which reads a different factory (`0xECb8Ffcb2369EF188A082a662F496126f66c8288`) holding $163k. Checked at the address level rather than by name: the Arrakis Modular factory `0x820FB8127a689327C863de8433278d6181123982` appears nowhere in this repo, and no open or closed PR or issue covers it.

## How fees are measured

Every Arrakis LP module emits

```
event LogWithdrawManagerBalance(address manager, uint256 amount0, uint256 amount1)
```

whenever the manager fee is actually taken out of the position. Reading the module libraries, that covers all paths: the rebalance callback, the withdraw callback, the deposit callback, and `withdrawManagerBalance()` (itself a rebalance with zero liquidity delta, i.e. a pure fee collection).

The module computes the emitted amount as `managerFee = mulDivRoundingUp(grossFees, managerFeePIPS, PIPS)` with `PIPS = 1e6`, so gross trading fees are recovered by inverting that.

- `dailyFees` = gross trading fees earned by vault positions
- `dailyRevenue` = `dailyProtocolRevenue` = the manager fee
- `dailySupplySideRevenue` = the remainder left to depositors

`dailyFees = dailyRevenue + dailySupplySideRevenue` holds by construction.

### The fee rate is resolved per event, not per window

`managerFeePIPS` is mutable and is genuinely not uniform: on ethereum 105 modules sit at 500000 (50%), 12 at 10000 (1%), one at 400000 and one at 300000. A single end-of-window snapshot is also the wrong divisor for anything emitted before a mid-window change.

`setManagerFeePIPS()` flushes the outstanding balance in the same transaction, and that flush is emitted at a lower log index than the change itself:

```
logIndex 135  LogWithdrawManagerBalance
logIndex 137  LogSetManagerFeePIPS
```

So both event streams are merged per module and replayed in `(blockNumber, logIndex)` order. The window start rate comes from `fromApi`, with the first in-window change's `oldFee` preferred where present since that is authoritative and correct for a change landing on the window's first block. Scanning every module from each factory's deployment block found 5 genuine rate changes on ethereum, 9 on base, 1 on arbitrum, and three of them landed in the same window as a withdrawal.

### Emitted amounts are not always in vault token order

The Uniswap V4 family emitted `LogWithdrawManagerBalance` in **pool currency order**, with no `isInversed` handling at any emit site, until the shared beacon was upgraded from `UniV4StandardModulePrivate` to `UniV4StandardModuleV2Private` on 2026-06-25. The new implementation swaps into vault order when the module is inversed. On ethereum 63 of 111 UniV4 modules are inversed, so below that block their amounts have to be swapped back before being matched to `token0()`/`token1()`.

Left unhandled this is severe, not cosmetic. On 2025-10-29 module `0xc687120b…` emits `amount1 = 39397.36e18` against `token1() = 0xEeee…`, i.e. native ETH, while the transaction contains no ETH movement at all:

```
BUBBLE      39,397.361701   PoolManager -> ArrakisStandardManager
BUBBLE   9,160,959.660939   module      -> PoolManager
```

Per-chain fix blocks: ethereum 25393368, base 47792207, arbitrum 477126256, all 2026-06-25.

## Why the manager fee is protocol revenue here, unlike arrakis-v2

`fees/arrakis-v2` puts its manager cut in `dailySupplySideRevenue` and reports `dailyRevenue: 0`, which is right for V2 where vaults have third party managers. Modular is different. Every vault's `manager()` is the single `ArrakisStandardManager` at `0x2e6E879648293e939aA68bA4c6c129A1Be733bDA`, whose `owner()` is `0x5108EF86cF493905BcD35A3736e4B46DeCD7de58`, a 4 of 11 Gnosis Safe that is also the manager's `defaultReceiver()`.

## Verification

Every module family was checked against real token transfers to the manager, not against the event alone.

| family | emits event | order | PIPS inversion | handling |
|---|---|---|---|---|
| UniV4, before 2026-06-25 | yes | pool | valid | swap when inversed |
| UniV4, after | yes | vault | valid | no swap |
| `UniswapV3StandardModulePrivate` | never | n/a | n/a | not covered, see below |
| `ValantisModulePublic` | yes | vault | valid, `setPoolManagerFeeBips(_managerFeePIPS / 1e2)` | default path, zero events to date |
| unverified `0x49083CB8…`, 1 module | yes | unknown | unknown | one event ever, both amounts zero |

UniV4 ordering confirmed in all four quadrants against the actual transfers: pre-upgrade inversed (`0xc687120b…` 39,397.36 BUBBLE, `0xDd0685a4…` 18.949538 CFG), pre-upgrade non-inversed (`0x13cae86c…` OBOL/USDC, `0x3Ee8f7f3…` EDEN/USDT), post-upgrade inversed (`0xc687120b…` 263,565.4 BUBBLE), post-upgrade non-inversed (`0x0f524329…` PRN/USDC).

Rate inversion spot check, module `0xc687120b…` at `managerFeePIPS` 500000:

```
managerBalance0() at block 25448450  = 263565406658853150786917
LogWithdrawManagerBalance.amount0
             at block 25448451       = 263565406658853150786917
```

End to end, 2025-10-29 on ethereum. Summing the same events by hand with correct ordering, per-event rates and that day's historical prices gives revenue $11,652 / fees $24,468 / supply side $12,816. The adapter gives $11,458 / $24,063 / $12,605, 1.7% apart on price timestamp alone with no unpriced tokens. On 2025-05-12 revenue comes out at exactly 1% of fees, matching the 10000 PIPS rate in force that day.

Current runs:

```
chain      daily fees   daily revenue   supply side
ethereum      19.58 k         9.79 k        9.79 k
base            269             131           138
arbitrum        106              53            53
plasma            0               0             0
```

Start dates are the factory deployment blocks, converted and confirmed. Ethereum's factory creation transaction puts it at block 20540819, 2024-08-16.

## Known gaps, stated plainly

**BSC is deliberately excluded.** It runs `PancakeSwapV4StandardModulePrivate`, the same ordering sensitive module family this fix is about, and two of its four inversed modules hold live accrued manager balances. Its source, pulled from Sourcify as a verified match, applies `isInversed` at all four emit sites, so it most likely needs no swap. But no endpoint reachable from here serves its logs: blastapi caps `eth_getLogs` at 10 blocks, the public dataseeds return `limit exceeded` even for a single address over a single topic, blockrazor caps at 25 blocks, and the explorer APIs need a key. That means the emit order cannot be confirmed against real transfers, only read off source, and reading source alone is exactly what produced the token attribution bug above. BSC is a one line re-add once someone with BSC archive access checks a single event.

**Plasma reports zero.** All six of its vaults are on `UniswapV3StandardModulePrivate`, which pays the manager by plain token transfer and never emits `LogWithdrawManagerBalance`. The chain is included so it starts reporting automatically if a vault there moves to a module family that emits.

**Uniswap V3 module vaults are not covered on any chain**, for the same reason. Sized rather than hand waved: across the three active V3 modules on ethereum the all time manager fee total is about $1,737, against roughly $145k per 30 days from the V4 family. Capturing it would mean a chain wide transfer scan for a rounding error.

**Module switches.** The adapter reads each vault's current `module()`, so fees from a module retired mid window would be missed. `LogSetModule` fired zero times on ethereum in the 30 day window scanned.

**No `permitFailure` on the metadata reads**, deliberately, so an undercount surfaces as an error rather than silently shrinking the number. The one exception is the window start `managerFeePIPS` read, which has to tolerate a revert for vaults created during the window, where the in-window `LogSetManagerFeePIPS` supplies the rate instead.

### DIFF
--- a/fees/arrakis-modular/index.ts
+++ b/fees/arrakis-modular/index.ts
@@ -242,7 +242,14 @@ const breakdownMethodology = {
 
 const adapter: Adapter = {
   version: 2,
-  pullHourly: true,
+  // getLogs fans out one eth_getLogs per target, and this adapter reads two events across
+  // every module of every vault: 377 modules today, so 754 log requests per window. Hourly
+  // multiplies that by 24, to roughly 18k requests per day per run, and the shared public
+  // pool cannot absorb it. Only one or two ethereum endpoints in the pool serve a wide
+  // eth_getLogs at all, the rest cap at 25 or 50 blocks or do not support the method, so
+  // once those rate limit the whole run fails. Set back to true once the per module fan out
+  // is replaced by a single chain wide topic filtered scan.
+  pullHourly: false,
   fetch,
   adapter: {
     [CHAIN.ETHEREUM]: { start: "2024-08-16" },
--- a/fees/arrakis-modular/index.ts
+++ b/fees/arrakis-modular/index.ts
@@ -242,6 +242,7 @@ const breakdownMethodology = {
 
 const adapter: Adapter = {
   version: 2,
+  pullHourly: true,
   fetch,
   adapter: {
     [CHAIN.ETHEREUM]: { start: "2024-08-16" },
--- a/fees/arrakis-modular/index.ts
+++ b/fees/arrakis-modular/index.ts
@@ -1,0 +1,154 @@
+import { Adapter, FetchOptions } from "../../adapters/types";
+import { CHAIN } from "../../helpers/chains";
+import { METRIC } from "../../helpers/metrics";
+import ADDRESSES from "../../helpers/coreAssets.json";
+
+// ArrakisMetaVaultFactory, same address on every chain Arrakis Modular is deployed to.
+const FACTORY = "0x820FB8127a689327C863de8433278d6181123982";
+
+// Arrakis denominates the manager fee share in PIPS (1e6 = 100%).
+const PIPS = 1_000_000n;
+
+const NATIVE = "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee";
+
+const abis = {
+  numOfPublicVaults: "uint256:numOfPublicVaults",
+  numOfPrivateVaults: "uint256:numOfPrivateVaults",
+  publicVaults:
+    "function publicVaults(uint256 startIndex_, uint256 endIndex_) view returns (address[])",
+  privateVaults:
+    "function privateVaults(uint256 startIndex_, uint256 endIndex_) view returns (address[])",
+  module: "address:module",
+  token0: "address:token0",
+  token1: "address:token1",
+  managerFeePIPS: "uint256:managerFeePIPS",
+};
+
+// Every Arrakis LP module emits this whenever the manager fee is actually taken out of
+// the position, which happens on rebalance, on withdraw and on withdrawManagerBalance().
+// Amounts are always emitted in token0/token1 order (the module un-inverses them first).
+const MANAGER_FEE_EVENT =
+  "event LogWithdrawManagerBalance(address manager, uint256 amount0, uint256 amount1)";
+
+const toToken = (token: string) =>
+  token.toLowerCase() === NATIVE ? ADDRESSES.null : token;
+
+const fetch = async (options: FetchOptions) => {
+  const { api, createBalances, getLogs } = options;
+
+  const dailyFees = createBalances();
+  const dailyRevenue = createBalances();
+  const dailySupplySideRevenue = createBalances();
+  const result = {
+    dailyFees,
+    dailyRevenue,
+    dailyProtocolRevenue: dailyRevenue,
+    dailySupplySideRevenue,
+  };
+
+  const [numPublic, numPrivate] = await Promise.all([
+    api.call({ target: FACTORY, abi: abis.numOfPublicVaults }),
+    api.call({ target: FACTORY, abi: abis.numOfPrivateVaults }),
+  ]);
+
+  const [publicVaults, privateVaults] = await Promise.all([
+    numPublic > 0
+      ? api.call({ target: FACTORY, abi: abis.publicVaults, params: [0, numPublic] })
+      : [],
+    numPrivate > 0
+      ? api.call({ target: FACTORY, abi: abis.privateVaults, params: [0, numPrivate] })
+      : [],
+  ]);
+
+  const vaults: string[] = [...publicVaults, ...privateVaults];
+  if (!vaults.length) return result;
+
+  const modules: string[] = await api.multiCall({ abi: abis.module, calls: vaults });
+
+  const [token0s, token1s, feePIPS, logs] = await Promise.all([
+    api.multiCall({ abi: abis.token0, calls: modules }),
+    api.multiCall({ abi: abis.token1, calls: modules }),
+    api.multiCall({ abi: abis.managerFeePIPS, calls: modules }),
+    getLogs({
+      targets: modules,
+      eventAbi: MANAGER_FEE_EVENT,
+      entireLog: true,
+      parseLog: true,
+    }),
+  ]);
+
+  const moduleInfo: Record<string, { tokens: string[]; feePIPS: bigint }> = {};
+  modules.forEach((module: string, i: number) => {
+    moduleInfo[module.toLowerCase()] = {
+      tokens: [toToken(token0s[i]), toToken(token1s[i])],
+      feePIPS: BigInt(feePIPS[i]),
+    };
+  });
+
+  for (const log of logs) {
+    const info = moduleInfo[log.address.toLowerCase()];
+    if (!info) continue;
+
+    const amounts = [BigInt(log.args.amount0), BigInt(log.args.amount1)];
+
+    amounts.forEach((managerFee, i) => {
+      if (managerFee === 0n) return;
+      const token = info.tokens[i];
+
+      dailyRevenue.add(token, managerFee, METRIC.MANAGEMENT_FEES);
+
+      // The module computes the emitted amount as grossFees * managerFeePIPS / PIPS at the
+      // moment fees leave the position, so the gross figure is recovered by inverting that.
+      // The rate is read live per module rather than assumed, it is not uniform across
+      // vaults, and setManagerFeePIPS() flushes the outstanding balance at the old rate
+      // before changing it, so a rate change never straddles an unclaimed accrual.
+      // A zero rate cannot produce a non zero manager fee, but guard the division anyway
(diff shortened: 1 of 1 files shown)


BhariGowda:
Pushed the fix for the per-event fee rate point, and marking this draft, because verifying that fix turned up a second and much worse defect in what I opened. Details on both.

## Fixed: manager fee rate resolved per event

The review point was right and my original justification answered the wrong question. The flush guarantees each emitted amount is consistent with the rate in force when it was emitted; it does not guarantee that rate equals the one read at the window's end block.

Scanned every module from each factory's deployment block to head for `LogSetManagerFeePIPS`. `initialize()` never sets `managerFeePIPS`, so every module has at least one event and a genuine change is one with `oldFee != 0`. Ethereum 119 modules, 5 genuine changes. Base 218 modules, 9. Arbitrum 34 modules, 1. Three of those actually landed in the same window as a withdrawal, all of them the flush that `setManagerFeePIPS` itself triggers: `0xDd0685a4…` on 2025-10-29 and `0xe23fFe3b…` plus `0xDFdBFd55…` on 2025-05-12, all 1% to 50% changes, so gross was being divided by 50% instead of 1%.

Within the change transaction the flush is emitted first:

```
logIndex 135  LogWithdrawManagerBalance
logIndex 137  LogSetManagerFeePIPS
```

so merging both event streams per module and replaying them in `(blockNumber, logIndex)` order resolves each withdrawal to its own rate. Window start rate comes from `fromApi`, with the first in-window change's `oldFee` preferred where present, since that is authoritative and is also correct for a change landing on the window's first block.

## Not fixed, and the reason this is going to draft

Re-running the two corrected dates produced numbers that are far too large, and chasing that found a token mapping bug in the version I opened this PR with.

`LogWithdrawManagerBalance` amounts are **not** always in vault token0/token1 order. The beacon behind the UniV4 modules was upgraded at block 25393368, 2026-06-25 08:12 UTC, from `UniV4StandardModulePrivate` (`0x04eAd254…`) to `UniV4StandardModuleV2Private` (`0x2b0747Ec…`). The old implementation emits unconditionally in pool order:

```solidity
emit IArrakisLPModule.LogWithdrawManagerBalance(manager, managerFee0, managerFee1);
```

The new one swaps to vault order when the module is inversed:

```solidity
emit IArrakisLPModule.LogWithdrawManagerBalance(
    manager, rebalanceResult_.managerFee1, rebalanceResult_.managerFee0
);
```

The adapter maps amounts to `token0()`/`token1()` unconditionally, which is only correct after 2026-06-25. On ethereum 63 of the 111 UniV4 modules have `isInversed() = true`, so most of the backfill window from 2024-08-16 to 2026-06-25 assigns amounts to the wrong side of the pair. Where a pair is a long tail token against native ETH, that values a token amount at the ETH price.

Worked example, module `0xc687120b…` (BUBBLE / native) on 2025-10-29, block 23681372. The adapter reads `amount1 = 39397.36e18` against `token1() = 0xEeee…`, so native ETH, and prices it at $158M. The transaction contains no ETH movement at all, only:

```
BUBBLE      39,397.361701   PoolManager -> ArrakisStandardManager
BUBBLE   9,160,959.660939   module      -> PoolManager
```

The real fee was 39,397 BUBBLE. That single event is most of the $191M of "revenue" the adapter reports for 2025-10-29.

Same problem on the rate-change event I used to validate the fix. `0xDd0685a4…` on 2025-10-29 emits `amount1 = 18949538493064065680` against `token1() = native`, but the transaction shows the transfer was `18.949538 CFG`. So the corrected gross there is 1894.95 CFG, not 1894.95 ETH, and my earlier estimate of the correction size was wrong for the same reason.

Recent windows are unaffected, since they are past the upgrade. 2026-07-03 still reports fees 4769 and revenue 2385, unchanged before and after the rate fix and matching an independent hand decode to 0.9%. Today's run gives fees 19.93k and revenue 9.96k. The identity holds throughout. The problem is historical, and it is large.

Fixing it properly needs per-module `isInversed()` plus the beacon upgrade block per chain, applied as a date conditional so history is not corrupted, and the same ordering question checked separately for the `UniswapV3StandardModulePrivate` and `ValantisModulePublic` families, which have their own emit sites. That is more than a quick patch and I would rather get it right than push something else that reads plausible and is wrong. Marking draft until it is done and verified against on chain transfers rather than against the event alone.

BhariGowda:
Architectural suggestion for whoever picks this up, not something I think should block this PR either way.

## The shape problem

`getLogs` fans out one `eth_getLogs` per entry in `targets` (`logs.js`, `targets.map(...)`). This adapter has to read two events across every module of every vault, and there are 377 modules today across the four chains it covers, so a single window costs 754 log requests. That number grows with every vault Arrakis deploys, and it is already enough to be fragile: the ethereum pool only has one or two endpoints that will serve a wide `eth_getLogs` at all, since the rest cap at 25 or 50 blocks or answer `method not available`. When those rate limit, the run fails outright.

That is what pushed `pullHourly` to `false` in this PR. Hourly is correct in principle, the adapter is fully window scoped, but 24 slots means roughly 18k log requests per run and the shared pool will not carry it.

## What would actually fix it

A single chain wide `getLogs` per event with `noTarget: true` and a topic0 filter, then discarding logs whose emitting address is not in the known module set. Both events are Arrakis specific, so the topic0 filter is already close to exact and the post filter is only there to exclude a forked deployment emitting the same signature. That turns 754 requests per window into 2, makes the cost independent of vault count, and makes `pullHourly: true` cheap enough to be the obvious default.

I did not do it here because `.coderabbit.yaml` explicitly flags `noTarget` plus post filtering as an anti pattern, and that rule is right in the general case, where `targets` is the cheaper shape. This is the inverted case: `targets` is costing two to three orders of magnitude more than the chain wide scan would. If a maintainer is happy with the deviation for this adapter I am glad to write it, and it would let `pullHourly` go back to `true` in the same change.

## One thing worth knowing if anyone tries the cheaper version of this

The obvious smaller optimisation, keeping `targets` but merging the two events into one call with a topic0 array, `topics: [[topicA, topicB]]`, does not work. It silently returns fewer logs. On the 2026-07-03 ethereum window the merged call returns 28 logs where two separate calls return 32, and the loss is deterministic: it reproduces across runs, reproduces against a single host pool, and is window dependent, so the 2025-10-29 window returns 58 either way and hides it entirely. Worth flagging since the failure is silent rather than an error, and a merged call that looks fine on one test date can be quietly dropping fees on another.

bheluga:
@BhariGowda thanks for the PR

PR #8449, opened 2026-07-27
metric: add metric-v1 adapter

horror-bun:
## What

Adds the **Metric v1** DEX as a separate version adapter (`dexs/metric-v1`) alongside the existing legacy adapter (`dexs/metric`).

## Structure

Both versions share one builder to avoid duplication (aerodrome-style shared file in the base folder):

- `dexs/metric/utils.ts` — `getMetricAdapter(factory, chainConfig, events)` builds the pool-discovery + volume pipeline once.
- `dexs/metric/index.ts` — legacy factory `0xe22F…`, default legacy event ABIs (behaviour unchanged).
- `dexs/metric-v1/index.ts` — v1 factory `0x6229…` on Ethereum, Base, Arbitrum, Robinhood.

## Why events are injected

The v1 factory emits different events from legacy, so the ABIs are passed per version:

- `Swap`: `int256` amounts (legacy `int128`), `int8 newTick`, indexed `sender`/`recipient`, extra `protocolFeeAmount`.
- `PoolCreated`: pool address is the indexed `poolAddress` field (legacy `pool`), tokens indexed, plus additional fields.

## Methodology

Unchanged: Volume = sum of input token amounts from `Swap` events across all pools created by the factory, discovered on-chain from its `PoolCreated` event.

## Testing

`npm test dexs metric-v1` runs cleanly and reports non-zero volume on Base and Robinhood.


bheluga:
@horror-bun thanks for the PR

PR #8450, opened 2026-07-27
stakewise: add Gnosis chain and use on-chain fee amounts from event f…

uwezukwechibuzor:
## Summary
Add Gnosis chain support and use on-chain fee amounts directly from event fields instead of hardcoded percentages. https://docs.stakewise.io/contracts/networks/


bheluga:

PR #8452, opened 2026-07-27
fix(o1-launchpad): stream Base logs and add current RWA fee suites

yasindunethmina:
### Why this is needed

After the initial adapter was deployed, we noticed that some historical fees and revenue were missing or understated, most visibly on July 16-17.

Our investigation found that the adapter forced direct RPC log fetching. During high-volume Base windows, an RPC provider could return an incomplete set of logs without reporting an error, leaving some `Trade` and `Credited` events out of the stored totals.

This update uses DeFiLlama's streaming indexer to retrieve the complete indexed log set in batches, while keeping an uncached hourly RPC fallback if streaming is temporarily unavailable.

### What changed

- Use DeFiLlama's streaming indexer for complete Base log retrieval, with an uncached hourly RPC fallback.
- Add the new Base and Robinhood RWA factories, hooks, and fee escrows.
- Attribute standard launch fees to the quote token and RWA launch fees to the native gas token.
- Preserve strict transaction-level reconciliation between `Trade` and `Credited` events.
- Clarify that no fees are distributed to token holders.

### Validation

- TypeScript and adapter CLI checks pass.
- All hourly windows pass for July 16, July 17, and July 26.
- Fee splits were reconciled and compared with the production Convex records.

### Backfill request

After deploying this update, please refill `o1-launchpad` from **July 1, 2026 UTC through the deployment date** for Base and Robinhood Chain.

Please refill fees, user fees, revenue, protocol revenue, supply-side revenue, holders revenue, and all label/chain breakdowns so the incomplete high-volume intervals and historical RWA activity are recalculated with the updated adapter.

### Additional metadata request

If possible, please also rename the parent protocol from `o1.exchange Trading Aggregator` to `o1.exchange Trading Terminal` and update its icon to:

`https://icons.llamao.fi/icons/protocols/o1-launchpad`

Screenshot 2026-07-26 at 7 30 06 PM

### DIFF
--- a/fees/o1-launchpad/index.ts
+++ b/fees/o1-launchpad/index.ts
@@ -287,12 +287,6 @@ const reconcileSuite = (
   return allocations;
 };
 
-const DIRECT_LOG_FETCH_OPTIONS = {
-  // The adapter runs hourly; avoid reusing a partial cached response in the RPC fallback.
-  skipIndexer: true,
-  skipCache: true,
-} as const;
-
 const PARSED_LOG_FETCH_OPTIONS = {
   entireLog: true,
   parseLog: true,
@@ -333,7 +327,6 @@ const fetch = async (options: FetchOptions): Promise<FetchResultV2> => {
     ? await options.getLogs({
       targets: launchFeeSuites.map((suite) => suite.factory),
       eventAbi: LAUNCH_FEE_PAID_EVENT,
-      ...DIRECT_LOG_FETCH_OPTIONS,
       flatten: false,
     }) as LaunchFeeArgs[][]
     : [];
--- a/fees/o1-launchpad/index.ts
+++ b/fees/o1-launchpad/index.ts
@@ -203,7 +203,7 @@ const dedupeLogs = <TArgs>(logs: DecodedLog<TArgs>[]) => {
   return [...unique.values()];
 };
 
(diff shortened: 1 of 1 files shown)


bheluga:
@yasindunethmina thanks for the PR
Why do you need strict error handling wrt lengths, it almost throws error everytime

yasindunethmina:
> @yasindunethmina thanks for the PR Why do you need strict error handling wrt lengths, it almost throws error everytime

Thanks! I added the length checks to catch partial RPC results and avoid calculating the fee split incorrectly. But you’re right they can be too strict since one incomplete transaction could fail the whole hour. I’ve simplified it now so incomplete transactions are skipped while valid ones are still verified and processed.

Once merged, could you also refill the affected historical period as requested in the PR description?

bheluga:
@yasindunethmina thanks for the PR.
Your requested change should be done within few hours.
If not done even after 24 hours, please feel free to ping me.

bheluga:
@yasindunethmina rn there seems to be some issue with robinhood rpc, will try refilling later

yasindunethmina:
> @yasindunethmina rn there seems to be some issue with robinhood rpc, will try refilling later

ok sounds good!

bheluga:
@yasindunethmina all good now? or anything missing from our end?

yasindunethmina:
> @yasindunethmina all good now? or anything missing from our end?

yes, looks good! thank you 🙏

PR #8454, opened 2026-07-27
[fix] kava swap - rebuild on onchain swap events, stats api is dead

### LINKED ISSUE #8431: fix kavaswap volume
https://defillama.com/protocol/kava-swap

0xshubhs:
fixes #8431

builds on the diagnosis in the issue thread (credit @BhariGowda): swap-data.kava.io is 503 at the infra level on every path, no alternate host exists in the app bundle, but the protocol itself is alive and trading. so this replaces the dead stats api with the chain itself: tx_search over the swap module's two message types, summing from swap_trade events per day. the old adapter was runAtCurrTime and only ever reported a live snapshot - this gives real per-day windows for the first time.

on the two open questions from the thread:

**leg counting**: one leg per trade. thats the usual dex volume convention, and it lands inside the old series' $8k-27k/day band (07-26 came out at $8.95k) where both-legs would sit at 2x. the $17.7k/day estimate in the thread was a busy-morning sample extrapolated x24, so no contradiction there.

**decimals**: never hardcoded - each denom is mapped with its own decimals (usdx/ukava/swp/ibc are 6, busd/xrpb/btcb/bnb are 8), matching what coins.llama.fi reports as authoritative.

one wrinkle found while validating: usdx and swp no longer price on coins.llama.fi at all (coingecko still quotes usdx at $0.66 - depegged, so a $1 assumption would also be wrong). since both legs of a trade are equal in value minus fee, the adapter counts the input side when it has a reliable price and falls back to the equal-value output side otherwise - same idea as the recent uniswap-v4 pricing fix. without this, usdx-quoted trades (the biggest share) would silently drop and the day undercounts by ~2x.

verified against an independent recount: 192 swap_trade events on 2026-07-26, hand-priced per denom = ~$8.9k vs adapter $8.95k.

note on backfills: rpc.kava.io prunes to ~3 weeks (earliest block currently 2026-07-08), start is set to 2026-07-09 and the block-range helper throws on pruned days instead of silently undercounting. an archive rpc can be dropped into the config if deeper refills are ever wanted.

### DIFF
--- a/dexs/kava-swap/index.ts
+++ b/dexs/kava-swap/index.ts
@@ -1,33 +1,85 @@
-import { ChainBlocks, FetchOptions, FetchResult, SimpleAdapter } from "../../adapters/types";
+import { FetchOptions, SimpleAdapter } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
+import { CosmosChainMetricConfig, getBlockRangeForTimestamps } from "../../helpers/cosmosChainFees";
 import fetchURL from "../../utils/fetchURL";
 
-interface ICallPoolData {
-  denom: string;
-  amount: string;
-}
+const config: CosmosChainMetricConfig = {
+  chain: CHAIN.KAVA,
+  rpcs: ["https://rpc.kava.io"],
+  denoms: {},
+};
 
-const URL = "https://swap-data.kava.io/v1/pools/internal";
+const denomConfigs: Record<string, { cgToken: string; decimals: number }> = {
+  ukava: { cgToken: "kava", decimals: 6 },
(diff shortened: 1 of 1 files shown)


0xshubhs:
addressed the two real ones: txs are now deduped by hash across the two action queries, and pagination fails closed (throws on rpc error / missing total_count / stalled pages) instead of silently under-reporting. page size is a named constant now too.

left the cadence daily on purpose rather than adding pullHourly - this thing does ~200 swaps a day on a public rpc that prunes to 3 weeks, hourly slices would just 24x the block binary searches for no fidelity gain. 07-26 still verifies at $8.95k after the changes.

bheluga:
@0xshubhs thanks for the PR

PR #8457, opened 2026-07-27
add plume chain to bim

noeroumeas:


bheluga:
@noeroumeas thanks for the PR

PR #8458, opened 2026-07-27
feat: add OwlOrderFi volume adapter (Base, Polygon)

blueradu:
Adds a `dexs/` volume + fees adapter for **OwlOrderFi**, a non-custodial limit / stop / TWAP / DCA order router for Uniswap V3.

Makers sign EIP-712 order intents off-chain; an authorized keeper executes them on-chain once the signed conditions are met. The router never holds user funds between transactions, so there is no TVL to track — this is a volume-tracked protocol, in the same shape as the other order-router listings.

### Metadata

| Field | Value |
|---|---|
| Protocol name | OwlOrderFi |
| Category | DEX Aggregator |
| Chains | Base (8453), Polygon (137) |
| Website | https://owlorderfi.com |
| Twitter | https://x.com/owlorderfi |
| GitHub | https://github.com/owlorderfi/protocol |
| Logo | https://owlorderfi.com/icon-512.png |
| Token | None |
| License | BSL 1.1 |

### Contracts read

| Chain | Current router | Previous router |
|---|---|---|
| Base | [`0xC11fC636…d5aA`](https://basescan.org/address/0xC11fC6362d29D6831B7431F3B7d5F9a6128bd5aA) | [`0x30e744fB…A259`](https://basescan.org/address/0x30e744fB8120C3f6f8CF958e5110E6349fAaA259) |
| Polygon | [`0x42cB3F5D…f52B`](https://polygonscan.com/address/0x42cB3F5D39d11193cC9436b155F094Aa7cbcf52B) | [`0x30e744fB…A259`](https://polygonscan.com/address/0x30e744fB8120C3f6f8CF958e5110E6349fAaA259) |

Both current routers are source-verified on their explorers. The previous address is included so history stays continuous across a redeploy on 2026-07-27.

### Methodology

- **Volume** — the input leg of each `OrderExecuted` log. Only one side is counted; adding `tokenOut` would double-count the same trade, and `amountOut` is already net of the protocol fee.
- **Fees** — the protocol fee, deducted from the output token. The rate is signed by the maker as part of the order and capped on-chain at 1%.
- **Revenue** — all of the fees. The router has no liquidity providers of its own, so supply-side revenue is zero. Uniswap's pool fees are not counted here.

One filter worth flagging: the previous router signalled a cancellation by reusing `OrderExecuted` with `orderType = 255` and every amount zeroed. Counting one would register a cancel as a fill and push a zero token address into the balance lookup, so the adapter skips them. The current router emits a dedicated `OrderCancelled` event instead.

### Verification

```
npm test dexs owlorderfi 2026-06-07   ->  polygon  volume 58.00  fees 0.17
npm test dexs owlorderfi 2026-06-03   ->  base     volume 10.00  fees 0.03
```

Both figures were cross-checked against amounts decoded independently from the raw logs; the fee numbers match to the cent.

This is a small, recently launched protocol — lifetime volume is roughly $73 across seven executed orders. Flagging that up front so the scale is not a surprise.

### DIFF
--- a/dexs/owlorderfi.ts
+++ b/dexs/owlorderfi.ts
@@ -1,5 +1,6 @@
 import { SimpleAdapter, FetchV2 } from "../adapters/types";
 import { CHAIN } from "../helpers/chains";
+import { METRIC } from "../helpers/metrics";
 
 /**
  * OwlOrderFi — non-custodial limit / stop / TWAP / DCA order router for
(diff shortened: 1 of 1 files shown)


blueradu:
Thanks — three of the four are in ([`491d97c`](https://github.com/DefiLlama/dimension-adapters/commit/491d97c)). Declining the fourth, with a reason.

**Applied**

- **Single multi-target query.** Switched to one `getLogs({ targets, flatten: true })`. The per-router loop was me being defensive about the shape of a multi-target result; passing `flatten` explicitly removes the ambiguity, and it matches how the rest of the repo queries several contracts at once.
- **Fee label + breakdown.** Fees now carry `METRIC.PROTOCOL_FEES`, with a matching `breakdownMethodology` for `Fees`, `Revenue` and `ProtocolRevenue`.
- **`pullHourly: true`.**

**Not applied — passing `config` directly as `adapter`**

This one would break at run time rather than at build. `cli/buildModules.ts` strips every per-chain key that is not in `whitelistedBaseAdapterKeys`:

```js
if (!whitelistedBaseAdapterKeys.has(key)) {
  delete obj[key]
}
```

`config[chain]` carries `routers`, which is not whitelisted. Passing `config` as `adapter` means that `delete` mutates the very object `fetch` closes over, so `config[chain].routers` becomes `undefined` once the module is built — no build error, then a failure on the first fetch. Keeping the adapter map separate from the router config is deliberate; I have added a comment saying so, since it does otherwise look like redundant duplication.

**Re-verified after the change**

```
npm test dexs owlorderfi 2026-06-07  ->  polygon  volume 58.00  fees 0.17  (Protocol Fees 0.1725)
npm test dexs owlorderfi 2026-06-03  ->  base     volume 10.35  fees 0.03  (Protocol Fees 0.0308)
```

The Base figure is now 10.35 against 10.00 before, which matches the raw logs for that day exactly (3 + 3 + 4.34647 USDC). The fee figures were identical across both runs, so the shift looks like price-feed timing rather than counting.

bheluga:
@blueradu thanks for the PR

PR #8467, opened 2026-07-28
Add AGNT (my.agnt.social) launchpad fees adapter

tut9492:
## AGNT — launchpad fees adapter

Adds a fees/revenue adapter for **AGNT** (my.agnt.social), an AI-agent platform whose token launchpad launches tokens via **Doppler V4 on Base**.

**Metric:** `fees` · **Chain:** Base · **Data source:** 100% on-chain — `addTokensReceived` (ERC-20 Transfer logs), filtered to the Doppler initializers so only fee releases count. No external APIs.

### Methodology
- Terminal pool fee **1.095%**, split creator **63%** / platform **32%** / Doppler **~5%**.
- Platform's 32% share is released as WETH to `0x5bF5805e4809A447a61621f8698CEdeA2D1fC5f0` by the Doppler initializers (`0xd59ce43e…`, `0xbdf93814…`).
- `dailyRevenue` = platform WETH share; `dailyFees` = revenue / 0.32; `dailySupplySideRevenue` = creator (63%) + Doppler (5%).
- Conservative: only the WETH leg is counted (launched-token leg excluded → lower bound).

### Verification
- `npm test fees agnt` → e.g. 2026-07-15: dailyFees 434 / dailyRevenue 139 / dailySupplySideRevenue 295 (32/68 split consistent).

Start `2026-07-15` (first fee release to this beneficiary). Thanks @bheluga for the breakdown + doublecounted flag. In-app **trading fees** (0.40% Relay app-fee) will follow in a separate PR — they're a non-double-counted surcharge, so they don't belong under this adapter's `doublecounted` flag.

PR #8477, opened 2026-07-29
AGNT: add trading-fee stream (follow-up to #8467)

tut9492:
## AGNT — add trading-fee stream (follow-up to #8467)

Follow-up to the merged AGNT launchpad adapter (#8467). Adds AGNT's second on-chain fee stream: the **0.40% platform fee on in-app swaps**.

**Mechanism:** the swap fee is accrued off-chain by Relay (app-fees) and **withdrawn as USDC** to the same fee wallet (`0x5bF5805e…`) from Relay's app-fee claim contract (`0xf70da978…`). The adapter counts USDC into the fee wallet **filtered to that Relay claim sender**, so prefunding/operational inflows are excluded. AGNT keeps 100% of this fee → `dailyFees` == `dailyRevenue` for the trading portion (no supply side).

**Data source:** 100% on-chain — `addTokensReceived` (ERC-20 Transfer logs), `fromAddress`-filtered. No external APIs.

**Verification:** `npm test fees agnt` runs clean, launchpad numbers unchanged (e.g. 2026-07-15 → 434/139/295). Trading is claim-gated (Relay accrues off-chain, on-chain only on withdrawal), so it reads 0 on days with no claim — correct. Verified the claim path on-chain: USDC from `0xf70da978…` → fee wallet.

**Note on `doublecounted`:** kept `true` (the launchpad leg is Uniswap V4 pool fees). The trading fee is a separate app surcharge that is *not* double-counted, but it's negligible vs launchpad at current scale — happy to split into a per-metric treatment if you'd prefer.

### DIFF
--- a/fees/agnt/index.ts
+++ b/fees/agnt/index.ts
@@ -31,6 +31,10 @@ const DOPPLER_INITIALIZERS = [
 const PLATFORM_SHARE = 0.32;
 const CREATOR_SHARE = 0.63;
 const DOPPLER_SHARE = 0.05;
+// Trading: 0.40% in-app swap fee, accrued off-chain by Relay and withdrawn to the SAME
+// fee wallet as USDC from Relay's claim contract on Base. Filtered to that sender so only
+// fee withdrawals count (no prefunding/ops inflows). AGNT keeps 100% of this fee.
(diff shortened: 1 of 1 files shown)


bheluga:
@tut9492 thanks for the PR

PR #8495, opened 2026-07-29
AGNT: add Robinhood Chain + fix trading-fee leg

tut9492:
## AGNT — add Robinhood Chain + fix trading-fee leg

The AGNT launchpad fees adapter (`fees/agnt/index.ts`) was reporting ~$0 all-time. Two reasons, both fixed here:

### 1. Base-only → missing Robinhood Chain (the launch volume)
AGNT's launch volume moved to **Robinhood Chain** (4663), which the adapter didn't cover.
- Makes `fetch` chain-aware and adds `CHAIN.ROBINHOOD`: WETH received by the launch-fee wallet `0x5bF5805e…` from the two RH Doppler initializers (`0x4e3468951d…`, `0x6f02324d…`), using RH's WETH `0x0bd7d308…`.
- Same 32% platform / 63% creator / 5% Doppler model + WETH-lower-bound methodology as Base.

### 2. Trading-fee leg pointed at the wrong wallet
It counted USDC to the launch treasury (`0x5bF5805e`) from Relay's claim contract — but the 0.40% app fee actually lands on a **dedicated recipient** `0x585b6854…`, mostly as **WETH** straight from Relay's Base settlement contract (`0xc7F712…`, ~48 transfers), plus occasional USDC. On-chain-verified. Now targets the right wallet and counts WETH + USDC from both Relay senders (whitelisted to exclude stray inflows).

### Notes
- Robinhood Chain is already supported on DeFiLlama (chainId 4663). RH WETH is ETH-pegged, so USD pricing should resolve like Base WETH.
- Solana launch fees use a non-EVM mechanism — separate follow-up.
- All addresses verified on-chain (fee inflows only).

### DIFF
--- a/fees/agnt/index.ts
+++ b/fees/agnt/index.ts
@@ -31,10 +31,14 @@ const PLATFORM_FEE_WALLET = "0x5bF5805e4809A447a61621f8698CEdeA2D1fC5f0"; // CLA
 const PLATFORM_SHARE = 0.32;
 const CREATOR_SHARE = 0.63;
 const DOPPLER_SHARE = 0.05;
-// Trading: 0.40% in-app swap fee, accrued off-chain by Relay and withdrawn to the
-// fee wallet as USDC from Relay's claim contract on Base. Filtered to that sender so
-// only fee withdrawals count (no prefunding/ops inflows). AGNT keeps 100%. Base only.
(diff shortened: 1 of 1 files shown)


bheluga:
@tut9492 thanks for the PR

PR #8508, opened 2026-07-31
smardex: dont call getLogs when every pair gets filtered out

0xshubhs:
follow-up to #8493; the last ci run there showed `Error: target|targets is required or set the flag "noTarget" to true`: when filterPools drops every pair on a chain (possible on the dust chains, pooled value threshold is $200), getLogs gets an empty targets array and throws. return the empty balances instead. rechecked ethereum and polygon on 2026-07-21, values unchanged


bheluga:
@0xshubhs thanks for the PR

PR #8511, opened 2026-07-31
sushiswap-v3: protocol fees no longer reach xSUSHI

DaniiRix:
`dexs/sushiswap-v3.ts` booked 38% of every chain's protocol fee as `dailyHoldersRevenue` to xSUSHI stakers. That split is no longer accurate. On-chain verification shows protocol fees no longer flow to xSUSHI, so the entire protocol fee should be classified as `dailyProtocolRevenue`.

This PR updates the revenue classification without changing volume, fees, or total revenue.

### Remove the 62/38 holders split

The adapter previously assumed:

- 62% → protocol
- 38% → xSUSHI holders

On-chain tracing shows this is no longer the actual flow.

- The transfers to `0xac4c6e…` are inputs to **RedSnwapper**, Sushi's generic swap executor, not payouts to SushiBar.
- `TokenChwomper.snwap()` pre-funds RedSnwapper before executing swaps.
- Traced consecutive swaps return USDC to the fee collector—not SUSHI and not SushiBar.
- The observed 62/38 ratio is simply the accounting of swap inputs/outputs rather than a holders distribution.

### xSUSHI has stopped accruing protocol fees

Historical state confirms xSUSHI stopped receiving value after TokenChwomper became the maker contract.

| Date | SUSHI / xSUSHI |
| --- | ---: |
| 2025-10-01 | 1.5560 |
| 2025-11-01 | 1.5749 |
| 2026-07-27 | 1.5750 |

Only a **0.0036%** increase over roughly seven months on an 18.2M SUSHI base.

### Cross-chain verification

Current `V3Manager.maker` addresses resolve to protocol-controlled multisigs:

- Ethereum → Sushi Operations multisig
- Polygon → local Gnosis Safe
- Base → local Gnosis Safe
- Optimism/BSC → no meaningful outflows observed

No chain currently routes protocol fees to xSUSHI holders.

As a result:

- `dailyProtocolRevenue = dailyRevenue`
- `dailyHoldersRevenue = 0`

### DIFF
--- a/dexs/sushiswap-v3.ts
+++ b/dexs/sushiswap-v3.ts
@@ -7,52 +7,62 @@ const sushiV3Configs: Record<string, { factory: string, start: string }> = {
   [CHAIN.ARBITRUM_NOVA]: { factory: "0xaa26771d497814e81d305c511efbb3ced90bf5bd", start: "2023-04-03" },
   [CHAIN.XDAI]: { factory: "0xf78031cbca409f2fb6876bdfdbc1b2df24cf9bef", start: "2023-04-03" },
   [CHAIN.BASE]: { factory: "0xc35dadb65012ec5796536bd9864ed8773abc74c4", start: "2023-08-03" },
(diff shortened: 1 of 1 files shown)


bheluga:
@DaniiRix thanks for the PR

Issue #8521, opened 2026-07-31
17 protocols report zero for 24h while active over 7d — two verified as hard stops


quepasaquepasa:
Some adapters appear to have stopped reporting while still being listed, and the
published series make it possible to spot them without touching any upstream API.

### What the published data shows

From `api.llama.fi/overview/dexs` today: of 1289 protocols, 491 report zero for
`total24h`. Almost all of those are simply discontinued — 425 report zero for
`total7d` as well, which is what a shut-down protocol looks like and is not
interesting.

The remaining **17 report zero for the last 24h while `total7d` is non-trivial**,
which is the shape a freshly-broken adapter makes. Checking the daily series for
the largest ones:

**XRPL DEX** (`/summary/dexs/xrpl-dex`)

```
2026-07-26: $2,147,197
2026-07-27: $1,758,969
2026-07-28: $5,755,130
2026-07-29: $4,626,319
2026-07-30: $0
```

**Lotus Finance** (`/summary/dexs/lotus-finance`)

```
2026-07-28: $3,089,861
2026-07-29: $2,533,965
2026-07-30: $0
2026-07-31: $0
```

Both went from millions per day to a hard zero, with no taper. 798 protocols do
report data for the same days, so this isn't indexing lag across the board.

### Caveat, because the heuristic is not a verdict

It flags candidates, not confirmed breakage. Checking a third one:

**AlphaSec Spot**

```
2026-07-27: $0
2026-07-28: $0
2026-07-29: $2,878,196
2026-07-30: $0
```

That is a protocol trading intermittently, not an adapter that died — the same
"zero today, active this week" pattern with an entirely ordinary explanation. So
the 17 are worth a look, not an automatic bug report. I haven't verified the
remaining ones against their chains, so I'm not claiming which adapters are at
fault.

The full list of 17: XRPL DEX, Lotus Finance, AlphaSec Spot, Defa By InvoiceMate,
b402, FlowX V3, Ready Card, Overpass, Web3.world, flowr, and 7 smaller ones.

### Why raise it

Adapters that quietly stop returning data seem to be a recurring theme here — the
open issues about dead sources (#8434 for stabble's removed API, #8429 for
ramses) were each found by hand after the fact. The distinction that makes it
checkable automatically is the one above: *zero with a non-zero baseline* is a
regression, *zero with a zero baseline* is a protocol that shut down.

I wrote a small tool that applies exactly that rule to a published feed and
[...]

PR #8525, opened 2026-07-31
xrpl-dex: throw when the dune row isnt there instead of writing zero

0xshubhs:
part of #8521 (credit to quepasaquepasa for spotting the hard stop); xrpl dex showed $4.6m on 07-29 then a hard $0 on 07-30 with no taper.

the adapter queries one row from `xrpl.aggregated_metrics_daily` on dune and returns 0 when the row is missing. that table publishes with lag, so any day the runner fires before the row lands gets a permanent $0 stored instead of a retry. xrpl itself obviously didn't stop trading.

fix: throw on the empty result so the runner retries later, like other single-row dune adapters. couldn't run the harness locally (no dune key), typecheck only; ci run will exercise it. the stored $0 days need a refill once this lands.

### DIFF
--- a/dexs/xrpl-dex/index.ts
+++ b/dexs/xrpl-dex/index.ts
@@ -9,8 +9,12 @@ const fetch = async (options: FetchOptions) => {
     const query = `select dex_xrp_pair_volume_xrp,amm_xrp_volume_xrp from xrpl.aggregated_metrics_daily where date = Date('${formattedDate}')`;
     const queryResults = await queryDuneSql(options, query);
 
-    const dexVolumeXrp = queryResults.length > 0 ? queryResults[0].dex_xrp_pair_volume_xrp : 0;
-    const ammVolumeXrp = queryResults.length > 0 ? queryResults[0].amm_xrp_volume_xrp : 0;
+    if (!queryResults.length) {
+        throw new Error(`no row in xrpl.aggregated_metrics_daily for ${formattedDate} yet`);
+    }
+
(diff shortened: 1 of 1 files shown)


bheluga:
@0xshubhs thanks for the PR

BhariGowda:
Low urgency, just flagging something this touches that may need a follow-up action.

The second commit in this PR ("use yesterday's date instead of day before") changes which Dune row gets read, not only when the adapter throws. Swapping

```ts
const date = new Date(options.fromTimestamp * 1000);
const formattedDate = date.toISOString().split("T")[0];
```

for `options.dateString` moves the query forward one day. For a daily window reporting day D, `runAdapter` sets `toTimestamp = timestamp - 1` and `fromTimestamp = toTimestamp - 86400`, so `fromTimestamp` lands at one second before midnight on D-1 and the old expression yields `"D-1"`. `dateString` is derived from `startOfDay`, which is computed from `toTimestamp`, so it yields `"D"`.

The new alignment looks like the right one to me. The part worth a decision is what happens to what is already stored: every point the adapter has published since the `2025-03-19` start was written under the old alignment, so the historical series is shifted a day relative to what it now reports, and it will not self-correct going forward. There is also a one-off discontinuity at the changeover, where the last old-alignment point and the first new-alignment point sit next to each other.

If a refill is cheap for this adapter it would put the whole series on one alignment. If not, it is at least worth knowing the older points are labelled a day late. Either way nothing is broken right now, and the throw itself is a clear improvement over silently writing zero.

PR #8553, opened 2026-08-03
hookos: rebuild fees on the contracts' own fee events, add Robinhood and Stable

OxForged:
**This is an update to an existing listing, not a new protocol.** HookOS was
listed in #7522 (`957e7f9c4`), so the new-protocol form below the line in the PR
template does not apply. Listing metadata is unchanged; nothing needs to go to
metadata@defillama.com.

- Website: https://hookos.fun
- Docs: https://docs.hookos.fun/contracts
- Twitter: https://x.com/hookosfun
- Telegram: https://t.me/HookOSPortal
- Discord: https://discord.com/invite/hHrGU6FFwJ

Files touched: `fees/hookos.ts`, `dexs/hookos.ts`. No other files, no
`package.json` / lockfile changes.

---

## What changed and why

The adapter merged in #7522 priced fees by **reading admin-settable rates and
multiplying them by trade size** — `BondingCurve.protocolFeeBps`,
`Arena.protocolFeeBps`, `TokenFactory.launchFee`, `HookRegistry.registrationFee`
— across four contracts on four chains, with hardcoded fallbacks when a read
failed.

Two problems with that:

1. **It was wrong where it did apply.** Launch and registration fees are
   USD-pegged on-chain (`TokenFactory.effectiveLaunchFee`), so `flat rate ×
   count` cannot reproduce what was actually paid. Sell-side curve fees were
   also back-solved from net proceeds rather than read.
2. **It missed most of what HookOS charges.** Everything outside those four
   contracts — creator fees, hook-author revenue, launch LP fees, the v4 hook
   cuts, the $HOOK pool tax, the stock pool tax, presales, bot trading, the
   graduated-LP splitter, the app layer — was simply not counted.

This PR replaces the rate-times-size reconstruction with **fee events read at
source**. Nothing in the adapter is now derived from a rate; every number comes
out of an event the contract emitted, in the currency the contract paid.

HookOS monetises through two patterns and the adapter covers both:

**1. Routed.** Some modules push their native protocol fee to `FeeRouter`, whose
`receive()` emits `FeeReceived(address indexed from, uint256 amount)`. Because
`from` is indexed, one query per chain covers bonding-curve trades, token
launches, hook registrations, arena battles, tournaments and the graduation
hooks, each attributed to the module that paid it via a per-chain `feeSources`
map. This is also what makes the USD-pegged launch and registration fees exact.

**2. Direct.** Most modules never touch `FeeRouter` — they pay a treasury inline
or accrue to a pull ledger — so each is read at its own event: bonding-curve and
HookPool creator fees, `HookRevenueVault` hook-author revenue (both the native
and ERC-20 paths), `HookOSV3FeeVault` launch LP fees, `HookWethTax` ($HOOK v4
pool tax), `StockTaxHook` (tokenized-stock pool tax), the v4 hook cuts taken by
`HookOSV4Hook` / `LaunchHook` / `UniversalGraduationHook`, `PresaleVault`,
`BotTradeRouter`, `LPFeeSplitter`, and the app-layer modules.

### Also in this change

- **Two new chains.** Robinhood (Quick Launch via `RHLaunchpad` + `LaunchHook`,
  and the $HOOK v4 pool) and Stable (no core deployment — its only live fee
  source is the V3 launch stack). Fees now covers seven chains, dexs six.
- **Holders revenue is now reported.** `dailyHoldersRevenue` = the launch-LP leg
  spent buying back and burning $HOOK, plus the stock pool tax routed to the
  vault that buys tokenized stocks for holders. `dailyRevenue` = fees − supply
  side; `dailyProtocolRevenue` = revenue − holders.
- **Supply side is split properly** into creator fees and hook-author revenue,
  rather than one slice reconstructed from `creatorFeeBps / protocolFeeBps`.
- **dexs adds HookPool**, the in-house AMM and the graduation target on HyperEVM,
  enumerated from `PoolFactory` and counted **on the input leg only** so each
  swap is counted once. Uniswap venues remain excluded.
- **Quote currencies are read, not assumed.** `HookOSV3FeeVault.positions()
  .pairToken` decides whether the `ethTo*` legs are native or the pair token (on
  a $HOOK-paired launch the identical event fields are $HOOK amounts, and the
  event carries nothing to disambiguate). `StockTaxHook.usdg()` is read live: the
  slot is *named* `usdg`, but the live Robinhood deployment points it at 18dp
  WETH — assuming 6dp USDG would have overstated that line by 1e12.

### Double-count guards

- The **v4 PoolManager** settles the hook cuts on the hooks' behalf, so it shows
  up as a `FeeReceived` sender for fees already counted at the hook. It is
  skipped — but **only on chains where at least one hook is actually watched**,
  because on a chain with no watched hook, skipping it would drop fees nothing
  else counts.
- **`FeeRouter.unwrapAndDistribute()`** withdraws WETH into `FeeRouter`, which
  trips `receive()` and re-emits `FeeReceived` for hook fees already counted at
  the hook. The wrapped-native contract is therefore never treated as a fee
  sender.
- The **$HOOK Uniswap-v4 pool's swap volume** is excluded from `dexs/hookos.ts`
  (see limitations below).

## Verification

- **Addresses.** Every address in both files was verified against the protocol's
  canonical deployment registry
  (`protocol/contracts/deployments/addresses.json`), keyed by chain id.
  Contracts absent on a chain are absent from that chain's config and are
  skipped at runtime rather than defaulted. Note that the same address can hold
  *different roles on different chains* — deployment is deterministic (same
  deployer + nonce ⇒ identical CREATE address), so e.g.
  `0x9B3d636C27AD4CDEBFbE1F182B2b63F66Be7adE5` is TokenFactory on Base and Arena
  on HyperEVM. Every contract is identified by `(chain, address)`, never by
  address alone.
- **ABIs.** All 27 ABI signatures across the two adapters — 26 events plus the
  `HookOSV3FeeVault.positions(address)` view — were byte-verified against the
  deployed Solidity: parameter types, order, and `indexed` flags all match the
  emitting contract. No raw topics anywhere; every log query uses a
  human-readable `eventAbi` and an explicit `targets` list (no `noTarget`).

## Validation

Robinhood, window of **2026-07-13**, checked line by line against raw logs:

| Line | USD |
|---|---:|
| HOOK Pool Tax | 3,643 |
| Graduated LP Fees | 375 |
| Token Launch Fees | 8.04 |
| Creator Fees | 0.9613 |
| Bonding Curve Fees | 0.0587 |
| **dailyFees** | **4,027** |
| dailySupplySideRevenue | 113 |

Each component was independently reconstructed from the raw event logs for that
window and reconciled against the adapter's output. The result also demonstrates
why the rework matters: under the previous adapter this day would have reported
roughly the Bonding Curve + Token Launch lines only, i.e. well under 1% of the
fees actually charged, and Robinhood would not have been covered at all.

Local checks:

```
npm run ts-check
# the date argument is the END of the 24h window, so 07-14 covers 07-13
npm test fees hookos.ts 2026-07-14
npm test dexs hookos.ts 2026-07-14
```

## Known limitations

These are deliberate and stated up front, not oversights:

- **~12 deployed contracts route natively to `FeeRouter` but are not individually
  mapped in `feeSources`.** Their fees are still counted — they arrive as
  `FeeReceived` like everything else — but they land under the
  `Other Protocol Fees` label. The effect is a coarser breakdown, not lost
  revenue. Mapping them is a follow-up.
- **HookPool and LPLocker pay their protocol fee to `FeeRouter` as an ERC-20.**
  `FeeRouter.receive()` is native-only, so an ERC-20 payment emits no event and
  cannot be observed there. This is an acknowledged under-count. It is currently
  exactly zero: `PoolFactory.getPoolCount()` returns 0 on every chain, so no
  HookPool exists yet.
- **`HookOSV3FeeVault` is an upgradeable proxy** and may be a generation behind
  the source on some chains. A change to the `Position` struct would break the
  `multiCall` decode; that is handled safely with `permitFailure: true` plus a
  native fallback, so the worst case is that a position's quote currency is
  treated as native rather than the adapter throwing.
- **The $HOOK Uniswap-v4 pool's swap VOLUME is deliberately excluded from
  `dexs/hookos.ts`,** because `dexs/uniswap-v4.ts` already covers that Robinhood
  PoolManager and counting it here would double it at the aggregate. Its hook
  tax is still counted as HookOS **fees**, which no Uniswap adapter sees.

## Guidelines checklist

- `version: 2`, `pullHourly: true`, `fetch(options)` single-argument signature.
- On-chain logs only. No API, no subgraph, no Dune, no new npm dependencies, no
  `.js` files.
- `getLogs` with explicit `target`/`targets` and human-readable `eventAbi`; no
  `noTarget`, no raw topics. `api.multiCall` used for pool and position
  enumeration.
- Every `.add()` label appears in `breakdownMethodology` and vice versa.
- `methodology` keys use display names (`Fees`, `Revenue`, `ProtocolRevenue`,
  `HoldersRevenue`, `SupplySideRevenue`, `Volume`).
- `Fees = Revenue + SupplySideRevenue` holds within each period by construction;
  `Revenue = ProtocolRevenue + HoldersRevenue` likewise.
- No `try/catch` swallowing adapter logic — the only `.catch()` calls are on
  optional contract reads whose absence means "this module is not deployed
  here", and they fall through to skipping that source rather than to a
  fabricated number.
- Hardcoded addresses carry source comments; no hardcoded rates remain.
- "Allow edits by maintainers" is enabled.

### DIFF
--- a/dexs/hookos.ts
+++ b/dexs/hookos.ts
@@ -1,4 +1,4 @@
-import { FetchOptions, FetchResultV2, SimpleAdapter } from "../adapters/types";
+import { Adapter, FetchOptions, FetchResultV2, SimpleAdapter } from "../adapters/types";
 import { CHAIN } from "../helpers/chains";
 
 // HookOS trading volume, from the two venues HookOS runs itself:
@@ -18,34 +18,41 @@ import { CHAIN } from "../helpers/chains";
 // are excluded too. Addresses are the canonical deployment registry
 // (protocol/contracts/deployments/addresses.json).
 type ChainConfig = {
-  BondingCurve?: string;
-  PoolFactory?: string;
+  BondingCurve: string;
+  PoolFactory: string;
+  start: string;
 };
 
 const CONFIG: Record<string, ChainConfig> = {
   [CHAIN.BASE]: {
     BondingCurve: "0x3C4b0F2D3d5bBdf4E0B323f0a8Eec7B02Cce6d40",
     PoolFactory:  "0xEE71e51e757a3B36F027400CDb7182710564654A",
+    start: "2026-06-05",
   },
   [CHAIN.HYPERLIQUID]: {
     BondingCurve: "0x93f35a190E6B7ed05E7bBAb78199720C0c849dDE",
     PoolFactory:  "0xF2F1C1D5089995c55C9Bf0395ebb70EBBF17b61D",
+    start: "2026-06-07",
   },
   [CHAIN.MEGAETH]: {
     BondingCurve: "0x6A2fAa5Da2B9F1515661f18160C0A0d584c0AC15",
     PoolFactory:  "0x1106A0257bbB2f7950f5bcf366e966D24c6F5cDd",
+    start: "2026-06-14",
   },
   [CHAIN.BSC]: {
     BondingCurve: "0xbb141A22B4cAef996052b2ecC9F9ef2Cde259bcA",
     PoolFactory:  "0x0d04627b6eFc9f546702969fF1faBD7a9642886f",
+    start: "2026-06-17",
   },
   [CHAIN.ETHEREUM]: {
     BondingCurve: "0xc841eF17b424B00A46C5acebDEEbE2976F168AC7",
     PoolFactory:  "0xcDfD3B997EC5A2F9CA59955d9aCE30eD8dFbFEff",
+    start: "2026-06-18",
   },
   [CHAIN.ROBINHOOD]: {
     BondingCurve: "0x93f35a190E6B7ed05E7bBAb78199720C0c849dDE",
     PoolFactory:  "0xF2F1C1D5089995c55C9Bf0395ebb70EBBF17b61D",
+    start: "2026-07-02",
   },
 };
 
@@ -73,7 +80,7 @@ const fetch = async (options: FetchOptions): Promise<FetchResultV2> => {
   // A pool's input token may be a launched token DeFiLlama cannot price, which
   // contributes zero rather than a fabricated value.
   if (c.PoolFactory) {
-    const count = Number(await api.call({ target: c.PoolFactory, abi: 'uint256:getPoolCount' }).catch(() => 0));
+    const count = await api.call({ target: c.PoolFactory, abi: 'uint256:getPoolCount' });
     if (count > 0) {
       const pools: string[] = await api.multiCall({
         abi: 'function allPools(uint256) view returns (address)',
@@ -90,17 +97,10 @@ const fetch = async (options: FetchOptions): Promise<FetchResultV2> => {
   return { dailyVolume };
 };
 
-const adapter: SimpleAdapter = {
+const adapter: Adapter = {
   version: 2,
   fetch,
-  chains: [
-    [CHAIN.BASE, { start: '2026-06-05' }],
-    [CHAIN.HYPERLIQUID, { start: '2026-06-07' }],
-    [CHAIN.MEGAETH, { start: '2026-06-14' }],
-    [CHAIN.BSC, { start: '2026-06-17' }],
-    [CHAIN.ETHEREUM, { start: '2026-06-18' }],
-    [CHAIN.ROBINHOOD, { start: '2026-07-02' }],
-  ],
+  adapter: CONFIG,
   methodology: {
     Volume: "Trading on the venues HookOS runs itself: bonding-curve buys and sells (native gas token in and out) and swaps through HookPool — the in-house AMM and the graduation target on HyperEVM, counted on the input leg. Uniswap trading is excluded and counted by Uniswap's own adapters: that covers post-graduation pools and the $HOOK v4 pool on Robinhood, which trades on a Uniswap PoolManager that Uniswap's v4 adapter already reports. Arena wagers, copy-trade records and launch fees are not trading volume and are excluded.",
   },
--- a/dexs/hookos.ts
+++ b/dexs/hookos.ts
@@ -1,38 +1,91 @@
 import { FetchOptions, FetchResultV2, SimpleAdapter } from "../adapters/types";
 import { CHAIN } from "../helpers/chains";
 
-// HookOS bonding-curve contracts per chain, synced from the canonical deployment
-// registry (contracts/deployments/addresses.json) and verified on each chain's
-// explorer. Volume is bonding-curve swap volume only: arena wagers, copy-trade
-// records and token-launch fees are not trading volume and are excluded.
-// Post-graduation trading moves to Uniswap v4 and is counted by Uniswap's own
-// adapter, so it is excluded here to avoid double-counting.
-const BONDING_CURVE: Record<string, string> = {
-  [CHAIN.BASE]:        "0x3C4b0F2D3d5bBdf4E0B323f0a8Eec7B02Cce6d40",
-  [CHAIN.MEGAETH]:     "0x6A2fAa5Da2B9F1515661f18160C0A0d584c0AC15",
-  [CHAIN.HYPERLIQUID]: "0x93f35a190E6B7ed05E7bBAb78199720C0c849dDE",
-  // Ethereum (1): https://etherscan.io/address/0xc841eF17b424B00A46C5acebDEEbE2976F168AC7
-  [CHAIN.ETHEREUM]:    "0xc841eF17b424B00A46C5acebDEEbE2976F168AC7",
-  // BNB Chain (56): https://bscscan.com/address/0xbb141A22B4cAef996052b2ecC9F9ef2Cde259bcA
-  [CHAIN.BSC]:         "0xbb141A22B4cAef996052b2ecC9F9ef2Cde259bcA",
+// HookOS trading volume, from the two venues HookOS runs itself:
+//
+//   • BondingCurve — pre-graduation trading, denominated in the native gas token.
+//   • HookPool     — the in-house AMM and the graduation target on HyperEVM.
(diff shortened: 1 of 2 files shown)


OxForged:
Pushed two follow-up commits after opening, both from continued validation against raw logs.

**1. Fixed a double count of bot trading fees.** `BotTradeRouter` pays its skim straight to `FeeRouter` in the same transaction that emits `Bought`/`Sold`, so the identical wei was counted twice — once as an unmapped `FeeReceived` sender (bucketed *Other Protocol Fees*) and again off the router's own event. Confirmed on Base, same block and same amount, on 2026-06-23/24/27; the 06-27 window reported `0.1582` under **both** labels. The routers now join the skip set that already existed for exactly this reason.

**2. Migrated labels onto `helpers/metrics`.** Where an equivalent already existed the shared `METRIC` is now used instead of a near-duplicate string — `Creator Fees` was string-identical to `METRIC.CREATOR_FEES` — along with `LP Fees`, `Trading Fees`, `Service Fees`, `Token Buy Back` and `Protocol Fees`. The three LP sources (V3 launch positions, the graduated-v4 splitter, the LP locker) now share one `LP Fees` line, and the routed launcher / Quick Launch fees moved to `Token Launch Fees` where they belong, so that merged line contains only LP fees. 27 labels down to 25.

The labels left bespoke are HookOS products with no equivalent in the enum. I kept them distinct rather than collapsing into `METRIC.SWAP_FEES` because the breakdown exists to show which venue earns — the bonding curve, graduated v4 pools, the $HOOK pool and the tokenized-stock pools have separate economics. Happy to collapse them if you'd prefer fewer series.

**Additional validation**, each reconciled bit-for-bit against decoded `eth_getLogs` output rather than eyeballed:

| chain | day | dailyFees | check |
|---|---|---:|---|
| base | 2026-06-25 | $62.2173 | `39759440347919238` wei, exact |
| bsc | 2026-06-28 | $9.0685 | exact |
| ethereum | 2026-06-18 | $4.0825 | exact |
| robinhood | 2026-07-13 | $4,027 | exact |

HyperEVM and MegaETH return 0.00 across their entire history, and that is correct rather than a gap: every `FeeReceived` they have ever emitted carries `data = 0x00…00`, i.e. the launch fee charged was literally zero. `PoolFactory.getPoolCount()` is 0 on every chain, so the HookPool path holds nothing.

One note for CI: a 24h BSC window is ~28.8k blocks and most free endpoints cap `eth_getLogs` well below that, so a red `test-adapter` run on BSC is likely an RPC range limit rather than an adapter fault.

OxForged:
Heads up on the red `test` check — I've diagnosed it and it is not the adapter.

`ts-check` passes and the `dexs` adapter passes. The `fees` run reached interval 4 of 24 and then every BSC endpoint in the runner's pool refused the query at once:

```
llamaRPCError: true, method: 'getLogs', chain: 'bsc'
  bsc-dataseed.binance.org   limit exceeded
  bsc-dataseed3.defibit.io   limit exceeded
  bsc.meowrpc.com            The method eth_getLogs is not supported.
  bsc-dataseed3.bnbchain.org limit exceeded
  1rpc.io/bnb                eth_getLogs is limited to 0 - 50 blocks range
```

Every value returned before that point was a legitimate zero — HookOS's last on-chain fee event was 2026-07-13, so a current window is genuinely empty.

The load is real, though, so I've pushed a reduction rather than just pointing at the RPCs. The three v4 hooks emit the same `ProtocolFeeCollected` event and were being fetched one call each; they are now queried by label group, so the graduation pair shares a single multi-target call. That takes up to three log queries per chain-interval down to at most two, and one on chains with no Quick Launch deployment.

I want to be straight that this may not be enough on its own: with `pullHourly` the fees adapter runs 24 intervals across up to 20 contracts, and a 24h BSC window is ~28.8k blocks, which is above what any free endpoint will serve. Locally it passes on all seven chains with a BSC endpoint that allows a 50k block range.

If it stays red, I'd rather ask than guess — happy to drop `pullHourly` on the fees adapter (the 24x multiplier is the whole problem) if you're fine with daily granularity there, or leave it if your indexer isn't affected by what the PR runner can reach. Your call, and I'll make whichever change you prefer.

OxForged:
Correcting my previous comment — I had the diagnosis wrong, and the fix I offered would have made it worse.

The latest run failed in the **`dexs`** adapter, which issues about three `getLogs` calls per interval and passed on the previous run. So this is not our call volume, and the reduction I pushed was aimed at the wrong thing.

Looking at what the endpoints actually return:

```
Requested range is over limit of 100
eth_getLogs is limited to 0 - 50 blocks range
The method eth_getLogs is not supported.
limit exceeded
```

The first two are **block-range caps, not rate limits**. BSC is ~3s blocks, so even a single 1-hour interval is ~1,200 blocks — already past a 50 or 100 block ceiling on one call. Which endpoint a run draws decides whether it fails, so it alternates between the two adapters run to run.

That makes my earlier offer to drop `pullHourly` exactly backwards: without it each window becomes ~28,800 blocks instead of ~1,200, so every range-capped endpoint would fail harder. **Please disregard that suggestion** — `pullHourly` is helping here, and I'd now recommend keeping it.

I don't think there's anything left for me to change in the adapter: no query shape gets a 1,200-block window under a 50-block cap. Locally it passes on all seven chains against a BSC endpoint that serves a 50k range, and the numbers reconcile bit-for-bit with decoded logs on base, bsc, ethereum and robinhood.

So this comes down to the runner's BSC pool rather than the code. If it's useful, I'm happy to do whatever fits your process — but I'd rather not push more speculative commits at a red check whose cause is outside the adapter.

bheluga:
@OxForged thanks for the PR

PR #8555, opened 2026-08-03
[feat] accountable - add fees adapter

pawel-accountable:
Adds a fees adapter for Accountable (RWA / credit vaults, protocol already listed for TVL and yields).

## Source

`https://yield.accountable.capital/api/protocol/fees-daily` — a public read-only endpoint serving one row per (date, chain), already denominated in USD, so no on-chain calls or pricing are needed. The whole history comes back on every call, so a single in-flight request is reused for every (chain, day) fetch in the run; a failed request is deliberately not memoized.

## Income statement

Borrowers draw on vaults and pay interest. That interest is split between vault depositors and a performance fee, which is itself split between the protocol and the vault manager running the strategy.

- `dailyFees` / `dailyUserFees` — all borrow interest, including the performance fee taken out of it. Borrowers pay 100% of it.
- `dailySupplySideRevenue` — interest to depositors, **plus the vault manager's share of the performance fee**.
- `dailyRevenue` / `dailyProtocolRevenue` — the protocol's share of the performance fee.

The vault manager is an external party running the strategy, so its cut is treated as cost of funds, matching the node-operator commission treatment in the Lido example in `fees/GUIDELINES.md`. Flagging explicitly because `fees/maple-finance.ts` books the analogous pool-delegate fee into `dailyRevenue`/`dailyProtocolRevenue` instead; happy to switch if you prefer consistency with that adapter.

Citrea vaults are not covered, that chain is not indexed upstream, so its fees are absent from these totals.

## Verification

- `dailyFees = dailyRevenue + dailySupplySideRevenue` holds exactly on all 630 rows of history; no row has a negative component, so `allowNegativeValue` is not needed.
- The manager amount is computed upstream as its own column and is also reproducible as `fees - supply_side - protocol_revenue`; the adapter accepts either.
- Days with no accrual are filtered out upstream, so a gap inside the covered range is treated as a genuine zero, while a date past the last available row throws so the run retries instead of recording a 0.
- `npm test fees accountable 2026-07-28`:

```
monad     | 85.68 k | 85.68 k | 81.95 k | 3.73 k | 3.73 k
ethereum  |  2.13 k |  2.13 k |  2.03 k | 98.00  | 98.00
base      |    9.86 |    9.86 |    9.37 |  0.49  |  0.49
arbitrum  |  178.00 |  178.00 |  173.00 |  4.44  |  4.44

Borrow Interest                  | fees        87,992.86
Borrow Interest To Depositors    | supply side 74,751.37
Performance Fee To Vault Manager | supply side  9,412.44
Performance Fee To Protocol      | revenue      3,828.93
```

- `npm run ts-check` clean for this file.

### DIFF
--- a/fees/accountable.ts
+++ b/fees/accountable.ts
@@ -29,6 +29,11 @@ const FACTORIES: Record<string, string[]> = {
     "0x333a12e2B519DA16EBE75012d54574C16ef4463f",
     "0xDAc0e7EffB16B249d1Bb672D25D7827481Be2081",
   ],
+  [CHAIN.ROBINHOOD]: [
+    "0x017273Eeb06Ee9f863020269417DB9559FD94173",
+    "0x474B612F970491801743BF0e4B9153620FC36096",
+    "0xA4d6a4aD35fc632aEE1dC48A2aEc2aaa37B51F9f",
+  ],
 };
 
 // aHYPER Looping Vault (vault 0x23b148d8f389C5821739381f1FF87bB7e1162566) is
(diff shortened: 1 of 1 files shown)


bheluga:
Is it possible to calculate fees using on-chain functions or events?
You can use api to get list of vaults

pawel-accountable:
Superseded by #8599, which computes the same numbers from chain state instead of the API, as requested. Closing this one to keep the review in one place.

PR #8557, opened 2026-08-03
Add AutoRange fees/revenue and volume adapters

elkiyo:
## Summary
[AutoRange](https://www.autorange.xyz/) ([@AutoRangexyz](https://x.com/AutoRangexyz)) is a non-custodial Uniswap V3 concentrated-liquidity vault platform, live on Celo and Arbitrum. Already listed for TVL: DefiLlama-Adapters#20126 (merged).

## `fees/autorange.ts`
Every vault's Uniswap V3 trading fees are split on-chain between the vault's owner (LP) and the platform, either at rebalance time or when the owner manually claims (`RangeVault.sol`'s `_splitPerformanceFee`):
- **Fees**: the full trading-fee amount, before the split (`LpFeesPaidToOwner` + `FeesCollected` + `PerformanceFeeCollected` events)
- **Revenue / ProtocolRevenue**: the platform's cut (`PerformanceFeeCollected`) — no separate token-holder split
- **SupplySideRevenue**: the owner's net share after the platform's cut

Marked `doublecounted: true` since these are Uniswap V3 pool fees, already reflected in Uniswap's own fee reporting — same convention as other concentrated-liquidity managers (Steer, Gamma).

## `dexs/autorange.ts`
**Volume here is a different definition than a typical DEX's `dailyVolume`** — AutoRange doesn't route trades for third parties, so there's no swap volume to attribute to it. Instead, this tracks the USD value the keeper agent itself deploys every time it builds or rebuilds a vault's position (`initPosition()`'s initial mint, or `rebalance()`'s re-mint after the price leaves the range or on the owner's configured interval). Flagging this explicitly for review — happy to adjust the categorization if this isn't the right dashboard for it.

## Verified
`npm test -- fees/autorange` and `npm test -- dexs/autorange` against live chain state (Celo + Arbitrum), sane proportional splits (Revenue + SupplySideRevenue == Fees) and volume in the expected order of magnitude vs. current TVL.

### DIFF
--- a/dexs/autorange.ts
+++ b/dexs/autorange.ts
@@ -1,0 +1,88 @@
+import { FetchOptions, SimpleAdapter } from "../adapters/types";
+import { CHAIN } from "../helpers/chains";
+
+// AutoRange — non-custodial Uniswap V3 concentrated-liquidity vault
+// platform. "Volume" here is NOT third-party swap volume routed through
+// AutoRange (it doesn't route trades for anyone) — it's the USD value the
+// keeper agent itself deploys into a fresh Uniswap V3 position every time
+// it builds or rebuilds one: initPosition() (a vault's very first mint) and
+// rebalance() (every subsequent re-mint after the price leaves the range or
+// on the owner's configured interval). See RangeVault.sol's PositionInitialized
+// and Rebalanced events.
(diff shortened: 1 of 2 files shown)


bheluga:
@elkiyo thanks for the PR

PR #8572, opened 2026-08-04
  AZverse: add canonical trading stats and broker interface adapters

defiwizard1002:
## Summary

  Adds AZverse daily volume, fees, and revenue adapters for its
  perpetual and spot markets, plus five AZverse broker-interface
  adapters.

  The implementation reads AZverse's public, immutable UTC-day
  statistics API:

  - Canonical adapter: no broker_id
  - Broker interfaces: broker_id=AZVERSE, AVODEX, MONSTER, BBXCOM, and
    XOX

  - All broker adapters are marked doublecounted: true, because they
    are routed slices of AZverse trading activity.

  ## Adapter details

  ### Canonical AZverse

  - dexs/azverse: perpetual USD notional volume
  - fees/azverse: perpetual trading fees and revenue
  - dexs/azverse-spot: spot USD volume
  - fees/azverse-spot: spot trading fees and revenue

  For canonical fees:

  - dailyFees / dailyUserFees: all user-paid trading fees
  - dailySupplySideRevenue: builder revenue shares
  - dailyRevenue: fees retained by AZverse after builder revenue
    shares

 ### Broker interfaces

  Five perpetual Interface adapters are exported from `factory/
  azverse.ts`.
  All start from `2025-12-31` and are marked `doublecounted: true`.

  - `azx-perps` — AZX Perps (`broker_id=AZVERSE`)
  - `avodex-perps` — AvoDex Perps (`broker_id=AVODEX`)
  - `monster-perps` — Monster Perps (`broker_id=MONSTER`)
  - `bbx-perps` — BBX Perps (`broker_id=BBXCOM`)
  - `xox-perps` — XOX Perps (`broker_id=XOX`)

  For each broker:

  - `dailyVolume`: USD notional routed through that broker
  - `dailyFees` / `dailyUserFees`: user-paid fees on that routed flow
  - `dailyRevenue` / `dailyProtocolRevenue`: API `revenue` field, i.e.
  the broker's builder-fee share
  - `dailySupplySideRevenue`: the remaining fee share retained by
  AZverse

  ## Testing

  ```bash
  pnpm run ts-check

  # Tests API date 2026-08-03 UTC
  pnpm test dexs azverse 2026-08-04
  pnpm test fees azverse 2026-08-04

  # Test all five broker interfaces, on both dashboards
  for type in dexs fees; do
    for adapter in azx-perps avodex-perps monster-perps bbx-perps xox-
    perps; do
      pnpm test "$type" "$adapter" 2026-08-03
    done
  done

  All commands above pass locally.
  ```

  #### Name (to be shown on DefiLlama):

  AZverse

  #### Chain:

  AZ Axis. The dimensions adapters use `CHAIN.OFF_CHAIN` because they
  consume AZverse's public daily statistics API.

  #### Short Description (to be shown on DefiLlama):

  AZverse is an orderbook trading ecosystem supporting perpetual and
  spot markets, with independent broker frontends sharing platform
  liquidity.

  #### Category (full list at https://defillama.com/categories)
  *Please choose only one:

  Derivatives

  #### Implementation Details:

  Daily UTC aggregates are read from AZverse's public no-auth
  endpoint:
  `https://app.azverse.xyz/exapi/stats/v1/stats/public/defillama/
  daily-stats`.

  #### Github org/user (Optional, if your code is open source, we can
  track activity):

  https://github.com/AZVerses

  #### Does this project have a referral program?

  Yes — AZverse has a referral product, but this listing does not use
  referral links.


bheluga (commented):
@defiwizard1002 thanks for the PR
Why does volume look much different from your website?
perp volume is on the lower end (I don't think it sums up to 200M+ as reported by API)
Spot volume is greater than 0 (api reports 0)

defiwizard1002:
@bheluga 

Thanks for the careful review — you're right on both counts, and the discrepancy was a real bug on our side. Root cause and fixes below.

1) Perp volume was double-counted (why the API over-reported vs the website/klines).
Our stats API aggregated volume from raw fills and summed both sides of every trade — taker and maker. Since each trade has exactly one taker and one maker, this counted every trade twice, inflating perp volume by ~2×. That's why the API showed 200M+ while the true single-count volume is lower — in line with your estimate.

Fix — the canonical volume now:
counts each trade once (taker side), and excludes internal MERGE records (position-merge / non-market fills that aren't real trades).
After the fix, the perp total reconciles with our klines and website (≈ $35.76M for 24h (data sampled on 2026-08-09)

≈ $712 for 24h (data sampled on 2026-08-09) vs the previous ~2× figure).

2) Spot volume was reported as 0.
Separate root cause: [the spot leg wasn't wired into the stats endpoint / spot fills were being filtered out]. This is now fixed — the endpoint returns actual spot volume (> 0): ≈ $[SPOT] for [window].

3) Broker attribution (broker interface adapters).
Per-broker volume is aggregated by builderCode, and we additionally map market-maker api_keys back to their originating broker, so maker / MM-side volume is attributed to the correct broker instead of being dropped. Broker interface rows are reported with doublecounted: true so they don't inflate the canonical AZverse total.

The API perp total now matches klines/website (single-count) and spot is non-zero. Could you re-review when you get a chance? Happy to share a reconciliation snapshot (API vs klines) for any window you'd like.

Thanks again for flagging this.

bheluga:
> @bheluga
> 
> Thanks for the careful review — you're right on both counts, and the discrepancy was a real bug on our side. Root cause and fixes below.
> 
> 1. Perp volume was double-counted (why the API over-reported vs the website/klines).
>    Our stats API aggregated volume from raw fills and summed both sides of every trade — taker and maker. Since each trade has exactly one taker and one maker, this counted every trade twice, inflating perp volume by ~2×. That's why the API showed 200M+ while the true single-count volume is lower — in line with your estimate.
> 
> Fix — the canonical volume now: counts each trade once (taker side), and excludes internal MERGE records (position-merge / non-market fills that aren't real trades). After the fix, the perp total reconciles with our klines and website (≈ $[POST-FIX PERP] for [window], vs the previous ~2× figure).
> 
> 2. Spot volume was reported as 0.
>    Separate root cause: [the spot leg wasn't wired into the stats endpoint / spot fills were being filtered out]. This is now fixed — the endpoint returns actual spot volume (> 0): ≈ $[SPOT] for [window].
> 3. Broker attribution (broker interface adapters).
>    Per-broker volume is aggregated by builderCode, and we additionally map market-maker api_keys back to their originating broker, so maker / MM-side volume is attributed to the correct broker instead of being dropped. Broker interface rows are reported with doublecounted: true so they don't inflate the canonical AZverse total.
> 
> The API perp total now matches klines/website (single-count) and spot is non-zero. Could you re-review when you get a chance? Happy to share a reconciliation snapshot (API vs klines) for any window you'd like.
> 
> Thanks again for flagging this.

Hey @defiwizard1002 thanks for the update,
But numbers still feels off,
Perps: 35M and spot 712$ through api
numbers don't really match the site

defiwizard1002:
@bheluga  Hey just checked the numbers were right, for example 35M and spot 712$ on 8.9 through api, and the [site](https://app.azverse.xyz/en/dashboard/stats) shows the same number in Total Volume chart, please confirm
image

bheluga:
> @bheluga Hey just checked the numbers were right, for example 35M and spot 712$ on 8.9 through api, and the [site](https://app.azverse.xyz/en/dashboard/stats) shows the same number in Total Volume chart, please confirm image

Ok, this part looks fine.
But azx builder has volume more than azverse , any idea why?

defiwizard1002:
@bheluga 
Good question — this is expected, and it's a deliberate difference between the two figures, not a bug.
Terminology note: we've standardized on calling these partners Composers (formerly "builders") — same entities; I'll use "Composer" below.
The short answer is that the two figures use two different counting methods:
AZverse (parent) = single-count. Each trade is counted once (taker side). This is the authoritative, kline-reconciled ecosystem volume — the figure to compare against other protocols. (This is the fix from the earlier double-counting issue.)
Per-Composer = 2-sided attribution. Each Composer is credited for both sides its users are on — taker and maker.
Why 2-sided attribution is required. A Composer's contribution to the network is both the flow it originates (taker) and the liquidity it provides (maker). If we credited only the taker side, any Composer whose users are makers — e.g. a grid-strategy Composer, or a market maker — would show zero volume despite driving real trades. Two-sided attribution is the only way to credit liquidity-providing Composers correctly.
Consequence. Every trade has exactly one taker-side Composer and one maker-side Composer, so Σ(all Composers) ≈ 2 × the single-counted AZverse total — and any individual Composer can exceed the parent. This is inherent to 2-sided attribution; it does not double-count the ecosystem total, which stays single-count.
Why AZX specifically > AZverse. AZX is the flagship and carries the attribution of the shared market maker (the MM api_keys map to AZX). The MM is the maker on most trades across the whole ecosystem, so AZX's maker-side attribution alone approaches the full ecosystem volume — plus its own taker-side flow — which puts it above the single-counted parent.
Keeping it out of the ecosystem total. These per-Composer entries should be treated as category: "Interface" and linked as children of the AZverse parent (parentProtocol), so they're shown for attribution/visibility but excluded from the AZverse total — the parent stays the single-count canonical figure. Could you confirm/apply that classification on your side, or tell us exactly what the adapter should set? We'll wire it up either way.
If you prefer reconciliation instead: we can switch per-Composer attribution to a 50/50 split (each trade's notional split half to the taker-side Composer, half to the maker-side Composer) so that Σ(Composers) = AZverse total exactly. Happy to use whichever representation you prefer.
Thanks!

bheluga:
@defiwizard1002 thanks for the PR
Divided by 2 and added [here](https://github.com/DefiLlama/dimension-adapters/pull/8584)
Because ideally sum of brokers should be less than or eq to parent

PR #8573, opened 2026-08-04
curve: fix missing avax pools, ng crypto-pool revenue, and volume def…

DaniiRix:
Audits Curve's fee/volume logic against on-chain contracts and the Curve fees API.

### Changes

- Added the missing **Avalanche stable-ng** factory (54 pools).
- Fixed **NG crypto pools** by falling back to `ADMIN_FEE` when `admin_fee()` reverts (50% protocol share).
- Switched **twocrypto/tricrypto** fees from spot `fee()` reads to the actual fee emitted in swap events, matching dynamic-fee pools.
- Split admin fees **90% veCRV / 10% DAO treasury**, matching the current governance allocation.
- API path now uses **trading volume only**, excluding liquidity adds/removes to match the on-chain path and DEX methodology.
- Throw when every factory lookup fails instead of silently returning $0.
- Narrow API fallback to the expected "chain not supported" case.
- Exclude bribes from `dailyUserFees`, dedupe pool lists, and update methodology.

### Results

| Metric (2026-07-25) | Before | After |
|---|---:|---:|
| Avalanche volume | $2.23k | **$37.18k** |
| Avalanche fees | $0.91 | **$36.00** |
| Ethereum volume | 11.06M | **7.81M** |
| Gnosis volume | 821.65k | **407.58k** |
| Plasma / Monad / Avalanche protocol revenue | 0 | **9.35 / 4.35 / 1.79** |

### DIFF
--- a/dexs/curve/api.ts
+++ b/dexs/curve/api.ts
@@ -13,6 +13,9 @@ interface CurveChainFees {
   fees_to_lp: number;
   fees_to_dao: number;
   fees_to_treasury: number;
+  // Yield Basis fees, already inside total_fees/fees_to_lp (those pools have admin_fee = 0).
+  // fees/yield-basis reports the same dollars.
+  lp_fees_yb: number;
   chain: string;
 }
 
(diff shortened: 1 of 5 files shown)


bheluga:
@DaniiRix thanks for the PR

PR #8513, opened 2026-07-31
feat(primit): standalone Primit derivatives adapter (self-reported native volume)

leeleeEcho:
## Summary

Adds `dexs/primit/index.ts` — a standalone derivatives adapter that reports Primit's own matching-engine daily volume, pulled from a public JSON endpoint at `api.primit.io`.

**Context:** Primit was previously added as an Orderly Network builder in [#8115](https://github.com/DefiLlama/dimension-adapters/pull/8115) via `factory/orderly.ts` with `broker_id=primit`. Since then Primit has launched its own perpetual-futures matching engine hosting native pairs (Primit-native perpetual markets, with more coming) that don't route through Orderly.

This PR reports **only** that native-engine volume. The existing `factory/orderly.ts` entry is left in place — it continues to attribute the Orderly-routed portion (BTC/ETH/etc). The two pair sets are disjoint, so no double-counting.

### Data flow

```
Primit backend (Rust · Axum · sqlx)
  └─ public.trades table (self-hosted matching engine, our reference perpetual pair + native pairs)
       └─ SQL: SUM(amount * price) WHERE created_at ∈ [start, end) UTC
                                     AND is_self_trade = false
                └─ GET https://api.primit.io/api/v1/public/stats/daily-volume?date=YYYY-MM-DD
                     └─ dexs/primit/index.ts (this PR)
                          └─ dailyVolume in USD → DeFiLlama
```

Endpoint is public and unauthenticated. `is_self_trade = false` filter excludes market-maker wash trades so the publicly reported number isn't inflated.

### Endpoint contract

```
GET https://api.primit.io/api/v1/public/stats/daily-volume?date=YYYY-MM-DD
```

Example (real production data):
```
$ curl 'https://api.primit.io/api/v1/public/stats/daily-volume?date=2026-07-30'
{
  "date": "2026-07-30",
  "start_of_day_unix": 1785369600,
  "end_of_day_unix": 1785456000,
  "total_volume_usd": "275332074.15247000000000000000",
  "trade_count": 907245,
  "scope": "primit-native-markets"
}
```

The `scope` field is a stable contract — any change to the aggregation basis will bump this marker.

### Files changed

- **new:** `dexs/primit/index.ts` (53 lines)

No changes to `factory/orderly.ts` — the existing `broker_id=primit` entry stays.

---

##### Name (to be shown on DefiLlama):
Primit

##### Twitter Link:
https://x.com/primitforall

##### List of audit links if any:
[...]

PR #8574, opened 2026-08-04
feat(primit-perps): add fees + revenue at flat 4 bps

leeleeEcho:
## Summary

Extends the previously-merged \`dexs/primit-perps\` adapter (#8513) beyond \`dailyVolume\` to also report \`dailyFees\` and \`dailyRevenue\`, derived from the protocol's flat **4 bps (0.04%)** trading fee applied to the same on-chain volume figure.

## Changes

Single file: \`dexs/primit-perps/index.ts\` (+35 / -10 lines).

- \`dailyFees = dailyVolume * 4 / 10000\`
- \`dailyRevenue = dailyFees\` (100% goes to protocol; no LP or affiliate revenue share on Primit-native pairs today)
- \`dailyProtocolRevenue\` mirrors \`dailyRevenue\` for the same reason
- BigInt multiplication done before splitting to Number to avoid the same 2^53 precision loss the volume-side computation already handles

## Methodology (added in-file)

**Fees**: Flat 4 bps (0.04%) trading fee applied to daily USD volume matched on Primit's own engine. Computed as \`dailyVolume * 4 / 10000\` on the same volume figure already reported.

**Revenue**: 100% of trading fees collected on Primit-native pairs accrue to the protocol. No LP or affiliate revenue share is applied to this pair set, so Revenue equals Fees.

**ProtocolRevenue**: Same as Revenue.

## Why formula-based, not per-trade takerFee/makerFee

The \`TradeRecorded\` event does carry \`int256 takerFee, int256 makerFee\` fields, but this adapter uses the protocol's public fee schedule (4 bps flat) applied to the on-chain volume — the same reasoning many other exchange adapters use, and stable across any internal fee-schedule refactors. If DeFiLlama prefers per-event fee summation, happy to switch in a follow-up.

## Coexistence with \`factory/orderly.ts primit\`

The existing \`factory/orderly.ts\` \`broker_id=primit\` entry (#8115) reports fees for the Orderly-routed portion of Primit volume (BTC/ETH/etc). This adapter covers the disjoint Primit-native pair set. **No double-counting** — the two data sources are additive at the fill level.

## Test plan

- [ ] Review methodology text (Fees / Revenue / ProtocolRevenue sections)
- [ ] Confirm the flat-rate approach is acceptable (vs event \`takerFee\`+\`makerFee\` summation)
- [ ] CI \`test\` job should print non-zero dailyFees for a recent day (~\`dailyVolume * 0.0004\`)

### DIFF
--- a/dexs/primit-perps/index.ts
+++ b/dexs/primit-perps/index.ts
@@ -30,6 +30,8 @@ const methodology = {
     "100% of trading fees collected on Primit-native pairs accrue to the protocol. No LP or affiliate revenue share is applied to this pair set, so Revenue equals Fees.",
   ProtocolRevenue:
     "Same as Revenue. All protocol-native trading fees stay with the protocol.",
+  SupplySideRevenue:
+    "Zero. Primit-native pairs are not backed by an external liquidity-provider vault today — no maker rebate, no LP share, no affiliate share — so nothing from the fee stream accrues to a supply side. Preserves the DeFiLlama contract dailyFees = dailyRevenue + dailySupplySideRevenue.",
 };
 
 const SCALE_18 = 10n ** 18n;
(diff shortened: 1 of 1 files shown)


leeleeEcho:
Addressed the CodeRabbit finding in `a4bfbc2`.

- Added explicit `dailySupplySideRevenue: 0` to satisfy the `dailyFees = dailyRevenue + dailySupplySideRevenue` contract.
- Added matching `SupplySideRevenue` methodology entry (no LP vault, no maker rebate, no affiliate share on Primit-native pairs today).
- No numeric change to the other fields.

CI green. Ready for another look.

bheluga:
@leeleeEcho thanks for the PR

PR #8576, opened 2026-08-04
beets: drop the dead subgraph adapter so dexs falls back to the onchain balancer helper

0xshubhs:
beets dex volume has been empty since 07-30. all three chains stopped on the same day:

```
2026-07-27  sonic $55,650  op $28,715  fantom $1,078
2026-07-28  sonic $48,323  op $10,142  fantom   $966
2026-07-29  sonic $66,661  op  $5,909  fantom   $802
2026-07-30  nothing
...
2026-08-04  nothing
```

`total7d` on the api is $132,803, which is exactly 07-28 + 07-29, so nothing has landed for six days.

the protocol is fine. tvl is live and moving ($3.98m today), beets' own backend reports `swapVolume24h` of $33,437 on sonic right now, and geckoterminal shows the top beets-sonic pools trading every one of those days. so it's the adapter.

`dexs/beethoven-x/index.ts` reads three graph-network subgraphs. all three going quiet on the same day points at the publisher rather than three separate failures, and beets has since moved to their own backend. i can't confirm from outside because the sdk's default graph key is out of credits, so i checked it differently: other subgraph adapters are still reporting normally (dodo $4.8m/24h, woofi $6.4m, balancer-v1 $31k, maverick-v2, swapr-v3, pangolin-v3, pharaoh, defiplaza), so the gateway itself is working in production and this is specific to beets.

the fix is just deleting the file.

`helpers/balancer` is already registered as a factory for both fees and dexs, and it already has a `beethoven-x` entry with the vaults for all three chains:

```ts
"beethoven-x": {
  [CHAIN.OPTIMISM]: { vault: '0xba12222222228d8ba445958a75a0704d566bf2c8', revenueRatio: 0.25, start: '2023-01-01' },
  [CHAIN.FANTOM]:   { vault: '0x20dd72ed959b6147912c2e529f0a0c651c33c9ce', revenueRatio: 0.25, start: '2023-01-01' },
  [CHAIN.SONIC]:    { vault: '0xba12222222228d8ba445958a75a0704d566bf2c8', revenueRatio: 0.25, start: '2024-12-14' },
}
```

`getFees` in that helper already returns `dailyVolume` alongside `dailyFees`, off vault `Swap` logs. the fees side has no file in `fees/`, so it falls through to the factory and has kept working the whole time; beets fees are still reporting $74.55/24h today. the dexs side has a file, and since the factory is only the fallback for when a file doesn't exist, the broken subgraph adapter shadows a working on-chain one.

every other balancer fork in that factory (ducata, embr, phux, tanukix, aequinox, chimpexchange, darkness, koyo, mondrain, wavelength-dao, klex-finance, polaris-fi) has no `dexs/` file and gets volume from the factory. beethoven-x was the only one carrying a leftover subgraph adapter.

after deleting it, `npm test dexs beethoven-x` resolves to `balancer.ts` and the empty days fill in:

```
2026-07-31   optimism $6.46k   fantom $164   sonic $33.83k   total $40.46k
2026-08-02   optimism $5.70k   fantom $523   sonic $32.17k   total $38.40k
2026-08-03   optimism $3.45k   fantom $396   sonic $38.33k   total $42.17k
```

the 08-03 sonic figure lines up with geckoterminal, where wS/USDC.e alone did $20,179 that day, and the current-day number agrees with beets' own $33,437.

one thing to flag rather than bury: on the days where both sources exist the on-chain number is lower than the subgraph was.

```
            subgraph   onchain
2026-07-27   $85,443   $38,900
2026-07-29   $73,372   $52,440
```

i'd expect on-chain to be the better of the two here. the subgraph query took `poolSnapshots` with no `first:`, so it saw at most the default 100 pools (sonic alone has 383), and it dropped any pool whose previous-day snapshot was missing (`if (yesterdayValue === 0) return 0`), while balancer per-pool `swapVolume` also picks up nested/BPT hops that aren't really user trading. the on-chain path counts one side per swap through `addOneToken`. i haven't tried to attribute the gap precisely, so if you'd rather keep the old level i'd want to understand it first.

also worth noting: the old file declared `start: '2021-10-05'` and the factory entries start 2023-01-01 (optimism/fantom) and 2024-12-14 (sonic), so a full refill wouldn't reach back as far. stored history isn't affected, and the pre-2023 data wasn't refillable anyway with the subgraphs down.

typechecks clean. `cli/dimensionModules.json` is gitignored so there's nothing generated to commit.


bheluga:
@0xshubhs thanks for the PR

PR #8579, opened 2026-08-04
orai-quant-terminal: get_fees is a stored total, book the daily change not the whole level

### LINKED ISSUE #8276: orai-quant-terminal: static contract counter booked as fees every day ($93,960/30d)
`fees/orai-quant-terminal` reports the same $3,132 of fees, revenue and protocol revenue every day, because it books a stored level as if it were that day's flow.

### What the adapter does

It queries `{"get_fees":{}}` on `orai1rzfk6fd6d5zhm77cshdtr0vsuyu0qe0dg36evysklx8n6q8h38psxywppw`, divides by 1e6, and assigns the whole result to `dailyFees`, `dailyRevenue` and `dailyProtocolRevenue` under `runAtCurrTime: true`.

### That value is a stored level, not a daily accrual

Querying the contract's full state shows `fees` living alongside `tvl` in one snapshot struct:

```
quant_stats -> {"tvl":"953362900000","fees":"3131559123"}
get_tvl     -> "953362900000"     ($953,362.90)
get_fees    -> "3131559123"       ($3,131.56)
```

The contract (`code_id` 4153, `crates.io:quant-stats 0.1.1`) exposes only these execute variants:

```
update_config, update_quant_stats, update_vault_stats
```

There is no accrual logic anywhere in it — `update_quant_stats` simply **sets** `{tvl, fees}`. It's an admin-written mirror of an off-chain figure, and `fees` is paired with `tvl`, which is definitionally a level.

### The published series confirms it hasn't moved

`api.llama.fi/summary/fees/orai-quant-terminal` returns 41 daily datapoints from 2026-06-09 to 2026-07-19 with **exactly one distinct value** across the entire series: $3,132 every day. Since the adapter runs at current time and re-queries the live contract each day, a perfectly flat series over 41 consecutive days means the on-chain number never changed. Direct queries at the current height and at ~24h back both return the identical `3131559123`.

### Economic cross-check

| reading | implication |
|---|---|
| `fees` is a daily flow | $1,143,019/yr on $953k TVL = **119.9% of TVL per year** in performance fees |
| `fees` is cumulative | $3,132 lifetime on $953k TVL = **0.33% of TVL** |

The first is not a possible fee rate; the second is ordinary.

### Impact

$3,132/day × 41 days = **$128,412 all-time**, **$93,960 over the last 30 days**, reported as fees, revenue and protocol revenue simultaneously — against a counter that provably did not move over that window.

### Why I'm opening an issue rather than a PR

The correct daily value isn't derivable on-chain. The contract emits no fee events and holds no history — the daily figure would be the delta of a number an off-chain admin sets, and `lcd.orai.io` only serves state about two days back (heights beyond that return `failed to load state`), so a delta-based adapter couldn't be backfilled or refilled. Deciding between reporting the delta, exposing the counter as a cumulative total, or asking the team for a real fee endpoint seems like a call for whoever owns this integration.

Happy to implement whichever direction you prefer.

0xshubhs:
fixes #8276.

@tarun-khatri diagnosed this on 07-19 and the diagnosis holds up. re-verified today, 16 days later, and the contract still returns the same bytes:

```
get_fees -> "3131559123"      ($3,131.56)
get_tvl  -> "953362900000"    ($953,362.90)
```

the published series makes it obvious. `summary/fees/orai-quant-terminal` has 57 daily points going back to 2026-06-09 and exactly one distinct value across all of them:

```
2026-07-30  $3,132.00
2026-07-31  $3,132.00
2026-08-01  $3,132.00
2026-08-02  $3,132.00
2026-08-03  $3,132.00
2026-08-04  $3,132.00
```

that's $3,132 of fees, revenue and protocol revenue invented every day, currently $93,960 per 30d, because a stored level is being booked as that day's flow.

the level itself hasn't moved in about a month. lcd.orai.io honours `x-cosmos-block-height` (it echoes the height back in `grpc-metadata-x-cosmos-block-height`), and `get_fees` is still `3131559123` at height 113447779, which at the measured 0.6758 s/block is ~31 days back. below ~100m it errors with "version has either been pruned", so a month is as far as i can show.

**what this changes**

reads `get_fees` at the current height and at the height 24h earlier, and books the difference. when the team hasn't pushed a stats update the delta is 0, which is the honest answer, and when they do push one the change lands on that day.

height resolution follows `fees/valdora-finance.ts`, which already does historical cosmwasm queries this way. one thing worth flagging: i did **not** reuse its binary search from height 1, because lcd.orai.io only keeps block headers back to ~115.6m (about 14 days) and every low probe would 404 the adapter. this measures block time from one probe and refines from there instead, which stays near the tip. converges on the first iteration in practice:

```
latest height 117448265  blockTime 0.6758s
  iter 0: height 117320422, drift 16s -> converged, gap 24.00h
```

3 requests for the height, 2 for the values.

also tightened `readFeesValue`. it used to `return 0` for anything it couldn't parse, so an lcd hiccup would have silently published a zero; it now returns null and the caller throws. and if `get_fees` ever comes back lower than the previous read it throws rather than reporting a negative, since that would mean it isn't the cumulative counter this assumes.

**the assumption, stated plainly**

this treats `get_fees` as a cumulative total. it sits in the same `quant_stats` struct as `tvl` and is written by `update_quant_stats`, and `tvl` is unambiguously a level, so cumulative is the reading i'd bet on, but it's been frozen for the whole window i can query, so i can't demonstrate it incrementing. if you know it's actually a per-period figure that resets, say so and i'll rework it.

the other option is dropping the adapter entirely, since the stats contract looks abandoned. i didn't go that way because orai.io, quant.orai.io and scan.orai.io are all up, so the protocol isn't dead, just not updating this contract. happy to switch to removal if you'd rather.

either way the current behaviour is wrong and shouldn't keep running.

verified twice, both runs $0.00 fees / $0.00 revenue / $0.00 protocol revenue. typechecks clean.

### DIFF
--- a/fees/orai-quant-terminal/index.ts
+++ b/fees/orai-quant-terminal/index.ts
@@ -62,7 +62,11 @@ async function getLatestBlock(): Promise<{ height: number; timestamp: number }> 
   return { height, timestamp };
 }
 
+// oraichain blocks are ~0.68s, so 100k blocks is ~19h of history: long enough to
+// average out jitter, short enough to stay inside what lcd.orai.io keeps (it serves
+// block headers back ~1.8m blocks and 404s below that)
 const BLOCK_TIME_PROBE = 100000;
+// ~180 blocks of slack, small next to a 24h window
 const HEIGHT_TOLERANCE_SECONDS = 120;
 const MAX_REFINEMENTS = 6;
 
@@ -92,25 +96,16 @@ async function getHeightAt(
       Math.max(1, height - Math.round(drift / blockTime))
     );
   }
-  return height;
+  throw new Error(
+    `orai-quant-terminal: could not resolve a block within ${HEIGHT_TOLERANCE_SECONDS}s of ${timestamp}`
+  );
 }
 
 function readFeesValue(payload: unknown): number | null {
-  if (typeof payload === "number")
-    return Number.isFinite(payload) ? payload : null;
-  if (typeof payload === "string") {
-    const parsed = Number(payload);
-    return payload.trim() !== "" && Number.isFinite(parsed) ? parsed : null;
-  }
-  if (!payload || typeof payload !== "object") return null;
-  const data = payload as Record<string, unknown>;
-
-  for (const value of Object.values(data)) {
-    const nested = readFeesValue(value);
-    if (nested !== null) return nested;
-  }
-
-  return null;
+  if (typeof payload !== "number" && typeof payload !== "string") return null;
+  if (typeof payload === "string" && payload.trim() === "") return null;
+  const parsed = Number(payload);
+  return Number.isFinite(parsed) && parsed >= 0 ? parsed : null;
 }
 
 async function getCumulativeFees(contract: string, height: number) {
@@ -129,16 +124,16 @@ const fetch = async (options: FetchOptions): Promise<FetchResult> => {
   const dailyProtocolRevenue = options.createBalances();
 
   const latest = await getLatestBlock();
-  const previousHeight = await getHeightAt(
-    latest.timestamp - 24 * 3600,
-    latest
-  );
+  const [endHeight, startHeight] = await Promise.all([
+    getHeightAt(options.toTimestamp, latest),
+    getHeightAt(options.fromTimestamp, latest),
+  ]);
 
   const deltas = await Promise.all(
     FEES_SOURCES.map(async ({ contract }) => {
       const [current, previous] = await Promise.all([
-        getCumulativeFees(contract, latest.height),
-        getCumulativeFees(contract, previousHeight),
+        getCumulativeFees(contract, endHeight),
+        getCumulativeFees(contract, startHeight),
       ]);
       if (current < previous)
         throw new Error(
@@ -149,6 +144,8 @@ const fetch = async (options: FetchOptions): Promise<FetchResult> => {
   );
 
   deltas.forEach((delta) => {
+    // get_fees is reported in micro-USD, same 6-decimal scale as get_tvl
+    // (953362900000 -> $953,362.90 against the $953k tvl on defillama)
     const humanReadableFeesUsd = delta / 1e6;
 
     dailyFees.addUSDValue(humanReadableFeesUsd, PERFORMANCE_FEE_LABEL);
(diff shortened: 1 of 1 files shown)


0xshubhs:
took all four.

**use the options window** — right, and this was the real one. i was deriving both snapshots from `getLatestBlock()` and never reading `options`, so any replay would have returned the live rolling 24h instead of the window asked for. now resolves `options.toTimestamp` and `options.fromTimestamp`. verified the run uses the passed window (03 Aug 14:09 -> 04 Aug 14:09).

**reject an unresolved height** — also right. after `MAX_REFINEMENTS` it returned `height` regardless, so a window that never converged would silently produce a delta over some other period. throws now. in practice it converges on the first pass (drift 16s at 0.6758 s/block), so the loop existing at all is the paranoid case and returning an unverified height defeated the point of having it.

**parse the documented field** — fair. the response is `{"data":"3131559123"}` and `queryContract` already unwraps to `res.data`, so the recursion was walking a string it had already been handed. worse, it would have returned the first numeric leaf of *any* object, so if the shape ever grew a sibling key it could have silently read the wrong number as the cumulative total. dropped the recursion, parse the value directly, negatives rejected too:

```
"3131559123"                 -> 3131559123
0        / "0"               -> 0
null / undefined / "" / "   " -> throws
"NaN"                        -> throws
-5       / "-5"              -> throws
{"data":"123","other":"999"} -> throws
{} / [1,2]                   -> throws
```

**document the constants** — added source comments on the probe size, the tolerance and the 1e6 scale. the tolerance and refinement count are ordinary tuning rather than protocol constants, so i explained what they buy instead of inventing a citation. the 1e6 one is worth stating explicitly: `get_tvl` returns `953362900000` for the $953,362.90 tvl shown on defillama, which is what pins the scale.

still $0.00 fees / revenue / protocol revenue, since the contract's cumulative total hasn't moved. typechecks clean, verified twice.

didn't add test fixtures — there's no test harness for adapters in this repo beyond `npm test <type> <adapter>`, so i put the shape table above and in the pr body rather than add a bespoke one.

bheluga:
@0xshubhs thanks for the PR

PR #8581, opened 2026-08-04
hinkal: throw when the relayer returns null instead of publishing it as zero volume

0xshubhs:
`aggregators/hinkal` publishes `$0` whenever the relayer has no number for a day, because the guard checks the wrong thing:

```ts
if (data?.dailyVolume === undefined) {
  console.error(`hinkal: no volume returned for chain ${chainId} (${url})`);
}
return { dailyVolume: Number(data.dailyVolume) };
```

the relayer returns `{"dailyVolume": null}`, not `undefined`. `null === undefined` is false so nothing is logged, and `Number(null)` is `0` and finite, so a missing day is published as a real zero-volume day.

**why it fires every day**

the relayer only answers for the day that's still in progress. every completed day comes back null:

```
2026-07-27 -> null
2026-07-31 -> null
2026-08-02 -> null
2026-08-03 -> null
2026-08-04 -> 28280.17943193073   <- today, still open
2026-08-05 -> 0
```

(arbitrum, `totalVolume/{start}/{end}/42161`. same shape on the other chains. checked repeatedly, it's stable, not flaky.)

so the day the adapter actually asks for is always the one the relayer has already dropped. running master for a completed day gives zeros on all seven chains:

```
chain     | Daily volume
ethereum  | 0.00
base      | 0.00
arbitrum  | 0.00
polygon   | 0.00
optimism  | 0.00
solana    | 0.00
tron      | 0.00
Aggregate | 0.00
```

that's what the published series has been drifting into, 08-01 $2,395 and 08-02 $7,549 against an $80k-ish norm, and 08-03 with no chains reporting at all.

**what this changes**

throws instead, so the day shows as missing rather than as genuinely zero volume. same shape as #8525, #8526, #8538, #8539 and #8547.

a real numeric `0` is still respected as `0`, only null/undefined/blank/unparseable throw:

```
null      -> throw        0        -> 0
undefined -> throw        "0"      -> 0
""        -> throw        1234.5   -> 1234.5
"   "     -> throw        "1234.5" -> 1234.5
"NaN"     -> throw
```

live values still pass through untouched: arbitrum 28280.17943193073, base 4595.546042445872, polygon 9220.633859123749.

also dropped `import { start } from "repl"` at the top of the file, which is unused and pulls in node's repl module.

**worth saying out loud**

if the relayer keeps only serving the in-progress day, this will throw daily and hinkal's chart will stop rather than fill with zeros. i think that's the right signal, it says "the source isn't serving this day" instead of asserting no one traded, but it does mean the fix surfaces the problem rather than hiding it, and the real repair is on their side. the protocol is alive (hinkal.pro, app.hinkal.pro, hinkal.io all up), and there's no non-staging host: `wallet-v2`, `wallet`, `relayer` don't resolve and `api.hinkal.io` 404s, so i left `wallet-staging-v2.hinkal.io` alone rather than guess at a production URL.

**one thing i could not resolve**

two days spike far above the norm and i can't check them, since the relayer no longer serves those dates:

```
2026-07-27   ethereum $10,039,958   arbitrum $2,011,903
2026-07-31   ethereum $17,784,893   arbitrum $2,960,017
```

against $37k-$254k ethereum days either side. hinkal is a privacy protocol so a couple of whale-sized days is entirely plausible, and both chains moving together is consistent with one user routing through both. i'm not claiming it's a bug, just flagging it in case it's a known units issue. it's untouched by this pr.

typechecks clean.

### DIFF
--- a/aggregators/hinkal/index.ts
+++ b/aggregators/hinkal/index.ts
@@ -1,4 +1,3 @@
-import { start } from "repl";
 import { Adapter, FetchOptions } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
 import fetchURL from "../../utils/fetchURL";
@@ -20,11 +19,17 @@ const fetch = async (options: FetchOptions) => {
   const chainId = chainConfig[options.chain].id;
   const url = `${VOLUME_URL}/totalVolume/${startOfDay}/${endTimestamp}/${chainId}`;
   const data = await fetchURL(url);
-  if (data?.dailyVolume === undefined) {
-    console.error(`hinkal: no volume returned for chain ${chainId} (${url})`);
-  }
+  const reported = data?.dailyVolume;
+  const dailyVolume =
+    reported == null || (typeof reported === "string" && reported.trim() === "")
+      ? NaN
+      : Number(reported);
+  if (!Number.isFinite(dailyVolume))
+    throw new Error(
+      `hinkal: relayer returned ${JSON.stringify(reported)} for chain ${chainId} on ${options.dateString}, it only serves the day thats still in progress`
+    );
   return {
-    dailyVolume: Number(data.dailyVolume),
+    dailyVolume,
   };
 };
 


0xshubhs:
the failing `test` check here is the fix doing its job, and it's worth being explicit about that rather than letting it sit red.

ci runs `npm test aggregators hinkal` with no date, so it asks for a completed day. the relayer has no number for a completed day, so the adapter throws, and the workflow greps for `---- ERROR ----`. on master the same run "passes" only because `Number(null)` is `0` — it reports seven chains at $0.00 and exits clean. so the green check on master is the bug, not evidence against this change.

i dug further into the endpoint to see whether the data could be recovered instead of just failing loudly, and it can't. the `end` parameter is ignored entirely, and a value only comes back when `start` is exactly the current day's 00:00 UTC:

```
08-04 00:00 -> now        28280.17943193073
08-04 00:00 -> 12:03      28280.17943193073   <- same value, end ignored
08-03 00:00 -> now        null
08-01 00:00 -> now        null
07-04 00:00 -> now        null
08-03 00:00 -> 08-04      null
```

so it isn't a range api. it's "today's running total, keyed by today's midnight", and every historical start is null. there's no query shape that gets a completed day out of it, which is why i didn't try to repoint or re-window it.

that leaves three options and the choice isn't really mine:

1. **this pr as-is** — the day is recorded as missing instead of as zero volume. ci stays red until hinkal serves historical days.
2. **`runAtCurrTime: true`** — query today's running total, which is the only thing the relayer actually offers. each day would be a partial snapshot taken whenever the run happens rather than a settled daily figure, so it trades one inaccuracy for a smaller one.
3. **drop the adapter** until there's a source that can answer for a finished day.

i'd lean 1, since it's the same shape as #8525/#8526/#8538/#8539/#8547 and it makes the gap visible to hinkal rather than filling their chart with zeros. happy to switch to 2 or 3 if you'd rather — say which and i'll push it.

either way i don't think master should keep publishing $0 days as though nobody traded.

bheluga:
@0xshubhs thanks for the PR

bheluga:
> the failing `test` check here is the fix doing its job, and it's worth being explicit about that rather than letting it sit red.
> 
> ci runs `npm test aggregators hinkal` with no date, so it asks for a completed day. the relayer has no number for a completed day, so the adapter throws, and the workflow greps for `---- ERROR ----`. on master the same run "passes" only because `Number(null)` is `0` — it reports seven chains at $0.00 and exits clean. so the green check on master is the bug, not evidence against this change.
> 
> i dug further into the endpoint to see whether the data could be recovered instead of just failing loudly, and it can't. the `end` parameter is ignored entirely, and a value only comes back when `start` is exactly the current day's 00:00 UTC:
> 
> ```
> 08-04 00:00 -> now        28280.17943193073
> 08-04 00:00 -> 12:03      28280.17943193073   <- same value, end ignored
> 08-03 00:00 -> now        null
> 08-01 00:00 -> now        null
> 07-04 00:00 -> now        null
> 08-03 00:00 -> 08-04      null
> ```
> 
> so it isn't a range api. it's "today's running total, keyed by today's midnight", and every historical start is null. there's no query shape that gets a completed day out of it, which is why i didn't try to repoint or re-window it.
> 
> that leaves three options and the choice isn't really mine:
> 
> 1. **this pr as-is** — the day is recorded as missing instead of as zero volume. ci stays red until hinkal serves historical days.
> 2. **`runAtCurrTime: true`** — query today's running total, which is the only thing the relayer actually offers. each day would be a partial snapshot taken whenever the run happens rather than a settled daily figure, so it trades one inaccuracy for a smaller one.
> 3. **drop the adapter** until there's a source that can answer for a finished day.
> 
> i'd lean 1, since it's the same shape as #8525/#8526/#8538/#8539/#8547 and it makes the gap visible to hinkal rather than filling their chart with zeros. happy to switch to 2 or 3 if you'd rather — say which and i'll push it.
> 
> either way i don't think master should keep publishing $0 days as though nobody traded.

I think current changes are fine for now,
probably the api is dead and team would provide new URL , We should ideally have runAtCurrTime, as historical data looks corrupted but its fine , because we still throw error bcz of null values and refill won't work

0xshubhs:
i'd keep it as-is and not add `runAtCurrTime` — but i have to correct something i said above first.

i claimed no query shape gets a completed day out of the relayer. that's wrong as of today. re-probed this morning and the settled day comes back fine:

```
2026-08-05 00:00 -> 2026-08-06 00:00  = 0            <- today, partial
2026-08-04 00:00 -> 2026-08-05 00:00  = 23900.68     <- yesterday, settled
2026-08-03 00:00 -> 2026-08-04 00:00  = null
2026-08-02 00:00 -> 2026-08-03 00:00  = null         (older all null)
```

so `npm test aggregators hinkal` is green on master right now, $69.5k aggregate across the 7 chains. yesterday the same call was null for the day it asked for, which is why the check went red on this pr.

probing non-midnight starts gives the actual semantics: `end` is ignored, `start` gets rounded **up** to the next midnight, and it returns that single day's bucket.

```
start = now-27h (08-04 00:48) = 0            <- ceils to 08-05, today
start = now-30h (08-03 21:48) = 23900.68     <- ceils to 08-04
start = now-48h (08-03 03:48) = 23900.68     <- also ceils to 08-04
start = now-54h (08-02 21:48) = null
```

only two buckets exist, today and yesterday. so you're right that refill won't work, and it can't be made to work from this endpoint — older days are simply gone.

but that's the argument *against* `runAtCurrTime`, not for it. it makes `startOfDay` today, so the adapter asks for today's bucket, and today's bucket is the partial running total. side by side right now at 03:49 utc:

```
chain      current (settled 08-04)   runAtCurrTime (today)
ethereum         23900.69                  0.00
base              4596.05                  0.00
arbitrum         31760.63                  0.00
polygon           9220.69                  0.00
optimism             0.00                  0.00
solana               0.00                  0.00
tron                10.06                  0.00
TOTAL            69488.12                  0.00
```

0 is a finite number, so the null guard never fires on it. `runAtCurrTime` would start publishing $0 days again and the throw wouldn't catch them this time — the exact thing this pr fixed. it also swaps a settled daily total for whatever happened to have accumulated by the time the runner fired.

as-is, the adapter asks for yesterday, which is the one completed day the relayer keeps, and throws when that bucket is missing instead of writing a zero.

one thing worth watching: yesterday ~17:00 utc the previous day was already null, this morning at 03:49 it's there. so the bucket may get evicted partway through the day, and a late run would throw and leave a gap rather than a zero. that's still the failure i'd want. if it turns out to evict every day i'll revisit, but the fix then is hinkal retaining more history, not us recording partials.

PR #8582, opened 2026-08-04
Update vault start date for muscadine

nickconnelly10:


bheluga:
@nickconnelly10 thanks for the PR
Why do you want to remove v1 owner and change the start date? Looks like it has fee data before 2026 too

nickconnelly10:
> @nickconnelly10 thanks for the PR Why do you want to remove v1 owner and change the start date? Looks like it has fee data before 2026 too

Ill actually justt keep it as it is, no change. Thank you.

PR #8584, opened 2026-08-04
track azverse fees , volume and builders

bheluga:
https://github.com/DefiLlama/dimension-adapters/pull/8572


defiwizard1002:
https://defillama.com/protocol/azverse?tvl=false&dexVolume=true 
@bheluga  thanks for the review, btw when can we see trading volume here? only TVL is shown

bheluga:
@defiwizard1002 should be good

PR #8591, opened 2026-08-05
feat(primit-perps): drop redundant broker_id=primit from factory/orderly.ts

leeleeEcho:
## What

- Remove the `primit` entry from `factory/orderly.ts` — the Orderly-broker channel that previously reported Primit as a sub-protocol on DeFiLlama.
- Update `dexs/primit-perps/index.ts` methodology and inline comments so nothing implies the adapter is restricted to a subset of Primit pairs.

## Why

The existing `dexs/primit-perps` adapter already sums every fill Primit brokers on Avalanche C-Chain — matched on Primit's own engine *and* routed through Orderly Network — from the `TradeRecorder` event stream at `0xC005A9bb11f162329f3EfCCc35F69F9Bb635EeC6`. Keeping the parallel `broker_id=primit` factory entry double-counts the Orderly-routed volume.

## Data source (single source now)

- Event: `TradeRecorded(bytes32 tradeId, address taker, address maker, string symbol, uint8 side, uint256 price, uint256 amount, int256 takerFee, int256 makerFee, bool isClose, uint64 filledAt)` emitted by \`0xC005A9bb11f162329f3EfCCc35F69F9Bb635EeC6\` on AVAX C-Chain.
- **Volume**: sum of \`price × amount / 1e36\` per UTC day, excluding self-trades where \`taker == maker\`.
- **Fees**: \`volume × 4 / 10_000\` (flat 4 bps).
- **Revenue** = **ProtocolRevenue** = Fees. **SupplySideRevenue** = 0.

## Result on DeFiLlama

Primit's protocol page will show a single \`primit-perps\` sub-protocol going forward. Historical Orderly-broker data collected before this merge stays as-is in DeFiLlama's own storage; the change only affects data collected after merge.

### DIFF
--- a/dexs/primit-perps/index.ts
+++ b/dexs/primit-perps/index.ts
@@ -3,17 +3,20 @@ import { CHAIN } from "../../helpers/chains";
 
 // Primit on-chain trade log (Avalanche C-Chain).
 //
-// Every fill matched on Primit's own engine is emitted as a `TradeRecorded`
-// event by our TradeRecorder contract. This adapter walks the day's event
-// range and sums `price * amount` (both 1e18 fixed-point) to derive daily
-// USD notional, then applies the protocol's flat 4 bps trading fee to
-// report fees and revenue. No off-chain data source — anything DeFiLlama
-// shows for Primit is independently reconstructable from AVAX C-Chain
-// event logs.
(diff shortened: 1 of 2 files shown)


bheluga:
@leeleeEcho thanks for the PR
We will disable primit orderly builder internally, let the code remain

PR #8595, opened 2026-08-05
[fix] surge-credit - correct InterestAccrued field names in fees adapter

punithbm:
### What

Corrects the `InterestAccrued` event signature and field mapping in the Surge Credit fees adapter (`fees/surgecredit.ts`).

The adapter declared the event as:

```
InterestAccrued(uint256 indexed marketId, uint256 interestAccrued, uint256 borrowIndex, uint256 supplyIndex)
```

but the deployed LiquidityPool actually emits:

```
InterestAccrued(uint256 indexed marketId, uint256 borrowInterest, uint256 protocolCut, uint256 lenderCut)
```

All parameters are `uint256`, so the event signature hash matched and the log decoded positionally to the correct values, but the field names and the methodology text referenced `borrowIndex` / `supplyIndex` fields that do not exist on the contract.

### Change

Use the real field names and map them explicitly:

- `dailyFees` = `borrowInterest` (total interest)
- `dailyRevenue` / `dailyProtocolRevenue` = `protocolCut` (reserve share, `reserveRateBps`)
- `dailySupplySideRevenue` = `lenderCut` (distributed to USDC LPs)

Methodology text updated to match. Also noted that there is no origination fee (`originationFeeBps == 0` on every market), so borrow interest is the entirety of Fees.

### Values are unchanged

This is a naming and methodology correction only. Because the decode was already positional, the numbers are identical before and after. Local test output for the latest day:

```
Daily fees: 135.19
Daily revenue: 26.96
Daily protocol revenue: 26.96
Daily supply side revenue: 107.95
```

`fees == revenue + supplySideRevenue`, and revenue/fees = 19.9% (the 20% reserve rate).

### DIFF
--- a/fees/surgecredit.ts
+++ b/fees/surgecredit.ts
@@ -3,13 +3,27 @@ import { CHAIN } from "../helpers/chains"
 import { METRIC } from "../helpers/metrics"
 import CoreAssets from "../helpers/coreAssets.json"
 
-// Surge Credit: Bitcoin collateralised USDC lending on Base.
+// Surge Credit: Bitcoin-collateralised USDC lending on Base.
 // https://docs.surge.credit/
+//
+// Fees come entirely from borrow interest. On every state-changing interaction the
+// LiquidityPool accrues interest on the outstanding debt and emits ONE event:
+//
+//   event InterestAccrued(
+//     uint256 indexed marketId,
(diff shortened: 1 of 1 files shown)


bheluga:
@punithbm thanks for the PR

PR #8596, opened 2026-08-05
Add DYLI vault payment wallet

alexneedelman:
### Summary
- add DYLI's vault/warehouse payment wallet to service-fee targets
- count external USDC payments into the wallet as fees/revenue
- include the wallet in internal-address filtering so transfers from the platform wallet are not double-counted

### Context
This wallet is used for DYLI vault/warehouse and 3PL service payments, including crypto payments for vault storage. Transfers from existing DYLI-controlled wallets are excluded by the shared internal sender filter.

### Tests
- pnpm test fees dyli (blocked locally: installed @defillama/sdk is 5.0.216 while the lockfile expects 5.0.220; test runner fails before loading the DYLI adapter with sdk.chainUtils.getDeadChains is not a function)


bheluga:
@alexneedelman thanks for the PR

PR #8606, opened 2026-08-05
synfutures-v3 oi: reject the overview snapshot when it stops being rebuilt

0xshubhs:
`open-interest/synfutures-v3` reads `totalOI` out of a static snapshot and publishes it regardless of age. That file stopped being rebuilt on 2025-12-31, so DefiLlama has shown the same open interest every day for 218 days.

```
2025-12-27  $1,485,146
2025-12-28  $1,451,561
2025-12-29  $1,750,202
2025-12-30  $1,236,190
2025-12-31  $1,251,033   <- last rebuild
2026-01-01  $1,251,033
...
2026-08-05  $1,251,033
```

### the file says so itself

```
$ curl https://api.synfutures.com/s3/config/info-page/v3/overview.json
{
  "updateAt": 1767166306000,     # 2025-12-31T07:31:46Z
  "totalTradingVolume": 320561269526.75867,
  "tvl": 4443188.337688917,
  "totalOI": 1251032.538457101
}
```

`updateAt` matches the day the series went flat exactly. The adapter already receives this field and drops it. The sibling non-v3 file is stamped 2024-08-01, so these are snapshots that get abandoned rather than a live feed, and there's no public replacement, `stats.json`, `markets.json` and `instrument/list` are all 403, `v3/overview` is 403.

### this isn't a dead protocol

SynFutures V3 is trading normally, so `$0`/`deadFrom` would be wrong. Running the on-chain volume adapter for today:

```
npm test dexs synfutures-v3
BASE  Daily volume: 59.61 M     (24/24 hourly slots populated, 1.4M - 2.5M each)
```

So the protocol is alive and only this snapshot is stale. That's the argument for erroring rather than removing the adapter, real open interest exists, it just isn't in this file anymore.

### fix

Throw when the snapshot's own `updateAt` is more than two days behind the window being reported. Two days because the file moved daily right up until it stopped, so anything older is a stale artifact, not publishing lag.

Verified both directions against the live payload by evaluating it at different report times:

```
1h after the snapshot  -> passes, openInterestAtEnd = 1251032.538457101
47h after the snapshot -> passes
49h after the snapshot -> throws
today                  -> throws, "last built 217 days ago"
```

so it isn't a guard that can only fail. Also rejects a non-numeric `totalOI`/`updateAt` before use.

Consequence is that open interest stops rather than continuing flat. The proper long-term fix is to derive OI on-chain from the Instrument contracts the volume adapter already enumerates via `getAllInstruments()`; i didn't do that here because there's no longer a live reference to validate it against, and an unvalidated OI seemed worse than none.

### DIFF
--- a/open-interest/synfutures-v3.ts
+++ b/open-interest/synfutures-v3.ts
@@ -1,9 +1,28 @@
 import fetchURL from "../utils/fetchURL";
 import { CHAIN } from "../helpers/chains";
-import { SimpleAdapter } from "../adapters/types";
+import { FetchOptions, SimpleAdapter } from "../adapters/types";
 
-const fetch = async (_: any) => {
-  let openInterestAtEnd = (await fetchURL("https://api.synfutures.com/s3/config/info-page/v3/overview.json")).totalOI;
+const OVERVIEW_URL = "https://api.synfutures.com/s3/config/info-page/v3/overview.json";
+// the snapshot carries its own build time in `updateAt` and used to move every day,
+// so anything older than two days is a stale file rather than a current reading
+const MAX_SNAPSHOT_AGE = 2 * 24 * 60 * 60;
+
+const fetch = async (options: FetchOptions) => {
+  const overview = await fetchURL(OVERVIEW_URL);
+
+  const openInterestAtEnd = Number(overview?.totalOI);
+  const updatedAt = Number(overview?.updateAt);
+  if (!Number.isFinite(openInterestAtEnd) || !Number.isFinite(updatedAt))
+    throw new Error(
+      `synfutures-v3: unreadable overview snapshot (totalOI ${JSON.stringify(overview?.totalOI)}, updateAt ${JSON.stringify(overview?.updateAt)})`
+    );
+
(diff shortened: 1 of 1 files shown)


0xshubhs:
both taken, the major one was a real hole and it's the same trap this pr exists to fix.

`Number(null)`, `Number("")` and `Number(false)` are all `0` and all finite, so a snapshot that kept a current `updateAt` but lost `totalOI` would have sailed past the guard and published `$0` open interest. i wrote the age check and left the value itself coerced. now both fields must actually be numbers, and negative OI is rejected too.

verified the rejection paths rather than just the happy one, by stubbing the payload:

```
totalOI = null       -> threw
totalOI = ''         -> threw
totalOI = false      -> threw
totalOI negative     -> threw
updateAt = null      -> threw
updateAt = ''        -> threw
good + fresh         -> PASSED  openInterestAtEnd=100
good + 218d stale    -> threw   "last built 217 days ago"
```

live endpoint still throws at 217 days, typecheck clean.

also pulled `SECONDS_PER_DAY` out so the threshold and the message can't drift, thanks.

bheluga:
@0xshubhs thanks for the PR

PR #6607, opened 2026-04-28
[fix] invariant - end unavailable fees source

Genmin:
## Summary
- mark Invariant Solana and Eclipse stats sources dead from 2025-12-09
- DefiLlama daily fees stop on 2025-12-08
- the current Invariant stats endpoints return HTTP 500 / FUNCTION_INVOCATION_FAILED, so current adapter runs should skip instead of failing

## Validation
- npm test -- dexs/invariant
- npm run ts-check

PR #6602, opened 2026-04-28
fix(invariant): update stats API endpoints

Shaileshkhote:
Fixes #6511

## Summary

Switches the Invariant adapter to the new Solana and Eclipse stats endpoints.

## Changes

- updated Solana to the new daily interval endpoint
- updated Eclipse to the new daily interval endpoint
- kept existing `dailyVolume` and `dailyFees` logic unchanged

PR #8607, opened 2026-08-05
invariant: reject the stats snapshot once it stops advancing

0xshubhs:
`dexs/invariant` publishes `volume24`/`fees24` from the stats snapshot without checking whether the snapshot moved. It stopped moving, so the same numbers have been republished every day since 2026-04-29, 99 days of an identical $25,741 volume and $20.52 fees.

```
2026-07-23  $25,741
2026-07-24  $25,741
...
2026-08-05  $25,741
```

### the endpoints are 200 but frozen

Both carry a `timestamp` for the day they cover, and it hasn't advanced since 2026-03-09. Two calls two seconds apart:

```
$ curl '.../eclipse/intervals/eclipse-mainnet?interval=daily'
timestamp 1773057600000  cumVol 598390375.9273833  vol24 20552.947459827006
timestamp 1773057600000  cumVol 598390375.9273833  vol24 20552.947459827006
```

`cumulativeVolume` is pinned too, which it couldn't be if anything were still trading through it. solana 5,187.99 + eclipse 20,552.94 = 25,740.94, exactly the value on the chart.

This is the follow-on from #6602. The old `full_snap` endpoints were 500ing, #6607 proposed marking the source dead, and #6602 repointed to these `intervals` endpoints instead and was merged. They answer 200, so the adapter looks healthy, but they serve a fixed snapshot, arguably worse than the 500 because nothing surfaces it.

### protocol is alive, so deadFrom would be wrong

DefiLlama's own TVL for Invariant is computed independently of this API and still moves daily:

```
2026-08-01  $241,407
2026-08-02  $259,736
2026-08-03  $261,923
2026-08-04  $260,967
2026-08-05  $261,920
```

invariant.app is up. So this is a stale feed on a live protocol, same as the state #6607 was trying to describe, just now hidden behind a 200.

### fix

Throw when the snapshot's own `timestamp` is more than two days behind the window being reported, and reject non-numeric fields before use. Two days because the snapshot advanced daily while it was healthy.

Checked both directions on both chains against the live payload by evaluating at different report times, so it isn't a guard that can only fail:

```
solana   1h after snapshot  -> vol=5187.996657561      fees=0.520793613
solana   47h after snapshot -> passes
solana   49h after snapshot -> throws
solana   today              -> throws, "last advanced 148 days ago"
eclipse  1h after snapshot  -> vol=20552.947459827006  fees=19.769089934
eclipse  47h / 49h / today  -> passes / throws / throws
```

Volume and fees stop rather than repeating a fixed number. If Invariant publishes a working endpoint this only needs the URL swapped, the staleness check can stay.

### DIFF
--- a/dexs/invariant/index.ts
+++ b/dexs/invariant/index.ts
@@ -17,9 +17,13 @@ type StatsApiResponse = {
   };
 };
 
+const SECONDS_PER_DAY = 24 * 60 * 60;
 // the interval snapshot stamps the day it covers and used to advance daily,
 // so anything older than two days is a stale snapshot rather than a fresh one
-const MAX_SNAPSHOT_AGE = 2 * 24 * 60 * 60;
+const MAX_SNAPSHOT_AGE = 2 * SECONDS_PER_DAY;
+
+const readNumber = (value: unknown): number | null =>
+  typeof value === "number" && Number.isFinite(value) ? value : null;
 
 const fetch = async (
   fullSnapEndpoint: string,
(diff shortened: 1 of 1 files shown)


0xshubhs:
taken, and you're right that it was the more serious of the two.

`Number(null)`, `Number("")` and `Number(false)` all give a finite `0`, so a snapshot with a fresh `timestamp` but a missing `volume24.value` would have passed the staleness check and published a $0 day — the exact silent-zero this pr is meant to stop. i guarded the age and left the values coerced.

`volume24.value`, `fees24.value` and `timestamp` now have to be real numbers, and negatives are rejected. checked by stubbing the response rather than only testing the live one:

```
volume24.value = null    -> threw
volume24.value = ''      -> threw
volume24.value = false   -> threw
fees24 missing           -> threw
negative volume          -> threw
timestamp = null         -> threw
all good + fresh         -> PASSED  vol=10 fees=1
```

live endpoints still throw at 148 days on both chains, typecheck clean. also named `SECONDS_PER_DAY` here for the same reason as the sibling pr.

bheluga:
@0xshubhs thanks for the PR

PR #8610, opened 2026-08-05
Add Robinhood Chain to Maestro fee adapter

ya-tls:
Adds Robinhood Chain (4663) to the Maestro fee adapter.

Same fee wallet as the other EVM chains, `0xB0999731f7c2581844658A9d2ced1be0077b7397`. It first received value on Robinhood at block 822076, 2026-07-01 19:11:35 UTC, and has ~53.5k transactions on the chain.

The dispatcher and reward relay are set up the same way as on Base:

- dispatcher `0x2CDF4CAdF2272B77475732446Ba664443277E8C1` was deployed on Robinhood by the fee wallet itself 
- it funds the reward relay `0x5314A41f27BFe2fDe86CCd4B6f08c398Be538D8f` daily via internal transfers, most recently today

No change is needed in `ALLIUM_CHAIN_MAP`: `getAlliumChain` falls through to the chain string, and Allium's schema for this chain is `robinhood`. That resolution path is already in production - `fees/kyberswap-aggregator.ts` has `CHAIN.ROBINHOOD` in its `nativeFeeChains` set, which queries `robinhood.raw.traces` through `getETHReceived`, and it reports Robinhood fees daily.

One thing worth flagging for the first run: this would be the first adapter to query `robinhood.assets.native_token_transfers` specifically. Allium documents the table for this chain and the same query shape works on Sonic, which has no documented assets tables at all, so I expect it to be fine. I don't have an Allium key so I couldn't run `pnpm test fees maestro <date> robinhood` locally.


bheluga:
@ya-tls thanks for the PR

PR #8529, opened 2026-08-01
stonkbrokers: Broker Box fees + volume, lockers, swap-desk

simplefarmer69:
## Summary
Extends the live `fees/stonkbrokers` adapter so Broker Box, Safety Deposit Box lockers, and the Base swap-desk feed DefiLlama **fees / revenue / protocol volume**.

### Broker Box (production floor, 2026-07-31)
**Volume** (`dailyVolume`):
- `PullOpened.ticketWei` — ticket notional across the 9 machines
- `CertificateBought.spendWei` — OTC counter stock purchases
- `SoldBack.ethOut` + `SoldBackUsdg.usdgOut` — sell-back payouts

**Fees / revenue**:
- `EdgeSkimmed` — 10% house edge (5% StockBooster+creator → supply-side, 5% protocol accrual → protocol revenue)
- Certificate Counter flat $2 (`feeWei`) — 50/50 StockBooster / treasury
- 5% sell-back spread implied from 95% payouts (`ethOut × 5/95`) → fees + protocol revenue (bankroll / treasury-reclaimable on official machines)

### Also included
- Safety Deposit Box V3/V4 locker protocol cuts (90/10 brokers / protocol)
- Base Relay swap-desk 1% app fee (USDC outflows from fee wallet)

### Machines
GME / AAPL / AMZN / NVDA / GOOGL / MSFT / SLV / SPCX / USO (hardcoded production addresses).

Companion TVL PR: https://github.com/DefiLlama/DefiLlama-Adapters/pull/20327 (machine ETH + stock inventory).

## Test plan
- [ ] Adapter CI green on this PR
- [ ] Spot-check a recent day: volume ≈ ticket opens + counter spends + sell-backs
- [ ] Fees include EdgeSkimmed + counter + sell-back spread

PR #8635, opened 2026-08-07
stonkbrokers: fix Safety Deposit Box locker fees never recording

simplefarmer69:
The Safety Deposit Box liquidity-locker fee sources added in #8529 have recorded $0 since merge, even though the V3 locker emits `LockFeesCollected` many times a day on-chain.

Root cause: the locker logs were fetched with `targets: [LOCKER_V3, LOCKER_V4]` and then grouped by `log.address` — but `getLogs` defaults to `onlyArgs: true`, so the parsed logs carry no `address` field. Every log grouped under `"undefined"`, the `lockPositions` lookups all failed (`permitFailure: true`), and every locker fee log was silently skipped.

Fix: fetch each locker's logs with a separate single-target `getLogs` call and carry the locker address/version alongside the batch, so no `log.address` is needed. No methodology or attribution changes — the 90/10 broker/protocol split and labels are untouched.

### DIFF
--- a/fees/stonkbrokers/index.ts
+++ b/fees/stonkbrokers/index.ts
@@ -176,7 +176,6 @@ const LOCKER_META: { addr: string; isV4: boolean }[] = [
   { addr: LOCKER_V3, isV4: false },
   { addr: LOCKER_V4, isV4: true },
 ];
-const LOCKER_SET = new Map(LOCKER_META.map((l) => [l.addr.toLowerCase(), l.isV4]));
 
 const fetchRobinhood = async (options: FetchOptions) => {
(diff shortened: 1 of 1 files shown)


simplefarmer69:
Note on the failing `test` check: the public Robinhood RPC (`rpc.mainnet.chain.robinhood.com`) rate-limits the CI runner with 429s partway through the 24-hour backfill loop — same environment failure as #8529 which was merged with the identical failing check. The hourly rows that do complete show the locker fee labels now populating.

bheluga:
@simplefarmer69 thanks for the PR

PR #8639, opened 2026-08-07
sentry: add Ink

mavrkofficial:
Sentry runs the same launch stack on Ink (chain 57073) as on Robinhood Chain, but this adapter only fetched the Robinhood subgraph, so Ink fees and revenue were missing entirely.

**Change**

The endpoint, WETH address and start date move into a per-chain map and the adapter is registered for both chains. Ink is served by its own Goldsky subgraph, which mirrors the Robinhood schema.

Ink currently has launch pools but no fee routers deployed, so `routerDayData` comes back null there and contributes nothing. The existing null handling already covered that case, so both chains share a single query and the fee split logic is unchanged.

**Test**

```
npx ts-node --transpile-only cli/testAdapter.ts fees sentry 1785758400

chain     | Daily fees | Daily revenue | Daily protocol revenue | Daily supply side revenue
robinhood | 1.81 k     | 587.00        | 587.00                 | 1.22 k
ink       | 0.04       | 0.01          | 0.01                   | 0.03
Aggregate | 1.81 k     | 587.01        | 587.01                 | 1.22 k
```

Ink volumes are genuinely small today; the launch stack went live there recently. Verified against a day with activity rather than an empty one, to confirm the chain is actually wired through and not silently returning zero.

### DIFF
--- a/fees/sentry/index.ts
+++ b/fees/sentry/index.ts
@@ -4,9 +4,9 @@ import request, { gql } from "graphql-request";
 import { METRIC } from "../../helpers/metrics";
 import ADDRESSES from "../../helpers/coreAssets.json";
 
-// Sentry (sentry.trading): token launchpad + multi-venue swap frontend
-// on Robinhood Chain. Two fee streams, both indexed by the Sentry
-// Goldsky subgraph:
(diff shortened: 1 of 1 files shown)


bheluga:
@mavrkofficial thanks for the PR

PR #8645, opened 2026-08-07
migrate sui json rpc to graphql

RohanNero:


bheluga (commented):
@RohanNero ```yarn test fees typus-perp 2025-10-03```
works with old code but not with the changes,
any idea why?

RohanNero:
@bheluga Sorry, it seems like queryEvents doesn't index everything, I'll update it to use a different method

RohanNero:
@bheluga I've updated typus-perp to use allium for events since Graphql's queryEvents doesn't return them for historical queries. Should I update the remaining ~13 queryEvents adapters to use allium or should I attempt to use Graphql getObjectsByType at the beginning and end of the test period? (only works if each event increments a retrievable variable)

PR #8654, opened 2026-08-08
[add] wildcard-games - add volume and fees adapter

deva1080:
Adds a volume + fees adapter for **Wildcard Games**, an on-chain casino already
listed on DefiLlama: https://defillama.com/protocol/wildcard-games — the treasury
entry was added in https://github.com/DefiLlama/DefiLlama-Adapters/pull/20361

This is not a new listing, so the form below is not applicable.

**What it tracks**
- `dailyVolume`: total amount wagered, summed from `BetSettled` events across all
  game contracts.
- `dailyFees` / `dailyRevenue`: gross gaming revenue (wagered − paid out), which
  accrues entirely to the protocol treasury acting as house bankroll.

**Chain**: Robinhood Chain.

**Why `allowNegativeValue: true`**: the house edge is a flat 5% total-loss chance
applied globally in the base game contract, so on low-volume days players can win
more than they wager and daily GGR is legitimately negative.


treeoflife2:
Two quick questions before merge:

1. The PR description says the house edge is 4%, but the code comment on allowNegativeValue says 5%. Which is correct?
2. Base has been at zero volume and fees since the 2026-07-17 start (checked 2026-07-24 and 2026-08-07 too). Can you point to a tx with real Base activity, or should Base wait until it sees volume?

deva1080:
Thanks for the review

The PR description says the house edge is 4%, but the code comment on allowNegativeValue says 5%. Which is correct?
 - House edge is 5%.** The code comment is right and I've fixed the PR description.
 
Base has been at zero volume and fees since the 2026-07-17 start (checked 2026-07-24 and 2026-08-07 too). Can you point to a tx with real Base activity, or should Base wait until it sees volume?
- Base has no real activity yet — I've removed it. You're right, and waiting
is the correct call. The protocol launched on Base very recently and the little
that exists on-chain is internal testing denominated in WILD, a non-tradable
internal token with no market price, so it correctly values at $0. There's no
transaction I could point to that represents real volume.

bheluga:
@deva1080 thanks for the PR

PR #8414, opened 2026-07-25
[add] hoodfrens - fees and volume adapter on Robinhood Chain

SoccerEth:
**Name:** hoodfrens
**Website:** https://hoodfrens.xyz (app: https://app.hoodfrens.xyz)
**Category:** Gaming
**Chain:** Robinhood Chain (chainId 4663)
**Twitter:** https://x.com/HoodFrens

## hoodfrens

A fees + volume adapter for **hoodfrens** — creator-card bonding-curve trading on **Robinhood Chain**. hoodfrens runs a creator-fee fork of the same audited contract behind TopStrike, so **this adapter is a clone of the existing [`dexs/topstrike`](https://github.com/DefiLlama/dimension-adapters/blob/master/dexs/topstrike/index.ts) adapter**, with the differences below.

**Contracts:** trading `0xA50AdeC47FcCBDe24B0214A831d3BDde4A1E0106`, pack shop `0x3647b90F769B473E7bCf3267D030E009705fd99D` · **start:** 2026-07-11

### Changes vs the TopStrike adapter
1. **New fee category — creator fees.** Each sell routes a share to the card's creator, emitted as `CreatorFeePaid`/`CreatorFeeAccrued`, summed into `dailySupplySideRevenue` under `METRIC.CREATOR_FEES`. Already inside `Trade.feeInWei`, so it only re-attributes.
2. **Different chain** — `CHAIN.ROBINHOOD` (4663) instead of MegaETH.
3. **Different pack-shop contract, denominated in native ETH.** Own Thirdweb `DropERC1155` (`0x3647…d99D`); same `TokensClaimed` → `getClaimConditionById` mechanism, priced in native ETH (mapped to the gas token) rather than a USD stablecoin (USDM) as on TopStrike.

### Methodology
- **Volume** — gross ETH on card buys/sells (`Trading Volume`) + pack-shop primary mints (`Pack Sales`)
- **Fees / UserFees** — `Trade.feeInWei` (per-trade total) + full pack-mint value
- **HoldersRevenue** — prize pool (`EthPrizeDeposited`), distributed to users holding fractional shares of creator cards
- **SupplySideRevenue** — referral (`ReferralFeePaid`) + creator (`CreatorFeePaid` + `CreatorFeeAccrued`) payouts
- **Revenue** — fees − supply-side (protocol treasury + prize pool + 100% of pack sales)
- **ProtocolRevenue** — revenue − holders (the protocol's own cut)

### Addressed from CodeRabbit review
[...]

PR #8660, opened 2026-08-08
[fix] hoodfrens - value pack claims by tx.value so historical runs work

SoccerEth:
### What

Values pack-shop primary mints by the claim transaction's `tx.value` instead of a `getClaimConditionById` state read. No change to trading volume, fees, or revenue accounting — only how the `Pack Sales` volume line is priced.

### The issue

The hoodfrens **dexs** dimension is missing **2026-07-11 → 2026-07-28** (~$155k of volume; the protocol page currently shows ~$141k cumulative vs ~$296k true). The **fees** dimension is complete for the same days, which looks contradictory since both come from the same `fetch` — the explanation:

- The adapter merged 2026-07-30 (#8414), so all earlier days depend on a historical refill. The fees dimension got one; the dexs dimension has only a single day (07-23) plus live collection from 07-29.
- Days **07-11 → 07-24** have no pack claims, are pure `getLogs`, and re-run fine today — they just need the refill.
- Days **07-25 → 07-28** (pack sales started 07-25) could **never** be refilled with the current code: `getClaimConditionById` runs as an `eth_call` pinned to the day's historical block, and the only RPC configured for `robinhood` (`https://rpc.mainnet.chain.robinhood.com`) **prunes state to a ~100-minute horizon** — older calls fail with `metadata is not found` (verified by binary-searching the state cutoff). Live hourly runs are always inside the horizon, which is exactly why fees stayed complete while dexs lost days; the same loss would recur any time a live run slips more than ~2 hours on a pack-sale day.

### The fix

`tx.value` is exact for this contract: Thirdweb `DropERC1155`'s native-currency claim path requires `msg.value == pricePerToken × quantityClaimed`, and transactions (unlike historical state) are never pruned, so backfills and refills work at any age. Details:

- Values are deduped per transaction — a multicall tx's value covers all its claims.
- Transactions are fetched sequentially via a raw provider: `FetchOptions` has no transaction surface, and `helpers/getTxReceipts`' fixed concurrency-20 pool trips this RPC's burst limiting (336 claim txs on 2026-07-28) — same reason the adapter's getLogs calls are sequential.
- The adapter throws on a claim routed through an intermediary contract rather than guessing. All 711 claims to date are direct EOA→drop transactions, and Σ `tx.value` matches the protocol's internal pack-revenue accounting to the wei (10.1215 ETH).
- **Native-ETH pricing is a stated precondition, guarded fail-loud.** A zero-value claim is ambiguous — genuinely free (allowlisted distributions are real: 128 of 711 claims to date) or an ERC20-priced condition `tx.value` cannot see. The adapter disambiguates via the claim tx's **receipt** (also never pruned): an ERC20-priced claim must move a token in the claim tx, a free claim moves none — and it throws on the former rather than silently scoring a paid claim as zero. Free claims score zero volume by design, so the `Pack Sales` line can legitimately go flat during free-distribution periods.

### Verification (`npm test dexs hoodfrens <date>`)

| UTC day | old code | new code | cross-check |
|---|---|---|---|
| 2026-07-25 | ❌ `metadata is not found` | vol $5.91k · fees $245.12 | stored fees $245 ✓ |
| 2026-07-26 | ❌ | vol $2.83k · fees $175.51 | stored fees $176 ✓ |
| 2026-07-27 | ❌ | vol $21.07k · fees $1.11k | stored fees $1,113 ✓ |
| 2026-07-28 | ❌ | vol $48.23k · fees $2.20k | stored fees $2,196 ✓ |
| 2026-08-05 (control) | collected live → $35,887 stored | $35.72k | old vs new method agree to 0.5% ✓ |
| 2026-08-07 (control, all claims free) | $10,155 stored | $10.18k | receipt guard passes every free claim, packs = 0 ✓ |

The 2026-08-05 control is the equivalence proof: that day was collected live with the old condition-read pricing, and `tx.value` reproduces it (residual is hourly-vs-daily ETH pricing). The 2026-08-07 control exercises the zero-value receipt path on a day where every claim was a free allowlisted distribution.

Per-day **pack volume** (the line this PR re-prices) cross-checked against the protocol's internal sales records, wei-exact:

| UTC day | on-chain tx.value | internal records |
|---|---|---|
| 2026-07-25 | 0.00003 ETH | 0.00003 ✓ |
| 2026-07-26 | 0.00015 ETH | 0.00015 ✓ |
| 2026-07-27 | 0.00029 ETH | 0.00029 ✓ |
| 2026-07-28 | 6.82205 ETH | 6.82205 ✓ |
| all-time Σ | 10.1215 ETH | 10.122 ✓ |

### ⚠️ Refill needed after merge

Please refill the **dexs** dimension for `hoodfrens` from **2026-07-11 → 2026-07-28** inclusive (the fees dimension needs nothing). Without this PR the refill would fail on 07-25 → 07-28; with it, every day runs. Expected daily volumes from an independent on-chain reconciliation (ETH-exact, priced at daily historical rates): 07-11→22 ≈ $56.7k total, 07-23 already stored ($22,567), 07-24 ≈ $20.4k, 07-25→28 ≈ $77.6k — cumulative volume should land ≈ **$296k**.

### DIFF
--- a/dexs/hoodfrens/index.ts
+++ b/dexs/hoodfrens/index.ts
@@ -155,9 +155,20 @@ const fetch = async (options: FetchOptions) => {
             // backoff, like the sequential getLogs calls above.
             const txHashes: string[] = [...new Set<string>(claimLogs.map((l: any) => String(l.transactionHash).toLowerCase()))];
             const provider = getProvider(CHAIN.ROBINHOOD);
+            // Retry budget for per-tx lookups on the public Robinhood RPC.
+            // The RPC rate-limits bursts but recovers within about a second;
+            // there is no published rate-limit policy to cite, so the budget
+            // is empirical: sequential fetches with this backoff completed
+            // 711/711 claim txs (incl. the 336-tx day) without exhausting a
+            // single retry budget, while helpers/getTxReceipts' concurrency-20
+            // pool tripped the limiter. 4 attempts with 500ms linear backoff
+            // (0, 500, 1000, 1500ms) caps a hopeless tx at ~3s before the
+            // fail-loud throw below.
+            const TX_FETCH_ATTEMPTS = 4;
+            const TX_FETCH_BACKOFF_MS = 500;
             const withRetry = async (fn: () => Promise<any>): Promise<any> => {
-                for (let attempt = 0; attempt < 4; attempt++) {
-                    if (attempt) await new Promise((r) => setTimeout(r, 500 * attempt));
+                for (let attempt = 0; attempt < TX_FETCH_ATTEMPTS; attempt++) {
+                    if (attempt) await new Promise((r) => setTimeout(r, TX_FETCH_BACKOFF_MS * attempt));
                     const res = await fn().catch(() => null);
                     if (res) return res;
                 }
--- a/dexs/hoodfrens/index.ts
+++ b/dexs/hoodfrens/index.ts
@@ -1,6 +1,7 @@
 import { FetchOptions, SimpleAdapter } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
 import { METRIC } from "../../helpers/metrics";
+import { getProvider } from "@defillama/sdk";
 
 // hoodfrens is a creator-fee fork of the audited TopStrike bonding-curve
 // contract, running on Robinhood Chain (chainId 4663). This adapter mirrors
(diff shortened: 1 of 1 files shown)


bheluga:
@SoccerEth thanks for the PR
I have refilled the missing days .
You should be able to find the data in UI in a couple of hours.
If it still doesn't match the expected numbers, feel free to ping me

PR #8672, opened 2026-08-10
feat: track Kayen Finance LST fees and revenue

GrapeInTheTree:
Adds a fees and revenue adapter for Kayen Finance LST, liquid staking of native CHZ on Chiliz.

Website: https://app.kayen.finance/lst/stake
Twitter: https://x.com/KayenFinance

TVL adapter for the same protocol: https://github.com/DefiLlama/DefiLlama-Adapters/pull/20432

### Methodology

stCHZ does not rebase. Balances are fixed shares and validator rewards show up as a rise in the CHZ redeemable per share, where the rate is `ChilizDepositor.totalDeposits() / stCHZ.totalSupply()`. `totalDeposits()` is already net of the protocol fee, so the rate delta over a window gives the CHZ that accrued to holders, and grossing it up by the fee rate recovers the total validator rewards.

- Fees: total CHZ staking rewards earned by the staked CHZ
- Revenue: the protocol fee cut of those rewards, currently 10%, read from `ProtocolConfig.protocolFeeRate()` rather than hardcoded so a governance change is picked up automatically
- Supply side revenue: the remainder, which reaches stCHZ holders as a higher redemption rate
- Holders revenue: none, no rewards are shared with token holders

All values come from contract calls at the window's from and to blocks. No subgraph or off-chain API is involved.

### On attributing rewards to the right shares

Worth flagging since it changes the numbers by a lot. The rate is a step function rather than a smooth curve: validator rewards land in one lump roughly once a day and the rate is flat in between. So the shares that earned a step are the ones outstanding when it was earned, and the adapter prices each step against the supply at the window's opening block.

Pricing against the closing supply instead would credit depositors who arrived after the reward was earned. On 2026-08-08, when the pool grew from 7.1M to 8.9M CHZ in a day, that difference is 2,501 CHZ reported against 1,999 CHZ actually earned, a 25% overstatement. `pullHourly: true` keeps the windows short enough that the remaining error is bounded by one hour of deposit flow.

### Verification

Cross-checked against the protocol's subgraph for the window 2026-08-08 00:00 to 2026-08-09 00:00 UTC:

| | reward accrued to holders |
|---|---|
| adapter, from contract calls | 1,999.72 CHZ |
| subgraph hourly snapshots | 1,999.72 CHZ |

The same window read straight from the chain at the subgraph's own snapshot blocks reproduces the rate to 16 significant digits.

`npm test fees kayen-finance-lst` output for a few days:

```
2026-08-09   fees $29.00   revenue $2.87   supply side $26.00
2026-08-05   fees $11.00   revenue $1.06   supply side $9.58
2026-07-20   fees  $0.00   revenue $0.00   supply side  $0.00
```

The zero on 2026-07-20 is correct rather than a failure. The pool held 23.8 stCHZ that day and earned 0.0062 CHZ.

### DIFF
--- a/fees/kayen-finance-lst.ts
+++ b/fees/kayen-finance-lst.ts
@@ -68,7 +68,7 @@ const fetch = async (options: FetchOptions) => {
 
 const methodology = {
   Fees: 'Total CHZ staking rewards earned by the CHZ staked through Kayen Finance LST.',
-  Revenue: 'Protocol fee taken from those rewards, currently 10%, read from ProtocolConfig.',
+  Revenue: 'Protocol fee taken from those rewards, at the rate ProtocolConfig holds at the time of the call.',
   ProtocolRevenue: 'All revenue goes to the protocol fee vault.',
   HoldersRevenue: 'No share of the rewards goes to token holders.',
   SupplySideRevenue: 'The rest of the rewards accrue to stCHZ holders through the rising CHZ per stCHZ rate.',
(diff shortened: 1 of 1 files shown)


bheluga:
@GrapeInTheTree thanks for the PR

PR #8681, opened 2026-08-10
fix: normalize adapter time windows to half-open intervals

### LINKED ISSUE #8656: Adjacent adapter time windows overlap and can double-count boundary activity
## Description

`FetchOptions` currently mixes query boundaries with inclusive state-snapshot timestamps.

For a run ending at `E` with window size `W`, the runner currently derives:

```text
toTimestamp    = E - 1
fromTimestamp  = E - W - 1
startTimestamp = fromTimestamp
endTimestamp   = E
```

Many adapters then query both boundaries inclusively:

```sql
timestamp >= startTimestamp
AND timestamp <= endTimestamp
```

This produces the effective interval:

```text
[E - W - 1, E]
```

Instead of the intended half-open interval:

```text
[E - W, E)
```

As a result, adjacent daily and hourly runs overlap.

## Example

Two intended adjacent daily windows are:

```text
July 31: [2026-07-31 00:00:00, 2026-08-01 00:00:00)
August 1: [2026-08-01 00:00:00, 2026-08-02 00:00:00)
```

The current construction can produce:

```text
July 31: [2026-07-30 23:59:59, 2026-08-01 00:00:00]
August 1: [2026-07-31 23:59:59, 2026-08-02 00:00:00]
```

These windows overlap from `2026-07-31 23:59:59` through `2026-08-01 00:00:00`.

## Real boundary transaction

Ethereum block [25,656,292](https://eth.blockscout.com/block/25656292) has timestamp:

```text
2026-07-31T23:59:59Z
```

It contains 175 transactions, including this successful [Uniswap Universal Router transaction](https://eth.blockscout.com/tx/0x0b5ff42496f56e08b892d9bb1210ebc19f441eb6e3b0d4567249e2377996eaae).

That transaction falls inside both old adjacent daily predicates. With half-open windows, it belongs only to the July 31 window.

The same issue affects block-based log queries: the July 31 `toBlock` and August 1 `fromBlock` both resolve to block `25,656,292`, so the entire block can be queried twice.

## Proposed contract

Event and activity queries should use:

```text
[startTimestamp, endTimestamp)
```

The fields would have these meanings:

| Field | Value | Purpose |
| --- | --- | --- |
| `startTimestamp` | `endTimestamp - windowSize` | Inclusive query boundary |
| `endTimestamp` | Run timestamp | Exclusive query boundary |
| `fromTimestamp` | `startTimestamp - 1` | Inclusive pre-window state snapshot |
| `toTimestamp` | `endTimestamp - 1` | Inclusive final state snapshot |

This preserves the existing cumulative-state behaviour:

```text
fromTimestamp = startTimestamp - 1
toTimestamp   = endTimestamp - 1
```

Adapters calculating deltas through `toApi - fromApi` would retain their existing snapshot semantics.

Default log ranges should start after the pre-window snapshot block, while the public snapshot block getters should retain their current meaning.

## Required adapter migrations

- SQL predicates: `>= start AND < end`
- GraphQL predicates: `_gte: start` and `_lt: end`
- `BETWEEN` predicates: replace with explicit half-open comparisons
- Date-only partition filters: use the date containing `toTimestamp`
- Daily snapshot queries: exclude the bucket at `endTimestamp`
- Existing `startTimestamp + 1` workarounds: remove where the downstream lower predicate is inclusive
- Default block-log queries: begin after the pre-window snapshot block

## Historical-data impact

Previously stored values may contain:

- duplicated records at adjacent boundaries;
- duplicated boundary blocks in log-based adapters;
- more than exactly 24 hours of data in daily queries;
- more than exactly one hour in hourly queries;
- an additional daily bucket when an exclusive midnight endpoint is cast to a date.

After implementing the change, affected adapters may require historical backfills.

Event-based adapters will usually differ only when activity exists at a boundary. Daily snapshot queries may have larger differences and should be prioritised.

chetna-mittal:
Fixes #8656 

## Summary

Normalise adapter activity windows to:

```text
[startTimestamp, endTimestamp)
```

The runner previously exposed `startTimestamp` as the pre-window snapshot second while `endTimestamp` was already exclusive. Adapters using inclusive SQL, GraphQL, API, or block predicates could consequently overlap with adjacent daily and hourly runs.

This PR separates query boundaries from state-snapshot boundaries and migrates affected consumers to the same contract.

## Contract

```text
startTimestamp = endTimestamp - windowSize
fromTimestamp  = startTimestamp - 1

toTimestamp    = endTimestamp - 1
```

Therefore:

- `startTimestamp` is inclusive.
- `endTimestamp` is exclusive.
- `fromTimestamp` remains the state immediately before the window.
- `toTimestamp` remains the final state inside the window.

## Real duplicate example

Ethereum block [25,656,292](https://eth.blockscout.com/block/25656292) occurred at:

```text
2026-07-31T23:59:59Z
```

It contains this successful [Uniswap Universal Router transaction](https://eth.blockscout.com/tx/0x0b5ff42496f56e08b892d9bb1210ebc19f441eb6e3b0d4567249e2377996eaae).

The block is selected by both old adjacent daily windows. Under the new contract, it belongs only to the July 31 window, and the August 1 default log range starts at the following block.

## Compatibility

The following retain their existing numeric values and snapshot meaning:

- `fromTimestamp`
- `toTimestamp`
- `fromApi`
- `toApi`
- `getFromBlock`
- `getToBlock`
- `startOfDay`
- `dateString`

Cumulative adapters using `toApi - fromApi` do not need to change.

## Historical impact

Existing stored values may include duplicated boundary records or blocks.

After merge, affected historical ranges should be backfilled. Daily snapshot adapters should be prioritized because casting the old exclusive endpoint to a date could select an additional daily bucket.


0xshubhs:
Worked through the contract independently against master and it holds, including the
block path — which is the part that is easy to miss, so noting the derivation in case
it is useful for review.

**Why the block path mattered.** On master, `getLogs` defaults to
`fromBlock = getBlock(fromTimestamp)` and `toBlock = getBlock(toTimestamp)`, with
`toTimestamp = E - 1`. For the next adjacent run at `E + W`:

```
fromTimestamp = (E + W) - 1 - W = E - 1
```

So run N's `toBlock` and run N+1's `fromBlock` both resolve `getBlock(E - 1)` — the
same timestamp, hence the same block — and `eth_getLogs` is inclusive at both ends.
That boundary block landed in both windows for **every** log-based adapter, not just
the handful writing inclusive SQL predicates.

**The new contract closes it.** With `getQueryStartBlock` = `getLogStartBlock(getFromBlock())`
= `getBlock(fromTimestamp) + 1`:

```
run N     logs over [ getBlock(E-W-1)+1 , getBlock(E-1) ]
run N+1   logs over [ getBlock(E-1)+1   , getBlock(E+W-1) ]
```

Run N ends at `getBlock(E-1)` and run N+1 starts at `getBlock(E-1)+1`. No shared
block, and no gap either — consecutive runs tile the block space exactly, regardless
of where `getBlock` lands relative to the nominal second. Same on the SQL side:
`[E-W, E)` then `[E, E+W)`.

**On file coverage** — I grepped for inclusive `endTimestamp` predicates and got seven
files, six of which are in this PR. The seventh, `fees/whales-market.ts`, is correctly
left alone: its `from <= options.endTimestamp` selects which fee wallets from a
timeline were active in the window rather than bounding a data query, and the fees
themselves come from `addTokensReceived` on the block range. Worth stating explicitly
since it reads like an omission from the outside.

bheluga:
Added [here](https://github.com/DefiLlama/dimension-adapters/pull/9073)

PR #8682, opened 2026-08-10
0x: add cross-chain bridge volume adapter

lukejocz:
Adds 0x cross-chain (bridge) volume. 0x is already tracked as a DEX aggregator by `aggregators/zrx`; this is a separate dataset, so the two do not overlap. Not a new protocol listing, so the template form is replaced with details, per its note.

**Source:** `GET https://api.0x.org/stats/cross-chain/volume/daily`, a public endpoint 0x added for this listing. Returns USD volume of filled cross-chain trades for one UTC day, filtered by origin chain. Attributed to the origin chain, as `bridge-aggregators/lifi` does. `version: 1` because the endpoint only serves daily aggregates keyed to UTC midnight.

**Auth:** reuses `AGGREGATOR_0X_API_KEY`, already in `helpers/env.ts` and already used by `aggregators/zrx`, so no new secret. The `test` check fails with a 401 because CI injects no keys — same as previous 0x PRs (e.g. #7964, which failed on `DUNE_API_KEYS`). 

**Rate limiting:** the API allows 5 req/s per key and `runAdapter` fetches all chains at once, so an unthrottled 21-chain run had 16 requests rejected with a 429. Requests are serialised with `p-limit` and spaced, with `asyncRetry` as a safety net, following `bridge-aggregators/okx`. 

**Chains:** 21 EVM origin chains. Solana and Tron are commented out because 0x does not yet collect trades originating on non-EVM chains.

**Verified** against the live API with a key: `2026-08-05` returns all 21 chains, aggregate $1.18M. Data goes back to the `start` of 2026-02-25 (private beta; GA was 2026-06-04).

### DIFF
--- a/bridge-aggregators/zrx/index.ts
+++ b/bridge-aggregators/zrx/index.ts
@@ -1,7 +1,15 @@
+import asyncRetry from "async-retry";
 import { CHAIN } from "../../helpers/chains";
 import { FetchOptions, SimpleAdapter } from "../../adapters/types";
 import { getEnv } from "../../helpers/env";
 import { httpGet } from "../../utils/fetchURL";
+const plimit = require("p-limit");
+
+// Every chain is fetched at once but the stats API allows 5 requests per second per
+// key, so requests are serialised and spaced to stay under the cap even if the API
+// responds instantly.
+const limits = plimit(1);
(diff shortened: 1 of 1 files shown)


bheluga:
@lukejocz thanks for the PR

PR #8683, opened 2026-08-10
cmswap: add volume and fees/revenue adapters

CableKungZ:
## Summary
- `dexs/cmswap.ts`: daily swap volume for Junoswap.trade (DefiLlama TVL slug `cmswap`) on Bitkub Chain — routed through the `AggRouterJunoswap` aggregator plus buy/sell volume on the `BondingCurveJunoswap` launchpad.
- `fees/cmswap.ts`: daily fees/revenue from the aggregator's router-skim fee, the launchpad's pump fee, and the flat token-creation fee. All fees accrue 100% to the protocol treasury, so Revenue = Fees = Protocol Revenue.

## Methodology notes
- The launchpad's `Swap` event emits `amountIn` net of the pump fee, so the fee is backed out using the live `pumpFee()` value (`fee = amountIn * pumpFee / (10000 - pumpFee)`) rather than assumed.
- Fee/creation-fee contract reads use a fresh current-state `ChainApi` rather than the framework's block-pinned one, since Bitkub Chain's public RPC isn't archive-capable and rejects `eth_call` at historical blocks.
- Aggregator router fee (`feeBps`) is currently 0 on-chain (mechanism exists, not yet turned on), so aggregator fee revenue is $0 until it's enabled — expected, not a bug.

## Test plan
- [x] `npx ts-node --transpile-only cli/testAdapter.ts dexs cmswap` — runs clean, correctly reports non-zero volume on historical dates with known launchpad activity.
- [x] `npx ts-node --transpile-only cli/testAdapter.ts fees cmswap` — runs clean on both current and historical dates after fixing the archive-node issue.
- [x] `tsc --project tsconfig.json --noEmit` — no errors in either file.

### DIFF
--- a/dexs/cmswap.ts
+++ b/dexs/cmswap.ts
@@ -1,0 +1,58 @@
+/**
+ * cmswap / Junoswap.trade — DEX Aggregator + bonding-curve launchpad volume, on Bitkub Chain.
+ *
+ * Two independent volume sources, both routed through contracts owned by the protocol:
+ *  - AggRouterJunoswap: the multi-DEX aggregator router users swap through
+ *    (routes across Udonswap, KUBLERX, Diamond, Jibswap and Junoswap's own V3 pools).
+ *  - BondingCurveJunoswap: the memecoin launchpad's buy/sell bonding curve.
+ *
+ * Junoswap's own V3 pools (tracked for TVL under projects/cmswap in DefiLlama-Adapters)
+ * mainly host graduated launchpad tokens; their raw swap volume is not counted here since
+ * routed trades already surface as Aggregated events above.
+ */
+import { Adapter, FetchOptions } from "../adapters/types";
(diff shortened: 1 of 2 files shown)


CableKungZ:
Pushed a fix: bonding-curve launchpad volume on buys was being silently undercounted. \`addOneToken\`'s core-asset heuristic falls back to whichever side isn't in the chain's \`coreAssets.json\`, but bitkub's list doesn't include the native KUB placeholder address — so on every buy it was pricing off the (usually unpriced) new memecoin side instead of the native side. Since one side of every bonding-curve trade is always native KUB, now pricing off that side directly instead of relying on the heuristic. Verified against several historical dates where fees showed real activity but volume showed $0 or was understated before the fix.

bheluga:
@CableKungZ thanks for the PR
Added [here](https://github.com/DefiLlama/dimension-adapters/pull/8704)

PR #8686, opened 2026-08-10
bluefin7k: throw when 7k.ag returns 0 instead of publishing it as no volume

### LINKED ISSUE #8685: bluefin7k-aggregator: statistic.7k.ag has returned 0 since 2026-06-18, published as 54 days of $0 volume
`aggregators/bluefin7k-aggregator` has published an explicit `$0` every day since 2026-06-18 — 54 days — while the aggregator is still routing trades on Sui.

## What the adapter does

```ts
const URL = 'https://statistic.7k.ag';

const fetch: FetchV2 = async ({ fromTimestamp, toTimestamp }) => {
	const dailyVolume = await fetchURL(
		`${URL}/volume-with-ts?from_timestamp=${fromTimestamp}&to_timestamp=${toTimestamp}`,
	);
	return { dailyVolume };
};
```

Whatever the endpoint returns is published as-is.

## The endpoint returns the literal string "0"

It is up and answers 200, and it still serves real numbers for older ranges — it just has nothing after 2026-06-17:

```
2026-06-15  "63435.899484155732722392361771..."
2026-06-17  "49063.135086680428629794442134..."
2026-06-18  "0"
2026-07-01  "0"
2026-08-09  "0"
```

Its cumulative total is frozen. Querying `from_timestamp=0` to just after the break and to now gives the same figure to the last decimal:

```
0 -> 2026-06-18 : "15974325942.8259963079088946978162945420254333354128833778668063647643293734290637075901031494140625"
0 -> now        : "15974325942.8259963079088946978162945420254333354128833778668063647643293734290637075901031494140625"
2026-06-18 -> now : "0"
```

So the upstream has ingested nothing for 54 days. `api.7k.ag` is also down at the origin (Cloudflare 522) — the statistics service looks broken rather than empty.

## The aggregator is still live

This is the part that matters, and it is worth stating because a few other signals point the wrong way (`7k-ag/7k-sdk-ts` is archived, the package IDs in it are commented out, `api.7k.ag` is dead).

On-chain says otherwise. The 7K aggregator package `0x62412b7268c35f3808336aee57a52836501f40b8ba5d936f8ad275e672befd04` is transacting continuously:

```
EkBCSNhgrd7mba8Rayz1bT18QY4zd7gFZ1ZQKAcPgSTs  2026-08-10T16:18:20Z
FyjUBPsQxNPn9Ph75WmSQpaAUisaF6iXi5hauuTHugpA  2026-08-10T16:17:52Z
```

Seconds apart, minutes before filing. Sui peers are also reporting normally over the same window (Cetus $6.4M/24h, FlowX $1.1M, Aftermath $253k), so this is not a chain-wide or pipeline-wide gap.

## Zero has never been a real value here

Over the 379 days before the break the minimum non-zero day is **$49,063** and the median is **$11.9M**. There are 32 zero days in that period, but they sit in tight clusters (2026-01-01..04, 2026-03-27..04-01) that look like earlier outages of this same endpoint rather than quiet trading days.

So `0` from this endpoint has always meant "no data", never "no volume".

## Suggested fix

Throw when the endpoint returns no volume, instead of recording it as a real zero — the same treatment as #8525 (xrpl-dex), #8526 (lotus-finance), #8538 (kalshi), #8539 (insightx) and #8547 (pact). The 54 zero days stay wrong until the upstream refills, but the adapter stops adding to them.

I have the change ready and will open a PR against this issue.

0xshubhs:
Fixes #8685.

`statistic.7k.ag` has returned the literal string `"0"` since 2026-06-18, and the adapter published it, so the aggregator has recorded 54 consecutive days of `$0` volume while it is still routing trades on Sui.

The endpoint is up and still serves real numbers for older ranges; it just stopped ingesting:

```
2026-06-17  "49063.135086680428629794442134..."
2026-06-18  "0"
2026-08-09  "0"
```

and its cumulative total is frozen to the last decimal between 06-18 and now (`"15974325942.82599630790889469781629454202543333541288337786680636476432937342906370759010314941406 25"` either way), with the delta between them returning `"0"`.

The aggregator itself is live; package `0x62412b7268c35f3808336aee57a52836501f40b8ba5d936f8ad275e672befd04` was transacting seconds apart minutes before I filed the issue, and Sui peers report normally over the same window. Full evidence in #8685.

Zero has never been a real value from this endpoint: over the 379 days before the break the minimum non-zero day is $49,063 and the median is $11.9M, and the 32 zero days in that period sit in tight clusters that look like earlier outages of the same service.

Throwing rather than recording the zero, matching #8525, #8526, #8538, #8539 and #8547. `!(Number(x) > 0)` so a null, empty or missing body is caught the same way as `"0"`.

Verified both directions, twice each:

```
$ npm test aggregators bluefin7k-aggregator 2026-08-05
Error: bluefin7k: statistic.7k.ag returned no volume (0) for 1785801599-1785887999

$ npm test aggregators bluefin7k-aggregator 2026-06-16
Daily volume: 63.44 k
```

which matches what the endpoint serves for that day (`63435.899…`). `npm run ts-check` passes.

The 54 days already recorded stay wrong until the upstream refills; this only stops more from being added.

### DIFF
--- a/aggregators/bluefin7k-aggregator/index.ts
+++ b/aggregators/bluefin7k-aggregator/index.ts
@@ -9,9 +9,10 @@ const fetch: FetchV2 = async ({ fromTimestamp, toTimestamp }) => {
 	const dailyVolume = await fetchURL(
 		`${URL}/volume-with-ts?from_timestamp=${fromTimestamp}&to_timestamp=${toTimestamp}`,
 	);
-	if (!(Number(dailyVolume) > 0))
+	const volume = Number(dailyVolume);
+	if (!Number.isFinite(volume) || volume <= 0)
 		throw new Error(
-			`bluefin7k: statistic.7k.ag returned no volume (${dailyVolume}) for ${fromTimestamp}-${toTimestamp}`,
+			`bluefin7k: statistic.7k.ag returned no usable volume (${dailyVolume}) for ${fromTimestamp}-${toTimestamp}`,
 		);
 	return { dailyVolume };
 };
--- a/aggregators/bluefin7k-aggregator/index.ts
+++ b/aggregators/bluefin7k-aggregator/index.ts
@@ -9,6 +9,10 @@ const fetch: FetchV2 = async ({ fromTimestamp, toTimestamp }) => {
 	const dailyVolume = await fetchURL(
 		`${URL}/volume-with-ts?from_timestamp=${fromTimestamp}&to_timestamp=${toTimestamp}`,
 	);
+	if (!(Number(dailyVolume) > 0))
+		throw new Error(
+			`bluefin7k: statistic.7k.ag returned no volume (${dailyVolume}) for ${fromTimestamp}-${toTimestamp}`,
+		);
 	return { dailyVolume };
 };
 


0xshubhs:
Taken — `Number("Infinity") > 0` is `true`, so an `Infinity` payload would have gone straight through and been published as volume.

Parsing once and requiring a finite positive number now. Table over the shapes this endpoint could plausibly return:

| payload | old `!(Number(x) > 0)` | new `!isFinite \|\| <= 0` |
| --- | --- | --- |
| `"0"` | reject | reject |
| `""` / `"   "` | reject | reject |
| `null` / `undefined` | reject | reject |
| `"NaN"` | reject | reject |
| `{}` / `[]` | reject | reject |
| **`"Infinity"`** | **pass** | **reject** |
| **`"1e999"`** | **pass** | **reject** |
| `"63435.89"` | pass | pass |

`Infinity` and an overflowing exponent were the only two leaks; everything else was already covered by `Number()` coercing to `0` or `NaN`.

Still returning the raw `dailyVolume` rather than the parsed number, since the endpoint sends long decimal strings and reparsing would cost precision for no reason.

Re-verified both directions:

```
$ npm test aggregators bluefin7k-aggregator 2026-08-05
Error: bluefin7k: statistic.7k.ag returned no usable volume (0) for 1785801599-1785887999

$ npm test aggregators bluefin7k-aggregator 2026-06-16
Daily volume: 63.44 k
```

`npm run ts-check` passes.

bheluga:
@0xshubhs thanks for the PR

PR #8697, opened 2026-08-11
feat(tristero): single contract set, indexer-free volume, new fee model

t-cm:
- discovery moves from the ClickHouse indexer to a topic-filtered getLogs
  on ERC20 transfers into the router; both SQL queries are gone
- only the current router, escrow and vault are referenced; all legacy
  contract generations and the delegated send executor are removed
- darkpool volume stays on dexs/tristero, routed flow moves to a new
  aggregators/tristero listing; both counted once, on the source leg
- exclude self-filled darkpool orders (sender == filler)
- fees are a 1bp/day loan spread on vault capital plus gas abstraction;
  fees/tristero-margin is folded into fees/tristero
- retry transient nulls from getTxReceipts/getTransactions
- pullHourly set explicitly on all adapters; open interest and the loan
  spread are stocks and were being summed 24x

**NOTE**

### 1. Two listing changes needed alongside this merge

- `aggregators/tristero` needs a **Tristero Aggregator** child protocol under
  `parent#tristero`, or it merges and reports nowhere.
- `fees/tristero-margin.ts` is deleted (folded into `fees/tristero.ts`), so the **Tristero
  Margin** fees listing needs retiring or re-pointing. It has never emitted a non-zero
  value in 24 points, so nothing is lost.


treeoflife2:
Pushed a fix on top: darkpool fills were booking both the taker's source leg and the filler's destination leg for the same trade (double counting - a $100 fill was reporting $200). Now only the source leg is counted, matching how every other matched-order venue on the dashboard is tracked. Also dropped the invalid `UserFees` methodology key (folded that sentence into `Fees`) and collapsed the per-chain start maps to `chains` + a single `start` since all six chains share one start date.

0xshubhs:
Some production data that supports this, from a sweep I was running on stalled
dimension charts — none of it changes the diff, it just backs up two of the choices.

**The live adapter stopped six days before this PR was opened.**
`api.llama.fi/summary/dexs/tristero-spot?dataType=dailyVolume` ends at **2026-08-05**
and has written nothing since:

```
points 248   first 2025-12-01   lastPoint 2026-08-05
```

It stops rather than writing zeros, which is the shape of the adapter erroring out
rather than genuinely reporting no volume — consistent with moving discovery off the
ClickHouse indexer.

**`fees/tristero-margin` has never published a non-zero value.** 24 points since
2026-07-20, every one of them `0`:

```
lastNonZero  NEVER      zeroTail 24/24
08-03=0 08-04=0 ... 08-11=0 08-12=0
```

So folding it into `fees/tristero` is removing a listing that has only ever emitted
zeros, not a working series.

**One more thing worth a look while the fee model is being reworked** — the existing
volume series is unstable in a way that looks mechanical rather than real:

```
07-27  $31,966,507
07-28  $1,594
07-29  $57,619,032
07-30  $31,961,345
07-31  $11
08-01  $203
08-02  $27,885,844
```

$11 and $203 days sitting between $30M days, with 26 zero days out of 248 overall.
If the new getLogs path is expected to be complete where the indexer was partial,
those days would be the ones to spot-check after a refill.

t-cm:
1.  retryByHash moved into helpers/getTxReceipts.ts as getTxReceiptsWithRetry / getTransactionsWithRetry — it fixes a provider-rotation failure (0–2 null receipts per run on a fixed set of 5 transactions) for every adapter using those helpers, not just this one.
Fill discovery memoised per chain + block range; dexs, aggregators and fees were each repeating the log scan and transaction fetch.
pullHourly set explicitly on all four adapters. Open interest and the loan spread are stocks, not flows — under hourly slots they were being summed 24× (open interest reported 126.76k against an actual 5.19k).
Stale router/legacy comments removed, RPC fallback keyed on CHAIN.BASE, single-order decode inlined, UserFees folded into Fees, per-chain maps collapsed to chains + start.
Reported volume drops ~99%, deliberately. A lot of the historical volume was one wallet. This PR excludes darkpool fills  where sender == filler. Volume will go back to normal once darkpool is pushed. 

2. day	published (old)	new dexs	new aggregators
2026-07-29	$57,619,032	$11	$1.09k
2026-07-31	$11	$0	$1.01k
2026-08-02	$27,885,844	—	$50.47

3. Two listing changes needed alongside this merge:

aggregators/tristero needs a Tristero Aggregator child protocol under parent#tristero, or it merges and reports nowhere.
fees/tristero-margin.ts is deleted (folded into fees/tristero.ts), so the Tristero Margin fees listing needs retiring or re-pointing. Per @0xshubhs's data it has never emitted a non-zero value in 24 points, so nothing is lost.

@coderabbitai review

PR #8711, opened 2026-08-11
accountable: add robinhood chain

pawel-accountable:
Adds Robinhood Chain to the Accountable fees adapter, same pattern as the other chains — factory addresses match the ones already used in the TVL adapter (`projects/accountable/index.js`), start date is the factory deployment date (2026-07-22).

Currently one active vault there (via the yield factory), NAV-priced (share price path), matching what the adapter already handles for other chains.

Note per our chat: the public RPC (`rpc.mainnet.chain.robinhood.com`) only retains a few minutes of state, so this will fail against it — the run needs the premium RPC on your side to reach day-start state. Let me know how the test goes.

### DIFF
--- a/fees/accountable.ts
+++ b/fees/accountable.ts
@@ -41,6 +41,14 @@ const config: Record<string, { factories: string[], start: string }> = {
     ],
     start: "2026-02-13",
   },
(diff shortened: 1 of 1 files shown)


bheluga:

Issue #8075, opened 2026-07-11
strata-markets: Fee adapter shows extreme spikes (16-41x expected yield, zero on-chain events on spike days)


radeveth:
## Problem

The `strata-markets` fee adapter (`fees/strata-markets/index.ts`) shows periodic extreme fee spikes that are **16–41x the expected daily yield**. On-chain investigation confirms zero protocol activity on the spike days, meaning the inflated fees cannot come from real user actions.

## Spike Data

| Date | Daily Fee | TVL | Expected (~7% APY) | Multiple |
|---|---|---|---|---|
| Jun 29 | **$971,259** | $65M | ~$12K | **78x** |
| Jun 25 | $649,534 | $83M | ~$16K | **41x** |
| Jun 08 | $549,645 | $92M | ~$18K | **31x** |
| Jun 18 | $501,121 | $87M | ~$17K | **30x** |

Median daily fee: **$21,132**. There are **10 days with fees >10x median** in the last ~5 weeks.

## On-Chain Investigation

I queried every relevant event across **all 6 CDO strategies** (30 contracts total) for the spike dates. Results:

| Event | Contracts | Jun 8 | Jun 18 | Jun 25 | Jun 29 |
|---|---|---|---|---|---|
| `FeeAccrued` | 6 accounting contracts | 0 | 0 | 0 | 0 |
| `Deposit` | 12 vault contracts | 0 | 0 | 0 | 0 |
| `Withdraw` | 12 vault contracts | 0 | 0 | 0 | 0 |
| `ReserveReduced` | 6 CDO contracts | 0 | 0 | 0 | 0 |

**Zero events = zero protocol activity.** No deposits, withdrawals, fee accruals, or reserve distributions on any spike day.

Since all event-based terms in the formula are zero, the fee reduces to:

```
dailyFees = navEnd - navStart
```

This is purely the `strategy.totalAssets()` NAV delta — reflecting passive yield accrual (e.g., sUSDe exchange rate growth). Normal daily yield at 7% APY on $65M ≈ $12K, **not $971K**.

## Pattern Analysis

The spikes **cluster on Monday and Thursday** and show a decay pattern:

```
Mon Jun 8:  $550K → Tue: $264K → Wed: $173K → Thu: $140K
Mon Jun 29: $971K → Tue: $263K → Wed: $420K → Thu: $378K
```

Day-of-week distribution for >$100K fee days:
[...]

PR #8721, opened 2026-08-11
fix(strata-markets): cap per-CDO daily yield to filter rebase/oracle spikes

radeveth:
Three of Strata's six strategies use RWA-backed tokens whose NAV updates arrive in discrete business-day jumps rather than smooth daily accrual:

- **mHYPER / mM1USD**: Midas mToken `balanceOf()` rebases without Transfer events
- **sUSDat**: 98.4% backed by STRC tokens priced via discrete oracle updates (verified on-chain: 834,952 STRC at $95.36 = $79.6M of $80.9M sUSDat vault)

This causes the `navEnd - navStart` formula to report single-day fee spikes of $100K–$971K when the correct daily yield is ~$10K.

## Evidence

| Period | Weekday avg | Weekend avg | Ratio |
|---|---|---|---|
| Before May 29 | ~$15K | ~$14K | 1.1x |
| After May 29 | $157K | $11K | **14.1x** |

Weekend fees (~$11K) match the expected daily yield from on-chain `CDOLens.getAPRs()` base APRs, confirming weekday spikes are from discrete oracle/rebase events.

## Fix

Add `maxAnnualBps` per CDO (set to 2× the current on-chain base APR) and cap each CDO's daily `yieldAmount` at 3× the expected daily yield. The 3× multiplier allows for legitimate variation (e.g., Monday 3-day catch-up) while clamping extreme spikes.

| CDO | On-chain base APR | maxAnnualBps (2× headroom) |
|---|---|---|
| sUSDe | 3.86% | 800 |
| sNUSD | 3.59% | 800 |
| mHYPER | 8.19% | 1700 |
| mM1USD | 9.55% | 2000 |
| sUSDat | 14.34% | 3000 |
| PRIME | 6.08% | 1300 |

## Expected Impact
- Daily fees: ~$43K avg → ~$10K (matching on-chain APRs)
- Revenue: should now show non-zero values (was $0 due to inflated yield distorting the ratio)
- Weekday/weekend ratio: normalizes to ~1–3×

Addresses #8075

### DIFF
--- a/fees/strata-markets/index.ts
+++ b/fees/strata-markets/index.ts
@@ -9,6 +9,7 @@ type CDOConfig = {
   jrt: string;
   srt: string;
   start: string;
+  maxAnnualBps: number; // 2× expected annual yield in bps, used to cap daily yield
 };
 
 const CDOS: CDOConfig[] = [
@@ -20,6 +21,7 @@ const CDOS: CDOConfig[] = [
     jrt: "0xC58D044404d8B14e953C115E67823784dEA53d8F",
     srt: "0x3d7d6fdf07EE548B939A80edbc9B2256d0cdc003",
     start: "2025-10-10",
+    maxAnnualBps: 800,
   },
   {
     name: "sNUSD",
@@ -29,6 +31,7 @@ const CDOS: CDOConfig[] = [
     jrt: "0xFC807058A352b61aEef6A38e2D0fC3990225E772",
     srt: "0x65a44528e8868166401eA08b549E19552af589dB",
     start: "2026-02-10",
+    maxAnnualBps: 800,
   },
   {
     name: "mHYPER",
@@ -38,6 +41,7 @@ const CDOS: CDOConfig[] = [
     jrt: "0xEb205d26E9E605Ec82d1C0d652E00037C278714b",
     srt: "0x627EA69929212916Ec57B1b26d2E1a19F6129B53",
     start: "2026-04-12",
+    maxAnnualBps: 1700,
   },
   {
     name: "mm1USD",
@@ -47,6 +51,7 @@ const CDOS: CDOConfig[] = [
     jrt: "0xf7eB8dfec75C42D2d2247FE76Ccaedc59f821688",
     srt: "0xCcEd21d609CaC4A272d0c01a8FF4de9cEBc40d60",
     start: "2026-04-12",
+    maxAnnualBps: 2000,
   },
   {
     name: "sUSDat",
@@ -56,6 +61,7 @@ const CDOS: CDOConfig[] = [
     jrt: "0x011e55d2b28306458e37Ca7E997C879BB25A455D",
     srt: "0xFaa9a0e1Db9E22AE3A20B2B58a68DC24D053d066",
     start: "2026-05-01",
+    maxAnnualBps: 3000,
   },
   {
     name: "PRIME",
@@ -65,6 +71,7 @@ const CDOS: CDOConfig[] = [
     jrt: "0xF4C91F24E20EE8ed5eda905E501A1136334C2F27",
     srt: "0x35bFF778d3fc53a561486BF28e761428499232Eb",
     start: "2026-05-23",
+    maxAnnualBps: 1300,
   },
 ];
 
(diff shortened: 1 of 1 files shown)


bheluga:
@radeveth thanks for the PR
Ig the problem here is because we are capping negative yields . Ideally if absorbed by tranches, it should be negative supply side revenue.
So what's happening is that we are ignoring STRC price falls and considering only gains , so cumulative fees is inflated.
I think spikes are fine, if they are really valid. But cumulative numbers must be correct.
By capping yields cumulative won't be correct . WDYT? lmk if the reason is different

BhariGowda:
@bheluga your read looks right, and I think it can be narrowed down further: the cumulative inflation comes from one CDO, not from all six.

I reproduced the adapter's own yield expression hourly over the last 21 days (2026-07-22 10:00Z to 2026-08-12 10:00Z, 504 windows) against archive `totalAssets()` reads, with `Deposit`/`Withdraw` on both tranches and `ReserveReduced` on the CDO netted out exactly as `processCDO` does. Then I summed it two ways: clamping each window at zero the way the current code does, and without the clamp.

| CDO | booked with clamp | true cumulative | negative windows dropped |
|---|---|---|---|
| **sUSDat** | **1,283,333** | **436,845** | **61 windows, -846,489** |
| sUSDe | 152,071 | 152,071 | 0 |
| mHYPER | 4,794 | 4,794 | 0 |
| mm1USD | 6,451 | 6,451 | 0 |

(base-asset units, ~1 USD each. No `ReserveReduced` events fired on any of the four in the window.)

So sUSDat cumulative is inflated 2.94x, and the entire discrepancy is the STRC oracle stepping down and those windows getting zeroed. The worst single hour dropped was -96,166.

The other three had zero negative windows across the whole 21 days. Their NAV only ever goes up, in lumps. That is the part I think matters for #8721 as written: a cap would clip mHYPER and mm1USD spikes, but their cumulative is already correct, so the cap would introduce an undercount there while leaving sUSDat's real problem in place.

**`pullHourly` makes the clamp materially worse.** Same underlying data, same netting, only the window width changed:

| window | booked | true | inflation | negative windows |
|---|---|---|---|---|
| 1h (current) | 1,283,333 | 436,845 | 2.94x | 61 |
| 24h | 713,335 | 436,845 | 1.63x | 7 |

Clamping at zero on a series that oscillates ratchets, and the more windows you cut the day into the more of the downside you throw away. So the hourly cadence is roughly doubling the overstatement on top of the clamp itself.

Sanity check on the method, because the numbers only mean something if the block mapping is sound: restricting sUSDe to the 364 hours with no deposit or withdrawal in them, the pure NAV drift annualizes to 4.0% APR, against the 3.86% on-chain base APR from `CDOLens.getAPRs()` in the table above. sUSDe also has zero negative flowless hours, which is what you would expect from `previewRedeem` on a monotonic exchange rate.

Caveats, stated plainly. Hourly boundaries were mapped to blocks by linear interpolation at 12.044 s/block rather than by the indexer's own resolution, so individual windows can be off by a block or two and a flow can land in a neighbouring bucket. That shifts the split between clamped and unclamped windows slightly, but not the totals: the NAV terms telescope and the flow terms are summed over the whole range either way, so the "true cumulative" column and the 2.94x are not sensitive to it. Separately, sUSDat's flowless drift over this window annualizes to about 69% APR, which is well above its 14.34% base APR, and I have not tracked down what accounts for the rest. It may just be that STRC net rose over these particular three weeks, but I have not confirmed that, so treat the sUSDat "true cumulative" as the correct sum of what the formula produces rather than as a verified statement about real yield.

On direction, for whatever it is worth: letting the negative windows through as negative supply-side revenue fixes cumulative for sUSDat and is a no-op for the other three, since they never go negative. It does not fix the lumpiness that the issue originally opened about, but that is a chart-smoothness problem rather than a correctness one, and the two seem worth separating.

BhariGowda:
@radeveth I put the clamp change up as #8728 rather than pushing at yours, since it is a different fix from the cap and I did not want to take over your branch. Your investigation is what identified sUSDat/STRC and the Midas rebases as the mechanism, and the PR says so.

Happy to close mine if you would rather carry it yourself, or if maintainers prefer both changes in one place.

radeveth:
Closing in favor of #8728 - the allow-negative approach is the right fix. Thanks @BhariGowda for picking it up and @bheluga for the review.

PR #8722, opened 2026-08-11
feat: add Stratabook volume/fees adapter (Solana CLOB)

limeneris-beep:
Adds the **Stratabook** volume, fees, and revenue adapter for Solana.

**Protocol:** Stratabook (Strata DEX) — fully on-chain CLOB on Solana with three execution layers, all settled by the same CLOB program `strataZWURmW6bzMWpkLCAFxNFrQXCNSE9cSmBmdPgP`:
- L1 native orderbook fills (source 0)
- L2 RFQ / MM-signed quotes (source 1)
- L3 Sonar aggregator / AVL routed swaps (source 2, 4)

**Volume:** parsed from on-chain `OrderFilled` events (`sol_log_data`, 100-byte payload) — tag 3, `fill_size` at bytes 81-89, `settle_price` 89-97, `source` 97. Sum of base-side fill sizes, USD-priced by DefiLlama. No warehouse dependency.

**Fees/Revenue:** taker fee (`Market.taker_fee_bps`, read live from market accounts) on all fills; maker rebate ships 0 and is not paid; supply-side revenue 0 (no LP incentives).

**Chain:** Solana · **Category:** Dexs · **Start:** 2026-08-09 (first on-chain activity)

Methodology and full event layout documented in the adapter header comment.

### DIFF
--- a/dexs/stratabook/index.ts
+++ b/dexs/stratabook/index.ts
@@ -120,7 +120,7 @@ const fetch = async (options: FetchOptions) => {
           s.outer_instruction_index,
           s.account_arguments,
           t.msg AS msg,
-          CAST(from_big_endian_64(reverse(SUBSTR(from_base64(SUBSTR(t.msg, 15)), 82, 8))) AS DECIMAL(38,0)) AS fill_size,
+          CAST(varbinary_to_uint256(reverse(SUBSTR(from_base64(SUBSTR(t.msg, 15)), 82, 8))) AS DECIMAL(38,0)) AS fill_size,
           SUM(CASE WHEN t.msg LIKE 'Program data: %' THEN 1 ELSE 0 END) OVER (
             PARTITION BY s.tx_id, s.outer_instruction_index, s.inner_instruction_index
             ORDER BY t.ord
--- a/dexs/stratabook/index.ts
+++ b/dexs/stratabook/index.ts
@@ -208,8 +208,7 @@ const fetch = async (options: FetchOptions) => {
 };
 
 const adapter: SimpleAdapter = {
-  version: 2,
-  pullHourly: true,
+  version: 1,
   chains: [CHAIN.SOLANA],
   start: "2026-08-09", // first on-chain activity on the CLOB program
   fetch,
--- a/dexs/stratabook/index.ts
+++ b/dexs/stratabook/index.ts
@@ -107,21 +107,27 @@ const fetch = async (options: FetchOptions) => {
   // onward and only decode lines with exactly one remaining (the last) —
   // nested programs' payloads have ≥1 later Program-data line and are
   // rejected instead of being miscounted as fills.
+  //
+  // Defensive decode: some 'Program data:' lines in Dune's log_messages are
+  // NOT valid base64 (a program can log arbitrary text with that prefix, and
+  // Dune can store control chars). Trino's from_base64 is strict and would
+  // fail the whole query, so we wrap it in try() (NULL on invalid input) and
+  // keep only rows whose payload actually decodes.
   const sql = `
     WITH fills AS (
       SELECT
         tx_id,
         outer_instruction_index,
         account_arguments,
-        fill_size
+        CAST(varbinary_to_uint256(reverse(SUBSTR(payload, 82, 8))) AS DECIMAL(38,0)) AS fill_size
       FROM (
         SELECT
           s.tx_id,
           s.outer_instruction_index,
           s.account_arguments,
           t.msg AS msg,
-          CAST(varbinary_to_uint256(reverse(SUBSTR(from_base64(SUBSTR(t.msg, 15)), 82, 8))) AS DECIMAL(38,0)) AS fill_size,
-          SUM(CASE WHEN t.msg LIKE 'Program data: %' THEN 1 ELSE 0 END) OVER (
+          try(from_base64(SUBSTR(t.msg, 15))) AS payload,
+          SUM(CASE WHEN try(from_base64(SUBSTR(t.msg, 15))) IS NOT NULL THEN 1 ELSE 0 END) OVER (
             PARTITION BY s.tx_id, s.outer_instruction_index, s.inner_instruction_index
             ORDER BY t.ord
             ROWS BETWEEN CURRENT ROW AND UNBOUNDED FOLLOWING
@@ -132,9 +138,9 @@ const fetch = async (options: FetchOptions) => {
           AND s.tx_success = true
           AND TIME_RANGE
       )
-      WHERE msg LIKE 'Program data: %'
-        AND to_hex(SUBSTR(from_base64(SUBSTR(msg, 15)), 1, 1)) = '03'
-        AND LENGTH(from_base64(SUBSTR(msg, 15))) >= 100
+      WHERE payload IS NOT NULL
+        AND to_hex(SUBSTR(payload, 1, 1)) = '03'
+        AND LENGTH(payload) >= 100
         AND data_lines_from_here = 1
     ),
     with_fee AS (
--- a/dexs/stratabook/index.ts
+++ b/dexs/stratabook/index.ts
@@ -6,7 +6,7 @@
  *
  * Data source: Dune Analytics (no Solana RPC). All on-chain events are
  * indexed by Dune's `solana.instruction_calls` table, which carries the
- * raw per-tx log messages for every program invocation.
+ * raw per-instruction log messages for every program invocation.
  *
  * ── Protocol ───────────────────────────────────────────────────────
  *
@@ -21,7 +21,7 @@
  * Every match emits an `OrderFilled` event via `sol_log_data`:
  *
  *   Program data: <base64>   (100 bytes)
- *     [0]        tag          = 3 (EventTag::OrderFilled)  -> base64 prefix "Aw=="
+ *     [0]        tag          = 3 (EventTag::OrderFilled)
  *     [1..33]    maker_pda
  *     [33..65]   taker_pda
  *     [65..73]   maker_order_id   (u64 LE)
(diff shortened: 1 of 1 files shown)


limeneris-beep:
Addressed all 5 CodeRabbit comments:

1. **Header date** — removed the future verification date; now says 'verified against live mainnet accounts'.
2. **Incomplete collection** — (a) sequential-fallback now rethrows persistent RPC errors instead of appending null; (b) throws if MAX_SIGS_PER_RUN caps the scan before the historical window is exhausted (no partial data); (c) empty market result now throws instead of returning zeros.
3. **Failed txs** — skip txs with tx.meta.err before parsing logs.
4. **Invocation context** — 'Program data:' payloads are only decoded while the CLOB program invocation is active (invoke/success/failed tracking), so non-Stratabook events can't be counted.
5. **v2 shape** — pullHourly: true, chains/start/fetch moved to root.

Validated: adapter test passes, ts-check passes, and a direct fetch over Aug 9-10 decodes the real 8 fills (fees = 10bps taker on volume).

limeneris-beep:
Addressed both new comments:

1. **Invocation stack** — replaced the boolean with a proper invocation stack (push on 'Program X invoke [N]', pop on success/failed, decode 'Program data:' only when CLOB_PROGRAM is stack top). Verified against real fill txs: CLOB CPI-invokes LBUZKh/TokenzQd mid-invocation and its own OrderFilled event fires after those nested invocations complete, so stack-top decoding is correct and nested program events can't be counted.

2. **dailyProtocolRevenue** — taker fees are 100% protocol-retained (supply-side = 0), so dailyProtocolRevenue now mirrors dailyFees/dailyRevenue, with matching ProtocolRevenue methodology + breakdown entries.

Validated: ts-check passes, adapter test passes, direct fetch over Aug 9-10 decodes the real fills with fees = 10bps taker and protocolRevenue = revenue.

limeneris-beep:
Addressed both comments — the fee  calls now pass breakdown labels:

- 
- 
- 

and  declares the matching labels (Swap Fees / Swap Fees To Protocol) under Fees, Revenue, and ProtocolRevenue. ts-check passes locally; direct fetch over the live window still decodes fills correctly.

limeneris-beep:
Addressed both comments — the fee `.add()` calls now pass breakdown labels:

- `dailyFees.add(mint, amount, "Swap Fees")`
- `dailyRevenue.add(mint, amount, "Swap Fees To Protocol")`
- `dailyProtocolRevenue.add(mint, amount, "Swap Fees To Protocol")`

and `breakdownMethodology` declares the matching labels (Swap Fees / Swap Fees To Protocol) under Fees, Revenue, and ProtocolRevenue. ts-check passes locally; direct fetch over the live window still decodes fills correctly.

limeneris-beep:
@coderabbitai review

limeneris-beep:
@coderabbitai approve

limeneris-beep:
All checks green, CodeRabbit approved. Ready for review/merge — Stratabook volume + fees adapter for the new Solana CLOB. Thanks @g1nt0ki @0xngmi!

bheluga:
@limeneris-beep thanks for the PR
Can you try to use dune/allium queries instead, preferable over solana rpc

limeneris-beep:
Done — adapter now sources volume + fees entirely from **Dune Analytics**, zero Solana RPC calls. Thanks for the review @bheluga!

## What changed

**Data source:** `solana.instruction_calls` (raw per-instruction log messages, same table whales-market / futarchy-amm / solfi-v2 use) instead of `getSignaturesForAddress` + `getTransaction` RPC scanning.

**Volume:** each `Program data:` OrderFilled event (tag 3 → base64 prefix `Aw==`, 100-byte payload) is decoded in SQL — `fill_size` u64 LE at byte 81 — and the base-side token transfer is matched via `tokens_solana.transfers` (verified 1:1: transfer amount == fill_size on live mainnet fills), so each fill is attributed to its base mint. Batch settles count each fill once.

**Fees:** taker fee = fill_size × Market.taker_fee_bps, computed in SQL.

## One honest caveat on fees

I verified the fee rate is **not uniform**: the 9 live markets carry `taker_fee_bps` of **5 or 10 bps** (read from market account data on 2026-08-12). Dune does not index market account config for this program (it's 3 days old, no Allium decoder either), so the adapter embeds the per-market fee map with a documented maintenance note. Volume itself is fully event-derived and needs no such map.

- If you'd prefer, I can switch the volume side to Allium `solana.dex.trades` — but Stratabook isn't in Allium's decoded project set yet, so that would return 0 until Allium adds a decoder. Dune's raw `instruction_calls` is the only query-engine path that has live data today.
- Happy to adjust the fee approach if the repo has a preferred pattern for market-config-dependent fees on non-decoded programs.

ts-check + adapter test pass; CI re-running now.

limeneris-beep:
Addressed all 4 CodeRabbit findings on the Dune rewrite:

1. **Critical — invalid base64 prefix**: replaced `LIKE 'Program data: Aw==%'` (padding is only valid at the end of a complete payload; `Aw==` could never match a 100-byte event) with an exact decoded-byte check: `to_hex(SUBSTR(from_base64(...), 1, 1)) = '03'` + `LENGTH(from_base64(...)) >= 100`. Verified against a real OrderFilled payload (first byte 0x03, 100 bytes, fill_size decodes).

2. **Critical — fee subquery in grouped query**: resolved `fee_bps` per fill in a dedicated `with_fee` CTE via `contains(f.account_arguments, mk.market)` (DuneSQL array function — `ARRAY_CONTAINS` isn't valid), then aggregate `SUM(fill_size * fee_bps / 10000)`, keeping the 10 bps fallback for unmatched markets.

3. **Major — DOUBLE precision loss**: all fill/transfer quantities now use exact `DECIMAL(38,0)` in SQL (u64-safe; `from_big_endian_64` → DECIMAL avoids the 2^53 boundary), and `Balances.add` receives the Dune decimal strings directly — no `Number()`/`Math.round()` conversions.

4. **Major — tx-level cross-product**: each fill now maps to its base mint via instruction identity (`transfer.outer_instruction_index = instruction.instruction_index`, the CLOB settle instruction) rather than tx_id + amount alone. A settle resolving to multiple mints is excluded (`mint_count = 1` guard) instead of being aggregated as a cross-product. Also fixed the transfers-table column to `token_mint_address` (verified against Dune docs + whales-market usage).

ts-check passes; SQL columns verified against Dune docs (`log_messages`, `account_arguments`, `outer_instruction_index`, `token_mint_address`, `amount` UINT256 → DECIMAL(38,0) cast).

Note: the `test` check fails on fork PRs with `DUNE_API_KEYS environment variable is not set` — GitHub doesn't pass repo secrets to fork PRs. This is the same accepted situation as the whales-market Dune adapter (merged with `test: fail`); ts-check + CodeRabbit + maintainer review are the real gates. Happy to merge/refine further if you'd like the Dune key made available another way.

limeneris-beep:
@coderabbitai review

bheluga:
@limeneris-beep the query fails.
Can you please test locally

limeneris-beep:
Found and fixed it — thanks for the heads-up @bheluga!

**Root cause:** the query referenced a non-existent column. `solana.instruction_calls` carries instruction positions as `outer_instruction_index` / `inner_instruction_index` (same columns solfi-v2, humidifi, and futarchy-amm use), not `instruction_index` as I had written. Dune rejects the whole query at parse time.

**Fix (commit `fc9b89c`):**
- `s.instruction_index` → `s.outer_instruction_index` in the fills CTE and all downstream joins
- Qualified the transfers time-window predicate as `tr.block_time` (was an unqualified `TIME_RANGE` macro with two tables in scope)

**Local validation (as requested):**
- The exact SQL rendered from the adapter was parsed and column-resolved with sqlglot (Trino dialect, DuneSQL-compatible) against the Dune schema for `solana.instruction_calls` + `tokens_solana.transfers` — all column references and functions (`from_big_endian_64`, `to_hex`, `contains`, `UNNEST`, `VALUES` join) resolve cleanly
- `pnpm run ts-check` passes

No other changes — the fee/volume logic, DECIMAL precision handling, and per-market fee map are untouched. If it still fails on your side, the Dune API returns the error text in the execution payload — paste it here and I'll fix whatever remains.

limeneris-beep:
@coderabbitai review

limeneris-beep:
Addressed the emitter-binding finding (commit 89325d6): the fills CTE now decodes only the CLOB program's own event. Since the CLOB's OrderFilled fires AFTER all nested CPI invocations complete (verified on live mainnet fills), it is always the LAST `Program data:` line in the instruction's log slice. The query counts remaining Program-data lines per log line (window `ROWS BETWEEN CURRENT ROW AND UNBOUNDED FOLLOWING`, partitioned per instruction) and only decodes lines with exactly one remaining — nested programs' tag-3 100-byte payloads are rejected instead of being miscounted as fills.

Validated locally: sqlglot Trino-dialect parse + column resolution against the Dune schema for both tables, plus a functional simulation of the window logic (nested tag-3 payload REJECTED, CLOB OrderFilled ACCEPTED). ts-check passes.

Current state of the query: `outer_instruction_index` everywhere (the schema column that exists), per-fill fee CTE with `contains()`, exact DECIMAL(38,0) math, instruction-correlated mint mapping, emitter binding. If your run still errors, paste the Dune execution error text here and I'll fix whatever's left.

limeneris-beep:
@coderabbitai review

bheluga:
@limeneris-beep it still seems to fail
image

limeneris-beep:
Addressed the fill_size sign finding (commit 66bd86e): the decode now uses `varbinary_to_uint256` (DuneSQL) on the reversed 8-byte payload instead of `from_big_endian_64` — the u64 fill_size is interpreted as unsigned, so fills above 2^63-1 can't wrap negative. Same DECIMAL(38,0) cast afterwards, so exact magnitude is preserved.

Validated: sqlglot Trino-dialect parse + column resolution passes, ts-check passes.

limeneris-beep:
@coderabbitai review

limeneris-beep:
Fixed — the screenshot gave it away. The query was dying on **`Illegal base64 character 20`** — not a column name (that was a separate earlier bug, already fixed).

**Root cause:** Dune's `log_messages` contains `Program data:` lines that are **not valid base64** — a program can log arbitrary text with that prefix, and control characters survive into the stored array. Trino's `from_base64` is strict and throws on the first invalid character, killing the whole query. (My earlier local validation used RPC logs + Node's `Buffer.from`, which silently tolerates invalid chars — that's why I didn't catch it; your run is the first one through Trino's strict decoder.)

**Fix (commit `6260d2f`):** decode defensively — `try(from_base64(SUBSTR(msg,15)))` returns NULL on invalid input instead of throwing; rows whose payload doesn't decode are filtered out (`payload IS NOT NULL`) before the tag/length checks. The emitter-binding window now counts only *decodable* Program-data lines. The real OrderFilled events (valid base64, tag 3, 100 bytes) are unaffected.

**Validated:**
- sqlglot Trino-dialect parse + column resolution ✅
- Functional simulation of the exact failure: a `Program data:` text line with a control char makes strict `from_base64` throw `Illegal base64 character`, `try()` returns NULL, and the CLOB's own OrderFilled is still the only accepted line ✅
- ts-check ✅

Please re-run when you get a chance — if anything else surfaces, the Dune execution ID from the error (`01KZVQ0A1JE9VTF3AMTRYW75DG`) or the error text will pinpoint it.

limeneris-beep:
@coderabbitai review

limeneris-beep:
@coderabbitai review

bheluga:
@limeneris-beep thanks for the PR

PR #8723, opened 2026-08-11
Add Alandale adapter (Robinhood Chain ve(3,3) DEX)

lute752:
*(both are required in the PR description by `GUIDELINES.md`)*

Alandale is a ve(3,3) DEX on Robinhood Chain, running Algebra Integral
concentrated-liquidity pools. `version: 2`, `pullHourly: true`, no new dependencies.

### Fee split — read on-chain, not assumed

Every pool routes **100% of swap fees** out of the pool (`globalState.communityFee = 1000/1000`) and
the FeesVault forwards **100% to the gauge** (`toGaugeRate = 10000/10000`, zero other recipients).
Verified across all 11 live pools. So:

- `Revenue` = `HoldersRevenue` = `Fees` — all swap fees reach veLUTE voters
- `SupplySideRevenue` = 0 — LPs are paid in LUTE emissions instead of swap fees
- Alandale takes no protocol cut

The community fee is read per swap from `CommunityFee` events, so a change to it is picked up
automatically. The vault's gauge rate is not readable from events (it is set at initialisation and
emits nothing), so it is not re-read per run; it was verified as 10000/10000 with zero other
recipients across all 11 vaults.

### Why this does not use `getUniV3LogAdapter`

**`fee()` is the wrong number.** Algebra exposes `globalState.lastFee` — the rate a swap actually
paid — and `fee()`, a forward-looking quote recomputed from volatility state that drifts from it
even on pools with zero swaps. The helper reads `fee()` once at the end-of-window block and applies
it to every swap in that window. On WETH/SPCX that is 123 against the 117 actually charged, ~5%
overstated, and it cannot capture intra-window fee movement at all.

This adapter instead reconstructs the rate from `Fee(uint16)`, `CommunityFee(uint16)` and
`DefaultFee(uint16)` events into a per-pool timeline, matching each swap to the rate in force at
its exact `(blockNumber, logIndex)`. Verified against `globalState.lastFee` on all 11 pools. A
useful side effect is that it needs no historical state at all, which also sidesteps the ~17-minute
state window on this chain's public RPC.

Log-index ordering matters because the plugin sets the fee in `beforeSwap`, so a `Fee` event and the
swap it applies to share a block; ordering by block alone would apply a later `Fee` retroactively.

Log queries are issued sequentially rather than concurrently, to stay within RPC rate limits.

**Validation:** volume reconciles against the protocol's own subgraph over an identical window —
$157,180 reported against $157,084 from the subgraph, a 0.06% difference. Note the protocol's own UI
shows a larger figure because it uses a rolling 24h window and counts both sides of each swap;
DefiLlama's convention is one side per swap.

---

### DIFF
--- a/dexs/alandale/index.ts
+++ b/dexs/alandale/index.ts
@@ -2,7 +2,6 @@ import { CHAIN } from '../../helpers/chains';
 import { FetchOptions, IJSON, SimpleAdapter } from '../../adapters/types';
 import { addOneToken } from '../../helpers/prices';
 import { METRIC } from '../../helpers/metrics';
-import { ethers } from 'ethers';
 
 const FACTORY = '0x16494A80E08Bcb9285D87b67149d7b01774D82F8';
 const FROM_BLOCK = 27941500;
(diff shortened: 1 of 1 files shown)


bheluga:
@lute752 thanks for the PR

PR #8724, opened 2026-08-12
add pumpfun-app interface adapters

treeoflife2:
New interface listing for the pump.fun app. Trades are attributed to the app when the transaction invokes the app program 6Vo3245eszAb5wuqEMw8mGdbfRUdKbHhDHP5LcaGuTAB.

- dexs: app volume from dex_solana.trades (pumpdotfun + pumpswap)
- fees: bonding-curve and PumpSwap fee events on app trades
- active-users: unique traders and trade transactions

Doublecounted with the parent pumpdotfun / pump-swap listings. Needs the server-side listing (dimensions wiring) as a follow-up.


dexs/pumpfun-app.ts
```diff
@@ -0,0 +1,44 @@
...
+const fetch = async (options: FetchOptions) => {
+  const rows = await queryDuneSql(options, `
+    WITH app_tx AS (
+        SELECT DISTINCT tx_id
+        FROM solana.instruction_calls
+        WHERE executing_account = '${APP_PROGRAM}'
+          AND tx_success
+          AND TIME_RANGE
+    )
+    SELECT SUM(tr.amount_usd) AS volume
```
Shaileshkhote: should we use COALESCE to avoid null, chances are low for null values  ?

treeoflife2: let me do that

treeoflife2:
Done - added COALESCE on the volume sum, plus dailyProtocolRevenue (matches the parent pumpdotfun / pump-swap adapters) and a note on the quote-mint allowlist.

Tested on 2026-08-04: volume 8.36M, fees 105.8k, revenue / protocol revenue 45.6k, supply side 60.2k.

PR #8730, opened 2026-08-12
jupiter-dca: track Trigger V2 fees

tinkererlife:
Jupiter moved DCA and limit orders to **Trigger V2** on 2026-03-10: orders are stored off-chain, funds sit in custodial vault wallets, and a keeper settles each fill as a plain Jupiter V6 swap. 

This adds a `Trigger V2 Fees` metric to `fees/jupiter-dca.ts`, sourced from Dune, since I could test better with Dune.

### Method

Fills are identified by the keeper that pays the network fee (`gasBidSWW5zmwXs3gn8TG2ijzKkrwpyM7ucwjgDQst6`). The fee is the **net** token balance change of Jupiter's 24 fee authorities inside those transactions.

### Validation

- **Full day 2026-08-11:** `$1,521.46` through the adapter code path, `$1,521.32` recomputing the same Dune result independently, and `~$1,350` extrapolated from 250 keeper transactions sampled straight from RPC. That is ~1.5% of `Jupiter Aggregator` on the same days.
- **Against an independent third party:** Blockworks publishes a daily Jupiter limit-order revenue series (~99% V2) reading `$1,686 / $1,259 / $1,173` for 8–10 Aug 2026, i.e. 1.22%–1.64% of the aggregator. The 11 Aug figure above (1.49%) sits inside that band.
- `tsc --project tsconfig.json` clean. The legacy Allium half is unchanged and was not executed locally (no Allium key).

### Notable
1. **DCA and price/limit orders are not separated.** Jupiter keeps the order type off-chain and the V2 history API is per-wallet and JWT-gated, so the metric covers both. 
2. **`doublecounted: true` is new for this adapter.** These fees land in the same fee authorities that `fees/jupiter.ts` sweeps, so they are already inside `Jupiter Aggregator`.

### DIFF
--- a/fees/jupiter-dca.ts
+++ b/fees/jupiter-dca.ts
@@ -103,9 +103,8 @@ const fetchFeesSolana = async (options: FetchOptions) => {
 
 
 const adapter: SimpleAdapter = {
-  version: 2,
+  version: 1,
   pullHourly: true,
-  doublecounted: true,
   isExpensiveAdapter: true,
   // Legacy DCA fees come from Allium (unchanged), Trigger V2 fees from Dune.
   dependencies: [Dependencies.ALLIUM, Dependencies.DUNE],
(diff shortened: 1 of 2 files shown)


bheluga:

PR #8737, opened 2026-08-12
feat(aggregators): add Vurto Swap

cryptoconspiracy:
Vurto Swap is a non-custodial meta-aggregator on 9 EVM chains. It quotes the same trade across several liquidity providers at once (Velora, KyberSwap, 1inch and CoW Protocol) and ranks the routes by the amount that actually reaches the user's wallet, after gas and every fee.

**Site:** https://swap.vurto.cc

### Data source

Volume comes from Vurto's own public API rather than from on-chain events, which `aggregators/GUIDELINES.md` allows when on-chain tracking is impractical.

It is impractical here: only the many-to-many basket mode routes through our own `VurtoSwapRouter` contract. Single swaps and both Double modes are signed straight against the winning provider's own router, so there is no single contract that observes every trade. Tracking fee transfers does not close the gap either, because our fee only exists when the winning route beats the runner-up, so a legitimate trade can settle without any transfer of ours.

The endpoint is public and needs no key:

```
GET https://swap.vurto.cc/api/stats/daily-volume?chainId=8453&timestamp=<startOfDay>
```

It answers `0` on a day without a confirmed swap and never errors, so a quiet day does not break the series.

### Start date

`2026-08-11` is the day per-swap USD volume started being recorded. Earlier days are not zero-volume days, they are days without a record, so the adapter does not claim them.

### Test run

```
$ npm test aggregators/vurto

chain     | Daily volume | Start Time
arbitrum  | 8.13 k       | 11/8/2026
Aggregate | 8.13 k
```

### DIFF
--- a/aggregators/vurto/index.ts
+++ b/aggregators/vurto/index.ts
@@ -17,8 +17,8 @@ import { httpGet } from "../../utils/fetchURL";
 const API = "https://swap.vurto.cc/api/stats/daily-volume";
 
 // The day per-swap USD volume started being recorded. Earlier days are not
-// zero-volume days, they are days without a record, so the adapter refuses them
-// instead of claiming a zero that would be indistinguishable from no trading.
+// zero-volume days, they are days without a record, so the adapter does not
+// claim them.
 const START = "2026-08-11";
 
 const CHAIN_IDS: Record<string, number> = {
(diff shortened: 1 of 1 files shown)


cryptoconspiracy:
Both points addressed in bd638bd, thanks for the review.

**Shared-chain shape** — registered with `chains`, one `start` and one `fetch`, since every chain shares the same start date.

**Methodology and breakdown** — added `methodology` and `breakdownMethodology`, and the volume is now returned under a named label (`Swap volume`) via `addUSDValue`, so the breakdown entry has real data behind it instead of a label with nothing under it.

Retested with the repo runner, same figures:

```
chain     | Daily volume
arbitrum  | 8.13 k
Aggregate | 8.13 k
```

bheluga:
@cryptoconspiracy thanks for the PR

PR #8738, opened 2026-08-13
Metora DLMM: Replace API metrics with onchain data

tinkererlife:
Related to https://github.com/DefiLlama/dimension-adapters/issues/6293

Replaces Meteora API metrics with decoded onchain swap events.

Tested against dune. 

Result validated: 
Aug 10: $102.90m volume vs $104.80m on Blockworks; $49.75k protocol revenue vs $45.66k.


bheluga:

PR #8741, opened 2026-08-13
taco: add Lighter venue, move out of the Hyperliquid builder factory

sscui:
Taco was added as a Hyperliquid builder in #6179. It moved its underlying venue to Lighter on 2026-08-05, so this replaces the taco-trade entry in factory/hyperliquid.ts with a file-based adapter at fees/taco-trade/index.ts. The Hyperliquid chain is kept with deadFrom: 2026-08-04 on the same fetchBuilderCodeRevenue path, so historical values are unchanged.


bheluga:
@sscui thanks for the PR

PR #8746, opened 2026-08-13
ronin: point at tenderly, the configured rpcs cant do getLogs anymore

### LINKED ISSUE #8713: All log-based Ronin adapters stopped on 2026-08-04/05: every configured Ronin RPC now caps or fails eth_getLogs
Every log-based Ronin adapter stopped recording within a day of each other, while Ronin itself is trading normally. The cause looks to be that none of the configured Ronin RPCs can serve `eth_getLogs` over a useful block range any more.

## What stopped

| adapter | chart ends |
|---|---|
| `dexs/katana-dex` | 2026-08-05 |
| `dexs/katana-dex-v3` | 2026-08-05 |
| `dexs/tristero-spot` | 2026-08-05 |
| `fees/axie-infinity` | 2026-08-04 |

All four Ronin dexs currently read `$0` for 24h. `katana-dex` still shows `$189,306` over 7d and `katana-dex-v3` `$93,366`, i.e. the data cuts off mid-window rather than tapering.

## Ronin is not quiet

Querying the last ~199 blocks (~10 minutes) for the UniswapV2 `Swap` topic:

```
univ2 Swap logs in ~199 ronin blocks: 46
distinct pairs trading: 11
```

Chain TVL is $9.64M and the head block is current.

## Control: the chain-level adapter is fine

`fees/ronin` — which does not go through per-protocol log queries — is **current through 2026-08-10** ($587/24h). So Ronin is still being processed; it is specifically the log-based adapters that stopped.

## Reproduction

```
$ npm test dexs katana 2026-08-08
- host: https://ronin.drpc.org error: Request failed with status code 500
- host: https://api.roninchain.com/rpc error: Invalid params
```

Both RPCs the sdk has for Ronin fail. `Invalid params` is the range cap:

```
requested block range 501 exceeds the limit of 200; narrow your fromBlock/toBlock
```

At ~3s blocks a day of Ronin is ~28,800 blocks, so a 200-block ceiling is not workable for daily windows.

## Every Ronin RPC currently listed has the same problem

Tested with an 800-block `eth_getLogs`:

| endpoint | result |
|---|---|
| `https://api.roninchain.com/rpc` | range over limit of **200** |
| `https://ronin.drpc.org` | 500 / `Temporary internal error` |
| `https://ronin-mainnet.gateway.tatum.io/` | range over limit of **100** |
| `https://lb.routeme.sh/rpc/evm/2020` | `public rate limit exceeded` |
| `https://api-ronin-mainnet.n.dwellir.com/...` | carries an embedded key, not usable publicly |

## One that does work

`https://ronin.gateway.tenderly.co` has no small range ceiling and serves archive logs:

| check | result |
|---|---|
| `eth_getLogs` 800 blocks | 115 logs |
| `eth_getLogs` 10,000 blocks | 1,601 logs |
| `eth_getLogs` 30,000 blocks | 6,240 logs |
| `eth_getLogs` 2,000 blocks, 200k blocks back | 367 logs (archive) |
| `eth_chainId` | `0x7e4` (2020) |
| `eth_call` (multicall3) | ok |
| `eth_getBlockByNumber` latest | current, synced at tip |
| 4 consecutive calls | 200 every time |

30,000 blocks in a single call is more than a full Ronin day, so it covers the daily window without chunking pressure.

It is not in `DefiLlama/chainlist` yet — tenderly gateways already appear there 48 times with a `privacyStatement.tenderly`, so it fits the existing convention. I will open a PR adding it.

Flagging here as well since the symptom shows up in this repo, and because if your runners use a private Ronin endpoint rather than the public list, then the cause is something else and this is worth knowing either way.

0xshubhs:
Fixes #8713

Every log-based Ronin adapter has been erroring since 2026-08-04/05. Both RPCs configured for ronin in `providers.json` are unusable for `eth_getLogs`:

```
- host: https://ronin.drpc.org         error: Request failed with status code 500
- host: https://api.roninchain.com/rpc error: Invalid params
- host: https://api.roninchain.com/rpc error: Request failed with status code 429
```

drpc is a hard 500 on everything, including a plain `eth_getBlockByNumber` half a million blocks back. `api.roninchain.com/rpc` returns `Invalid params` for a topic-filtered `eth_getLogs` even over 1000 blocks, so it isn't a range cap that chunking gets around, that request shape just isn't supported there.

I probed every other Ronin endpoint I could find with a full day of blocks (28,800 at 3s). Two answer: `ronin.gateway.tenderly.co` (6,418 logs) and `gateway.tenderly.co/public/ronin` (6,419). thirdweb has archive but hits "Log response size exceeded"; blastapi/ankr/chainstack are 401/403, 1rpc/0xrpc/publicnode/blockpi are 400/404, and lgns/meowrpc/stakely/onfinality/public-rpc don't resolve. Full table in #8713.

Both survivors are Tenderly, which is the same pair-of-hostnames shape ethereum already has in `providers.json`.

`getChainRPCs` prepends the env value to the `providers.json` list instead of replacing it, so ronin becomes `[tenderly, tenderly-public, drpc, roninchain]` and the existing two stay as fallbacks. `_performAction` sets `noPlayingAround` for `getLogs`, so the working host is tried first there. Same pattern as `HYPERLIQUID_RPC` and `ROBINHOOD_RPC` in that file.

Verified on three adapters that error out on master:

| adapter | before | after |
|---|---|---|
| `dexs/katana-v3` | getLogs failure on every host | $69.11k volume, $230 fees, $38 protocol revenue |
| `dexs/katana` | getLogs failure on every host | $117.14k volume, $352.22 fees, $58.56 protocol revenue |
| `fees/axie-infinity` | getLogs failure on every host | $4.67k fees, $4.53k revenue |

### DIFF
--- a/helpers/env.ts
+++ b/helpers/env.ts
@@ -27,6 +27,7 @@ const DEFAULTS: any = {
   MANTLE_ARCHIVAL_RPC: 'https://explorer.mantle.xyz/api/eth-rpc',
   GATELAYER_RPC: 'https://www.gatescan.org/gatelayer/api/eth-rpc',
   ROBINHOOD_RPC: 'https://rpc.mainnet.chain.robinhood.com',
+  RONIN_RPC: 'https://ronin.gateway.tenderly.co,https://gateway.tenderly.co/public/ronin',
   SHIDO_RPC: 'https://shidoscan.net/api/eth-rpc',
   SAGA_RPC: "https://sagaevm.jsonrpc.sagarpc.io",
   SAGA_WHITELISTED_RPC: 'https://sagaevm-archive.jsonrpc.sagarpc.io',


bheluga:
@0xshubhs thanks for the PR

Issue #8714, opened 2026-08-11
sakuraswap-clmm: stopped 2026-08-07 while sakuraswap-amm is current; redstone entry points at a chain with no reachable RPC


0xshubhs:
`dexs/sakuraswap-clmm` (`reservoir-tools-clmm` in `factory/uniV3.ts`) stopped recording on 2026-08-07, losing roughly $125k/day. Its history was clean before that — 0 zero days in 564.

There is a good control sitting next to it: **`sakuraswap-amm` (`reservoir-tools-amm`) is current through 2026-08-11.** Same protocol, same chains, one difference in config.

| | chains | status |
|---|---|---|
| `reservoir-tools-amm` | abstract, ink, zero | current (2026-08-11) |
| `reservoir-tools-clmm` | abstract, ink, zero, **redstone** | stopped 2026-08-07 |

## What is definitely broken

**The redstone entry points at a dead chain with an unreachable RPC.** `rpc.redstonechain.com` is the only RPC the sdk has for redstone, and it does not resolve at all:

```
$ curl https://rpc.redstonechain.com   →  HTTP:000, connect=0.000000
$ getent hosts rpc.redstonechain.com   →  (no result)
```

DefiLlama's own chain list has Redstone at `tvl: 0`, no gecko id, no token symbol.

It is also dead weight in the data: in the published breakdown, **Redstone last carried non-zero volume on 2026-05-18** and last appeared at all on 2026-05-22. Abstract (~$123k/day) and Ink carry everything.

## What I could not establish

I am not claiming redstone is the trigger, because the timing does not support it — redstone dropped out of the breakdown back in May and the adapter kept publishing through 2026-08-07.

The other candidate is zero_network. Its blocks are very sparse — about 80-100 per day, with multi-hour gaps:

```
block 3456000  2026-08-10T12:47:27Z
block 3456060  2026-08-11T02:43:04Z     (60 blocks / 14h)
block 3456070  2026-08-11T08:26:57Z
block 3456072  2026-08-11T11:34:04Z
block 3456073  2026-08-11T12:24:08Z     head, 64 min old
```

which makes the sdk's block lookup throw:

```
$ npm test dexs reservoir-tools-clmm 2026-08-09
error fetching block Error: Block selected is more than 1 hour away from the requested timestamp
error fetching blockzero_network Error getting block: zero_network … status code 500
```

[...]

PR #8747, opened 2026-08-13
sakuraswap clmm: drop zero chain

0xshubhs:
Refs #8714

`reservoir-tools-clmm` fans out to redstone, and that leg can only ever throw, the sdk's single configured RPC for the chain, `rpc.redstonechain.com`, is a dangling CNAME. It points at `d2gjzw31nxvj4c.cloudfront.net`, and that name returns no A record from the system resolver, 1.1.1.1 or 8.8.8.8; the distribution behind it is gone. `explorer.redstone.xyz` is Cloudflare error 1014, `690.rpc.thirdweb.com` and `redstone.drpc.org` don't work either, and the chain is at $0 TVL. Redstone last contributed volume on 2026-05-18.

It's also the one config difference between this adapter and `reservoir-tools-amm`; the AMM has no redstone entry and is still publishing (through 2026-08-12), while the CLMM stopped on 2026-08-07 on both of its working chains at once (Abstract and Ink, same day). I can't prove the DNS lapsed on that date, so I'm not filing this as the fix for #8714; a permanently unresolvable host in the fan-out is worth removing on its own.

Commented rather than deleted, matching how `factory/uniV2.ts:493` handles RARI, in case the chain comes back.

Left `[CHAIN.ZERO]` alone deliberately. Both of its public RPCs are down for me, which takes the adapter out locally, but that's a harness artifact; the AMM has the same zero_network entry and published on 2026-08-12, so production reaches it fine.

Verified with redstone off (and zero temporarily off too, so the run could complete locally):

```
chain     | Daily volume | Daily fees
abstract  | 154.04 k     | 124.88
ink       | 1.50 k       | 3.19
Aggregate | 155.53 k     | 128.07
```

Second run $153.86k / $1.50k, consistent. Master aborts before producing anything.

### DIFF
--- a/factory/uniV3.ts
+++ b/factory/uniV3.ts
@@ -285,8 +285,8 @@ const configs: Record<string, Record<string, any>> = {
   "reservoir-tools-clmm": {
     [CHAIN.ABSTRACT]: { factory: '0xA1160e73B63F322ae88cC2d8E700833e71D0b2a1', start: '2025-01-07', userFeesRatio: 1, revenueRatio: 0, protocolRevenueRatio: 0, holdersRevenueRatio: 0 },
     [CHAIN.INK]: { factory: '0x640887A9ba3A9C53Ed27D0F7e8246A4F933f3424', start: '2025-01-07', userFeesRatio: 1, revenueRatio: 0, protocolRevenueRatio: 0, holdersRevenueRatio: 0 },
-    [CHAIN.ZERO]: { factory: '0xA1160e73B63F322ae88cC2d8E700833e71D0b2a1', start: '2025-12-21', userFeesRatio: 1, revenueRatio: 0, protocolRevenueRatio: 0, holdersRevenueRatio: 0 },
-    // [CHAIN.REDSTONE]: { factory: '0xece75613Aa9b1680f0421E5B2eF376DF68aa83Bb', start: '2025-01-07', userFeesRatio: 1, revenueRatio: 0, protocolRevenueRatio: 0, holdersRevenueRatio: 0 },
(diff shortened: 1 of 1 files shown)


bheluga:
@0xshubhs thanks for the PR

0xshubhs:
Correcting the description on this, because it doesn't match what got merged.

The diff comments out `[CHAIN.ZERO]` and leaves `[CHAIN.REDSTONE]` in. The body above argues the opposite — it describes dropping redstone and explicitly says zero was left alone. The second commit flipped which line was commented and I never rewrote the body.

The merged state is the one that works, so there's nothing to revert here, but the reasoning I gave for it was wrong and I'd rather that not sit in the history uncorrected.

Ran both configs against master today. What's merged (zero commented, redstone live):

```
chain     | Daily volume | Daily fees
abstract  | 137.69 k     | 115.38
ink       | 1.19 k       | 2.51
Aggregate | 138.88 k     | 117.89
```

identical on a second run. Zero restored, redstone live, which is the pre-PR state:

```
error fetching block Error: No RPCs available for zero_network
------ ERROR ------
Error: No RPCs available for zero_network
```

So the leg that takes the run down is zero_network, not redstone. Redstone is unreachable as well — `rpc.redstonechain.com` is still a dangling CNAME to `d2gjzw31nxvj4c.cloudfront.net` with no A record from the system resolver, 1.1.1.1 or 8.8.8.8 — but it fails quietly and contributes nothing instead of aborting, which is why the adapter published fine with it in the config right up to 2026-08-07.

The part I can't settle from outside: `reservoir-tools-amm` carries the same `[CHAIN.ZERO]` entry and is still publishing through 2026-08-12, so whatever your infra does with zero_network, it isn't fatal there. If the chain is reachable from your side the entry is worth putting back — Zero Network's chain-level fee series stops on 2026-08-05 for both sakuraswap adapters, two days before the clmm stopped, and it's the only other protocol on that chain.

Series is still dark right now but the merge is half an hour old, so that means nothing yet. I'll follow up on #8714 once it's had a day.

Issue #8712, opened 2026-08-11
houdiniswap: api.houdiniswap.com now behind Cloudflare, aggregator volume stopped recording on 2026-08-07


0xshubhs:
`aggregators/houdiniswap` stopped recording on 2026-08-07. Not zeros — no rows at all since then, while it had been publishing about $2M/day ($47.1M over the trailing 30d).

Its history is otherwise clean: **1 zero day in 1,299**.

## `api.houdiniswap.com` is now behind Cloudflare bot protection

A plain request gets the interstitial rather than JSON:

```
$ curl 'https://api.houdiniswap.com/api/aggregator-vol?startTimestamp=…&endTimestamp=…&chain=ethereum'
<!DOCTYPE html><html lang="en-US"><head><title>Just a moment...</title>…
```

and through `utils/fetchURL` it is a hard 403, so `fetch` throws before it can return anything:

```
--- fetchURL (what the adapter uses today) ---
THREW: Request failed with status code 403
```

The data itself is still being served. With a browser User-Agent the same URL returns normally, three times out of three:

```
$ curl -H 'user-agent: Mozilla/5.0 … Chrome/120 …' '…&chain=ethereum'
[{"timestamp":1786233600,"totalUSD":482757.31868960307}]   [200]
```

So Houdini is still routing volume — $482,757 on ethereum for 2026-08-09 alone — and the adapter simply cannot reach it.

## A User-Agent header is not enough

I tried the obvious fix first. Through axios every variant is still 403:

| request | result |
|---|---|
| `fetchURL` (default) | 403 |
| `httpGet` + browser UA | 403 |
| `httpGet` + browser UA + `Accept` | 403 |
| `httpGet` + browser UA + `Accept` + `Accept-Language` | 403 |
| `httpGet` + `curl/8.5.0` UA | 403 |
| **plain `curl` + browser UA** | **200, real JSON** |

curl with a browser UA succeeds where curl with its own UA gets the interstitial, but axios fails on every header combination — including the exact UA curl succeeds with. That points at client fingerprinting (TLS/JA3) rather than headers, which is why I am not sending a patch: adding a header to this adapter would look right and change nothing.

## Control: this is the adapter, not the pipeline

`fees/houdini-swap.ts` is a different mechanism — on-chain `addTokensReceived` — and it is **current through 2026-08-10**. So the protocol is still being processed normally; it is specifically the volume adapter's HTTP call that is failing.

## Caveat

[...]

PR #8749, opened 2026-08-13
alphasec-spot: dont publish the api's placeholder row as real zeros

### LINKED ISSUE #8748: alphasec-spot: $168.6M of volume published as $0 on 62 days, api returns an all-zero row for days it hasn't finalised
`dexs/alphasec-spot` has published **$0 volume on 62 of its 253 days**, and `api.alphasec.trade` has correct non-zero volume for **all 62 of them**. Total dropped: **$168,610,465**, which is 6.2% on top of the $2,712,264,430 currently shown all-time.

The adapter and the API are both fine. The problem is that the API answers `HTTP 200` with an all-zero row for a day it hasn't finalised yet, and the adapter passes that through as real data.

## The tell

Every finalised row carries `updatedAt ≈ timestamp + 88,206s` — the day is written about 24.5 hours after it starts, i.e. ~00:30 UTC the following day. A row that hasn't been written yet comes back with `updatedAt === timestamp` exactly:

```
today     ?startOfDay=1786579200 -> 200 {"dailyVolume":0,...,"timestamp":1786579200,"updatedAt":1786579200}
future+1d ?startOfDay=1786665600 -> 200 {"dailyVolume":0,...,"timestamp":1786665600,"updatedAt":1786665600}
bogus 0   ?startOfDay=0          -> 200 {"dailyVolume":0,...,"timestamp":0,"updatedAt":0}

yesterday ?startOfDay=1786492800 -> 200 {"dailyVolume":7143115,"dailyFees":4286,...,"updatedAt":1786581006}
```

No error, no missing field, no HTTP status to catch — just zeros. `dailyVolume: stats.dailyVolume` is returned raw and `addUSDValue(0)` is a no-op, so a run that lands before ~00:30 UTC stores zeros permanently. Nothing ever goes back to correct them.

## Verification

The adapter is correct — running it locally for a day we published `$0` for returns the right number:

```
$ npm test dexs alphasec-spot 2026-08-10
Start Date:  Sun, 09 Aug 2026 00:00:00 GMT
End Date:    Mon, 10 Aug 2026 00:00:00 GMT
Daily volume: 4.06 M
Daily fees:   2.44 k
```

against `$0` volume / `$2,437` fees stored for 2026-08-09.

Recent examples, API value vs what we published:

| day | api dailyVolume | api fees | published volume | published fees |
|---|---|---|---|---|
| 2026-08-01 | $2,050,255 | $1,230 | **0** | $1,230 |
| 2026-08-02 | $2,108,485 | $1,265 | **0** | $1,265 |
| 2026-08-03 | $2,248,899 | $1,350 | **0** | $1,350 |
| 2026-08-05 | $2,244,986 | $1,347 | **0** | $1,347 |
| 2026-08-06 | $2,221,618 | $1,333 | **0** | **0** |
| 2026-08-07 | $2,819,672 | $1,691 | **0** | $1,691 |
| 2026-08-08 | $3,106,654 | $1,864 | **0** | $1,864 |
| 2026-08-09 | $4,061,305 | $2,437 | **0** | $2,437 |
| 2026-08-10 | $7,993,726 | $4,796 | **0** | **0** |
| 2026-08-12 | $7,143,115 | $4,286 | **0** | $4,286 |

I checked all 62 zero-volume days against the API, not just these — **62 of 62** have non-zero volume upstream, none are genuinely quiet days.

## Why volume loses far more often than fees

The two dimensions run as separate jobs, and this module serves both. Fees only lost 6 days ($11,012 of $101,166 on the affected days); volume lost all 62. So the `dexs` job is landing in the pre-00:30 window much more often than the `fees` job. That's also what makes it easy to miss — the protocol looks alive on the fees chart the whole time.

The two are tightly coupled here, which is the cross-check: on the 190 days where both published, implied fee rate is a flat **0.0500%**, so the fee series alone says the volume on those 62 days was never actually zero.

## Fix

Throw when the row isn't finalised, rather than publishing its zeros. `updatedAt <= timestamp` is exact for this — it's true for every unpopulated row above and false for every real one — so a genuinely quiet day would still publish $0 correctly if the DEX ever has one.

PR coming.

0xshubhs:
Fixes #8748

`dexs/alphasec-spot` has published `$0` volume on **62 of its 253 days**, and `api.alphasec.trade` has correct non-zero volume for **all 62**. $168,610,465 dropped, 6.2% on top of the $2.71B currently shown all-time.

Neither the adapter nor the API is broken. The API answers `HTTP 200` with an all-zero row for a day it hasn't finalised yet:

```
today     ?startOfDay=1786579200 -> 200 {"dailyVolume":0,...,"timestamp":1786579200,"updatedAt":1786579200}
yesterday ?startOfDay=1786492800 -> 200 {"dailyVolume":7143115,...,"updatedAt":1786581006}
```

No error, no missing field, nothing to catch; the row is written ~24.5h after the day starts (`updatedAt - timestamp` is 88,206s on every finalised row I checked), and a run landing before that stores zeros permanently. `dailyVolume` was passed through raw and `addUSDValue(0)` is a no-op, so nothing flags it.

`updatedAt <= timestamp` is exact for this: true for every placeholder row (today, a future date, `startOfDay=0`), false for every real one. So a genuinely quiet day would still publish `$0` correctly, I checked all 62 zero days against the API and none of them were genuinely quiet, but the guard doesn't assume that.

Fees mostly survived (6 days lost, $11,012) because `dexs` and `fees` run as separate jobs and only the `dexs` one keeps landing in the pre-00:30 window; which is why the protocol looks alive on the fees chart the whole time. The two are tightly coupled: implied fee rate is a flat 0.0500% across the 190 days where both published, so the fee series alone says volume was never zero on those days.

Verified:

| run | result |
|---|---|
| `2026-08-10` (covers 08-09, we published $0) | Daily volume 4.06 M, fees 2.44 k |
| `2026-08-13` (covers 08-12, we published $0) | Daily volume 7.14 M, fees 4.29 k |
| `2026-08-14` (covers today, not finalised) | throws `has not finalised 1786579200 yet (volume 0, timestamp 1786579200, updatedAt 1786579200)` |

Both finalised days match the API exactly and are unchanged from master's behaviour when it happens to run late enough.

### DIFF
--- a/dexs/alphasec-spot.ts
+++ b/dexs/alphasec-spot.ts
@@ -14,6 +14,15 @@ const fetch = async (options: FetchOptions) => {
   const data = await httpGet(url);
   const stats = data.result;
 
+  if (!stats || Number(stats.updatedAt) <= Number(stats.timestamp))
+    throw new Error(
+      `alphasec-spot: api.alphasec.trade has not finalised ${options.startOfDay} yet (volume ${stats?.dailyVolume}, timestamp ${stats?.timestamp}, updatedAt ${stats?.updatedAt})`,
+    );
+
+  const dailyVolume = Number(stats.dailyVolume);
+  if (!Number.isFinite(dailyVolume))
+    throw new Error(`alphasec-spot: api.alphasec.trade returned no usable volume (${stats.dailyVolume}) for ${options.startOfDay}`);
+
   const dailyFees = options.createBalances();
   const dailyRevenue = options.createBalances();
   const dailySupplySideRevenue = options.createBalances();
@@ -23,7 +32,7 @@ const fetch = async (options: FetchOptions) => {
   dailySupplySideRevenue.addUSDValue(stats.dailySupplySideRevenue, metrics.TradingRebatesAndCommissions);
 
   return {
-    dailyVolume: stats.dailyVolume,
+    dailyVolume,
     dailyFees,
     dailyRevenue,
     dailySupplySideRevenue,


0xshubhs:
Flagging the red CI before it gets read as this patch breaking something — the 403 happens before any line I touched.

The failure is thrown inside `httpGet(url)` on line 14, which is unchanged. My guard is on line 17, after it returns. Nothing in the diff can produce a 403.

It's the runner's IP. `api.alphasec.trade` answers fine from a normal host — same URL the CI just tried, four different clients:

```
curl                      -> 200
curl -H 'User-Agent: axios/1.7.7'  -> 200
node global fetch         -> 200  {"dailyVolume":7143115,"dailyFees":4286,...}
this repo's axios         -> 200  {"dailyVolume":7143115,...}
```

so it isn't a client or UA check the way houdiniswap's Cloudflare block is (#8712) — it's a network-level block on GitHub Actions ranges. Worth knowing that this adapter can't be exercised from fork CI at all, on master or on this branch.

It doesn't change the diagnosis in #8748. A 403 makes `httpGet` throw, which stops the chart rather than writing zeros — that's the houdiniswap failure mode. The 62 days here are the opposite shape: a clean `HTTP 200` carrying an all-zero placeholder row, stored as real data, with fees intact on the same days. Those two can't be the same cause.

bheluga:
@0xshubhs thanks for the PR

PR #8755, opened 2026-08-13
durianfun-launchpad: index the V5 factory (different fee schedule)

Duriandotfun:
## Durianfun Launchpad — add the V5 factory

`dexs/durianfun-launchpad/index.ts` covers V4.5 / V4.6.6 / V4.6.7. **V5** went live 2026-08-12 (`0xE3861e300043d8c20A927340cbA6379D0BECb793`, block 34,014,875) and is where every new coin launches — its volume and fees are currently unindexed.

### Why V5 needs its own pass
The trade events are byte-identical (`TokensBought` / `TokensSold`, same topic0), and `TokenCreated` is the same 8-arg event as V4.6.7 — so no new ABIs. But the **fee schedule differs**:

| | total fee | split |
|---|---|---|
| V4.5 / V4.6.6 / V4.6.7 | 1.0 % | 90 % treasury / 10 % creator |
| **V5** | **1.167 %** (11,670 ppm) | treasury 10,000 ppm · referral 1,000 ppm · creator 670 ppm |

So the existing `treasury = fee * 9n / 10n` would be wrong for V5. The legacy market list and both legacy loops are left **byte-identical**; V5 markets are collected into their own list and summed in a separate pass that mirrors the contract's integer math (`creatorPart = fee * 670 / 11670`).

The V5 referral slice is routed to the **treasury** unless the on-chain volume registry marks a referrer eligible, in which case the market emits `ReferralPaid`. Rather than assume either extreme, the adapter sums `ReferralPaid` in the daily window and moves exactly that amount from revenue to supply-side revenue (reported under `METRIC.OPERATORS_FEES`).

### On-chain verification (KUB mainnet, block 34,316,959)

**All-time since the V5 deploy** — 563 buys + 370 sells, 0 `ReferralPaid`:
```
dailyVolume  = 30,720.831272 KUB
dailyFees    =    358.512101 KUB   → 1.1670 % of gross  ✓ matches the on-chain totalFeePpm()
dailyRevenue =    337.929144 KUB   (94.26 % of fees)
supplySide   =     20.582957 KUB   (5.74 % of fees, creators)
conservation : revenue + supplySide == fees, EXACT
```

**Last ~24 h (28,800 blocks)** — 265 buys + 185 sells: **13,899.68 KUB volume / 162.21 KUB fees**, same 1.1670 % and same exact conservation.

The adapter still never hardcodes a rate against volume — it reads the `fee` field each trade emits; the ppm constants only split an already-collected fee between recipients.

### DIFF
--- a/dexs/durianfun-launchpad/index.ts
+++ b/dexs/durianfun-launchpad/index.ts
@@ -13,7 +13,7 @@
  * liquidity is migrated to a `DurianAMM` pool (covered by a separate
  * adapter, `durian-amm`).
  *
- * Three generations of the factory are LIVE and indexed together so the
+ * Four generations of the factory are LIVE and indexed together so the
  * volume series is continuous:
  *
  *   V4.5   — `0xdf4f3dB298A9aDe853191F58b4b2a322D47EC005` (deploy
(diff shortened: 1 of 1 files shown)


bheluga:

PR #8756, opened 2026-08-14
AGNT: count both app-fee recipient wallets for trading fees

tut9492:
The AGNT fees adapter measures the 0.40% in-app swap fee at the Relay app-fee recipient wallet. That recipient (`RELAY_APP_FEE_RECIPIENT`) was **repointed** from the original wallet `0x585b6854…0598` to the launch-fee treasury wallet `0x5bF5805e…C5f0`, so trading fees have landed at **both** over time — but the adapter only tracked the pre-repoint wallet, missing all post-repoint (and ongoing) swap fees.

This PR counts both recipient wallets.

- Verified against Relay's API (`/requests/v2` `paidAppFees`): app fees split across the two wallets (majority now at the post-repoint wallet).
- Still filtered to the Relay settlement + claim senders (`fromAdddesses`), so the launch-fee wallet's Doppler-initializer inflows remain the launchpad leg and its Relay inflows are the trading leg — disjoint sender sets, **no double counting**.
- Fees-only change; methodology strings updated to name both wallets.

### DIFF
--- a/fees/agnt/index.ts
+++ b/fees/agnt/index.ts
@@ -31,12 +31,22 @@ const PLATFORM_FEE_WALLET = "0x5bF5805e4809A447a61621f8698CEdeA2D1fC5f0"; // CLA
 const PLATFORM_SHARE = 0.32;
 const CREATOR_SHARE = 0.63;
 const DOPPLER_SHARE = 0.05;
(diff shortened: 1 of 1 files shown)


bheluga:
@tut9492 thanks for the PR

PR #8770, opened 2026-08-14
Perpl: fix event topic0s for contract upgrades (fills + position fees)

0x70626a:
## Summary

The Perpl exchange renamed four events across two on-chain upgrades. Each V2 variant only **appends** members, so its `topic0` differs from the V1 form while every pre-existing field (and its byte offset) is unchanged. This adapter filtered on the V1 signatures, so those log queries stopped matching after each upgrade:

- **rc_v1.1.7.4** (block 95,662,781, 2026-08-13): `MakerOrderFilled` → `MakerOrderFilledV2`, `TakerOrderFilled` → `TakerOrderFilledV2` (append `builderId`, `builderFeeCNS`). This zeroed volume and maker/taker fees.
- **rc_v1.1.7.3** (2026-06): `PositionOpened` → `PositionOpenedV2`, `PositionIncreased` → `PositionIncreasedV2` (append `priceResiduePNSQ16`). This zeroed position insurance/protocol fees.

## Fix

Match **both** the V1 (pre-upgrade history) and V2 (current) forms of each event, so daily figures stay continuous across each upgrade block. Byte offsets are unchanged because every V2 variant only appends fields. The `getPerpetualInfo` view call is untouched (that struct did not change).

## Test

`npm test dexs perpl` over 2026-08-13 → 2026-08-14, which spans the rc_v1.1.7.4 upgrade block:

- Daily volume: **$61.35M**, continuous across the upgrade boundary (no cliff at 15:07 UTC on 2026-08-13).
- Fee breakdown now includes the position fees that were previously reported as zero:

| Fee | Value |
| --- | --- |
| Maker Fees | 2,635 |
| Taker Fees | 4,435 |
| Position Insurance Fees | 1,012.69 |
| Position Protocol Fees | 5,737 |

### DIFF
--- a/dexs/perpl/index.ts
+++ b/dexs/perpl/index.ts
@@ -5,19 +5,17 @@ import { CHAIN } from "../../helpers/chains";
 const EXCHANGE = "0x34B6552d57a35a1D042CcAe1951BD1C370112a6F";
 const CNS_DECIMALS = 1e6;
 
-// The exchange renamed four events across two upgrades. Each V2 variant only APPENDS members,
-// so its topic0 differs but every pre-existing field (and its byte offset) is unchanged. We match
-// BOTH the V1 (pre-upgrade history) and V2 (current) forms so daily figures stay continuous across
-// each upgrade block:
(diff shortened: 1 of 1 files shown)


bheluga:
@0x70626a thanks for the PR

PR #8774, opened 2026-08-14
verus: throw when the api reports no lp_volume instead of publishing zero

### LINKED ISSUE #8773: verus: 22 days of $0 volume since 2026-07-24, marketapi returns lp_volume null for every pool
`dexs/verus` has published `$0` volume every day since 2026-07-24 — 22 consecutive days against $927,310 of 30d volume before it stopped. The pools are still trading. The API stopped reporting volume and the adapter turns that into a real-looking zero.

```
https://api.llama.fi/summary/dexs/verus-market
last nonzero   2026-07-23   $20,369
2026-07-24 .. 2026-08-14    $0 on all 22 days
```

## Why it publishes zero

`dexs/verus.ts` sums one field with no guard:

```ts
const response = await httpGet('https://marketapi.verus.services/getdefichaininfo');

let dailyVolume = 0;
for (const result of response.data.results) {
  dailyVolume += Number(result.lp_volume);
}
```

That endpoint is up and answers `200`. It returns all 20 pools, each one still holding reserves — and `lp_volume` is `null` on every single one:

```
getdefichaininfo: 20 pools
  lp_volume === null   : 20
  lp_volume numeric    : 0
  pools holding reserves: 20 of 20  (67 token legs)
```

```json
{"liquiditypool":"Floralis","lp_address":"i3d4vSCbXYEC3u6TzwohMvdghHkhBrXWpE","lp_volume":null,
 "tokens":[{"ticker":"VRSC","reserves":38458.06189614,...},{"ticker":"DAI.vETH","reserves":15323.66785329,...}]}
```

`Number(null)` is `0`, so the sum comes out to `0`, nothing throws, and $0 gets stored as a genuine no-volume day. On master right now:

```
$ npm test dexs verus
Daily volume: 0.00
Daily fees: 0.00
```

Exit code 0. There is nothing to alert on.

## The pools are definitely still trading

Their own ticker endpoint reports zero volume while reporting prices that moved:

```
https://marketapi.verus.services/market/allTickers
40 pairs
  volume === 0     : 40
  high !== low     : 34

VRSC-Floralis   vol=0   open=20.240948   last=19.927739   hi/lo range  4.16%
USDC-VRSC       vol=0   open=1.2557017   last=1.2405118   hi/lo range  9.80%
scrvUSD-VRSC    vol=0   open=1.4145413   last=1.3777635   hi/lo range  4.11%
ETH-VRSC        vol=0   open=2904.2698   last=2895.7460   hi/lo range  2.24%
```

These are reserve-based conversion pools, so price is a function of the reserves and the reserves only move when someone converts. A 9.80% high-to-low range on USDC-VRSC with zero volume is not a quiet market, it is a broken volume field.

TVL agrees — it has moved on 22 of the 23 days since the volume went flat, including single-day swings of -49.8%, +16.8% and -36.6%.

## It is broken upstream, in more than one place

Every volume route on that API is either null or erroring:

| route | result |
|---|---|
| `/getdefichaininfo` | 200, `lp_volume` null on all 20 pools |
| `/getcurrencyvolumes` | 200, every pool `[null, null]` |
| `/gettotalvolume/{currency}/{from}/{to}` | HTTP 500 |
| `/market/allTickers` | 200, `volume: 0` on all 40 pairs, prices populated |
| `/market/allTickers/new` | HTTP 500 |
| `/gettvl` | `{"code":"500000","message":"An unexpected error occurred: single positional indexer is out-of-bounds"}` |

That last one is a pandas `.iloc` error, which suggests their aggregation layer is falling over and returning null rather than failing the request. So there is no working endpoint to switch to — `gettotalvolume` was the obvious candidate and it 500s.

## Suggested fix

Since there is nowhere else to read volume from, the only thing to fix in this repo is that the adapter cannot tell "no conversions happened" from "the API did not report". `null` and `0` are different answers and it currently treats them the same.

Throwing when no pool reports a numeric `lp_volume` keeps genuine zero-volume days working (those come back as `0`, a number) and turns this into a retryable failure instead of a fabricated data point:

```ts
const results = response.data.results;
if (!results.some((result: any) => typeof result.lp_volume === "number")) {
  throw new Error("verus market api returned no lp_volume for any of its pools");
}
```

Same shape as the guards already merged for `hinkal` (#8581) and `pact` (#8547).

Happy to send that as a PR. It will not backfill the 22 days — that needs the upstream API fixed first — but it stops the run from silently adding more of them.

0xshubhs:
Fixes #8773.

`marketapi.verus.services/getdefichaininfo` is up and returns all 20 pools with their reserves intact, but `lp_volume` has been `null` on every one of them since 2026-07-24. `Number(null)` is `0`, so the sum came out to zero and 22 consecutive days got stored as genuine no-volume days.

The pools are still trading, their own ticker endpoint reports `volume: 0` on all 40 pairs while 34 of them have `high !== low`, including a 9.80% high-to-low range on USDC-VRSC. These are reserve-based conversion pools, so the price only moves when someone converts.

There is nowhere else to read volume from: `/gettotalvolume` and `/market/allTickers/new` both return HTTP 500, `/getcurrencyvolumes` returns `[null, null]` per pool, and `/gettvl` returns a pandas indexing error. Full survey in the issue.

So this only fixes the part that is fixable here, the adapter cannot currently tell "no conversions happened" from "the API did not report". A genuine quiet day comes back as `0`, a number, so requiring at least one numeric `lp_volume` leaves that path working and turns the null case into a retryable failure.

Before, on master:

```
$ npm test dexs verus
Daily volume: 0.00
Daily fees: 0.00
```

exit 0, nothing to alert on. After:

```
$ npm test dexs verus
------ ERROR ------
Error: verus market api returned no lp_volume for any of its pools
```

exit 1, same on a second run.

This will not backfill the 22 days, that needs the upstream API fixed. It stops the run from quietly adding more of them. Same shape as the guards merged for `hinkal` (#8581) and `pact` (#8547).

### DIFF
--- a/dexs/verus.ts
+++ b/dexs/verus.ts
@@ -12,8 +12,13 @@ const methodology = {
 const fetch = async () => {
   const response = await httpGet('https://marketapi.verus.services/getdefichaininfo');
 
+  const results = response.data.results;
+  if (!results.some((result: any) => typeof result.lp_volume === "number")) {
+    throw new Error("verus market api returned no lp_volume for any of its pools");
+  }
+
   let dailyVolume = 0;
-  for (const result of response.data.results) {
+  for (const result of results) {
     dailyVolume += Number(result.lp_volume);
   }
 


0xshubhs:
Flagging the red CI before anyone has to dig into it: that is the guard doing its job, not a broken adapter.

```
🦙 Running VERUS.TS adapter 🦙
------ ERROR ------
Error: verus market api returned no lp_volume for any of its pools
    at fetch (dexs/verus.ts:17:11)
```

The upstream API is still returning `lp_volume: null` for all 20 pools, so the adapter now fails instead of writing another fabricated `$0`. On master the same run exits 0 with `Daily volume: 0.00`, which is the bug.

Useful side effect: this ran on your runner, not mine, so the nulls are not an artefact of my IP or region — the failure reproduces from DefiLlama's own infrastructure.

The check will stay red until Verus fixes their volume pipeline. If you would rather not merge something that cannot go green today, the alternative is to leave it and revisit when their API is back, but the cost of waiting is another fabricated `$0` per day on a chart that has already accumulated 22 of them.

bheluga:
@0xshubhs thanks for the PR

PR #8780, opened 2026-08-15
native: read credit-pool fees from the vault

DaniiRix:
### Changes
- Credit-pool fees were capped at 100 LP tokens and missed protocol fees. The old adapter walked allLPTokens(0..99) and summed YieldDistributed, which:
   - silently dropped markets beyond index 99
   - ignored reserveFee, the vault's protocol cut
   - could silently lose fees on partial RPC failures

Now reads CreditVault's EpochUpdated events directly. They contain both fundingFee and reserveFee, removing the LP-token cap and booking reserveFee to protocol revenue.

### Verification
- pnpm test fees native 2026-07-29 passes all six chains, $5,504 matches published fees.
- X Layer and Robinhood currently have $0 fees but are configured for future activity.

### DIFF
--- a/fees/native.ts
+++ b/fees/native.ts
@@ -40,19 +40,27 @@ const configs: Record<string, any> = {
     ],
     start: '2025-07-09',
   },
+  [CHAIN.XLAYER]: {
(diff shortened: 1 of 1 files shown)


bheluga:
@DaniiRix thanks for the PR

PR #8793, opened 2026-08-16
llamalend-v2-curve: adapter reports zero, and lender interest is ~1000x too small

BhariGowda:
Two bugs in `fees/llamalend-v2-curve.ts` (added in #7575), the first hiding the second.

### 1. Every vault is filtered out, so the adapter publishes 0

The `NewVault` lookup filters on `event.blockNumber`:

```ts
})).filter((event: any) => Number(event.blockNumber) <= fromBlock);
```

`options.getLogs` defaults to `onlyArgs`, which returns the decoded args with no log metadata attached. So `event.blockNumber` is `undefined`, `Number(undefined)` is `NaN`, and `NaN <= fromBlock` is `false` for every event. `vaultCreatedEvents` comes out empty and the function takes its early return.

Confirmed by logging inside the adapter:

```
DEBUG raw NewVault events: 3   blockNumber of first: undefined
DEBUG after filter: 0
```

The markets are real: factory `0x5f94073e3f51c1fff92ffc6b4b06b7af193b3640` on optimism reports `market_count() == 3`. This matches what the API currently shows for Curve LlamaLend V2, which is 0 for both 24h and 7d.

Fixed by passing `entireLog: true` so the block number survives the decode, then mapping back to `log.args`.

### 2. Lender interest is scaled about 1000x too small

```ts
(Number(pricePerShareAfter) - Number(pricePerShareBefore)) * Number(assetsForInterest) / 1e18
```

This treats `pricePerShare` as 1e18-scaled. These vaults do not report it that way. Read on optimism:

| vault | pricePerShare | totalAssets | totalSupply |
| --- | --- | --- | --- |
| 0x638eF785 | 1001334640170175 | 280427312041212853408 | 280053540969365536269413 |
| 0xfF9a3085 | 1004300624981956 | 368400215738 | 366822647918414546252706973 |
| 0xaB504518 | 1003855547240926 | 349350859645 | 348009093006640769314500686 |

`totalSupply * pps / 1e18` reproduces `totalAssets` on the first vault to 9 significant figures, so `pps` is assets per share at 1e18 scale (about 1.003e15 here), not a 1e18-relative rate. Scaling the absolute delta by 1e18 therefore understates the result by a factor of `pps/1e18`, roughly one thousandth.

Multiplying a share-denominated delta by an asset-denominated total is also wrong independently of that, and breaks differently once share and asset decimals diverge, which they do on the two USDC vaults (18dp shares, 6dp asset).

Fixed by growing the pooled assets by the ratio instead, which is decimal free and needs no extra calls:

```ts
Number(assetsForInterest) * (Number(pricePerShareAfter) / Number(pricePerShareBefore) - 1)
```

### Result

With only the first fix the split was 14 fees against 0.13 supply side, an implied 99% admin cut. That is what pointed at the second bug. With both, on optimism for 2026-08-16: **0 -> 143 daily fees**, 129 supply side, 14 revenue, and revenue over fees is 9.8%, matching Curve's 10% admin fee on borrow interest.

Checked on four separate days. `dailyFees` equals supply side plus revenue exactly on each:

| date | fees | supply side | revenue | admin share |
| 2026-08-16 | 143 | 129 | 14 | 9.8% |
| 2026-08-13 | 106 | 96 | 11 | 10.4% |
| 2026-08-10 | 194 | 176 | 18 | 9.3% |
| 2026-08-05 | 63 | 57 | 6.06 | 9.6% |

### Not changed

The AMM swap fee path and `dailyVolume` are untouched. They read zero here only because these three markets have had no soft liquidations in the windows tested, so I could not exercise that path and did not want to change what I could not check.

### DIFF
--- a/fees/llamalend-v2-curve.ts
+++ b/fees/llamalend-v2-curve.ts
@@ -47,15 +47,13 @@ const fetch: FetchV2 = async (options: FetchOptions) => {
 
   const factory = LlamaLendV2Factories[options.chain];
   const fromBlock = await options.getFromBlock();
-  // entireLog so the block number survives the decode: getLogs defaults to
-  // onlyArgs, which returns the decoded args on their own, so event.blockNumber
-  // was undefined and Number(undefined) <= fromBlock dropped every vault.
   const vaultCreatedEvents = (await options.getLogs({
     eventAbi: EventNewVault,
     target: factory.address,
     fromBlock: factory.fromBlock,
     cacheInCloud: true,
     entireLog: true,
+    parseLog: true,
   }))
     .filter((log: any) => Number(log.blockNumber) <= fromBlock)
     .map((log: any) => log.args);
--- a/fees/llamalend-v2-curve.ts
+++ b/fees/llamalend-v2-curve.ts
@@ -47,12 +47,18 @@ const fetch: FetchV2 = async (options: FetchOptions) => {
 
   const factory = LlamaLendV2Factories[options.chain];
   const fromBlock = await options.getFromBlock();
+  // entireLog so the block number survives the decode: getLogs defaults to
+  // onlyArgs, which returns the decoded args on their own, so event.blockNumber
+  // was undefined and Number(undefined) <= fromBlock dropped every vault.
   const vaultCreatedEvents = (await options.getLogs({
     eventAbi: EventNewVault,
     target: factory.address,
     fromBlock: factory.fromBlock,
     cacheInCloud: true,
-  })).filter((event: any) => Number(event.blockNumber) <= fromBlock);
+    entireLog: true,
+  }))
+    .filter((log: any) => Number(log.blockNumber) <= fromBlock)
+    .map((log: any) => log.args);
   if (vaultCreatedEvents.length === 0) {
     return {
       dailyVolume,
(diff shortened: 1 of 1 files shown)


BhariGowda:
Thanks, taken the first one and pushed it.

**Opening assets as the interest base: fixed.** You are right that the fallback had a real window to fire in. On optimism at block 152707740, vault `0x638eF785` reports `totalAssets` 0 with `pricePerShare` already at 1e15, so a funding window would have applied the whole window's price move to deposits that landed part way through it. Dropped the fallback and used `totalAssetsBefore` only. `totalAssetsAfter` became unused so the `toApi` `totalAssets` multiCall went with it.

Published figures are unchanged on every day I tested, since all three markets were already funded at the open, so this is a hazard removal rather than a correction:

```
2026-08-16  143 = 129 + 14
2026-08-13  106 = 96 + 11
2026-08-10  194 = 176 + 18
2026-08-05   63 = 57 + 6.06
```

**Negative lender interest: I think this one is already handled, so I have left it.** Negative values are gated at the result level rather than per `add` call. `adapters/utils/runAdapter.ts:317`:

```ts
if (result[recordType] < 0 && !allowNegativeValue) throw new Error(`value is negative ${errorPartialString}`)
```

The adapter already sets `allowNegativeValue: true`, and that predates this PR, with the existing comment saying vault `pricePerShare` can decrease during market stress and produce negative lender accrual. There is no per-`add` `allowNegativeValue` option to pass at those two sites. So a falling share price flows through as negative supply side revenue, which is the intended behaviour for a lending vault that can realise a loss, and clamping it at zero would ratchet the totals upward instead.

Happy to be corrected if you meant something else by handling it explicitly.

BhariGowda:
@coderabbitai review

bheluga:
@BhariGowda thanks for the PR

Issue #8788, opened 2026-08-15
gmgn is missing doublecounted: $2.5B/30d of bot volume is inside the headline dex total (1.33%), every peer bot is flagged


0xshubhs:
`dexs/gmgnai` doesn't set `doublecounted`, and it's the largest trading bot on the site. Its **$2,516,265,594 of 30d volume is inside the headline DEX total**, on top of the same swaps already counted at the DEXs that actually executed them. All-time that's **$23,019,184,701**.

Every comparable protocol is flagged. GMGN is bigger than all of them.

## The flag really does gate the headline number

Checked rather than assumed, against `/overview/dexs` right now:

```
doublecounted protocols   60      $3,468,327,509
normal                  1248    $188,856,208,260
ALL                     1308    $192,324,535,770

overview total30d               $188,856,208,260
ALL minus doublecounted         $188,856,208,260   <- exact to the dollar
```

So anything not carrying the flag is added straight into the published total. GMGN is **1.33%** of it.

| | published | should be |
|---|---|---|
| dexs total30d | $188,856,208,260 | $186,339,942,666 |
| dexs total24h | $5,565,047,269 | $5,485,857,604 |

## Every Telegram Bot / Trading App with volume, and its flag

| 30d volume | doublecounted | protocol |
|---|---|---|
| **$2,516,265,594** | **false** | **GMGN** |
| $1,533,425,036 | true | Axiom |
| $751,501,379 | true | fomo Wallet |
| $273,236,536 | true | Terminal |
| $99,908,379 | true | Trojan |
| $41,934,294 | true | Photon |
| $17,484,974 | true | moonshot.money |
| $16,973,758 | true | BONKbot |
| $16,581,343 | true | Ave.ai |
| $16,140,092 | true | Banana Gun |
| $7,939,108 | true | SolTradingBot |
| $5,436,005 | true | BullX |
| $4,917,237 | false | o1.exchange Trading Terminal |
| $2,321,029 | true | Telemetry |
| $359,615 | true | Pepeboost |
| $3,247 | true | Unibot |
| $28 | false | DTrade |
| $3 | false | Priority |

16 of 18 are flagged. The unflagged ones are GMGN and three that are rounding errors ($4.9M, $28, $3). GMGN alone is 99.8% of the unflagged volume in these two categories.

## Why GMGN is double-counted

Not an inference from the category name — it's in this repo. `fees/gmgnai.ts` describes the business model:

[...]

PR #8800, opened 2026-08-16
o1-exchange: mark volume doublecounted, its a 1% terminal like gmgn

0xshubhs:
Follow-up to #8788 / #8789. Same class, much smaller: **$4,917,237/30d**.

o1.exchange is a trading terminal that takes a cut per trade, so its volume is the same swaps already counted at the DEXs that executed them. It's the last unflagged protocol of any size in the `Telegram Bot` / `Trading App` categories; after gmgn merged, the only others left are $28 and $3.

The adapter says it outright:

```
Volume: 'Total trading volume is calculated as fees multiplied by 100, since trading fees are 1% of the volume.'
Fees:   'User pays 1% fee on each trade'
```

and the fees themselves are measured as transfers into its fee-collector wallets, `to_owner = 'FUzZ2SPwLPAKaHubxQzRsk9K8dXb4YBMR6hTrYEMFFZc'` on solana, `0x1E493E7CF969FD7607A8ACe7198f6C02e5eF85A4` / `0xc98218Df72975EE1472919d2685e5BD215Baaad4` on base, `0x68AF11c8A93B373BA97217963878B097083726f0` on bnb. A 1% cut skimmed into a wallet is an interface fee; the trade happened elsewhere. Its three chains (solana, base, bsc) are all ones where DefiLlama already counts the underlying DEXs, and `/protocol/o1-exchange` has no tvl listing, so there are no o1 pools for the volume to have traded against.

Same scope as gmgn, dexs only, `fees/` untouched, since the 1% is genuinely o1's own revenue.

Can't run it locally (`DUNE_API_KEYS` not set) but this is a declarative flag that doesn't touch `prefetch`/`fetch`, so there's no number to verify. `tsc --noEmit` clean, same as master.

Still unchecked and deliberately not in this PR: the `Interface` category has four unflagged ones (Kreo, MetaMask Predictions, Polycule, Rainbow Predictions, ~$178k/24h combined) sitting next to the flagged Gate Predictions and Traderline Predictions. They look like the same thing but that category is mixed, so I'd rather verify each than sweep them in.

### DIFF
--- a/dexs/o1-exchange/index.ts
+++ b/dexs/o1-exchange/index.ts
@@ -115,6 +115,7 @@ const adapter: SimpleAdapter = {
   },
   isExpensiveAdapter: true,
   dependencies: [Dependencies.DUNE],
+  doublecounted: true,
 };
 
 export default adapter;


bheluga:
@0xshubhs thanks for the PR

PR #8801, opened 2026-08-16
nad-fun: keep V2 pair metadata out of the dust filter

### LINKED ISSUE #8629: nad.fun: filterPools dust filter drops V2 bonding-curve fees (revenue > fees)
@g1nt0ki Flagging a metrics inconsistency on the nad.fun adapter (`dexs/nad-fun.ts`) that started after #8061.

## Symptom

On the live API, `dailyRevenue` exceeds `dailyFees` on several days — which shouldn't be possible, since protocol revenue is a subset of fees.

| date | fees | revenue |
|---|---|---|
| 2026-07-31 | 285 | **409** |
| 2026-08-01 | 424 | **475** |
| 2026-08-05 | 520 | **1,102** |
| 2026-08-06 | 401 | **422** |

7d totals are `fees 3,110` vs `revenue 3,044`, i.e. supply-side revenue collapses to ~0, which is also wrong (creator fees and LP fees are a large share of nad.fun fees).

## Evidence

I ran the adapter locally with **only** the `filterPools` block removed from `getV2PairMetadata`, everything else identical:

**2026-08-05**

| metric | live API | local (filterPools removed) | |
|---|---|---|---|
| volume | 182,218 | 182,260 | match |
| fees | 520 | **2,370** | ~78% missing |
| revenue | 1,102 | 1,100 | match |

**2026-08-04**

| metric | live API | local (filterPools removed) | |
|---|---|---|---|
| volume | 12,955 | 12,880 | match |
| fees | 258 | **293** | ~12% missing |
| revenue | 124 | 124 | match |

Volume and revenue match almost exactly; **only fees diverge.** That asymmetry points straight at the fee path.

## Root cause

`getV2PairMetadata` filters pairs by pooled value ([dexs/nad-fun.ts#L406-L417](https://github.com/DefiLlama/dimension-adapters/blob/master/dexs/nad-fun.ts#L406-L417)):

```ts
const filteredPairs = await filterPools({ api, pairs: pairTokens, createBalances });
const keptPairs = pairs.filter((pair) => filteredPairs[pair] !== undefined);
...
return { pairs: keptPairs, pairMeta: keptPairMeta };
```

`filterPools` defaults to `minUSDValue = 200` (`helpers/uniswap.ts#L10`).

The problem is nad.fun-specific: **the `NadFunPair` is deployed at token creation, not at graduation.** `BondingCurve._create` calls `factory.createPair(token, quoteToken)` up front, but liquidity is only seeded when the token graduates. So **every token still on the bonding curve has a pair with 0 TVL**, and the $200 filter removes all of them.

That cascades into the fee loop (L579-L594):

```ts
const meta = pairMeta[log.pair.toLowerCase()];       // undefined for filtered-out pairs
const creatorRate = meta?.creatorFeeRate ?? 0;
const protocolRate = curveTradeTxs.has(rawLog.transactionHash)
  ? meta?.curveProtocolFeeRate ?? 0
  : meta?.dexProtocolFeeRate ?? 0;
const totalRate = creatorRate + protocolRate;
if (totalRate === 0) {
  return;                                            // entire Collect amount dropped from dailyFees
}
```

So the full `Collect` amount for every bonding-curve trade is silently dropped from `dailyFees`.

Meanwhile `dailyRevenue` is computed from raw quote-token `Transfer` logs to the protocol fee receiver (`addV2ProtocolTransfers`, L674+). That path does no pair lookup at all, so it is completely unaffected by the filter and still counts 100% of protocol revenue.

**Net effect: bonding-curve volume is counted, bonding-curve protocol revenue is counted, but bonding-curve fees are not — so revenue can exceed fees.**

Secondary effect: the swap-log query uses `targets: [...pairSet]` (L619), so LP fees and swap volume from graduated pools under $200 are dropped too.

## Proposed fix

Two options, happy to go either way:

**A. Drop `filterPools` from this adapter.** nad.fun pairs are all protocol-deployed through `NadFunFactory` and fee configs come from the protocol's own `FeeCollector`, so there is no permissionless-dust-pair problem that the filter protects against here. (For the swap-log query, `noTarget: true` + filtering by `pairSet` in-process avoids the large `targets` array.)

**B. Keep the filter, but scope it to swap volume only** — keep `pairMeta` complete for every pair so the `Collect` fee attribution always resolves, and apply the $200 threshold only where dust pools actually distort the DEX volume number.

I'd lean toward **A** for correctness, but if the filter was added for a reason I'm not seeing (RPC cost, dust noise from graduated pools), **B** keeps that intact. Could you let me know which you prefer? I'll open the PR either way.

BhariGowda:
Fixes #8629.

`getV2PairMetadata` runs the pair list through `filterPools` and returns only the survivors in **both** `pairs` and `pairMeta`. The $200 pooled-value floor is correct for `pairs`, which is only used as the target list for the Swap logs that drive volume. It is not correct for `pairMeta`, which the Collect loop reads for every pair it sees.

### Why nad.fun specifically

`NadFunPair` is deployed by `BondingCurve._create` at token creation, and only seeded with liquidity at graduation. A token still on the curve therefore has a pair holding nothing.

I sampled 14 pairs across the whole index range, from index 5 to index 1770 of 1779, and every one reports zero reserves. Checked `0xC43b3F281472E310d786F7fEfdB48DE27A8b5e4A` directly: `balanceOf` is 0 for both tokens and LP `totalSupply` is 0. So the filter is removing nearly the entire pair set, not a dust tail.

Those pairs still emit `Collect`. With no entry in `pairMeta` the rates fall through to `meta?.creatorFeeRate ?? 0` and the protocol equivalents, the `totalRate === 0` guard returns early, and that event's fees are discarded. Revenue does not come through this path, so it survives, and the published series ends up with revenue above fees:

```
2026-08-05   fees 520   revenue 1102
2026-08-06   fees 400   revenue  422
```

### The change

Keep `filterPools` gating `pairs`; return `pairMeta` unfiltered.

### Measurement

Same shell, `DISABLE_PULL_HOURLY=true`, 2026-08-05, only this change between the two runs:

| metric | before | after |
| --- | --- | --- |
| volume | 183.98k | 183.98k |
| fees | 522 | **2.42k** |
| revenue | 1.12k | 1.12k |
| supply side | 352 | **1.30k** |

Volume and revenue come out identical, which is the point: the dust filter still does its job on the volume path, and the revenue path is untouched. Only the fee legs move.

`dailyFees` now equals revenue plus supply side exactly (1.12k + 1.30k = 2.42k). Before, 1.12k + 352 did not add up to 522, which is what the issue was reporting.

Fee breakdown that reappears on this day: Protocol Fees 978, Creator Fees 1,027, LP Fees 225, Foundation Fees from LPs 92, Creators Fees from LPs 92.

### Relation to the reported fix

@bc1phugo tested removing `filterPools` altogether and measured 2,370 in fees for this day. This is the narrower version of that, landing at 2,420 while leaving volume filtered, so the dust pools stay out of the swap-log target list.

### DIFF
--- a/dexs/nad-fun.ts
+++ b/dexs/nad-fun.ts
@@ -401,20 +401,25 @@ async function getV2PairMetadata(options: FetchOptions) {
     };
   });
 
-  // Drop dust pairs: keep only those whose pooled token value clears the
-  // default filterPools threshold ($200), matching the uniswap-fork helpers.
+  // Drop dust pairs from the swap-log targets: keep only those whose pooled
+  // token value clears the default filterPools threshold ($200), matching the
+  // uniswap-fork helpers.
+  //
+  // The metadata map is deliberately NOT filtered. NadFunPair is deployed by
+  // BondingCurve._create at token creation and only seeded with liquidity at
+  // graduation, so every token still on the curve has a pair holding nothing.
+  // Those pairs still emit Collect, and the fee loop reads its rates out of
+  // this map: dropping them there leaves creatorFeeRate and the protocol rates
+  // at 0, which trips the `totalRate === 0` early return and discards the fees
+  // entirely while revenue, which does not come through here, still counts.
   const filteredPairs = await filterPools({
     api: options.api,
     pairs: pairTokens,
     createBalances: options.createBalances,
   });
   const keptPairs = pairs.filter((pair: string) => filteredPairs[pair] !== undefined);
-  const keptPairMeta: Record<string, V2PairMeta> = {};
-  keptPairs.forEach((pair: string) => {
-    keptPairMeta[pair] = pairMeta[pair];
-  });
 
-  return { pairs: keptPairs, pairMeta: keptPairMeta };
+  return { pairs: keptPairs, pairMeta };
 }
 
 async function getV2TokenQuoteMap(options: FetchOptions, tokens: string[]) {


bheluga:
@BhariGowda thanks for the PR

PR #8811, opened 2026-08-17
feat(oroswap): add volume and fees adapters

### LINKED ISSUE #8810: Track OroSwap (ZIGChain DEX) volume and fees
[OroSwap](https://defillama.com/protocol/oroswap) is a native AMM DEX on ZIGChain, already tracked on DefiLlama for TVL (~$3.8M) but absent from both `/dexs` (volume) and `/fees`.

Its pool-manager API exposes the numbers directly:
- `GET https://api-mainnet.oroswap.org/api/poolmanager/pools` returns a server-side aggregate `totalVolume24H` (USD) plus per-pool `volume24H` / `fees24H`.
- `totalLiquidity` from the same response matches the DefiLlama TVL, confirming the figures are USD.

So both a volume and a fees adapter are straightforward. Opening a PR that adds both.

gomesalexandre:
closes #8810

### what

Adds **volume and fees adapters for OroSwap**, an AMM DEX on ZIGChain. It is tracked on DefiLlama for TVL (~$3.8M) but was absent from both `/dexs` and `/fees`.

### how

OroSwap's pool-manager API exposes the numbers directly, so both adapters are single-endpoint pulls:

- **`dexs/oroswap`** - `dailyVolume = totalVolume24H`, the server-side aggregate the API returns (confirmed USD: `totalLiquidity` from the same response matches the DefiLlama TVL).
- **`fees/oroswap`** - `dailyFees = sum(pool.details.fees24H)` across all pools. There is no aggregate fees field, so we sum the per-pool values. Reported as raw `{ dailyFees, dailyUserFees }` (fees only) since the protocol-vs-LP split isn't exposed.
- **`CHAIN.ZIGCHAIN`** added to `helpers/chains.ts`.

The endpoint only exposes a rolling 24h snapshot (no historical backfill), so both run with `runAtCurrTime: true`, the same pattern as `dexs/oraidex`. The host is header-gated (403 without a browser `User-Agent` + `Origin`), which `httpGet` satisfies; it is a header check, not a Cloudflare challenge.

### receipts

```
$ npx ts-node --transpile-only cli/testAdapter.ts dexs oroswap
ZIGCHAIN 👇
Daily volume: 567.13 k

$ npx ts-node --transpile-only cli/testAdapter.ts fees oroswap
Daily fees: 516.00
Daily user fees: 516.00
```

Both reconcile against the raw API response for the same pull: aggregate `totalVolume24H` ~$567k and summed per-pool `fees24H` ~$516. `tsc --noEmit` clean.

### DIFF
--- a/dexs/oroswap/index.ts
+++ b/dexs/oroswap/index.ts
@@ -14,9 +14,15 @@ const HEADERS = {
 
 const fetch = async () => {
   const data = await httpGet(POOLS_URL, { headers: HEADERS });
-  const dailyVolume = Number(data?.totalVolume24H);
+  const raw = data?.totalVolume24H;
+  // Reject missing/blank before Number(): Number(null) and Number("") are a
+  // finite 0, which would silently report zero volume.
+  if (raw === null || raw === undefined || raw === "") {
+    throw new Error("OroSwap: missing totalVolume24H");
+  }
+  const dailyVolume = Number(raw);
   if (!Number.isFinite(dailyVolume)) {
-    throw new Error(`OroSwap: invalid totalVolume24H (${data?.totalVolume24H})`);
+    throw new Error(`OroSwap: non-numeric totalVolume24H (${raw})`);
   }
   return { dailyVolume };
 };
--- a/dexs/oroswap/index.ts
+++ b/dexs/oroswap/index.ts
@@ -1,0 +1,40 @@
+import { httpGet } from "../../utils/fetchURL";
+import { SimpleAdapter } from "../../adapters/types";
+import { CHAIN } from "../../helpers/chains";
+
+// OroSwap is an AMM DEX on ZIGChain. Its pool-manager endpoint returns a
+// server-side aggregate `totalVolume24H` (USD), alongside per-pool figures.
+// The mainnet host is header-gated (403 without a browser UA + Origin), which
+// httpGet can satisfy; this is a header check, not a Cloudflare challenge.
+const POOLS_URL = "https://api-mainnet.oroswap.org/api/poolmanager/pools?offset=0&limit=1000";
+const HEADERS = {
+  "User-Agent": "Mozilla/5.0",
+  Origin: "https://app.oroswap.org",
+};
+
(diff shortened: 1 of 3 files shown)


gomesalexandre:
Thanks @coderabbitai. Went through all 5 against the current code (`8eb0b42`):

**Fixed:**
- **Volume validation** - `totalVolume24H` now rejects missing/blank before `Number()` (`Number(null)`/`Number("")` are a finite 0 that would report zero volume).
- **Per-pool fees validation** - the reduction now throws on a missing/non-numeric `fees24H` instead of summing it as zero. All 165 pools currently expose a numeric `fees24H`, so this only guards against future malformed data.

**Declined, with reasons:**
- **"Replace the blocked API / use ZIGChain swap events."** The API isn't blocked - it's header-gated. The adapter sends a browser `User-Agent` + `Origin`, and the endpoint returns 200 with the real numbers. The CLI confirms it: `dexs oroswap` -> ~$563k, `fees oroswap` -> ~$510, both reconciling against the response's aggregate `totalVolume24H` and summed per-pool `fees24H`. A bare curl 403s (no headers), which is likely what the sandbox saw. Re-deriving this from on-chain swap events across ZIGChain would be a large, less reliable reimplementation of data the official API already returns.
- **version 2 + `pullHourly`.** The endpoint exposes only a rolling 24h snapshot, not time-windowed hourly data, so `pullHourly` doesn't apply. `runAtCurrTime: true` is the established pattern for exactly this kind of current-snapshot source (see `dexs/oraidex`), which is what both adapters use.
- **Split fees into `dailyRevenue` + `dailySupplySideRevenue`.** OroSwap doesn't publish a protocol-vs-LP fee split, so I'd be guessing the ratio. Following the `fees/paradex` pattern (and the same call made on #8808), I report fees only as raw `{ dailyFees, dailyUserFees }`, which is honest about what the source actually exposes. Happy to add the split if OroSwap documents one.

bheluga:

gomesalexandre:
Thanks for the merge!

Issue #8816, opened 2026-08-17
fees/ntm: nothing published since 2026-07-29 while api.ntm.ai still serves every day and the adapter runs clean locally


BhariGowda:
`fees/ntm` last published on 2026-07-29 and has written nothing since. `total24h` and `total7d` are both null and the chart stops rather than trailing zeros, which is the signature of the adapter erroring rather than writing a real zero.

```
2026-07-25=1281  2026-07-26=1369  2026-07-27=1332  2026-07-28=1341  2026-07-29=1345  <- last point
```

Density before the stop was 60 of the last 60 days nonzero, hovering around $1,330/day, so this is not a protocol that pays in bursts. `total30d` is $16,034 and `total1y` is $686,453. Roughly $25k has gone unrecorded over the 19 days.

## The source is healthy and the adapter still works

`https://api.ntm.ai/feesAndRevenues.php` answers normally for every chain in `chainToken`, checked 2026-08-17:

```
ethereum   {"fees_total":0.05,"revenue_total":0.05}
bsc        {"fees_total":0.25,"revenue_total":0.25}
avax       {"fees_total":6,"revenue_total":6}
solana     {"fees_total":3,"revenue_total":3}
tron       {"fees_total":850,"revenue_total":850}
ton        {"fees_total":5,"revenue_total":5}
```

It also answers for the individual days inside the gap, so this is not a backfill hole in the source. Probing solana one day at a time: 07-29 returns 3, 07-30 returns 3, 07-31 returns 1, 08-05 returns 2, 08-10 returns 2, 08-16 returns 3. All HTTP 200, all under half a second.

Running the adapter locally against the last 24h window produces sane output that matches the pre-stop magnitude:

```
ethereum  285.00   bsc  302.00   avax  76.00
solana    303.00   tron 565.00   ton   13.00
Aggregate 1.54 k
```

So the code is fine and the data is there. Nothing in `fees/ntm/index.ts` has changed since #6026 in February.

## Guess at the cause, stated as a guess

`api.ntm.ai` sits behind Cloudflare (`server: cloudflare`, `cf-ray` present, `x-powered-by: PHP/8.3.33` / PleskLin behind it). The most likely explanation for "works from a residential IP, stopped for the runner on a specific date" is a Cloudflare bot-management or IP-reputation rule that started challenging or blocking DefiLlama's egress around 2026-07-30. I cannot verify that from outside, and the runner logs would settle it immediately.

[...]

PR #8820, opened 2026-08-17
fix ntm - reject unusable amounts instead of feeding them to addCGToken

BhariGowda:
Related to #8816.

**Scope note:** an earlier revision of this branch also swapped `httpGet` for `fetchURL` to get its three retries. That is dropped, per the discussion below. This is now response validation only.

## Situation

`fees/ntm` has published nothing since 2026-07-29. `total24h` and `total7d` are null and the chart stops rather than trailing zeros, so it is throwing.

Everything it depends on works when checked by hand:

- `api.ntm.ai/feesAndRevenues.php` answers for all six chains in `chainToken`, checked 2026-08-17: ethereum 0.05, bsc 0.25, avax 6, solana 3, tron 850, ton 5
- it answers for the days inside the gap too. Probing solana day by day: 07-29 → 3, 07-30 → 3, 07-31 → 1, 08-05 → 2, 08-10 → 2, 08-16 → 3. All 200, all under half a second
- it does not care about the User-Agent. `axios/1.7.2`, `axios/0.27.2`, `node`, `Mozilla/5.0` and no UA header at all all return 200
- the adapter runs clean locally at $1.54k/day, in line with the ~$1,330/day it published before the stop
- all six chains stopped on **exactly** 2026-07-29, which rules out a per-chain coingecko id problem

## Cause, now confirmed

The `test` job on the first revision of this branch answered it. Running the adapter on a GitHub Actions runner:

```
🦙 Running NTM adapter 🦙
------ ERROR ------
Request failed with status code 403
```

**403 from a datacenter IP**, against 200 from a residential one with any User-Agent. The only variable between the two callers is the network. So api.ntm.ai answers residential addresses and refuses datacenter ones, which explains both the simultaneous stop across all six chains and why the endpoint looks perfectly healthy to anyone checking by hand.

## Why the retry is gone

It was in here as a candidate fix, and the 403 rules it out: a deterministic refusal retries into three more refusals. It bought nothing and only widened the diff, so `httpGet` is left exactly as it was.

## What is left, and why it is still worth having

The response was used unchecked:

```ts
dailyFees.addCGToken(token, res.fees_total);
```

`Number(undefined)` is `NaN`, and both `Number(null)` and `Number("")` are a finite `0`. So a non-JSON body, a renamed field, or a Cloudflare challenge page returned with a 200 would land as an unreadable NaN or a **silent zero** rather than an error. Both amounts now go through a guard that throws naming the field and the endpoint, and a chain with no coingecko id mapped throws instead of calling `addCGToken(undefined, ...)`.

The silent-zero half is the part I would argue for on its own merits, independent of this outage: a 200-with-a-challenge-body is exactly the shape that writes a false zero into a series, which is the failure mode that is hard to notice afterwards.

## Verification

No behaviour change on valid data. Identical run before and after:

```
ethereum  | 286.00
bsc       | 302.00
avax      |  76.00
solana    | 303.00
tron      | 565.00
ton       |  13.00
Aggregate |  1.54 k
```

`ts-check` clean. The `test` job will stay red for as long as the runner is refused, and that red is the 403 above, not this diff.

## Not fixed by this

The data. That needs an allowlist request to NTM covering the runner's egress range, or routing this one adapter off a bare datacenter IP. Neither is a code change, so #8816 stays open for it.

### DIFF
--- a/fees/ntm/index.ts
+++ b/fees/ntm/index.ts
@@ -23,12 +23,15 @@ const fetchFeesAndRevenues = async (options: FetchOptions) => {
   const res = await httpGet(url);
 
   // Reject anything that is not a real number before it reaches addCGToken.
-  // Number(undefined) is NaN and Number(null)/Number("") are a finite 0, so a
-  // non-JSON body or a changed field name would otherwise land as a silent zero
-  // or an unreadable NaN instead of an error naming the endpoint.
+  // Number(undefined) is NaN and Number(null)/Number("")/Number(" ")/Number([])
+  // are all a finite 0, so a non-JSON body or a changed field name would
+  // otherwise land as a silent zero or an unreadable NaN instead of an error
+  // naming the endpoint. Only a number or a non-blank string is accepted.
   const readAmount = (raw: unknown, field: string): number => {
-    if (raw === null || raw === undefined || raw === "") {
-      throw new Error(`ntm: missing ${field} from ${url}`);
+    const usable =
+      typeof raw === "number" || (typeof raw === "string" && raw.trim() !== "");
+    if (!usable) {
+      throw new Error(`ntm: missing or non-numeric ${field} (${JSON.stringify(raw)}) from ${url}`);
     }
     const value = Number(raw);
     if (!Number.isFinite(value) || value < 0) {
--- a/fees/ntm/index.ts
+++ b/fees/ntm/index.ts
@@ -19,14 +19,31 @@ const fetchFeesAndRevenues = async (options: FetchOptions) => {
   const endTime = new Date(options.endTimestamp * 1000)
     .toISOString()
     .split(".")[0];
-  const res = await httpGet(
-    `${endpoint}start_date=${startTime}&end_date=${endTime}&chain=${options.chain}`,
-  );
+  const url = `${endpoint}start_date=${startTime}&end_date=${endTime}&chain=${options.chain}`;
+  const res = await httpGet(url);
+
+  // Reject anything that is not a real number before it reaches addCGToken.
+  // Number(undefined) is NaN and Number(null)/Number("") are a finite 0, so a
+  // non-JSON body or a changed field name would otherwise land as a silent zero
+  // or an unreadable NaN instead of an error naming the endpoint.
+  const readAmount = (raw: unknown, field: string): number => {
+    if (raw === null || raw === undefined || raw === "") {
+      throw new Error(`ntm: missing ${field} from ${url}`);
+    }
+    const value = Number(raw);
+    if (!Number.isFinite(value) || value < 0) {
+      throw new Error(`ntm: unusable ${field} (${JSON.stringify(raw)}) from ${url}`);
+    }
+    return value;
+  };
+
   const token = chainToken[options.chain];
+  if (!token) throw new Error(`ntm: no coingecko id mapped for chain ${options.chain}`);
+
   const dailyFees = options.createBalances();
   const dailyRevenue = options.createBalances();
-  dailyFees.addCGToken(token, res.fees_total);
-  dailyRevenue.addCGToken(token, res.revenue_total);
+  dailyFees.addCGToken(token, readAmount(res?.fees_total, "fees_total"));
+  dailyRevenue.addCGToken(token, readAmount(res?.revenue_total, "revenue_total"));
 
   return { dailyFees, dailyRevenue };
 };


BhariGowda:
CI just settled the open question in this PR, and it changes the conclusion.

The `test` job fails with:

```
🦙 Running NTM adapter 🦙
------ ERROR ------
Request failed with status code 403
```

**403, from the GitHub Actions runner.** From my machine the same URL returns 200 with every User-Agent I tried, including none at all. The only difference between the two callers is the network they sit on, and a GitHub runner is a datacenter IP in the same class as the production runner.

So the Cloudflare-egress theory in #8816 is no longer a guess. api.ntm.ai answers residential IPs and 403s datacenter IPs, which is exactly why this adapter has published nothing since 2026-07-29 while the endpoint looks perfectly healthy to anyone checking by hand.

Two consequences, being straight about both:

**The retry half of this PR will not fix it.** A deterministic 403 is not transient, so three attempts just means three 403s. I said in the description that the retry might not be the answer; this confirms it is not.

**The validation half is now the point, and CI is showing why it is needed.** Before this change the 403 surfaced as `undefined / Request failed with status code 403` with nothing naming the adapter's own expectations. That is still what you see above, because a 403 throws inside `fetchURL` before any response reaches my guard. The guard only covers the case where the host answers with an unusable body.

Given that, I am happy for this to be judged as what it is: a small hardening change that does not restore the data. The actual fix is either an allowlist request to NTM for the runner's egress range, or routing this one adapter through something that is not a bare datacenter IP. That is a call for someone with access to the runner's networking, not something I can do in a PR.

If you would rather not carry the retry change given it demonstrably does not help here, say so and I will strip it back to just the response validation.

BhariGowda:
Stripped the retry as offered above. Force-pushed, so the branch is now response validation only.

`httpGet` is back exactly as it was. The 403 makes the retry pointless, and there is no reason to carry a wider diff than the change needs.

What remains is the guard on `res.fees_total` / `res.revenue_total`, plus a throw when `chainToken` has no entry for the chain. Title and description updated to match.

The part I would still argue for on its own merits, separate from this outage: `Number(null)` and `Number("")` are both a finite `0`, so a 200 carrying a challenge page or an error body would write a **false zero** into the series rather than failing. That is the failure mode that is hard to spot after the fact, and it is one line of guarding to remove.

Verified unchanged on valid data: ethereum 286, bsc 302, avax 76, solana 303, tron 565, ton 13, aggregate 1.54k. `ts-check` clean. The `test` job stays red while the runner is refused, and that red is the 403, not this diff.

bheluga (commented):
@BhariGowda thanks for the PR
I checked the logs on-prod, this is failing due to 403.
So, i don't think these changes will have any effect

BhariGowda:
Thanks, that matches what I saw. The `test` job on the first revision of this branch failed the same way:

```
------ ERROR ------
Request failed with status code 403
```

so prod and CI agree, and that 403 is what made me drop the retry from this PR.

To be clear about scope: this does not bring the data back and I don't want to imply it does. api.ntm.ai answers a residential IP with any User-Agent, including none at all, and refuses from a runner, so the fix for the outage is on the network side, not in this file. #8816 stays open for it.

What is left in the diff is response validation. api.ntm.ai sits behind Cloudflare and the response was used unchecked:

```ts
dailyFees.addCGToken(token, res.fees_total);
```

`Number(null)`, `Number("")` and `Number(" ")` are all a finite 0, so a challenge or interstitial body returned with a 200 would write a silent zero into the series instead of erroring. That is the part I would argue for on its own merits, independent of this outage: a 403 announces itself, a false zero does not. Latest commit also rejects booleans and arrays, per the review comment.

Two things that would help if you have a minute:

- can someone ask NTM to allowlist the egress range? It is a rule on their Cloudflare, nothing on our side changes it.
- worth one test from prod whether a browser User-Agent plus `Referer: https://ntm.ai/` gets through. I cannot test that from here, since a residential request gets 200 with or without those headers. If it works from your egress it is a two line change and I will send it.

If you would rather not carry the validation at all, say so and I will close this.

BhariGowda:
@coderabbitai review

The type-aware check you asked for landed in 7f994fb: `readAmount` now accepts only a `number` or a non-blank `string`, so booleans, arrays, objects and whitespace-only strings are rejected before `Number()` runs. The finite and non-negative checks are unchanged.

bheluga:
@BhariGowda thanks for the PR

PR #8822, opened 2026-08-17
pulsechain: reject the explorer's stalled totalfees instead of publishing a zero-fee day

### LINKED ISSUE #8799: fees/pulsechain: explorer totalfees returns 0 then null since 2026-08-08, adapter publishes a false zero then crashes
`fees/pulsechain.ts` published a false zero on 2026-08-08 and has produced no datapoints at all since. The chain itself is fine; its blockscout indexer is what stopped.

### The adapter's source

```ts
const feesData = await fetchURL(`https://api.scan.pulsechain.com/api?module=stats&action=totalfees&date=${dateStr}`);
const dailyFees = options.createBalances();
dailyFees.addGasToken(feesData.result);
```

That endpoint degrades in two stages, and returns HTTP 200 with `status: "1"` throughout, so nothing signals a failure:

```
2026-08-07   result "219371981151759992142032798"    last real day
2026-08-08   result "0"
2026-08-10   result null
2026-08-14   result null
2026-07-01   result "210879002233319646814275276"    older dates still answer
```

On 08-08 the `"0"` was booked as a real day, so the published series carries a zero-fee day for the chain. From 08-09 or so `result` is `null`, `addGasToken(null)` throws inside the sdk, and the run dies before it reaches the base-fee loop:

```
Error: Invalid balance: null - 0x0000000000000000000000000000000000000000
    at isValidNumber (@defillama/sdk/build/util/common.js:146:19)
    at sumSingleBalance (@defillama/sdk/build/util/common.js:114:5)
```

Reproduced locally on 2026-08-14.

### It is the explorer, not the chain

The same explorer's stats service agrees that its own indexing stopped on the same date. `/api/v2/stats/charts/transactions`:

```
2026-08-08   tx_count 0
2026-08-07   tx_count 633,419
2026-08-06   tx_count 451,188
```

The chain is producing blocks and transactions normally right now, read from `rpc.pulsechain.com`:

```
block 27300281   15 txs   2026-08-16T06:53
block 27299781   40 txs   2026-08-16T05:26
block 27295281   32 txs   2026-08-15T16:30
```

roughly 10.4s blocks, about 8,300 blocks a day. Chain TVL over the same window went from $51.3M to $64.8M, so activity is up, not gone.

### Suggested handling

Two separate things, and the first is worth doing regardless of the second:

1. Guard the response. A `null` result should raise something that names the endpoint rather than surfacing as `Invalid balance: null` from inside the sdk, and a `"0"` result should not be published as a real zero-fee day for a chain that is demonstrably processing 600k transactions a day. The adapter already computes `totalBaseFees` on chain further down, which is a lower bound on total fees, so `dailyFees` coming back below `dailyRevenue` is a self-check available from data the adapter already has.

2. A replacement source, if the explorer does not recover. `helpers/blockscoutFees.ts` already covers about 40 chains and supports an `allStatsApi` stats service; pulsechain is not in it and has this bespoke adapter instead. I could not find a working stats path on `api.scan.pulsechain.com` (`/api/v1/lines/...` wants the legacy module/action params, `/api/v2/stats` returns a 500), so migrating it is not a drop-in and I have not tried to.

I am happy to send the guard in (1). Holding off on it only because "throw instead of publishing" leaves a permanent gap unless (2) lands too, and which of those you would rather have is your call.

BhariGowda:
Fixes #8799.

The blockscout indexer behind `api.scan.pulsechain.com` stalls while still answering 200 with `message: "OK"`. Still true today:

```
2026-08-07  {"message":"OK","result":"219371981151759992142032798","status":"1"}
2026-08-08  {"message":"OK","result":"0","status":"1"}
2026-08-10  {"message":"OK","result":null,"status":"1"}
2026-08-16  {"message":"OK","result":null,"status":"1"}
```

The chain kept producing 15-40 transactions a block throughout, and its TVL went from $51.3M to $64.8M over the same stretch, so none of this is a quiet chain.

Neither shape was checked, so each failed differently. The `"0"` was booked as a real zero-fee day. The nulls reached `addGasToken(null)` and died with `Invalid balance: null - 0x0000...`, which says nothing about where it came from.

## The zero case proves itself wrong for free

The adapter already sums `baseFeePerGas * gasUsed` over every block in the window for revenue, and base fees are burnt out of what users paid, so they are a subset of the total. **A zero total against a positive burn is arithmetically impossible.**

Master's own output for that day says so out loud: it published `Daily fees: 0.00` next to `Daily revenue: 726.00`. Revenue above fees, in the same run.

So the check needs no new data source. Both totals being zero is a genuinely idle window and is still allowed through.

## Changes

- reject null / undefined / empty / non-numeric / negative `result`, with the date and the endpoint in the message
- reject a zero total when the window's blocks burnt a positive amount, quoting the burn
- the explorer call and the block loop now run in parallel, since the check needs both. That is the only structural change; a healthy day passes through exactly as before

## Verification, all three cases

**2026-08-07, the last good day.** Unchanged, invariant silent:

```
Daily fees: 2.08 k
Daily revenue: 804.00
Daily holders revenue: 804.00
```

**2026-08-08, the false zero.** Master publishes `fees 0.00` with `revenue 726.00`. After:

```
Error: pulsechain: explorer reports 0 totalfees for 2026-08-08 while the blocks
in that window burnt 77602864779703528647002247 base fees
```

**Current window, the null.** Master dies with `Error: Invalid balance: null - 0x0000000000000000000000000000000000000000`. After:

```
Error: pulsechain: explorer returned no totalfees for 2026-08-16
(https://api.scan.pulsechain.com/api?module=stats&action=totalfees&date=2026-08-16)
```

`ts-check` clean.

## What this does not do

It does not bring the data back. The explorer is still returning null as of 2026-08-16 and that is on the pulsechain side. What it does is keep the zero out of the series and make the failure say what happened, which is the difference between a chain that looks like it earned nothing and one that is visibly not being indexed.

### DIFF
--- a/fees/pulsechain.ts
+++ b/fees/pulsechain.ts
@@ -6,13 +6,31 @@ import fetchURL from "../utils/fetchURL";
 
 const CONCURRENCY = 25;
 
-const fetch = async (options: FetchOptions) => {
-    const dateStr = options.dateString;
-    const feesData = await fetchURL(`https://api.scan.pulsechain.com/api?module=stats&action=totalfees&date=${dateStr}`);
+const TOTAL_FEES_URL = (dateStr: string) =>
+    `https://api.scan.pulsechain.com/api?module=stats&action=totalfees&date=${dateStr}`;
 
-    const dailyFees = options.createBalances();
-    dailyFees.addGasToken(feesData.result);
+// The blockscout indexer behind this endpoint can stall while still answering
+// 200 with message "OK". It returned "0" on 2026-08-08 and null every day from
+// 2026-08-09, on a chain that was producing 15-40 transactions a block the
+// whole time. Unchecked, the first shape published a real zero-fee day and the
+// second reached addGasToken(null) and died without saying why.
+const readTotalFees = (raw: unknown, dateStr: string): bigint => {
+    if (raw === null || raw === undefined || raw === "") {
+        throw new Error(`pulsechain: explorer returned no totalfees for ${dateStr} (${TOTAL_FEES_URL(dateStr)})`);
+    }
+    let value: bigint;
+    try {
+        value = BigInt(String(raw));
+    } catch {
+        throw new Error(`pulsechain: explorer returned a non-numeric totalfees ${JSON.stringify(raw)} for ${dateStr}`);
+    }
+    if (value < 0n) {
+        throw new Error(`pulsechain: explorer returned a negative totalfees ${value} for ${dateStr}`);
+    }
+    return value;
+};
 
+const sumBaseFees = async (options: FetchOptions): Promise<bigint> => {
     const fromBlock = await options.getFromBlock();
     const toBlock = await options.getToBlock();
     const provider = getProvider(CHAIN.PULSECHAIN);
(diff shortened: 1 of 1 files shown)


BhariGowda:
The red `test` job is this change working, not failing. Worth saying so before it reads as broken.

```
------ ERROR ------
Error: pulsechain: explorer returned no totalfees for 2026-08-16
(https://api.scan.pulsechain.com/api?module=stats&action=totalfees&date=2026-08-16)
    at readTotalFees (fees/pulsechain.ts:19:15)
    at fetch (fees/pulsechain.ts:63:23)
```

That is the new guard firing on exactly the condition it was added for. The explorer is still returning `{"message":"OK","result":null,"status":"1"}` for every date since 2026-08-09, so any honest run of this adapter has to fail today.

For comparison, master on the same window fails too, just uselessly: `Error: Invalid balance: null - 0x0000000000000000000000000000000000000000`, which names neither the adapter, the date, nor the source.

So the choice is not between green and red. It is between a red that says which upstream endpoint stalled and on what date, and a red that says a balance was null somewhere. Plus the case that was worse than either: on 2026-08-08 the explorer returned `"0"` and master published it as a real zero-fee day, sitting in the series next to a revenue figure of 726.00 from the same run.

`ts-check` is green. The adapter will go green again on its own the moment pulsechain's indexer catches up.

bheluga:
@BhariGowda thanks for the PR

PR #8833, opened 2026-08-17
fix(rocketpool): count megapool rewards dropped since saturn 1

### LINKED ISSUE #8812: fees/rocketpool: Saturn 1 megapool reward deposits are skipped — 78.23 ETH of rETH reward deposits hit the minipool-only continue branch
### Summary

`fees/rocketpool.ts` reads `EtherDeposited` on rETH, then for each unique `fromAddress` calls `getNodeDepositBalance()` and `getNodeFee()` with `permitFailure: true`, and skips the event when either returns null:

```ts
if (nodeBalance === null || nodeFeeRate === null) {
  continue
}
```

Since Saturn 1, node operators can run **megapools** instead of minipools. Megapools deposit their reward share into rETH through the same `EtherDeposited` event, but they do not implement those two minipool functions. On **all 31** megapool addresses in the total below, both calls revert, so those deposits take the `continue` branch.

Measured at 2026-08-17: **78.234366859157413918 ETH** across **53** deposits from **31** megapool addresses are currently being dropped this way. That is a lower bound (see below).

The current minipool-based adapter was introduced in #2765, merged 2025-04-03; `revenue = 0` was added later in #3519 on 2025-06-22. Both predate the Saturn 1 activation, which is when `RocketStorage` began serving those contracts at block 24,479,994 (2026-02-17).

### The two calls revert on megapools, and the selectors are fine

Positive control first, using Rocket Pool's own registry `rocketMinipoolManager` (`getMinipoolCount()` = 42,317), sampling index 0, midpoint and last:

| minipool | `getNodeDepositBalance()` | `getNodeFee()` |
|---|---|---|
| `0x56b678c9…3916` (index 0) | 8.0 ETH | 14.0% |
| `0x5077d219…8697` (index 21,158) | 8.0 ETH | 14.0% |
| `0x3b7fe930…22fb` (index 42,316) | 8.0 ETH | 5.0% |

3 of 3 answer. The same two calls against **every one of the 31 megapool addresses** that make up the total below: **31 of 31 revert on both**, no exceptions. `getMinipoolExists()` returns `0` for them — they are not minipools.

### The deposits are real and they land on rETH

Queried per address against rETH `EtherDeposited`, `topic0` `0xef51b4c870b8b0100eae2072e91db01222a303072af3728e58c9d4d2da33127f`, from block 0:

- 31 megapool addresses, **53** matching logs, **78.234366859157413918 ETH** total. No address reached the explorer's 1000-log page limit.
- Positive control: the same query for minipool `0x3b7fe930…22fb` returns 11 logs, so the query is not simply failing to match.

Cross-check from the other side: summing `rethAmount` over the 53 `RewardsDistributed` logs emitted by those megapools gives **78.234366859157413918 ETH** — the same figure, residual **0 wei**.

The two sets also pair one-to-one, not just in aggregate: matching on `(address, transactionHash, amount == rethAmount)` gives **53 of 53** matched, with no unmatched `RewardsDistributed` and no leftover `EtherDeposited`. So in this window each megapool → rETH deposit is a reward distribution rather than a capital repayment.

That matters for a fix: the adapter's `>= 24e18` / `>= 16e18` subtraction assumes a minipool repaying borrowed capital in the same transfer, which none of these 53 deposits is.

### This number is a lower bound

My seed set is the megapools that have emitted `RewardsDistributed` at least once — I did not enumerate the full megapool registry. Any megapool outside that set is not counted here, so the skipped total is at least this figure, not exactly it.

### Separately, the methodology strings no longer describe the reward path

Saturn 1 added a voter-share leg taken on the borrowed portion. Read at head 25,772,201 via `RocketStorage` `0x1d8f8f00cfa6758d7bE78336684788Fb0ee0Fa46`:

```
rocketDAOProtocolSettingsNetwork 0x67fd03a5095197d1ad1f932bc55e022c420b1153
  getVoterShare()        -> 90000000000000000   (9%)
  getNodeShare()         -> 50000000000000000   (5%)
  getProtocolDAOShare()  -> 0

rocketNetworkRevenues 0x9d9708da8e0200dd8dd9ad09e0aaf184ad260842
  getCurrentVoterShare() -> 90000000000000000   (9%)
```

Across the same 53 distributions to block 25,772,216: `voterAmount` **8.187317462004845634 ETH**, `rethAmount` 78.234366859157413918, `nodeAmount` 17.544251704296097817, `protocolDaoAmount` **0** — the last matching `getProtocolDAOShare()` returning 0. `Σ voterAmount / Σ rethAmount` = `0.104651162791`, exactly `9/86` to 12 digits, and the borrowed portion works out to 87.5000% of the distributed total, consistent with the 4/32 bond.

The voter share leaves to `rocketVault` under the name `rocketRewardsPool`: `Σ EtherDeposited − Σ EtherWithdrawn == rocketVault.balanceOf("rocketRewardsPool")` with residual 0 wei (`8.187317462004845634 − 4.437058101524732668 = 3.750259360480112966`), and `getPendingVoterShare()` independently returns the same `3750259360480112966`.

The `SupplySideRevenue` string — *"Total ETH staking rewards are distributed to rETH stakers and minipool depositors"* — still describes the pre-Saturn-1 split. Whether the voter share belongs under `Revenue` is a separate taxonomy question I am not trying to settle here.

The current API response also carries two further methodology strings that appear to describe a protocol fee, from `https://api.llama.fi/summary/fees/rocket-pool?dataType=dailyFees` read 2026-08-17:

```
UserFees          Percentage of rewards paid to protocol
ProtocolRevenue   Fee from users rewards
Revenue           Rocket Pool protocol doesn't take any fees or rewards cut.
```

Only `Fees`, `Revenue` and `SupplySideRevenue` appear in `fees/rocketpool.ts`. I did not trace where the other two originate, so I am not relying on them for anything above.

### One filtering trap, if someone implements the voter-share side

`RewardsDistributed(uint256,uint256,uint256,uint256,uint256)` is not unique to Rocket Pool. Scanning that topic0 without filtering the emitter picks up an unrelated contract emitting 5 logs with the same signature, whose second field sums to 426.928010500 ETH — unfiltered `435.115327962004845634` vs filtered `8.187317462004845634`, a 53.1x overstatement. The filter I used is the protocol's own flag rather than an address list: `rocketStorage.getBool(keccak256("megapool.exists" + address))`, which passes 31 of 32 emitters.

### What would show I am wrong

- either call succeeding on any of the **31 addresses** that make up the 78.234366859157413918 ETH figure — each such address would have to come out of that total;
- the adapter already accounting for megapool deposits through a path I did not read — I only traced `fetch()` in `fees/rocketpool.ts` at master;
- rETH `EtherDeposited` from those 31 addresses totalling something other than `78.234366859157413918 ETH` over the same range, or the 53-to-53 pairing not reproducing;
- `getVoterShare()` at block 25,772,201 not returning `90000000000000000`, or the vault identity at block 25,772,216 not closing at 0 wei.

### Scope

Ethereum mainnet only, read at head 25,772,201 and block 25,772,216 on 2026-08-17. The numerical checks are on-chain reads plus explorer log queries; the reward formula comes from Rocket Pool contract source; the code path comes from `fees/rocketpool.ts` at master; the methodology comparison comes from that file and your published API response. I have not run the adapter, audited the beacon-chain reward reconstruction, or looked at the MEV/smoothing-pool branch.

gomesalexandre:
closes #8812

## the bug

`fees/rocketpool.ts` reads rETH `EtherDeposited` and, for each sender, calls the minipool getters `getNodeDepositBalance()`/`getNodeFee()`; if either is null it `continue`s. Since Saturn 1, node operators can run **megapools**, which deposit their rETH reward share through the same event but do not implement those getters, so they revert and hit the `continue` - their rewards are silently dropped. #8812 measured 78.23 ETH dropped across 53 deposits / 31 megapools.

## the fix

After the minipool loop, pick megapools up from their own `RewardsDistributed` event:

1. Filter the `EtherDeposited` senders to confirmed megapools via `rocketStorage.getBool(keccak256("megapool.exists", addr))`.
2. Read `RewardsDistributed(nodeAmount, voterAmount, rethAmount, protocolDaoAmount, time)` scoped to those addresses.
3. Sum the four amount legs into `dailyFees`.

The four legs sum **exactly** to the full pre-split beacon reward: on-chain the node leg is defined as the residual (`nodeAmount = reward - the other three`), so no bond/commission gross-up is needed and it is robust to bond-ratio variation across megapools. This is the megapool analogue of the minipool `rewardFromBeacon`, so mixing both in `dailyFees` is apples-to-apples.

The minipool loop keeps dropping megapool `EtherDeposited` (getters revert), so each megapool distribution is counted once, here, at full value - no double count.

## why scoped, and why no principal leaks in

- `RewardsDistributed` is a plain five-uint256 signature that unrelated contracts also emit, so it is scoped to confirmed megapool emitters (the already-collected `EtherDeposited` senders), never an unscoped topic scan.
- A validator **exit** does not route through this event: `RewardsDistributed` is emitted at a single site (`_distributeAmount`), which is guarded `require(numExitingValidators == 0)` + `require(numLockedValidators == 0)`, and exit capital is dispersed by the separate `_notifyFinalBalance` path (which emits no `RewardsDistributed`). So the leg-sum is pure reward, never ~32 ETH of principal.

## scope

This counts the previously-dropped megapool rewards, consistent with the existing minipool treatment (full reward to `dailyFees` -> supply side, revenue stays 0). Saturn 1 added a voter share (~9%) and protocol-DAO share (0% today) that are arguably protocol revenue; reclassifying them is left to a follow-up per the reporter's request, and noted in a code comment.

## receipts (on-chain, 2026-08-17)

Validated the exact adapter logic against mainnet via an archive RPC:

- `megapool.exists` -> **true** for a real megapool (`0x3fe4...46f4`), **false** for a minipool and for rETH itself (detection, no false positives).
- A real `RewardsDistributed` log summed `nodeAmount + voterAmount + rethAmount + protocolDaoAmount = 1.5755 ETH`, with `voterAmount / rethAmount = 0.104651 = 9/86` exactly - matching the ground-truth ratio in #8812, confirming the leg-sum and that these are genuine Saturn-1 distributions.

The adapter can't be run end-to-end locally (the pre-existing MEV branch needs a Llama indexer key), so the new megapool path was validated in isolation as above. tsc clean.

### DIFF
--- a/fees/rocketpool.ts
+++ b/fees/rocketpool.ts
@@ -140,10 +140,12 @@ async function fetch(options: FetchOptions): Promise<FetchResultV2> {
   const megapoolExistsKeys = minipoolAddresses.map((address) =>
     ethers.solidityPackedKeccak256(['string', 'address'], ['megapool.exists', address])
   );
+  // getBool returns false for an unset key and never reverts, so unlike the
+  // minipool getters above it needs no permitFailure - letting a genuine RPC
+  // failure throw is preferable to silently dropping a megapool's rewards.
   const isMegapool = await options.api.multiCall({
     abi: 'function getBool(bytes32) view returns (bool)',
     calls: megapoolExistsKeys.map((key) => ({ target: rocketStorage, params: [key] })),
-    permitFailure: true,
   })
   // Scope RewardsDistributed to confirmed megapool emitters. Its topic0 is a
   // plain five-uint256 signature that unrelated contracts also emit, so an
@@ -156,6 +158,11 @@ async function fetch(options: FetchOptions): Promise<FetchResultV2> {
       eventAbi: 'event RewardsDistributed(uint256 nodeAmount, uint256 voterAmount, uint256 rethAmount, uint256 protocolDaoAmount, uint256 time)',
     })
     for (const log of rewardsDistributedLogs) {
+      // The full beacon reward is counted as fees and flows to supply side, the
+      // same as the minipool path above. Saturn 1 added a voter share (~9%) and
+      // a protocol-DAO share (currently 0%) that are arguably protocol revenue
+      // rather than supply-side; reclassifying them is left to a follow-up so
+      // this change stays scoped to the dropped-fees fix (see #8812).
       const beaconReward = Number(log.nodeAmount) + Number(log.voterAmount) + Number(log.rethAmount) + Number(log.protocolDaoAmount)
       dailyFees.addGasToken(beaconReward, METRIC.STAKING_REWARDS)
     }
--- a/fees/rocketpool.ts
+++ b/fees/rocketpool.ts
@@ -1,4 +1,5 @@
 import * as sdk from "@defillama/sdk";
+import { ethers } from "ethers";
 import ADDRESSES from '../helpers/coreAssets.json'
 import { Adapter, FetchOptions, FetchResultV2 } from "../adapters/types";
 import { CHAIN } from "../helpers/chains";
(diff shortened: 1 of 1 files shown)


gomesalexandre:
Heads up on the red `test` job: it fails in the pre-existing MEV / smoothing-pool block on `sdk.indexer.getTransactions` with `Llama Indexer URL/api key is not set` - the indexer key isn't available to fork-PR CI. That call is unchanged by this PR (it's on master today), and it runs *after* the new megapool code, which uses only RPC `multiCall`/`getLogs` and executes fine before the run hits the indexer.

Since I can't set that key from a fork, the new megapool path was validated directly on-chain instead (receipts in the description): `megapool.exists` true/false detection, and a real `RewardsDistributed` log whose four legs sum with `voter/reth = 9/86`. Happy to adjust anything if you'd rather the megapool rewards be sourced differently.

bheluga:

gomesalexandre:
Thanks for the merge!

PR #8839, opened 2026-08-17
feat: add Prism Network fees adapter

prismnetworksupport:
Prism Network rents GPUs by the second on Robinhood Chain (`CHAIN.ROBINHOOD`, already tracked by DefiLlama). A renter funds an escrow up front and the escrow finalises the lease on chain.

## Methodology

One event carries all three metrics, so nothing is double counted and no price lookup is needed beyond USDG, which DefiLlama already prices on this chain:

```
LeaseFinalized(leaseId, charged, fee, providerPaid, refunded, receiptHash)
```

`charged` is what the renter paid for the time actually consumed, and it is exactly `fee + providerPaid`. Deposits are taken up front, but the unused remainder is refunded in the same transaction and is not counted as a fee.

| Metric | Source |
| --- | --- |
| `dailyFees` | `charged` |
| `dailyRevenue`, `dailyProtocolRevenue` | `fee`, currently 10% |
| `dailySupplySideRevenue` | `providerPaid`, to the GPU operator |

## Testing

```
ROBINHOOD_RPC="https://rpc.mainnet.chain.robinhood.com" pnpm test fees prism-network
ROBINHOOD_RPC="https://rpc.mainnet.chain.robinhood.com" pnpm test fees prism-network 2026-08-15
```

Both pass. The backfill run was checked against an independent tally of the same logs: for 2026-08-14 both report 3.54 charged, 0.35 protocol, 3.19 supply side across 17 leases.

Escrow: `0x62C042265991bEa17B07229322A01850974626dA`
Start date: 2026-08-13, the first finalised lease.

### DIFF
--- a/fees/prism-network/index.ts
+++ b/fees/prism-network/index.ts
@@ -1,0 +1,52 @@
+import { Adapter, FetchOptions } from "../../adapters/types";
+import { CHAIN } from "../../helpers/chains";
+
+// Prism Network rents GPUs by the second. A renter funds an escrow up front and
+// the escrow finalises the lease onchain, splitting what was actually consumed
+// between the protocol and the operator whose machine served it. Unused deposit
+// is refunded in the same transaction and is not a fee.
+const ESCROW = "0x62C042265991bEa17B07229322A01850974626dA";
+const USDG = "0x5fc5360D0400a0Fd4f2af552ADD042D716F1d168";
+
+// `charged` is what the renter paid for the time consumed, and it is exactly
+// `fee + providerPaid`, so the three metrics come from one event with no
+// double counting and no price lookup of our own.
+const LEASE_FINALIZED =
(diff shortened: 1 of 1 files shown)


bheluga:
@prismnetworksupport thanks for the PR

prismnetworksupport:
Thanks for merging this.

One thing on the protocol entry it created: it is named `prism` with slug `prism`, which resolves to `icons.llamao.fi/icons/protocols/prism`. That icon already belongs to PRISM Protocol (slug `prism-protocol`), so our page currently shows their logo.

Could the entry be renamed to **Prism Network** with slug `prism-network`? That matches the adapter module name and removes the collision. Happy to open a PR wherever the protocol metadata lives if you point me at it, or to supply an icon once the slug is distinct.

For reference: https://prismnetwork.tech, and the entry is id 8446, currently `module: dummy.js` with empty `chains` (it reports on Robinhood Chain).

realdealshaman:
Hi @prismnetworksupport the icon is now fixed, please clear your browser cache to see the new one

PR #8846, opened 2026-08-18
machinex-cl: read volume from peaq logs, their day series stopped on 08-13

### LINKED ISSUE #8823: machinex-cl and machinex-legacy dark since 2026-08-13: their analytics api reports lastSynced=now but dayData stops at 08-13
Both `dexs/machinex-cl.ts` and `dexs/machinex-legacy.ts` have published nothing since 2026-08-13, in both dimensions. `total24h` is null on all four series and the charts stop with no trailing zeros, so the adapters are throwing rather than writing zeros.

```
              machinex-cl          machinex-legacy
              volume     fees      volume    fees
2026-08-08     8,755       88       1,108    0.57
2026-08-09     8,955       90         459    0.23
2026-08-10     7,532       75         635    0.37
2026-08-11    22,788      164       1,166    0.73
2026-08-12    11,971      120         645    0.34
2026-08-13    21,713      217         680    0.37   <- last point
```

30d totals at the stop: machinex-cl $520,448 volume / $5,103 fees, machinex-legacy $41,753 / $25.

## Cause

Both adapters read the same endpoint and index it by day:

```ts
const data = await fetchURL('https://machinex-api-production.up.railway.app/analytics')
const record = data.dayData.find((day: any) => day.timestamp === startOfDay)

return {
  dailyFees: record.cl.feesUSD,
  dailyVolume: record.cl.volumeUSD,
}
```

The endpoint is up — 200 on `/analytics`, `OK` on `/health` — and its own freshness field is live. Two reads twenty minutes apart:

```
"lastSynced": "2026-08-17T09:20:40.436690"
"lastSynced": "2026-08-17T09:40:40.773372"
```

But the last entry in `dayData` is `timestamp: 1786579200`, which is 2026-08-13. There is no 08-14, 08-15 or 08-16 row. `find()` returns undefined, `record.cl.feesUSD` throws `TypeError: Cannot read properties of undefined`, and the chart stops.

The mapping itself is fine. `dayData` for 08-13 carries cl `volumeUSD` 21713.49653821108 and legacy `volumeUSD` 680.25568297798, which are exactly the last two points published.

## Controls

- **peaq is not halted.** Five independent providers — quicknode1/2/3.peaq.xyz, onfinality, publicnode — agree on head 11240346 and advance ~15 blocks over 90s.
- **the protocol is still trading.** GeckoTerminal's `machinex` dex on peaq lists 12 pools with $7,447 of volume over the last 24h: ACU/WPEAQ $4,234, wstPEAQ/WPEAQ $1,084, USDT/WPEAQ $633, AUKI/WPEAQ $510, USDC/WPEAQ $380.
- **the TVL adapter is unaffected** and still reports daily, $550,900 on peaq today.

Chain live, protocol live, and only the daily aggregation behind `/analytics` is stuck.

## Fix

Nothing in this repo can read rows that do not exist, so this needs one of:

1. someone pinging the MachineX team to restart the day-data job. `lastSynced` advancing on schedule while `dayData` sits four days behind reads like one aggregation task died rather than the service going down.
2. moving both adapters on-chain. MachineX CL is a uni-v3 style CL dex on peaq and all 12 pools are indexable, so a logs-based rewrite is straightforward — happy to send that if you would rather not depend on their api.

Either way `record` is worth guarding, so that a missing day throws something readable instead of a TypeError on undefined.

0xshubhs:
Fixes #8823.

`dexs/machinex-cl` has published nothing since 2026-08-13. It reads `dayData` off MachineX's api and indexes it by `startOfDay`, and that series stops on 08-13, so `find()` returns undefined and `record.cl.feesUSD` throws. On master today:

```
TypeError: Cannot read properties of undefined (reading 'cl')
    at fetch (dexs/machinex-cl.ts:10:23)
```

Every volume field on that host is stuck, not just the one the adapter reads, `poolDayData` ends on the same day and `stats.last_24h_vol` is 0 on all 19 pairs, while the current-state fields are live, which is why the TVL adapter kept working. So there is no other endpoint there to repoint at.

## the change

Read peaq directly. MachineX CL is a uni-v3 fork (`fee()`, `tickSpacing()`, `token0()`, `token1()` all present, one factory at `0x2646dcbe025d21a2925fdaceb639e998e17d6060`), so `getUniV3LogAdapter` does the work. It is given an explicit `pools` list rather than a `factory`, because the factory path reads pool-creation logs from the R2 cache and there is no TVL adapter populating one for this protocol; the `pools` path reads `fee`/`token0`/`token1` on chain instead.

The pool list comes from `/data` on the same api, which is the live half of it and is already what `DefiLlama-Adapters/projects/machinex-cl` uses for TVL. New pools are picked up without a code change.

## the numbers change, upward

Rebuilding 2026-08-13 from chain gives **$54,210** against the **$21,713** the api reported and llama published, so the upstream has been understating by about 2.5x. GeckoTerminal independently gives **$46,198** for that day across 5 of its 7 indexed pools; the adapter covers 16, so the remainder closes the gap.

| day | this adapter | machinex api |
|---|---|---|
| 2026-08-13 | $54,210 | $21,713 |
| 2026-08-14 | $18,220 | nothing |
| 2026-08-15 | $17,720 | nothing |
| 2026-08-16 | $8,370 | nothing |
| 2026-08-17 | $72,240 | nothing |

## fees are dropped, deliberately

The old adapter also returned `dailyFees`. I left fees out rather than guess the split: the pools carry real accrued `protocolFees()` on both tokens, so "all to LPs" is wrong, and `slot0().feeProtocol` reads 160, 0 on token0 and 10 on token1 under the standard packing, which is fork-specific enough that I would only be guessing at the ratio. Volume alone is a strict improvement on an adapter that currently publishes nothing, and I will add fees the moment someone can state the split.

## verification

`npx ts-node --transpile-only cli/testAdapter.ts dexs machinex-cl 2026-08-14` twice, $54.21k both times.

Worth flagging for anyone testing this locally: free peaq RPCs return an empty log set instead of an error when they rate limit, and that empty result gets cached, so back-to-back runs report `0.00` from the same command. Space the runs out and clear `@defillama/sdk/build/util/local_cache/zlib-1.0/event-logs/peaq` between them.

`machinex-legacy` is broken by the same frozen series and is untouched here; it is a solidly-style v2 factory at `0xa3f356f0403b4f10345cd95e0c80483fddd63ebd` with 3 pairs, and I would rather send it separately than double the review surface.

### DIFF
--- a/dexs/machinex-cl.ts
+++ b/dexs/machinex-cl.ts
@@ -9,19 +9,24 @@ async function fetch(options: FetchOptions) {
   const { pairs } = await getConfig('machinex-cl-peaq', poolsEndpoint)
   const pools = pairs.filter((pair: any) => !pair.hasOwnProperty('stable')).map((pair: any) => pair.id)
 
-  const { dailyVolume } = await getUniV3LogAdapter({ pools })(options)
+  const { dailyVolume, dailyFees } = await getUniV3LogAdapter({ pools })(options)
 
-  return { dailyVolume }
+  return { dailyVolume, dailyFees }
 }
 
 const adapter: SimpleAdapter = {
   version: 2,
-  adapter: {
-    [CHAIN.PEAQ]: { fetch },
-  },
+  pullHourly: true,
+  fetch,
+  chains: [CHAIN.PEAQ],
   methodology: {
     Volume: "Swap events on the MachineX concentrated liquidity pools, read on chain.",
+    Fees: "Fees from swap events on the MachineX concentrated liquidity pools, read on chain.",
   },
+  breakdownMethodology: {
+    "Token Swaps": "Fees from swap events on the MachineX concentrated liquidity pools, read on chain.",
+  },
+  skipBreakdownValidation: true,
 }
 
 export default adapter
--- a/dexs/machinex-cl.ts
+++ b/dexs/machinex-cl.ts
@@ -1,19 +1,27 @@
-import { FetchOptions } from "../adapters/types";
+import { FetchOptions, SimpleAdapter } from "../adapters/types";
 import { CHAIN } from "../helpers/chains";
-import fetchURL from "../utils/fetchURL";
+import { getConfig } from "../helpers/cache";
+import { getUniV3LogAdapter } from "../helpers/uniswap";
 
-async function fetch({ startOfDay }: FetchOptions) {
-  const data = await fetchURL('https://machinex-api-production.up.railway.app/analytics')
-  const record = data.dayData.find((day: any) => day.timestamp === startOfDay)
+const poolsEndpoint = 'https://machinex-api-production.up.railway.app/data'
 
-  return {
-    dailyFees: record.cl.feesUSD,
-    dailyVolume: record.cl.volumeUSD,
-  }
+async function fetch(options: FetchOptions) {
+  const { pairs } = await getConfig('machinex-cl-peaq', poolsEndpoint)
+  const pools = pairs.filter((pair: any) => !pair.hasOwnProperty('stable')).map((pair: any) => pair.id)
+
+  const { dailyVolume } = await getUniV3LogAdapter({ pools })(options)
+
+  return { dailyVolume }
 }
 
-export default {
-  version: 1,
-  fetch,
-  chains: [CHAIN.PEAQ],
+const adapter: SimpleAdapter = {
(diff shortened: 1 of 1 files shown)


bheluga:
@0xshubhs thanks for the PR

PR #8851, opened 2026-08-18
PRIZM: fees adapter (Solana trading app)

MBrassey:
## New listing: PRIZM

**Name:** PRIZM
**URL:** https://www.prizm.trading
**Category:** Trading App
**Chain:** Solana
**Logo:** https://www.prizm.trading/icon.png (512x512)
**Token:** none
**Twitter:** none yet

PRIZM is a non-custodial Solana trading terminal: spot, limit and DCA orders routed through Jupiter, plus liquid staking, lending on Kamino and marginfi, and Meteora DLMM liquidity. Live since 2026-08-08.

## Methodology

Two fee flows, both measured as value received at PRIZM's on-chain fee accounts:

1. **Swaps (0.85%)** — collected through Jupiter's referral mechanism in the swap's output mint. Targets are the two referral token accounts (USDC and wSOL).
2. **Staking / lending / liquidity (0.33%)** — an SPL transfer carved off atomically with the action, landing in the fee wallet's associated token accounts. Targets are the wallet and its token accounts.

All targets are also passed as `blacklists` so referral claims and sweeps between PRIZM's own accounts are never counted twice. There are no referral or cashback payouts, so `dailyRevenue = dailyFees`.

Follows the same `getSolanaReceived` pattern as other Solana trading apps in `fees/` (e.g. `stbot.ts`).

## Testing

`npm test fees prizm` compiles and runs; the query itself needs an Allium key locally ("Ignore this error for github bot"), so numbers come from CI. The fee accounts are active on-chain — recent transactions visible on each target address.

### DIFF
--- a/fees/prizm.ts
+++ b/fees/prizm.ts
@@ -1,0 +1,54 @@
+import { Dependencies, FetchOptions, SimpleAdapter } from "../adapters/types";
+import { CHAIN } from "../helpers/chains";
+import { getSolanaReceived } from "../helpers/token";
+
+// Jupiter referral token accounts that collect PRIZM's platform fee on routed
+// swaps. The fee is taken in the swap's output mint, so there is one account
+// per fee mint.
+const REFERRAL_FEE_ACCOUNTS = [
+  '9knAMZ43uqWTdvA5uovwmiG2J35eBUv587E3LBqTLuJe', // USDC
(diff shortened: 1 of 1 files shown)


bheluga:
@MBrassey thanks for the PR

PR #8861, opened 2026-08-19
ntm: route the fee api through the proxy, it 403s datacenter ips

### LINKED ISSUE #8816: fees/ntm: nothing published since 2026-07-29 while api.ntm.ai still serves every day and the adapter runs clean locally

BhariGowda:
Fixes #8816.

`fees/ntm` has published nothing since 2026-07-29. `total24h` and `total7d` are both null and the chart stops rather than trailing zeros, which is the adapter erroring rather than writing a real zero. Density before the stop was 60 of the last 60 days nonzero at roughly $1,330/day, so this is not a protocol that pays in bursts. `total30d` is $13.5k and `total1y` is $678k.

## The source is healthy, the caller's IP is the problem

`api.ntm.ai/feesAndRevenues.php` answers 200 for every chain in `chainToken` and for the individual days inside the gap, so this is not a backfill hole.

The diagnosis came out of CI on #8820, not from guessing. That PR's `test` job ran the adapter on a GitHub Actions runner:

```
🦙 Running NTM adapter 🦙
------ ERROR ------
Request failed with status code 403
```

The identical request from a residential IP returns 200 with any User-Agent, including no User-Agent header at all. `axios/1.7.2`, `axios/0.27.2`, `node` and `Mozilla/5.0` all return 200. So the block is on the caller's IP, not on the request shape, and a datacenter runner is exactly what production is.

## Why the proxy

`proxiedFetch` is already the repo's treatment for this failure mode:

- `aggregators/houdiniswap` switched to it in #8745 ("cloudflare blocks node"). It had a gap over 08-08 to 08-11, the change merged 08-14, and it has published every day since, currently $1.94M on 08-18.
- `fees/space-and-time` uses it and is current through 08-18.

`proxiedFetch` falls back to a direct request when `PROXY_AUTH` is unset, so local runs and CI behave exactly as they do today. The `readAmount` guard added in #8820 still applies to whatever comes back, so a bad body still throws rather than publishing a zero.

## Test

Run locally, which takes the fallback path:

```
chain     | Daily fees | Daily revenue
ethereum  | 287.00     | 287.00
bsc       | 301.00     | 301.00
avax      | 76.00      | 76.00
solana    | 231.00     | 231.00
tron      | 566.00     | 566.00
ton       | 13.00      | 13.00
Aggregate | 1.47 k     | 1.47 k
```

In line with the pre-stop level of roughly $1,330/day.

## What this does not prove

I cannot verify from here that the proxy's egress IP is not also blocked by ntm, since I do not have `PROXY_AUTH`. What is established is that the block is IP-based rather than header-based, and that this same proxy clears the same class of block for houdiniswap. If ntm blocks the proxy too, the adapter keeps erroring and keeps publishing nothing, which is the current behaviour, so this cannot make things worse.

### DIFF
--- a/fees/ntm/index.ts
+++ b/fees/ntm/index.ts
@@ -1,7 +1,12 @@
 import { CHAIN } from "../../helpers/chains";
 import { FetchOptions } from "../../adapters/types";
-import { httpGet } from "../../utils/fetchURL";
+import { proxiedFetch } from "../../utils/fetchURL";
 
+// api.ntm.ai answers normally from a residential IP and returns 403 to a
+// datacenter IP, which is why nothing has been published since 2026-07-29.
+// Routed through the shared proxy, the same treatment aggregators/houdiniswap
+// and fees/space-and-time already use. proxiedFetch falls back to a direct
+// request when PROXY_AUTH is unset, so local runs and CI are unchanged.
 const endpoint = "https://api.ntm.ai/feesAndRevenues.php?";
 const chainToken: Record<string, string> = {
   [CHAIN.TON]: "the-open-network",
@@ -20,7 +25,7 @@ const fetchFeesAndRevenues = async (options: FetchOptions) => {
     .toISOString()
     .split(".")[0];
   const url = `${endpoint}start_date=${startTime}&end_date=${endTime}&chain=${options.chain}`;
-  const res = await httpGet(url);
+  const res = await proxiedFetch(url);
 
   // Reject anything that is not a real number before it reaches addCGToken.
   // Number(undefined) is NaN and Number(null)/Number("")/Number(" ")/Number([])


BhariGowda:
The red `test` job here is the diagnosis reproducing, not a broken change:

```
------ ERROR ------
Request failed with status code 403
```

That is the same 403 the runner produced on #8820, and it is the whole reason this PR exists. `PROXY_AUTH` is not set on a fork PR runner, so `proxiedFetch` takes its documented fallback and issues the request directly, from the datacenter IP that ntm blocks. Identical outcome to master, since master issues that same direct request unconditionally.

So CI cannot demonstrate the fix from here. What it does demonstrate, again, is that the block is real and IP-based: the identical URL returns 200 from a residential IP with any User-Agent, including none.

The change is a no-op wherever `PROXY_AUTH` is unset and only differs where it is set, which is production. If ntm blocks the proxy egress as well, the adapter keeps erroring and keeps publishing nothing, which is exactly what it does today.

Happy to close this if maintainers would rather not point another adapter at the proxy.

BhariGowda:
The observation is accurate. `proxiedFetch` does set `rejectUnauthorized: false` on its agent:

```ts
const client = axios.create({
  httpsAgent: new https.Agent({ rejectUnauthorized: false }),
});
```

I am not changing it in this PR, and I want to be explicit about why rather than just declining.

It is pre-existing shared infrastructure, not something this PR introduces. The diff here is a one-line swap of `httpGet` for `proxiedFetch` in a single adapter. `aggregators/houdiniswap` and `fees/space-and-time` already route through the same helper, and the TVL repo carries its own copy in `projects/helper/http.js` (which was itself touched as recently as 9d3683a for a port fix), so the blast radius of editing it is every proxied adapter across two repos, not this one.

The fix you are describing needs an explicit trusted proxy CA, which means knowing what CA the proxy actually presents. That lives with `PROXY_AUTH`, which I do not have. I cannot test the change, and a wrong CA bundle fails closed on three adapters at once with no local signal that it happened — `proxiedFetch` swallows proxy errors and falls back to a direct request, so the symptom would be silent reversion to the blocked path rather than an obvious failure.

So this belongs with whoever owns the proxy credentials, as its own change, not bundled into a one-line adapter repoint. Flagging it as a legitimate separate issue rather than pretending it is out of scope entirely.

bheluga:
hmm not sure if this really works,
lets see

PR #8862, opened 2026-08-19
yield-basis: mark the fees adapter doublecounted, its fees come from Curve pools

### LINKED ISSUE #8622: dexs/curve and fees/yield-basis both count the Yield Basis pools' trading fees, $11k-50k/day
`dexs/curve` books the Yield Basis pools' trading fees inside `total_fees`, and `fees/yield-basis` reports the same pools separately. On the four days I sampled that overlap is $11.3k to $49.8k a day, 34% to 81% of everything Curve reports.

The overlap is already known in code. #8573 added the field and labelled it, but nothing nets it out:

```ts
// dexs/curve/api.ts:16-18
// Yield Basis fees, already inside total_fees/fees_to_lp (those pools have admin_fee = 0).
// fees/yield-basis reports the same dollars.
lp_fees_yb: number;
```

`dexs/curve/index.ts:565` then books the unadjusted total:

```ts
dailyFees.addUSDValue(chainData.total_fees);
```

Ethereum resolves through the API path, not the on-chain fallback, so this is the live path for the chain where all of the Yield Basis volume sits.

## What each side counts

`fees/yield-basis` walks the YB factory `0x370a449FeBb9411c95bf897021377fe0B7D100c0`, reads each market's `lt`, and builds `dailyFees` from `pricePerShare` growth plus the `FundEpoch` admin-fee distributions (`fees/yield-basis/index.ts:134` and `:146`). Its own methodology calls this "Net trading fees distributed to users (pricePerShare growth + admin fees to veYB)".

Those markets are Curve cryptopools. Their trading fees are what `lp_fees_yb` measures on the Curve side. Same pools, same underlying fee stream.

## The numbers

From `prices.curve.finance/v1/chains/fees`, aggregated, and `api.llama.fi/summary/fees/<slug>?dataType=dailyFees`:

| day | curve `total_fees` | `lp_fees_yb` | share | published Curve DEX | published Yield Basis |
|---|---:|---:|---:|---:|---:|
| 2026-08-02 | 61,626 | 49,840 | 80.9% | 63,958 | 35,426 |
| 2026-08-03 | 53,611 | 20,481 | 38.2% | 57,281 | 3,377 |
| 2026-08-04 | 33,235 | 11,253 | 33.9% | 37,690 | -27,921 |
| 2026-08-05 | 50,559 | 24,584 | 48.6% | 56,606 | 11,505 |

`lp_fees_yb` is entirely Ethereum on every day sampled, no other chain carries any.

On the aggregate side, `Curve DEX` sits under `parent#curve-finance` while `Yield Basis` has no `parentProtocol`, so the two are summed as independent protocols.

## Where I am not confident

The comment says the two report "the same dollars", and I do not think that is exactly right. The magnitudes do not line up, and Yield Basis published **-27,921** on 08-04 against an `lp_fees_yb` of 11,253 the same day.

That is consistent with the two sides measuring the same fee stream at different points. Curve counts the gross fee charged on the swap; Yield Basis counts what is left after Cryptoswap rebalancing and the cost of maintaining 2x leverage, which is why it can go negative. So this is a genuine overlap in source, not a clean line item that can be subtracted one for one.

I have not tried to establish the exact overlapping amount, because which number that even is depends on the answer below.

## The question

Which adapter should net the shared dollars out?

Three directions I can see, and I do not think the chain data picks between them:

1. Subtract `lp_fees_yb` from `total_fees` in `dexs/curve`, treating Curve as the venue and Yield Basis as the issuer of that fee stream.
2. Leave `dexs/curve` alone and reparent or flag `fees/yield-basis`, treating Curve's pool-level accounting as the source of truth.
3. Leave both, on the view that gross pool fees and net levered LP yield are different metrics that are not actually the same dollars.

Option 3 is defensible given the negative day, which is why I would rather not pick. Happy to open the PR once there is a direction.

cc @DaniiRix @bheluga

BhariGowda:
Fixes #8622.

`fees/yield-basis` books the downstream share of trading fees earned in the Yield Basis pools, which are Curve pools. `dexs/curve` already reports those trading fees. Counting both in the aggregates counts the same flow twice.

This sets `doublecounted: true` on `fees/yield-basis`. `dexs/yield-basis` already carries the same flag for the volume side of the same overlap, so this makes the two halves consistent.

The adapter's own output is unchanged; the flag only affects aggregation.

Curve is left alone deliberately. Those trading fees really were paid in Curve pools, and Curve should report them gross.

### DIFF
--- a/fees/yield-basis/index.ts
+++ b/fees/yield-basis/index.ts
@@ -206,6 +206,12 @@ const adapter: SimpleAdapter = {
     methodology,
     breakdownMethodology,
     allowNegativeValue: true,
+    // The yield and admin fees booked here are the downstream share of trading
+    // fees earned in the Yield Basis pools, which are Curve pools. dexs/curve
+    // already reports those trading fees, and reports them gross, so counting
+    // this adapter in the aggregates as well would count the same flow twice.
+    // dexs/yield-basis carries the same flag for the volume side of this overlap.
+    doublecounted: true,
 }
 
 export default adapter;


BhariGowda:
The red `test` job is pre-existing and unrelated to this change:

```
Error: curve: every factory returned no pools on robinhood - treating this as zero volume would be wrong
```

That is the on-chain path failing on robinhood. This PR only touches `fetchFromApi`, which robinhood never reaches, and it changes two arithmetic expressions inside it.

I confirmed it reproduces on unmodified master before opening: stashing the change and running `npm run test dexs curve` locally gives the identical error. Robinhood's public RPC keeps only a few thousand blocks of state, so a factory scan there returns nothing from any environment that is not reading the tip.

The path this PR does touch was tested by narrowing the chain list to ethereum and running the same window twice:

```
before   fees 5.61k   supply side 4.30k   revenue 1.31k   volume 20.28M
after    fees 2.59k   supply side 1.28k   revenue 1.31k   volume 20.28M
```

`revenue + supply side == fees` both times, and volume, revenue, protocol revenue and holders revenue are unchanged.

BhariGowda:
Taken, pushed. You are right that I was trusting the invariant rather than enforcing it.

`ybFees` is now clamped to `[0, fees_to_lp]`. One bound covers both legs: `total_fees` is the sum of the lp, dao and treasury legs and none of them go negative, so `fees_to_lp <= total_fees` always, and bounding by `fees_to_lp` keeps the adjusted fee leg non-negative too.

The identity is unaffected for any bound value, since `revenue + supplySide = dao + treasury + fees_to_lp - b` and `fees = total_fees - b`, so the same `b` cancels.

On the cases you asked about. There is no unit test harness for adapters in this repo (`npm run test` runs the adapter live against the real API), so I walked them directly instead of adding tests that would have nowhere to live:

| `lp_fees_yb` | bounded | fees | supplySide | result |
|---|---|---|---|---|
| 27,917 (== `fees_to_lp`) | 27,917 | 8,448 | 0 | non-negative, identity holds |
| 99,999 (> `fees_to_lp`) | 27,917 | 8,448 | 0 | non-negative, identity holds |
| -5 | 0 | 36,365 | 27,917 | non-negative, identity holds |
| absent / null | 0 | 36,365 | 27,917 | non-negative, identity holds |

And against the API's exact values for the last 8 days, the bound is a **no-op on every one** — it never changed a figure, no leg went negative, and the identity error stayed at float epsilon (worst 1.6e-11). So this is purely a guard on a state the API says cannot happen, not a behaviour change.

Re-ran the ethereum path after the change: volume 21.65M, fees 2.90k, revenue 1.46k, supply side 1.44k, and revenue + supply side = fees.

BhariGowda:
Two notes on the state of this PR.

The review comment asking for the deduction to be bounded was already addressed. It was posted at 05:21 against the first commit; 7e0257d ("bound the Yield Basis deduction instead of trusting it raw") went up at 05:53 and does exactly that:

```ts
const ybFees = Math.min(Math.max(chainData.lp_fees_yb ?? 0, 0), chainData.fees_to_lp);
```

`fees_to_lp <= total_fees` always, since `total_fees` is the sum of the lp, dao and treasury legs and none of them go negative, so the single bound keeps both adjusted legs non-negative.

The red `test` check is not this change. It fails in pool discovery on robinhood, before anything in `fetchFromApi` runs:

```
Error: curve: every factory returned no pools on robinhood - treating this as zero volume would be wrong
    at getAllPools (helpers/curve/helpers.ts:169:11)
```

I ran `npm run test dexs curve` on this branch and on unmodified master in a separate checkout. Identical failure, identical line, both exit 1. It is a pre-existing condition of that guard on robinhood, not something this PR introduces. The diff here touches only `dexs/curve/index.ts` inside `fetchFromApi`, which is the path ethereum resolves through.

bheluga:
@BhariGowda thanks for the PR
Yield basis pools are curve pools too right?
Why don't we just add double counted flag in yield basis adapter instead?

BhariGowda:
You are right, and it is worse than a preference between two equivalent fixes. I checked before switching and my original premise does not survive the check.

I had assumed the two adapters report the same dollars. They do not. For 2026-08-19, ethereum:

```
curve API   lp_fees_yb        +631,901     gross trading fees in the YB pools
fees/yield-basis                -63,806    = -100,233 asset yield + 36,427 protocol fees
```

Different measurements, and on that day opposite signs. `fees/yield-basis` reads `pricePerShare` deltas on the LT contracts, so what it books is the yield actually delivered to ybBTC holders, net of the rebalancing and volatility decay costs of the 2x leverage. That is why it carries `allowNegativeValue`. `lp_fees_yb` is the gross fee traders paid. The first is a downstream slice of the second, not the same number.

So subtracting `lp_fees_yb` from Curve would have removed real gross trading fees from the venue that earned them, while the YB adapter reported a negative number for the same day. That is not a fix. Thanks for pushing back on it.

The flag is also the internally consistent choice. `dexs/yield-basis` already sets `doublecounted: true` for the volume side of this exact overlap, with the comment "all volume and fees are on Curve DEX". Handling the fee side of the same overlap by arithmetic in the Curve adapter instead would have left the two halves on different mechanisms.

Repushed as a single change: `doublecounted: true` on `fees/yield-basis`, Curve untouched. Title and description updated to match, since the old ones asserted the claim above.

Worth noting how large this has got since I opened the PR. The Yield Basis share of Curve's ethereum LP fees:

| day | fees_to_lp | lp_fees_yb | share |
|---|---|---|---|
| 2026-08-17 | 33,913 | 25,977 | 76.6% |
| 2026-08-18 | 27,917 | 20,403 | 73.1% |
| 2026-08-19 | 652,240 | 631,901 | 96.9% |
| 2026-08-20 | 231,764 | 216,015 | 93.2% |
| 2026-08-21 | 492,030 | 465,787 | 94.7% |

Separately, and not something I am touching here: `fees/yield-basis` returned -100,233 of Assets Yields on 08-19 against 36,427 of protocol fees. The negative is by design given the leverage decay, but the size of it next to the gross fees on the same day may be worth a look on its own.

PR #8836, opened 2026-08-17
feat(curators): count RockawayX's Kamino kvaults on Solana

gomesalexandre:
closes #8708

## what

RockawayX curates two Kamino kvaults on Solana (RWA USDC + RockawayX SOL) that the curator framework had no path to, so ~$3.8k/day of gross yield and its 5% curator revenue went uncounted. This adds Kamino (Solana) support to the framework and attributes those two vaults to RockawayX.

## how

`factory/curators.ts` auto-generates a fees adapter per curator via `getCuratorExport`, which was EVM-only (Morpho/Euler). This change:

1. Adds an optional `kaminoVaults?: string[]` field to the per-chain `CuratorConfig`.
2. Adds a `getKaminoVaultFee` helper that reads `api.kamino.finance/kvaults/vaults/<addr>` + `/metrics/history`, takes the cumulative `interest` delta over the window as gross yield, books `performanceFeeBps` as curator revenue and the rest as supply side. This mirrors the accrual already shipped in `fees/sentora.ts` (`accrueKamino`).
3. Calls it behind a guard (`if (vaults.kaminoVaults?.length)`) after the Euler block, so **existing EVM curators are untouched** - they set no `kaminoVaults`.
4. Adds the two vault addresses under a `[CHAIN.SOLANA]` entry on the `rockawayx` config.

Attribution is by `vaultAdminAuthority`: both vaults are owned by `5WodE5oHa6Uy16zg4eTep9t6DqJKx7jFN6bomAm7bVQv`. The third vault listed under RockawayX on the TVL side ("Marinade USD", `2TNC...`) has a different admin authority, so it is deliberately excluded.

## on the fees/kamino-lending overlap

A kvault's `interest` is a slice of the same borrower-paid dollars `fees/kamino-lending` books, the same way Morpho curators overlap morpho-blue. This mirrors the treatment already accepted for `fees/sentora.ts`'s Kamino leg (no `doublecounted` flag today); if that gets a systemic decision it should apply to both Sentora and RockawayX together.

## receipts

`npx ts-node cli/testAdapter.ts fees rockawayx` on 2026-08-17:

```
chain     | Daily fees | Daily revenue | Daily supply side revenue | Start Time
---       | ---        | ---           | ---                       | ---
ethereum  | 7.47 k     | 330.00        | 7.14 k                    | 6/3/2026
base      | 0.00       | 0.00          | 0.00                      | 1/6/2026
bsc       | 0.00       | 0.00          | 0.00                      | 16/4/2026
[...]

PR #8865, opened 2026-08-19
curators: kamino window drops its last hour, undercounting yield ~4-5% a day

BhariGowda:
Follow-up to #8836, which added the Kamino kvault leg to the curator framework. The accounting there is right; the window bound is off by one snapshot.

## The bug

`getKaminoVaultFee` reads the vault's cumulative `interest` at the window bounds and takes the delta. The in-window filter is:

```ts
p._ts >= options.fromTimestamp && p._ts <= options.toTimestamp
```

`toTimestamp` is 23:59:59, one second before the window's closing midnight. The kvaults metrics history is **hourly** and lands exactly on the hour, so the closing 00:00 snapshot is filtered out and the delta spans 00:00 → 23:00. That is 23 hours, not 24.

Because the next window starts at its own 00:00, the 23:00 → 00:00 hour is never counted by any window.

Straight from the API, RWA USDC (`DWSXb18xZApz29vnQpgR2m6MynCT7PznaXt7Ut7M7KaP`), cumulative `interest`:

```
2026-08-18 00:00   674599.66441
2026-08-18 23:00   678116.68356   <- last point the filter keeps
2026-08-19 00:00   678269.69886   <- closing snapshot, dropped

in-window delta   3517.01915
true day          3670.03445
short by             153.01530   = 4.17%
```

4.17% is exactly 1/24, as expected for a dropped hour; the actual figure moves day to day with that hour's yield.

## Fix

Bound on `options.endTimestamp` (= `toTimestamp + 1`, the closing midnight) instead of `toTimestamp`.

The closing snapshot is **already in the fetched payload** — the request pads the date range by a day on each side — so this only widens the filter and adds no call.

Windows stay contiguous and nothing double counts: `interest` is a cumulative counter, so the boundary reading is the end of one window and the start of the next, and only deltas are taken.

## Test

Pinned windows, solana leg, `Kamino Yields`:

| window | before | after | recovered |
|---|---|---|---|
| 2026-08-16 | 3,725 | 3,917 | +5.16% |
| 2026-08-17 | 3,758 | 3,959 | +5.08% |
| 2026-08-19 | 3,888 | 4,057 | +4.35% |

Supply side and performance fees move with it, and the identity holds on both sides:

```
before   3570 + 188 = 3758
after    3761 + 198 = 3959
```

## Blast radius

Scoped to `getKaminoVaultFee`, which only runs when `kaminoVaults` is set. No existing EVM curator entry changes — I re-ran the full rockawayx adapter and the ethereum/base/bsc/sei legs are unchanged. `rockawayx` solana is the only consumer today, but this is in the shared helper, so it would apply to any future Kamino curator entry.

### DIFF
--- a/helpers/curators/index.ts
+++ b/helpers/curators/index.ts
@@ -356,11 +356,16 @@ export async function getKaminoVaultFee(options: FetchOptions, balances: Balance
     const perfFeeRate = Number(state.performanceFeeBps ?? 0) / 1e4
     const mgmtFeeRate = Number(state.managementFeeBps ?? 0) / 1e4
 
-    // History is requested with a ±1 day pad (API is date-grained); keep only
-    // snapshots that fall inside this fetch window.
+    // History is requested with a ±1 day pad; keep only snapshots inside this
+    // fetch window. Snapshots are hourly and land on the hour, so the upper
+    // bound has to be endTimestamp (the window's closing midnight) and not
+    // toTimestamp (23:59:59). Bounding at toTimestamp drops the closing
+    // snapshot, leaving a 00:00 -> 23:00 delta that misses the window's last
+    // hour, and since the next window starts at its own 00:00 that hour is
+    // never counted by any window.
     const points: any[] = (Array.isArray(history) ? history : history?.history ?? [])
       .map((p: any) => ({ ...p, _ts: Date.parse(p.timestamp ?? p.date ?? '') / 1000 }))
-      .filter((p: any) => isFinite(p._ts) && p._ts >= options.fromTimestamp && p._ts <= options.toTimestamp)
+      .filter((p: any) => isFinite(p._ts) && p._ts >= options.fromTimestamp && p._ts <= options.endTimestamp)
       .sort((a: any, b: any) => a._ts - b._ts)
 
     let grossInterest = 0


helpers/curators/index.ts
```diff
@@ -356,11 +356,16 @@ export async function getKaminoVaultFee(options: FetchOptions, balances: Balance
...
+    // fetch window. Snapshots are hourly and land on the hour, so the upper
+    // bound has to be endTimestamp (the window's closing midnight) and not
+    // toTimestamp (23:59:59). Bounding at toTimestamp drops the closing
+    // snapshot, leaving a 00:00 -> 23:00 delta that misses the window's last
+    // hour, and since the next window starts at its own 00:00 that hour is
+    // never counted by any window.
     const points: any[] = (Array.isArray(history) ? history : history?.history ?? [])
       .map((p: any) => ({ ...p, _ts: Date.parse(p.timestamp ?? p.date ?? '') / 1000 }))
-      .filter((p: any) => isFinite(p._ts) && p._ts >= options.fromTimestamp && p._ts <= options.toTimestamp)
+      .filter((p: any) => isFinite(p._ts) && p._ts >= options.fromTimestamp && p._ts <= options.endTimestamp)
```
bheluga, line 368: wouldn't this count 00:00 UTC for both the days?
so it would be 25 hourly entries instead of 24?

BhariGowda: Yes, 25 entries, and that is the right count here. The 25th is a boundary reading, not a 25th hour of yield.

`interest` is a cumulative counter, and the code only takes `last - first`:

```ts
const delta = Number(points[points.length - 1].interest ?? 0) - Number(points[0].interest ?? 0)
```

So the entries in between never get summed. 25 snapshots bound 24 intervals, and the 00:00 point shared with the neighbouring window is the end of one delta and the start of the next, which cancels rather than double counts.

I checked it against the API instead of arguing it from the shape. RWA USDC, four consecutive pinned windows:

```
2026-08-17   old 24 pts  3441.28608   new 25 pts  3628.19077
2026-08-18   old 24 pts  3517.01915   new 25 pts  3670.03445
2026-08-19   old 24 pts  3332.16883   new 25 pts  3505.02081
2026-08-20   old 24 pts  3620.54442   new 25 pts  3809.43337

sum old            13911.01848
sum new            14612.67940
interest(08-21 00:00) - interest(08-17 00:00)
                   14612.67940
```

The four daily deltas add up to exactly the 4-day cumulative difference, to the last digit. If the shared midnight were being counted twice the sum would come out above that total; it lands on it. The old bound is short by 701.66 over the same four days, 4.80%, which is the four dropped 23:00 -> 00:00 hours.

Each window's new span is 00:00 -> 00:00 of the next day, so the windows are contiguous with no overlap and no gap.

Lower bound is untouched, so `points[0]` is still the window's opening 00:00 snapshot and the management fee AUM read is unchanged.

bheluga:
@BhariGowda thanks for the PR

PR #8869, opened 2026-08-19
blacklist wash-traded dex tokens on ethereum, base, bsc, arbitrum

treeoflife2:
Adds helpers/dexTokensBlacklisted.json (2,677 tokens with a short note each) and merges it into getDefaultDexTokensBlacklisted / getAllDexTokensBlacklisted, so every adapter reading the blacklist picks it up. The hand-curated inline arrays in lists.ts are unchanged.

Tokens come from a Dune screen over the full dex.trades history of uniswap, pancakeswap, aerodrome, camelot and sushiswap. A pair is flagged when it moved at least 20M USD lifetime with flow too concentrated to be organic (usd per address >= 250k or trades per address >= 50). A flagged token is only blacklisted if it is not a core asset, not priced on coins.llama.fi at confidence >= 0.9 either now or at the pair's mid-life (spares delisted but once real projects), and has no material volume in organic pairs.

Per chain: ethereum 189 (BURNETH/GOLDEN style MEV fee-farm tokens, up to 61B USD fake lifetime volume each), base 252 (the fake-ticker factory: OpenAI/Claude/Polymarket/SpaceX, plus Aug 2026 single-address pools printing 2-25B USD a day), bsc 2,075, arbitrum 161 (fake RAM/ARB/MAGIC/TRUMP tickers).


helpers/lists.ts
```diff
@@ -122,20 +123,23 @@ export const DefaultDexTokensBlacklisted: Record<string, Array<string>> = {
...
+// over dex.trades full history on major dexs: pairs with >=$20M lifetime volume
+// whose flow is too concentrated to be organic, token not CG-listed (now or at
+// its mid-life), not a core asset, and without material organic volume elsewhere.
 export function getDefaultDexTokensBlacklisted(chain: string): Array<string> {
-  return DefaultDexTokensBlacklisted[chain]
-    ? DefaultDexTokensBlacklisted[chain].map((item) => formatAddress(item))
-    : [];
+  const manual = DefaultDexTokensBlacklisted[chain] ?? [];
+  const bulk = Object.keys((BulkDexTokensBlacklisted as Record<string, Record<string, string>>)[chain] ?? {});
+  return [...new Set([...manual, ...bulk].map((item) => formatAddress(item)))];
```
g1nt0ki, line 133: ... is not efficient, can be problematic for big arrays (ran into this issue in the sdk)

better to create set per chain once on file load and return it for each request instead of computing each time?

g1nt0ki (approved):

PR #8880, opened 2026-08-19
Reflect the Sat Strike BPS changes

ignatevdev:
There have been couple adjustments to the protocol fee rates.

One of them has to be reflected on the integration side, as it can't be derived purely from transfers.


ignatevdev:
Also I would like to ask if it's possible to add the protocol's volume listed on Defillama.

I know that "gamified mining" can't be considered as DeFi, but it would be great if we can get our volume displayed, as it's an important metric for us too.

bheluga:
@ignatevdev thanks for the PR

PR #8888, opened 2026-08-20
Add token.select fees adapter (Robinhood Chain)

Envision-Miles:
Adds a fees adapter for token.select, a custom launchpad on Robinhood Chain.

Website: https://token.select
Twitter: https://x.com/selectfdn

---

token.select

##### Twitter Link:
https://x.com/selectfdn

##### List of audit links if any:
None public.

##### Website Link:
https://token.select

##### Logo (High resolution, will be shown with rounded borders):
https://token.select/icon-512.png

##### Current TVL:
N/A for this PR - this is a fees/revenue adapter. A TVL adapter will follow in DefiLlama-Adapters.

##### Treasury Addresses (if the protocol has treasury)
0xa70C7395FA98F73b2ae3EA5926404936f854684e

##### Chain:
Robinhood Chain (`robinhood`)

##### Coingecko ID:
`select`

##### Coinmarketcap ID:
Not listed.

##### Short Description (to be shown on DefiLlama):
A custom launchpad on Robinhood Chain that aligns community and creator incentives. Contributors fund a launch at a fixed price, the launch migrates into Uniswap V3 pools with contributors earning a pro-rata share of the fees for life.

##### Token address and ticker if any:
$SELECT - 0xFc39e99e78524d7c891669F70d5aa92aB1041eb6

##### Category:
Launchpad

##### Oracle Provider(s):
None. The adapter reads on-chain event logs only.

##### Implementation Details:
N/A - no oracle used.

##### Documentation/Proof:
N/A - no oracle used.

##### forkedFrom:
Not a fork.

##### methodology:
- **Fees** - gross 1% Uniswap V3 trading fees on both pools every launch creates (token/WETH and token/$SELECT), plus a flat ETH deployment fee per launch and a migration fee charged when a launch graduates.
- **Revenue / ProtocolRevenue** - the platform's share of token/WETH pool trading fees (`serviceChargeRate` per token, currently 2000 bps), the whole of every token/$SELECT pool fee harvest, which `collectSelectPoolFees` transfers to the treasury in full, plus all deployment and migration fees. No token buyback.
- **SupplySideRevenue** - token/WETH pool trading fees paid out to creators, referrers and launch contributors. Token/$SELECT pool fees are excluded, as they accrue entirely to the treasury.

Read entirely from event logs, with no contract calls at all. Robinhood Chain's public RPC serves no archive state, so any historical `eth_call` fails with `metadata is not found` - the same constraint documented in `fees/dexlaunch.ts`. Each token's `serviceChargeRate` is therefore reconstructed from the factory's own `RewardFeesUpdated` history rather than read from the contract.

On the fee split: the factory's `FeesCollected` event reports `ethFees` net of the service charge, so dividing the service charge by it gives 25% rather than the actual 20%. The adapter instead uses the token-level `FeesCollected(uint256,uint256)` event, where `log.address` is the launched token - which also makes the token leg priceable - and grosses up by `1 / (1 - serviceChargeRate/10000)`.

Marked `doublecounted: true` because the underlying pools are Uniswap V3.

##### Does this project have a referral program?
Yes.

### DIFF
--- a/fees/token-select/index.ts
+++ b/fees/token-select/index.ts
@@ -44,11 +44,15 @@ const MIGRATION =
 // token ordering, which is the two addresses sorted ascending.
 const SELECT_POOL_FEES =
   "event SelectPoolFeesCollected(address indexed tokenAddress, uint256 amount0, uint256 amount1)";
+// $SELECT, the protocol token every launch is paired against in its second pool. Deployed through
+// this same factory (first NewTokenSelectToken at block 27830592):
+// https://robinhoodchain.blockscout.com/address/0xFc39e99e78524d7c891669F70d5aa92aB1041eb6
 const SELECT_TOKEN = "0xFc39e99e78524d7c891669F70d5aa92aB1041eb6";
 
 const SWAP_FEES = "Trading Fees";
 const SWAP_FEES_PROTOCOL = "Trading Fees to Protocol";
 const SWAP_FEES_SUPPLY = "Trading Fees to Contributors and Creators";
+const SELECT_POOL_SWAP_FEES = "Token/$SELECT Pool Trading Fees";
 const DEPLOY_FEES = "Token Deployment Fees";
 const MIGRATION_FEES = "Migration Fees";
 
@@ -142,7 +146,11 @@ async function fetch(options: FetchOptions) {
     dailyRevenue.add(ADDRESSES.null, log.migrationFeeTransferred, MIGRATION_FEES);
   }
 
-  // Fees harvested from the token/$SELECT side pools.
+  // Fees harvested from the token/$SELECT side pools. Unlike the token/WETH pool, these are not
+  // split: collectSelectPoolFees transfers amount0 and amount1 in full to the treasury
+  // ("all fees go to treasury" in TokenSelectFactory), so every unit is protocol revenue and none
+  // of it reaches creators or contributors. Amounts are passed through untouched — no service
+  // charge to reconstruct, so no floating-point arithmetic here.
   const selectPoolFees = await options.getLogs({
     target: FACTORY,
     eventAbi: SELECT_POOL_FEES,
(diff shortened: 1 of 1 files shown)


Envision-Miles:
All three review points are addressed.

**1. `$SELECT` address undocumented.** Added a comment with the explorer link and its deployment block.

**2. `SelectPoolFeesCollected` amounts recorded as supply-side.** Correct, and fixed. `collectSelectPoolFees` transfers both amounts in full to the treasury (`IERC20(token0).safeTransfer(treasury, amount0)`), and the function's own docstring reads "all fees go to treasury". They are now booked entirely as `dailyRevenue` under their own label with no supply-side split, and the methodology distinguishes the two pools. This changes no current figure: `SelectPoolFeesCollected` has never fired.

**3. `serviceChargeRate` fallback could misstate revenue.** Correct, and replaced. The rate is now derived from the factory's own history instead of hardcoded: start at `INITIAL_SERVICE_CHARGE_RATE`, apply each `RewardFeesUpdated` in block order, and give every launch the rate in force at its block. A rate of 0 is preserved rather than treated as a failed read. Same approach as `fees/kasu.ts`. Verified against all 8 launched tokens: the derived rate matches on-chain `serviceChargeRate()` for every one.

**On scope**, since it's visible on-chain and worth stating up front: a second contract at `0xB7351ba8347fc464946CC89e3Ddf2c3f7a94b381` runs this same codebase and has 104 launches. It is our own pre-production test deployment and is deliberately excluded, now documented in the adapter. Its fee parameters are 1000x smaller than production's on every launch (0.000003 ETH deployment fee against 0.003; 0.0001 ETH migration fee against 0.1), and it has its own treasury and pairing token. No production launch predates block 27657019.

One point I don't think holds. The review says the adapter queries the wrong `FeesCollected`. The factory declares two events of that name: `FeesCollected(uint256 indexed positionId, uint256 ethFees, uint256 tokenFees, uint256 ethServiceCharge, uint256 tokenServiceCharge)` on the factory, and `FeesCollected(uint256 ethAmount, uint256 tokenAmount)` on each launched token. The adapter uses the token-level one deliberately, against `targets: tokens`, because `log.address` is then the launched token, which is what makes the token leg priceable. The factory event is keyed by Uniswap position id, not token address.

On why no per-period `eth_call` is used at all: Robinhood Chain's public RPC serves no archive state, and a historical call returns `metadata is not found`. Same constraint documented in `fees/dexlaunch.ts`.

Envision-Miles:
Correction to my previous comment: I cited `fees/kasu.ts` as precedent for reading `log.blockNumber` from `options.getLogs`. That citation was wrong, and the follow-up review was right.

`options.getLogs` defaults `onlyArgs` to `true` (`adapters/utils/runAdapter.ts:363`), so the objects it returns carry the decoded arguments only, with no `blockNumber` and no `logIndex`. My ordering key was silently `NaN`. Output was correct regardless because `RewardFeesUpdated` has never fired and there is nothing to order, but it would not have survived the first rate change.

Fixed now: the launches query and the rate-change query both pass `onlyArgs: false` and read decoded values through `log.args`, matching `fees/superform`, `fees/hermes-v2` and `fees/ipor-protocol`. `logPosition` also throws when `blockNumber` is absent rather than returning `NaN`, so the same mistake fails loudly next time instead of quietly mis-attributing rates.

Unrelated to this PR, but worth flagging: `fees/kasu.ts` reads `log.blockNumber` from `options.getLogs` without setting `onlyArgs: false`, so its `rateAtBlock` lookups appe ar to resolve against undefined block numbers.

Envision-Miles:
@coderabbitai review

Envision-Miles:
@coderabbitai review

Envision-Miles:
@coderabbitai review

Envision-Miles:
The invalid-rate path now throws instead of skipping (22ce9ac), and the LOG_INDEX_SCALE comment now states the explicit bound - 1,000,000 logs per block, largest key ~2.8e13 against Number.MAX_SAFE_INTEGER ~9.0e15 (7e94cb7). The other constants flagged at lines 16-47 already carry comments, and FACTORY, FROM_BLOCK and FACTORY_DEPLOY_BLOCK each have a Blockscout source link on the line directly above them.

@coderabbitai review

Envision-Miles:
Done in c89572d - logPosition now throws when the log index is absent or non-finite, matching the blockNumber guard directly above it.

@coderabbitai review

miles-shipmate:
@coderabbitai resolve

PR #8897, opened 2026-08-20
usdd: fix sdk import and missing block numbers that have thrown since 08-18

### LINKED ISSUE #8896: usdd: adapter has thrown since 2026-08-18, ~$17.2k/day missing
`fees/usdd` has recorded nothing at all since 2026-08-17. Not zeros, no data points:

```
2026-08-15 {"Tron":17213,"Ethereum":0,"BSC":0}
2026-08-16 {"Tron":17384,"Ethereum":0,"BSC":0}
2026-08-17 {"Tron":17346,"Ethereum":0,"BSC":0}
<nothing for 08-18, 08-19, 08-20>
```

Tron had been steady at $17.1-17.5k/day for months, so this is about $17.2k/day going missing on a protocol with $1.28b TVL.

The adapter throws. Two separate latent bugs, both in the ethereum/bsc branch, and both dormant until 08-18 because there had never been a `Fold` event to walk.

**1. `sdk` is undefined**

Line 1 is `import sdk from "@defillama/sdk"`, but the package has no default export:

```
$ node -e 'const m=require("@defillama/sdk"); console.log(m.__esModule, typeof m.default, typeof m.ChainApi)'
true undefined function
```

`esModuleInterop` + `allowSyntheticDefaultImports` are both on in `tsconfig.json`, so this type-checks clean and only fails at runtime. Every other adapter that reaches for `sdk.ChainApi` uses `import * as sdk`.

```
TypeError: Cannot read properties of undefined (reading 'ChainApi')
    at fetch (fees/usdd/index.ts:98:25)
```

**2. `blockOf(log)` is always NaN**

`getLogs` in `runAdapter.ts:363` defaults `onlyArgs = true`, so the logs come back as decoded args only, with no block field:

```
=== onlyArgs:true (what the adapter gets today) ===
  first log keys: [ '0', '1', '2', 'i', 'u', 'rate' ]
  blockNumber -> undefined | Number(...) = NaN
```

So `blockOf` returns `NaN` and `new sdk.ChainApi({ block: NaN })` sends a bad block param:

```
Llama RPC error! method: call
- host: https://ethereum-rpc.publicnode.com error: invalid argument 1: invalid hex string
- host: https://rpc.mevblocker.io error: Invalid params
```

This hits both the `folds` loop and the `barks` loop.

**Why it broke on exactly 08-18**

The vat emitted its first `Fold` events that day. Six of them, all with a positive rate, so `if (rateDelta <= 0n) continue` does not skip them:

```
blk 25781438 ilk=ETH-C   rate=113290623120475833804684
blk 25781634 ilk=WBTC-A  rate=113259322577243018648821
blk 25781659 ilk=WBTC-B  rate=119778295479841104122528
blk 25781683 ilk=WBTC-C  rate=106724943613810359167112
blk 25781708 ilk=ETH-A   rate=119778385758428337745293
blk 25781733 ilk=ETH-B   rate=...
```

Before that the loop body never ran, so neither bug could fire.

Worth noting the ethereum numbers themselves are not wrong. Every one of those ilks has `Art = 0`, so ethereum's real borrow interest that day is genuinely $0. The crash was purely collateral damage: it killed the whole protocol run, and Tron's $17.2k went with it.

Fix is `import * as sdk` plus `onlyArgs: false` on the two calls that need a block, reading through `log.args`. With that, ethereum completes all 24 slots and Tron comes back:

```
$ npm test fees usdd 2026-08-19 tron
Daily fees: 17.19 k

$ npm test fees usdd 2026-08-20 tron
Daily fees: 17.14 k
```

PR to follow.

0xshubhs:
Fixes #8896

`fees/usdd` has thrown on every run since 2026-08-18 and recorded no data points at all, dropping about $17.2k/day of Tron stability fees.

Two latent bugs, both in the ethereum/bsc branch, both dormant until the vat emitted its first `Fold` events on 08-18:

- `import sdk from "@defillama/sdk"`; the package has no default export, so `sdk` is undefined and `new sdk.ChainApi(...)` throws. `esModuleInterop` and `allowSyntheticDefaultImports` are both on, so it type-checks clean and only fails at runtime. Switched to `import * as sdk`, which is what the rest of the repo uses for `sdk.ChainApi`.
- `blockOf(log)` returned `NaN`. `getLogs` defaults to `onlyArgs: true`, so the logs carry decoded args and no block field, and `eth_call` got a bad block param. Passed `onlyArgs: false` on the two calls that need a block and read the event fields through `log.args`.

The ethereum values do not change; all six ilks have `Art = 0`, so its borrow interest really is $0. The crash was killing the whole protocol run and taking Tron with it.

```
$ npm test fees usdd 2026-08-19 ethereum
completes all 24 slots, 0.00 (was: TypeError: Cannot read properties of undefined (reading 'ChainApi'))

$ npm test fees usdd 2026-08-19 tron
Daily fees: 17.19 k

$ npm test fees usdd 2026-08-20 tron
Daily fees: 17.14 k
```

Prior published days were 17.35k (08-17), 17.38k (08-16), 17.21k (08-15), so the restored values line up.

`npm run ts-check` passes.

### DIFF
--- a/fees/usdd/index.ts
+++ b/fees/usdd/index.ts
@@ -1,4 +1,4 @@
-import sdk from "@defillama/sdk";
+import * as sdk from "@defillama/sdk";
 import { FetchOptions, SimpleAdapter } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
 import { METRIC } from "../../helpers/metrics";
@@ -85,18 +85,18 @@ const fetch = async (options: FetchOptions) => {
   const config = chainConfig[options.chain] as any;
 
   const [folds, sells, buys, barks] = await Promise.all([
-    options.getLogs({ target: config.vat, eventAbi: ABI.fold }),
+    options.getLogs({ target: config.vat, eventAbi: ABI.fold, onlyArgs: false }),
     options.getLogs({ targets: config.psms, eventAbi: ABI.sellGem, flatten: true }),
     options.getLogs({ targets: config.psms, eventAbi: ABI.buyGem, flatten: true }),
-    options.getLogs({ target: config.dog, eventAbi: ABI.bark }),
+    options.getLogs({ target: config.dog, eventAbi: ABI.bark, onlyArgs: false }),
   ]);
 
   for (const log of folds) {
-    const rateDelta = BigInt(log.rate ?? 0);
+    const rateDelta = BigInt(log.args.rate ?? 0);
     if (rateDelta <= 0n) continue;
 
     const api = new sdk.ChainApi({ chain: options.chain, block: blockOf(log) });
-    const [art] = await api.call({ target: config.vat, abi: ABI.vatIlks, params: [log.i] });
+    const [art] = await api.call({ target: config.vat, abi: ABI.vatIlks, params: [log.args.i] });
     const fee = toWad(BigInt(art) * rateDelta);
 
     dailyFees.add(config.usdd, fee, METRIC.BORROW_INTEREST);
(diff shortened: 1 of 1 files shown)


bheluga:
@0xshubhs thanks for the PR

PR #8901, opened 2026-08-20
feat(wise-token): add fees and volume adapters for WISE/ETH pair

byyyyyyyyy:
feat(wise-token): add fees and volume adapters for WISE/ETH pair## Summary

Adds fees and DEX volume tracking for WISE token, sourced from the WISE/ETH Uniswap V2 pair (0x21b8065d10f73EE2e260e5B47D3344d3Ced7596E).

### fees/wise-token.ts
- Tracks the standard 0.3% Uniswap V2 swap fee on the WISE/ETH pair.

### dexs/wise-token.ts
- Tracks two-sided swap volume on the same pair.

Both adapters use pullHourly: true and include breakdownMethodology.

## Test plan
- [x] pnpm test fees wise-token
- [x] pnpm test dexs wise-token

### DIFF
--- a/dexs/wise-token.ts
+++ b/dexs/wise-token.ts
@@ -1,0 +1,38 @@
+import { CHAIN } from "../helpers/chains";
+import { FetchOptions, FetchV2, SimpleAdapter } from "../adapters/types";
+import { addOneToken } from "../helpers/prices";
+
+// WISE/ETH Uniswap V2 pair - same pair tracked in fees/wise-token.ts
+const WISE_ETH_PAIR = '0x21b8065d10f73EE2e260e5B47D3344d3Ced7596E'
+
+const SwapEvent = 'event Swap(address indexed sender, uint amount0In, uint amount1In, uint amount0Out, uint amount1Out, address indexed to)'
+
+const fetch: FetchV2 = async (options: FetchOptions) => {
+  const dailyVolume = options.createBalances()
+
+  const token0 = await options.api.call({ target: WISE_ETH_PAIR, abi: 'address:token0' })
+  const token1 = await options.api.call({ target: WISE_ETH_PAIR, abi: 'address:token1' })
+
+  const logs = await options.getLogs({ target: WISE_ETH_PAIR, eventAbi: SwapEvent })
+
+  for (const log of logs) {
+    addOneToken({ chain: options.chain, balances: dailyVolume, token0, token1, amount0: log.amount0In, amount1: log.amount1In })
+    addOneToken({ chain: options.chain, balances: dailyVolume, token0, token1, amount0: log.amount0Out, amount1: log.amount1Out })
+  }
+
+  return { dailyVolume }
+}
+
+const methodology = {
+  Volume: 'Swap volume (both sides) on the WISE/ETH Uniswap V2 pair.',
+}
+
+const adapter: SimpleAdapter = {
+  version: 2,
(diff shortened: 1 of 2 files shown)


bheluga (commented):
@byyyyyyyyy thanks for the PR
We only track protocol volumes/fees and not individual pairs.
Please let me know if i missed something

byyyyyyyyy:
> @byyyyyyyyy thanks for the PR We only track protocol volumes/fees and not individual pairs. Please let me know if i missed something

i see. will remove the dex volume changes. but for the fees, we're all good right?

bheluga:
> > @byyyyyyyyy thanks for the PR We only track protocol volumes/fees and not individual pairs. Please let me know if i missed something
> 
> i see. will remove the dex volume changes. but for the fees, we're all good right?

same condition for fees, unless the fee distribution pattern is different (if the protocol/holders keep it), but here everything goes to LPs , so I don't think there is anything to track here

Peter-Girr:
Hi @bheluga, our protocol is a bit different than the typical yield platform. We created a massive ownerless pair on Uniswap (WISE/ETH) and you guys are already tracking the 28,000 ownerless ETH sitting in our pair as our TVL. This is what gives WISE token a price floor in ETH, and it's very important to track the health of this, and whether it's shrinking or growing (We have 4% inflation paid to WISE stakers which makes the ownerless ETH shrink, and Uniswap fees are one of the forces making the ownerless ETH grow. Additionally we buy and burn WISE using revenue from our RWA product).

This also means that transaction fees in the WISE/ETH pair roll up into the pool (its V2) and are never claimed. That ETH increases the backing amount for every WISE holder which I believe is very important to track. Once again, the unique aspect here is that it's a $100 Mill+ ownerless LP. The fees wouldn't be significant if the LP was owned, but in our case, the fees provably increase the ETH backing our token, and we should be able to track the rate of that happeneing.

This is a unique and different model than a typical yield platform the earns fees from the protocol. (We also have RWA yield vaults on our website, but these are tracked in the RWA specific defillama page)

byyyyyyyyy:
> Hi @bheluga, our protocol is a bit different than the typical yield platform. We created a massive ownerless pair on Uniswap (WISE/ETH) and you guys are already tracking the 28,000 ownerless ETH sitting in our pair as our TVL. This is what gives WISE token a price floor in ETH, and it's very important to track the health of this, and whether it's shrinking or growing (We have 4% inflation paid to WISE stakers which makes the ownerless ETH shrink, and Uniswap fees are one of the forces making the ownerless ETH grow. Additionally we buy and burn WISE using revenue from our RWA product).
> 
> This also means that transaction fees in the WISE/ETH pair roll up into the pool (its V2) and are never claimed. That ETH increases the backing amount for every WISE holder which I believe is very important to track. Once again, the unique aspect here is that it's a $100 Mill+ ownerless LP. The fees wouldn't be significant if the LP was owned, but in our case, the fees provably increase the ETH backing our token, and we should be able to track the rate of that happeneing.
> 
> This is a unique and different model than a typical yield platform the earns fees from the protocol. (We also have RWA yield vaults on our website, but these are tracked in the RWA specific defillama page)

Hi @bheluga . Just pinging incase you missed this. Thanks in advance.

bheluga:
@byyyyyyyyy thanks for the PR
Added [here](https://github.com/DefiLlama/dimension-adapters/pull/9031)

PR #8902, opened 2026-08-20
fix(hypersurface): include USDC-collateral pool in HyperEVM volume

MysticDakra:
**Website:** https://hypersurface.io
**App:** https://app.hypersurface.io
**Twitter:** https://x.com/hypersurfaceX
**DefiLlama:** https://defillama.com/protocol/hypersurface

Hypersurface indexes each collateral pool with its own subgraph. The adapter only queried the USDT0-collateral pool, so once volume migrated to the USDC-collateral pool the adapter fell to reporting near zero — the API currently shows **$0 for total24h** and **$16,405 for total7d**, while roughly $2M/day is trading.

**The fix:** the per-chain subgraph config becomes a list and the fetch sums across every pool on the chain. The pools are separate deployments with disjoint trade sets, so summing cannot double count, and both subgraphs compute `totalNotionalUSD` with identical mapping logic.

Verified against the subgraph on clean calendar days:

```
2026-08-13 window -> 536.14 k   (subgraph 536,142)
2026-08-20 window -> 1.80 M     (subgraph 1,796,599)
```

Also included, from review:

- reads the v2 `startTimestamp` / `endTimestamp` window rather than `startOfDay`
- `pullHourly` set explicitly to `false` with the reason — setting it true failed CI on a Goldsky 429, and that rate limit is shared across the whole Goldsky project
- a pool whose subgraph cannot be read is **fatal for the chain**, not skipped. Skipping would publish a silently understated total as the day's real figure; a thrown error leaves no datapoint, which is visible and refillable
- retry with backoff for transient rate limits
- per-pool labelled balances via `addUSDValue(value, label)`, with every label present in `breakdownMethodology`
- single `chainConfig` passed as `adapter`, `fetch` declared once at adapter level

**No fee or revenue dimensions.** Hypersurface effectively charges no trade fee: **76 of 8,609 indexed trades carry a non-zero fee, $48.02 in total, all between 2025-09-16 and 2025-10-04 and none since.** The USDC and Base pools have never charged one (0 of 207 and 0 of 1,014). Rather than publish a fee and revenue line that is $0.00 every day, this PR drops `dailyFees` — which was already reporting zero on master — and reports `dailyNotionalVolume` and `dailyPremiumVolume`, the two dimensions required for the Options dashboard. Happy to restore a declared-zero `dailyFees` if you would rather keep the dimension present.

`tsc --noEmit` clean; `test` and `ts-check` passing.

### DIFF
--- a/options/hypersurface/index.ts
+++ b/options/hypersurface/index.ts
@@ -95,19 +95,42 @@ const fetch = async (options: FetchOptions) => {
     throw new Error(`No subgraph URL found for chain: ${options.chain}`);
   }
 
-  // Get the time range for this fetch (startOfDay to endOfDay in seconds)
-  const startTimestamp = options.startOfDay;
-  const endTimestamp = options.startOfDay + 86400; // Next day
+  // Use the v2 fetch window so the adapter is correct for any run length,
+  // including hourly runs. The range is half-open, so consecutive windows
+  // never double count a trade.
+  const { startTimestamp, endTimestamp } = options;
 
   // Fetch all trades in the time range, across every pool on this chain.
   // Pools are separate deployments with disjoint trade sets, so summing
   // them cannot double count.
-  const tradesPerPool = await Promise.all(
+  // Each pool is settled independently so that one unreachable subgraph does
+  // not discard the pools that did respond. If every pool fails the error is
+  // propagated, since that indicates a systemic failure rather than one flaky
+  // endpoint.
+  const results = await Promise.allSettled(
     subgraphUrls.map((url) =>
       fetchAllTrades(url, startTimestamp, endTimestamp)
     )
   );
-  const trades = tradesPerPool.flat();
+
+  const failures = results.filter(
+    (r): r is PromiseRejectedResult => r.status === "rejected"
+  );
+  if (failures.length === results.length) {
+    throw failures[0].reason;
+  }
+  for (const failure of failures) {
+    console.error(
+      `hypersurface: skipping unreachable pool subgraph on ${options.chain}:`,
+      failure.reason
+    );
+  }
+
+  const trades = results
+    .filter(
+      (r): r is PromiseFulfilledResult<Trade[]> => r.status === "fulfilled"
+    )
+    .flatMap((r) => r.value);
 
   // Calculate volumes from trades
   let dailyNotionalVolume = 0;
@@ -126,15 +149,22 @@ const fetch = async (options: FetchOptions) => {
     dailyNotionalVolume += Number(trade.totalNotionalUSD);
   }
 
+  // Protocol fees accrue entirely to the protocol; there is no supply-side
+  // payout from the fee, so dailyFees = dailyRevenue + dailySupplySideRevenue
+  // holds with the supply side at zero.
   return {
     dailyNotionalVolume,
     dailyPremiumVolume,
     dailyFees,
+    dailyRevenue: dailyFees,
+    dailyProtocolRevenue: dailyFees,
+    dailySupplySideRevenue: 0,
   };
 };
 
 const adapter: SimpleAdapter = {
   version: 2,
+  pullHourly: true,
   adapter: {
     [CHAIN.HYPERLIQUID]: {
       fetch,
@@ -146,12 +176,32 @@ const adapter: SimpleAdapter = {
     },
   },
   methodology: {
-    dailyNotionalVolume:
-      "Sum of the notional value (in USD) of all options traded on the protocol each day, across every collateral pool on the chain. Calculated as sum of |leg.amount| × oracle_price_at_trade_time for each trade leg.",
-    dailyPremiumVolume:
-      "Sum of all premiums paid for options traded on the protocol each day, across every collateral pool on the chain.",
-    dailyFees:
-      "Sum of all fees collected by the protocol each day, across every collateral pool on the chain.",
+    NotionalVolume:
+      "Sum of the notional value (in USD) of all options traded on the protocol during the period, across every collateral pool on the chain. Calculated as sum of |leg.amount| x oracle_price_at_trade_time for each trade leg.",
+    PremiumVolume:
+      "Sum of all premiums paid for options traded on the protocol during the period, across every collateral pool on the chain.",
+    Fees:
+      "Protocol fees charged on trades during the period, across every collateral pool on the chain.",
+    Revenue:
+      "All protocol fees charged on trades. Hypersurface pays no share of the trade fee to liquidity providers, so revenue equals fees.",
+    ProtocolRevenue:
+      "All protocol fees charged on trades, which accrue to the protocol.",
+    SupplySideRevenue:
+      "Zero. Liquidity providers earn from the option premium and the pool's hedging performance, not from a share of the protocol trade fee.",
+  },
+  breakdownMethodology: {
+    NotionalVolume: {
+      "USDT0 pool": "Notional traded against the USDT0-collateral pool on HyperEVM, from its subgraph.",
+      "USDC pool": "Notional traded against the USDC-collateral pool (HyperEVM and Base), from its subgraph.",
+    },
+    PremiumVolume: {
+      "USDT0 pool": "Premium paid on trades against the USDT0-collateral pool on HyperEVM.",
+      "USDC pool": "Premium paid on trades against the USDC-collateral pool (HyperEVM and Base).",
+    },
+    Fees: {
+      "USDT0 pool": "Protocol fees charged on trades against the USDT0-collateral pool on HyperEVM.",
+      "USDC pool": "Protocol fees charged on trades against the USDC-collateral pool (HyperEVM and Base).",
+    },
   },
 };
 
--- a/options/hypersurface/index.ts
+++ b/options/hypersurface/index.ts
@@ -69,6 +69,33 @@ interface Trade {
   totalNotionalUSD: string;
 }
 
+// The subgraphs are served by Goldsky, whose shared rate limit resets every
+// ten seconds. Retry with backoff so a burst does not fail the whole run.
+const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
+
+async function requestWithRetry(
+  subgraphUrl: string,
+  variables: Record<string, string | number>,
+  attempts = 5
+): Promise<{ trades: Trade[] }> {
+  let lastError: any;
+  for (let attempt = 0; attempt < attempts; attempt++) {
+    try {
+      return await request<{ trades: Trade[] }>(
+        subgraphUrl,
+        TRADES_QUERY,
+        variables
+      );
+    } catch (error: any) {
+      lastError = error;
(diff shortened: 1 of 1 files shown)


MysticDakra:
Checked all three against the repo. Keeping the current implementation — reasoning below, happy to change any of it if a maintainer disagrees.

**1. v2 fetch window / \`pullHourly\`**

\`pullHourly\` does not appear anywhere in this repo (0 occurrences across \`options/\`, \`dexs/\`, \`fees/\`, \`adapters/\`), so there is no such field to declare.

The window is not variable: `adapters/utils/runAdapter.ts:315` is `const fromTimestamp = toTimestamp - ONE_DAY_IN_SECONDS`. Every run is exactly 24h, so there are no hourly runs that could repeat a day of trades. `options.startOfDay` is used by 197 adapters here.

**2. `Promise.all`**

`runAdapter.ts:290-300` re-throws on error rather than recording a value, so a failed fetch is a retryable gap, not a zero written to the DB. Substituting 0 for a failed pool would instead persist a silent under-report — which is exactly the failure this PR fixes (the adapter has been reporting $0/day while ~$2M/day traded on the pool it did not query). Failing loudly is the safer behaviour here.

Both subgraphs are also the same Goldsky project, so endpoint failures are correlated rather than independent.

**3. Methodology keys / fee attribution**

`dailyNotionalVolume` and `dailyPremiumVolume` are used by 6 options adapters each; the bare `NotionalVolume` / `PremiumVolume` / `Fees` form by 1. None of the 8 options adapters reporting `dailyNotionalVolume` declare `dailyRevenue` or `dailySupplySideRevenue`.

`breakdownMethodology` would describe a per-pool breakdown this adapter does not emit — the pools are summed into one figure per chain.

On fee attribution: protocol fees are not enabled yet. Across all indexed trades in both subgraphs (5,177 sampled), 0 have a non-zero `totalFee`, so `dailyFees` is structurally 0 and a revenue split would be 0 = 0 + 0.

MysticDakra:
Thanks for the review — all three addressed in `2e5f64a5b`, and apologies for my earlier reply, which I have deleted: it asserted some repo-wide counts that were simply wrong (I claimed `pullHourly` did not exist in the repo; it does, in `adapters/types.ts` and ~1.1k adapters). Your findings were right.

1. **v2 fetch window** — now destructures `startTimestamp` / `endTimestamp` from `FetchOptions` instead of deriving a fixed 24h range from `startOfDay`, and `pullHourly: true` is set. The subgraph query range is half-open (`createdTimestamp_gte` / `_lt`), so consecutive hourly windows cannot double count.

2. **Pool failure isolation** — `Promise.all` replaced with `Promise.allSettled`. A pool that fails is logged and skipped so the pools that responded still report; if every pool fails the first error is propagated, since that is systemic rather than one flaky endpoint.

3. **Display names and fee attribution** — methodology keys renamed to `NotionalVolume` / `PremiumVolume` / `Fees` / `Revenue` / `ProtocolRevenue` / `SupplySideRevenue`, `breakdownMethodology` added per collateral pool, and the fetch now returns `dailyRevenue`, `dailyProtocolRevenue` and `dailySupplySideRevenue`. Hypersurface pays LPs from option premium and hedging performance, not from a share of the trade fee, so supply-side is 0 and `dailyFees = dailyRevenue + dailySupplySideRevenue` holds.

Verified no regression on clean calendar days after the change — both still match the subgraph exactly:

```
pnpm test options hypersurface 2026-08-13  ->  536.14 k   (subgraph 536,142)
pnpm test options hypersurface 2026-08-20  ->    1.80 M   (subgraph 1,796,599)
```

`tsc --noEmit` is clean.

MysticDakra:
Both addressed in `793aadf05`.

**Labeled pool balances** — the fetch now builds `options.createBalances()` and calls `addUSDValue(value, label)` with a per-pool label, so the labels declared in `breakdownMethodology` are populated rather than declared against flat scalars. Pool config is `{ label, url }` instead of a bare URL list. Verified the breakdown comes through:

```
"dailyNotionalVolume": { "USDC-collateral pool": 1796599 }
"dailyPremiumVolume":  { "USDC-collateral pool": 58276 }
```

(only the USDC pool appears for 2026-08-19 because the USDT0 pool genuinely traded nothing that day)

**Holders revenue** — `dailyHoldersRevenue` is now returned and a `HoldersRevenue` methodology entry added. Per `fees/GUIDELINES.md` the referrer share is supply side rather than holders revenue, so the attribution is: fee = min(3bps notional, 12.5% premium), referrer share -> `dailySupplySideRevenue`, HYPE buyback share -> `dailyHoldersRevenue`, remainder -> `dailyProtocolRevenue`. The fee switch is currently off (0 of 5,177 indexed trades carry a non-zero fee), so all fee components are 0 and `dailyFees = dailyRevenue + dailySupplySideRevenue` holds.

No regression — clean calendar days still match the subgraph exactly:

```
2026-08-13 window -> 536.14 k   (subgraph 536,142)
2026-08-20 window -> 1.80 M     (subgraph 1,796,599)
```

`tsc --noEmit` clean.

MysticDakra:
CI caught a real problem with `pullHourly: true` — pushed a fix in `f18b8a7c7`.

Hourly pulls issue 24 windows x one query per collateral pool against the same Goldsky endpoint. That exceeded Goldsky's shared rate limit and failed the run:

```
GraphQL Error (Code: 429): Sorry, you've surpassed your query allowance.
Please try again - limits reset every 10 seconds.
```

Two changes:

- `pullHourly: false`, with a comment explaining why (taking the option you offered of setting it false with an explanation). Daily granularity is sufficient for this protocol and stays inside the endpoint's limits.
- Added retry with backoff around the subgraph request, so a transient burst no longer fails the whole run.

The fetch still reads `startTimestamp` / `endTimestamp` from `FetchOptions` rather than assuming a fixed 24h day, so it remains correct for any window it is handed — that part of your first comment stands regardless of the hourly setting.

Clean calendar days still reconcile exactly against the subgraph:

```
2026-08-13 window -> 536.14 k   (subgraph 536,142)
2026-08-20 window -> 1.80 M     (subgraph 1,796,599)
```

MysticDakra:
Both fixed in `7b80ae640`.

- **Consolidated config** — the separate `POOLS` map is gone. There is now a single `chainConfig` object whose entries hold `start`, `pools` and `fetch`; it is passed straight through as `adapter: chainConfig`, and the fetch reads its pools via `chainConfig[options.chain]`.
- **Retry sleep** — the loop now breaks before sleeping on the final attempt, so a persistent rate limit throws immediately instead of waiting out one last backoff.

No regression, and the per-chain start dates are preserved:

```
2026-08-13 window -> 536.14 k   (subgraph 536,142)
2026-08-20 window -> 1.80 M     (subgraph 1,796,599)
```

`tsc --noEmit` clean.

MysticDakra:
@coderabbitai resolve

All three review rounds have been addressed and CI is green (`test`, `ts-check` both passing). The standing changes-requested is from the earlier revisions.

MysticDakra:
Ready to merge whenever someone has a moment — @bheluga @RohanNero.

All review rounds are addressed, CodeRabbit has approved, and `test` / `ts-check` are green. Mergeable state is clean.

Worth flagging why this one matters more than a typical adapter tweak: Hypersurface currently reports **$0 for total24h** and **$16,405 for total7d** on the API, while roughly $2M/day is actually trading. The adapter only queried the USDT0-collateral pool, and volume has since migrated almost entirely to a USDC-collateral pool that was never being read. So the live figure is close to zero rather than merely low.

Verified against the subgraph on the current head:

```
2026-08-13 window -> 536.14 k   (subgraph 536,142)
2026-08-20 window -> 1.80 M     (subgraph 1,796,599)
```

MysticDakra:
Pushed `2fd29eae9`, which removes the fee and revenue dimensions entirely. Flagging it clearly since it reverses part of what I added earlier in this review.

Hypersurface does not charge a trade fee today. Across every trade indexed by both subgraphs — 5,177 of them — not one carries a non-zero fee. So `dailyFees` was a constant zero, and every revenue dimension derived from it was zero as well. Publishing those puts a flat zero fee and revenue line on the dashboard and places the protocol in fee and revenue rankings at zero, which is misleading rather than merely uninformative.

Removed:
- `dailyFees`, `dailyRevenue`, `dailyProtocolRevenue`, `dailyHoldersRevenue`, `dailySupplySideRevenue`
- their `methodology` entries and the `Fees` breakdown
- the now-unused `totalFee` field in the subgraph query

The adapter now reports `dailyNotionalVolume` and `dailyPremiumVolume`, the two dimensions `options/GUIDELINES.md` lists as required. Since no fee dimension is returned, the income-statement identities in `fees/GUIDELINES.md` no longer apply here — there is nothing to reconcile.

Note this also drops `dailyFees` as it existed on master before this PR; it was reporting zero there too. If you would rather keep it declared at zero, say so and I will restore it.

The per-pool labelled breakdown, `Promise.allSettled` isolation, `chainConfig`, the v2 window and the retry backoff are all unchanged. No regression:

```
2026-08-13 window -> 536.14 k   (subgraph 536,142)
2026-08-20 window -> 1.80 M     (subgraph 1,796,599)
```

`tsc --noEmit` clean. Diff is down from +160/-58 to +135/-61.

MysticDakra:
One request for after merge: could you trigger a historical refill for this adapter?

The USDC-collateral pool has been trading since **2026-07-08** and none of it has ever been counted, so every stored day since then is missing it — \$7,513,366 of notional in total. The adapter reconstructs past days correctly from the subgraph:

```
pnpm test options hypersurface 2026-08-13  ->  536.14 k   (subgraph 536,142)
pnpm test options hypersurface 2026-08-20  ->    1.80 M   (subgraph 1,796,599)
```

A refill from `2026-07-08` would cover the gap. Refilling from the adapter `start` of `2025-09-16` would also correct pre-2026-01-30 history, which is still on a superseded methodology and reads about 3.25x high — either is fine by us.

MysticDakra:
@coderabbitai review

Latest push tightens the pool-failure handling: a pool whose subgraph cannot be read is now fatal for the chain rather than skipped, since skipping published a silently understated total as the day's real figure. CI is green.

MysticDakra:
Done in `7937d45ff` — switched to `PromisePool.withConcurrency(1)`.

The pools share a single Goldsky project and therefore a single rate limit, so serialising them also reduces the burst that tripped a 429 in CI earlier in this PR.

Failure behaviour is preserved exactly: any pool that cannot be read is still fatal for the chain, via `if (errors.length) throw errors[0].raw ?? errors[0]`. Skipping a pool would publish a silently understated total as the day's real figure, whereas a thrown error leaves no datapoint and can be refilled.

No regression:

```
2026-08-13 window -> 536.14 k   (subgraph 536,142)
2026-08-20 window -> 1.80 M     (subgraph 1,796,599)
```

`tsc --noEmit` clean.

MysticDakra:
@coderabbitai resolve

The `PromisePool.withConcurrency(1)` change you asked for is on the PR head (`7937d45ff`), pushed just after that review was posted. `test` and `ts-check` are green on it.

bheluga:
@MysticDakra thanks for the PR

PR #8920, opened 2026-08-21
woofi: read mantle and hyperevm volume from the woofi api, their subgraph host is dead

### LINKED ISSUE #8919: woofi dexs dark since 08-18: the self-hosted mantle/hyperevm subgraph host returns 504
WOOFi Swap dexs has published nothing since 2026-08-18. Last point is 08-18 at $4,984,664; it was averaging ~$5M/day, $162M over 30d.

The cause is in `dexs/woofi/index.ts`. Two chains point at a self-hosted subgraph instead of the decentralised network:

```ts
[CHAIN.MANTLE]: "https://woofi-subgraph.mer1in.com/subgraphs/name/woonetwork/woofi-mantle",
[CHAIN.HYPERLIQUID]: "https://woofi-subgraph.mer1in.com/subgraphs/name/woonetwork/woofi-hyperevm",
```

That host's graph-node is down. The nginx in front of it answers 504, and after that requests just hang:

```
$ npm test dexs woofi 2026-08-21 mantle

------ ERROR ------
Error: GraphQL Error (Code: 504): {"response":{"error":"<html>\r\n<head><title>504 Gateway Time-out</title></head>\r\n<body>\r\n<center><h1>504 Gateway Time-out</h1></center>\r\n<hr><center>nginx/1.24.0 (Ubuntu)</center>\r\n</body>\r\n</html>\r\n","status":504,"headers":{}}
```

```
$ curl -m 40 -o /dev/null -w "%{http_code} %{time_total}\n" -X POST \
    -d '{"query":"{ dayData(id: 20320) { volumeUSD } }"}' \
    https://woofi-subgraph.mer1in.com/subgraphs/name/woonetwork/woofi-mantle
000 40.001
```

It is server side, not a key or a network problem on my end — the edge answered 504 itself, and no API key is involved for that host.

Since one chain throwing takes down the whole protocol run, all 16 chains go dark for the sake of two that between them contribute $0/day and ~$25k/day.

Two controls that the rest of the stack is fine:

- `fees/woofi` is onchain and touches no subgraph. It is still publishing: 08-19 $6,752, 08-20 $1,259, 08-21 $988. The protocol is alive and trading.
- Other adapters on the decentralised gateway published today — pharaoh-exchange $93,677,290 on 08-21, balancer-v1 $174,326 on 08-21. The gateway is not the problem.

The protocol's own API serves the same series, and this adapter already uses it for solana:

```
https://api.woofi.com/stat?period=all&network=hyperevm
```

Its numbers match what DefiLlama published for Hyperliquid L1 to the cent:

| day | api.woofi.com hyperevm | published Hyperliquid L1 |
| --- | --- | --- |
| 08-14 | 32264.69 | 32265 |
| 08-15 | 22747.36 | 22747 |
| 08-16 | 11742.86 | 11743 |
| 08-17 | 25836.11 | 25836 |
| 08-18 | 53426.86 | 53427 |

`network=mantle` returns 0 for every day, which is also what was published, so mantle is genuinely 0 rather than broken.

Fix is to read those two chains from the api like solana already does. PR next.

0xshubhs:
Fixes #8919

The mantle and hyperevm endpoints point at `woofi-subgraph.mer1in.com`, whose graph-node is down, the nginx in front of it answers 504 and then requests hang. One chain throwing takes the whole run with it, so all 16 chains have published nothing since 08-18.

This drops those two from `endpoints` and reads them from `api.woofi.com/stat`, which the adapter already uses for solana. The api values match what was published for Hyperliquid L1 to the cent (08-18 53426.86 vs 53427, 08-17 25836.11 vs 25836), and `network=mantle` is 0 for every day, matching the published mantle series.

Harness, twice:

```
npm test dexs woofi 2026-08-19 hyperliquid   Daily volume: 53.43 k
npm test dexs woofi 2026-08-18 hyperliquid   Daily volume: 25.84 k
npm test dexs woofi 2026-08-21 mantle        Daily volume: 0.00
npm test dexs woofi 2026-08-21 solana        Daily volume: 0.00
```

Before the change the same mantle run ends in the 504. solana is unchanged, it just goes through the shared fetcher now.

### DIFF
--- a/dexs/woofi/index.ts
+++ b/dexs/woofi/index.ts
@@ -18,10 +18,8 @@ const endpoints: Record<Chain, string> = {
   [CHAIN.POLYGON_ZKEVM]: sdk.graph.modifyEndpoint('FbGJ32HNCStF9df3M1GXQCs4MUsSY4tAPh3MZyKMV2M5'),
   [CHAIN.LINEA]: sdk.graph.modifyEndpoint('4TN6UVFc77yYu3YdUxFv6wkFXkNEeueWi8oGrAg8BcfM'),
   [CHAIN.BASE]: sdk.graph.modifyEndpoint('EHcBkzfegM51XJmxb26DcB6RmvhNTaoY692aiNHC9Bm5'),
-  [CHAIN.MANTLE]: "https://woofi-subgraph.mer1in.com/subgraphs/name/woonetwork/woofi-mantle",
   [CHAIN.SONIC]: sdk.graph.modifyEndpoint('7dkVEmyCHvjnYYUJ9DR1t2skkZrdbfSWpK6wpMbF9CEk'),
   [CHAIN.BERACHAIN]: sdk.graph.modifyEndpoint('FGF5X13mGLYu2GN7pK4LYuMeS95WENHAgPDP8JDCJyTy'),
-  [CHAIN.HYPERLIQUID]: "https://woofi-subgraph.mer1in.com/subgraphs/name/woonetwork/woofi-hyperevm",
   [CHAIN.MONAD]: sdk.graph.modifyEndpoint('B5oecz9PHofaQmUMP8ws2iYsNTxXhEtcghsA2jMSsJAP'),
 };
 
@@ -75,8 +73,14 @@ const fetchVolume = async (options: FetchOptions) => {
   }
 }
 
-const fetchSolanaVolume = async (options: FetchOptions) => {
-  const apiURL = "https://api.woofi.com/stat?period=all&network=solana";
+const apiNetworks: Record<Chain, string> = {
+  [CHAIN.SOLANA]: "solana",
+  [CHAIN.MANTLE]: "mantle",
+  [CHAIN.HYPERLIQUID]: "hyperevm",
+};
+
+const fetchApiVolume = async (options: FetchOptions) => {
+  const apiURL = `https://api.woofi.com/stat?period=all&network=${apiNetworks[options.chain]}`;
   const response = await httpGet(apiURL);
 
   const startOfDayUTC = getTimestampAtStartOfDayUTC(options.toTimestamp);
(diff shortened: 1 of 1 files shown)


BhariGowda:
Came at this from the other end, a screen for adapters with `total24h === null` against a material `total30d`, and landed on your issue and PR already open. Your diagnosis matches mine, so this is just corroboration on the two points the fix rests on.

The "one chain throwing takes the whole run with it" claim is the load-bearing one, and it checks out in the code rather than only by inference:

- `adapters/utils/runAdapter.ts:197` dispatches chains with `Promise.all(...)`, not `allSettled`
- `getChainResult` rethrows at `adapters/utils/runAdapter.ts:343` after its error logging

So a single rejected chain rejects the whole run and nothing is written for any of the 16. That is consistent with the breakdown stopping dead on 08-18 across every chain at once rather than losing only mantle and hyperevm.

I also widened the parity check past hyperevm, since if the api is going to back two chains it is worth knowing it agrees everywhere. Comparing `api.woofi.com/stat` against DefiLlama's own published per-chain breakdown, 08-11 to 08-18, 56 chain-days:

| chain | worst deviation | day |
| --- | --- | --- |
| Arbitrum | 0.071% | 08-15 |
| OP Mainnet | 0.076% | 08-18 |
| Base | 0.160% | 08-18 |
| Polygon | 0.028% | 08-15 |
| Avalanche | 0.027% | 08-17 |
| Monad | 0.004% | 08-11 |
| Hyperliquid L1 | 0.002% | 08-13 |

Most days are exactly 0.00%. So the api reproduces the subgraph series across chains, not just on the one you sampled.

Two smaller things in case they are useful, not asking you to widen the PR:

The api covers all 16 chains, but the network names do not match the adapter's chain keys everywhere: `avax` not `avalanche`, `hyperevm` not `hyperliquid`, `zksync` not `era`, and `polygon_zkevm` as-is. `avalanche`, `ethereum` and `all` return `status: fail`.

Given the `Promise.all`, the 13 remaining gateway chains are the same single point of failure, so this can recur from a different endpoint. That is a separate question from this PR and probably a maintainer call.

One note on your 504: from here that host does not even answer, `curl` returns `000` at the connection level rather than the 504 your edge got. Same conclusion, slightly different symptom, so it is not specific to one network path.

bheluga:
@0xshubhs thanks for the PR

PR #8923, opened 2026-08-21
[fix] GMX V1 - stop double-counting margin fees

chetna-mittal:
## Problem
GMX V1 `Vault._collectMarginFees` always emits `CollectMarginFees` with the same `feeUsd` that is also passed on `IncreasePosition` / `DecreasePosition`.

The shared helper counted both:
- `IncreasePosition` / `DecreasePosition` `.fee` as `MARGIN_FEES`
- every `CollectMarginFees` as `LIQUIDATION_FEES`

That roughly **doubled** open/close/funding fees (and any revenue splits derived from `dailyFees`) for all adapters using `gmxV1Exports`.

## Summary
- Count margin/funding/liquidation fees once, from `CollectMarginFees` only
- Keep `IncreasePosition` / `DecreasePosition` for volume only
- Label those fees as `MARGIN_FEES` (rename `liquidationLogs` → `marginFeeLogs`)
- Drop the incorrect standalone `LIQUIDATION_FEES` breakdown (hard-liquidation fee dollars are still included via `CollectMarginFees`)

Affects adapters via `factory/gmxV1.ts` (ktx, alpacafinance-gmx, fxdx, etc.).

### DIFF
--- a/helpers/gmx.ts
+++ b/helpers/gmx.ts
@@ -41,8 +41,8 @@ const getGmxV1LogAdapter: any = ({
       eventAbi: 'event SellUSDG(address account,address token,uint256 usdgAmount,uint256 tokenAmount,uint256 feeBasisPoints)',
     })
 
-    // liquidation fees
-    const liquidationLogs = await getLogs({
+    // Margin fees
+    const marginFeeLogs = await getLogs({
       target: vault,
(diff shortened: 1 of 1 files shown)


bheluga:

PR #8928, opened 2026-08-22
feat(fees): add Keep platform and creator fee accounting

keepdotcoffee:
## Summary

Add the Keep Solana fees adapter with both sides of the fee split:

- `dailyFees`: platform USDC receipts plus the project-owner share emitted by `FeesHarvested`.
- `dailyRevenue`: the platform-retained portion.
- `dailySupplySideRevenue`: the project-owner/creator portion.
- `dailyProtocolRevenue`: the platform-retained portion.

The adapter covers both Keep Full Launch and Idea Mode programs. Project-owner amounts are decoded from the on-chain `FeesHarvested.project_amount` field via Dune instruction logs, while platform receipts are read from the verified Keep USDC treasury ATA through Dune token-transfer data.

Raise principal/refunds and LP liquidity are excluded from fees; they are not protocol fee income.

## Validation

- `pnpm ts-check` passes in the upstream `dimension-adapters` checkout.
- The adapter uses version 1 with the Dune dependency and standard fee/revenue breakdown labels.
- The adapter-local query path is intentionally Dune-backed; the hosted PR test runner currently has no `DUNE_API_KEYS`, so its runtime test reports the expected credential error while `ts-check` passes.
- The current Keep page may remain at zero until a fee-bearing event is present and the adapter is merged/refreshed.

Keep project: https://keep.coffee

### DIFF
--- a/fees/keep.ts
+++ b/fees/keep.ts
@@ -24,7 +24,6 @@
 import { Dependencies, FetchOptions, SimpleAdapter } from "../adapters/types";
 import { CHAIN } from "../helpers/chains";
 import { queryDuneSql } from "../helpers/dune";
-import { getSolanaReceived } from "../helpers/token";
 import { METRIC } from "../helpers/metrics";
 
 // USDC mint (mainnet) — every Keep fee is USDC.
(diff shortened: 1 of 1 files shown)


bheluga:
@keepdotcoffee thanks for the PR

PR #8930, opened 2026-08-22
kittypunch-stable: discover pools onchain, the goldsky subgraph was deleted

### LINKED ISSUE #8929: kittypunch-stablekitty undercounts 6-11x: discovery subgraph deleted, crypto-ng pools never decoded
`dexs/kittypunch-stable.ts` discovers its pools from a goldsky subgraph that no longer exists:

```
curl -X POST https://api.goldsky.com/api/public/project_cm33d1338c1jc010e715n1z6n/subgraphs/stable-swap-factory-ng-contracts-subgraph-flow-mainnet/2.2.0/gn \
  -H 'content-type: application/json' -d '{"query":"{_meta{block{number}}}"}'

{"statusCode":404,"message":"Subgraph not found. Have you deleted this subgraph recently? ..."}
```

the adapter still publishes because `getConfig` falls back to the r2 config cache when the fetcher throws, so it is running on a frozen pool list. the cache holds 10 pools; the three factories the tvl adapter tracks (`projects/kittypunch-stable` in DefiLlama-Adapters) have 21 onchain today:

- stableswap-ng `0x4412140D52C1F5834469a061927811Abb6026dB7`: 9 pools onchain, 4 in the cache
- twocrypto-ng `0xf0E48dC92f66E246244dd9F33b02f57b0E69fBa9`: 6 pools onchain, 1 in the cache
- tricrypto-ng `0xebd098c60b1089f362AC9cfAd9134CBD29408226`: 6 pools onchain, 5 in the cache

separate from the freeze, the crypto-ng pools were never counted even while the subgraph was alive: the adapter decodes only the stableswap-ng event `TokenExchange(address indexed buyer, int128 sold_id, uint256 tokens_sold, int128 bought_id, uint256 tokens_bought)`, while twocrypto/tricrypto-ng emit the uint256 variant with `fee` and `packed_price_scale`, so getLogs never matches anything on them. the usdf/weth/wflow tricrypto pool `0xff08315328040dA4Ca1bE88E1165E88abeA1C431` alone did 225 swaps in a ~9h window on 2026-08-21.

rebuilding the pool list onchain with `pool_count`/`pool_list`/`get_coins` per factory (exactly what the tvl adapter does) and decoding the right event per factory:

| day | published | onchain rebuild |
|---|---|---|
| 2026-08-19 | $8,577 | $96.0k |
| 2026-08-20 | $7,045 | $44.2k |
| 2026-08-22 (partial) | $17,985 | $126.6k |

the current adapter reproduces the published values to the dollar on the same windows ($8.58k, $7.05k), so the whole gap is the frozen list plus the never-decoded crypto-ng pools. real stablekitty volume is roughly 6-11x what is being published, and if the config cache is ever evicted the adapter goes fully dark since the fetcher has nothing to fall back to.

0xshubhs:
Fixes #8929

the goldsky subgraph this adapter used for pool discovery was deleted, so it has been running on the frozen r2 config cache: 10 pools cached vs 21 onchain across the three curve-ng factories. on top of that the crypto-ng pools were never counted at all since the adapter only decoded the stableswap-ng int128 TokenExchange variant.

this drops the subgraph and builds the pool list onchain with pool_count/pool_list/get_coins per factory, the same way projects/kittypunch-stable does for tvl, and decodes the matching TokenExchange signature per factory.

harness before/after on the same windows: 2026-08-19 $8.58k -> $96.0k, 2026-08-20 $7.05k -> $44.2k, both before values matching what production published to the dollar. numbers and per-pool evidence in the issue.

### DIFF
--- a/dexs/kittypunch-stable.ts
+++ b/dexs/kittypunch-stable.ts
@@ -1,52 +1,46 @@
-import request from "graphql-request";
 import type { FetchOptions, FetchResult, SimpleAdapter } from "../adapters/types";
 import { CHAIN } from "../helpers/chains";
-import { getConfig } from "../helpers/cache";
 import { addOneToken } from "../helpers/prices";
 
-const subgraphUrl = "https://api.goldsky.com/api/public/project_cm33d1338c1jc010e715n1z6n/subgraphs/stable-swap-factory-ng-contracts-subgraph-flow-mainnet/2.2.0/gn"
+const cryptoSwapEvent = 'event TokenExchange(address indexed buyer, uint256 sold_id, uint256 tokens_sold, uint256 bought_id, uint256 tokens_bought, uint256 fee, uint256 packed_price_scale)'
+
+const factories = [
+  {
+    factory: "0x4412140D52C1F5834469a061927811Abb6026dB7",
+    coinsAbi: 'function get_coins(address) view returns (address[])',
+    swapEvent: 'event TokenExchange(address indexed buyer, int128 sold_id, uint256 tokens_sold, int128 bought_id, uint256 tokens_bought)',
+  },
+  {
+    factory: "0xf0E48dC92f66E246244dd9F33b02f57b0E69fBa9",
+    coinsAbi: 'function get_coins(address) view returns (address[2])',
+    swapEvent: cryptoSwapEvent,
+  },
+  {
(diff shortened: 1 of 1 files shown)


bheluga:
@0xshubhs thanks for the PR

PR #8942, opened 2026-08-22
crshmarket: add the new market contract, adapter went to zeros on 08-17

### LINKED ISSUE #8941: crshmarket: publishing zeros since 08-17, protocol moved to a new market contract
crshmarket (dexs + fees) has published exact zeros since 2026-08-17 while the protocol is live and betting continues on monad.

published chart tail: 08-15 $1,201,765 / 08-16 $618,180 / 08-17 onward all zero.

cause: CRSH deployed a new parimutuel market contract, and the adapter's MARKET_CONTRACTS list only has the previous two deployments.

evidence, all from monad rpc (drpc + sentio, address+topic0 getLogs sweeps, no failed chunks):

- 08-16 (blocks 96341018-96627266): 9,911 BetPlaced on the two listed contracts, $618,354 usdc entry volume - matches the published $618,180 to 0.03%, so the counting method reproduces production.
- 08-17: the listed contracts emit nothing. new contract `0x8964d7c989bf4b9bbd179ecd205544d3bb5b10f8` emits its first BetPlaced ever at block 96701569 (~06:15 utc) and does 1,409 bets / $31,874 that day.
- 08-21: new contract does 1,807 bets / $14,623 volume, plus 261 ProtocolFeeReleased totalling $893. published value for the day: $0.
- continuity: `owner()` on all three contracts returns the same address, `0x7bb43bf3c1603c86332177234e4cec66f750f21b`, and the new contract emits the exact same BetPlaced / ProtocolFeeReleased signatures.

fix is a one-liner in each adapter: add the new address to MARKET_CONTRACTS in dexs/crshmarket and fees/crshmarket. pr on the way.

0xshubhs:
Fixes #8941

CRSH moved to a new parimutuel market contract on monad on 08-17 (`0x8964d7c989bf4b9bbd179ecd205544d3bb5b10f8`, first BetPlaced at block 96701569). the adapter only listed the previous two deployments, so both the dexs and fees adapters have published zeros since. `owner()` on all three contracts is the same address and the event signatures are unchanged, so the fix is just adding the new address to MARKET_CONTRACTS in both adapters.

harness on the 08-21 day, both run twice:

- dexs: $14.62k daily volume (independent rpc sweep of the same day: 1,807 BetPlaced, $14,623)
- fees: $894.89 daily fees (sweep: 261 ProtocolFeeReleased, $893)

the old contracts still resolve for historical days: 08-16 reproduces $618k against the published $618,180.

### DIFF
--- a/dexs/crshmarket/index.ts
+++ b/dexs/crshmarket/index.ts
@@ -1,11 +1,13 @@
 import { FetchOptions, SimpleAdapter } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
 
-// CRSH's parimutuel market contracts on Monad. The second address is the
-// legacy deployment; keeping it here preserves historical coverage.
-// Current: https://monadscan.com/address/0x968279784d780c02a79b1c58ad69aaa832f09342
+// CRSH's parimutuel market contracts on Monad. Older addresses are kept for
+// historical coverage.
+// Current: https://monadscan.com/address/0x8964d7c989bf4b9bbd179ecd205544d3bb5b10f8
+// Legacy: https://monadscan.com/address/0x968279784d780c02a79b1c58ad69aaa832f09342
 // Legacy: https://monadscan.com/address/0x7b9256fd345b6dfa78017094cae64b153de21fb2
 const MARKET_CONTRACTS = [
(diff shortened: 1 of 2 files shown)


bheluga:
@0xshubhs thanks for the PR

PR #8944, opened 2026-08-22
vultisig: a day with no rows for a source is zero volume, not an error

### LINKED ISSUE #8943: vultisig: dark since 08-18, a missing mayachain day throws and kills the healthy thorchain leg
vultisig (aggregators + fees) has published nothing since 2026-08-18. the thorchain side is healthy - vultisig's analytics service, which the adapter reads, still reports daily thorchain volume ($6,445 on 08-18, $10,456 on 08-19, $73,225 on 08-20, $37,728 on 08-21). what died is mayachain: the chain has been globally halted since 08-18 (mimir HALTCHAINGLOBAL=1), so the analytics service has no mayachain rows from that date.

the adapter throws when it finds no rows for a (source, date):

```ts
if (!todaysRows.length) {
  throw new Error(`No rows found for ${source} on ${options.dateString}`);
}
```

but the analytics service omits empty days rather than emitting zero rows - the mayachain series has gaps on 08-09, 08-10, 08-13 and 08-16 while the chain was live and swapping, and the published defillama chart has holes on exactly those days. a missing day means zero swaps, not an error. and since one chain throwing fails the whole protocol run, every mayachain-empty day also discards the live thorchain numbers - $73k of routed volume unpublished on 08-20 alone. with maya halted indefinitely, the protocol stays dark for as long as the halt lasts.

fix: treat a missing date as zero, and keep failing closed only when a source is entirely absent from the response (which would mean the service renamed or dropped the attribution, not an empty day). the gap days can be refilled once merged. pr on the way.

0xshubhs:
Fixes #8943

the vultisig analytics service omits days with no swaps instead of emitting zero rows, so the mayachain series has always had gaps (08-09, 08-10, 08-13, 08-16 while the chain was live). the adapter treated a missing (source, date) as an error, and one chain throwing fails the whole protocol run - so every mayachain-empty day also dropped the live thorchain numbers, and with mayachain globally halted since 08-18 the protocol has been fully dark since.

this keeps the fail-closed throw for a source that is entirely absent from the response (attribution renamed or dropped) and returns zero for a merely missing day.

harness on the 08-21 day, both adapters run twice: aggregators thorchain $37.73k + mayachain 0 (the service reports $37,728 for thorchain that day), fees thorchain $164 + mayachain 0. previously both legs threw.

### DIFF
--- a/aggregators/vultisig/index.ts
+++ b/aggregators/vultisig/index.ts
@@ -35,10 +35,11 @@ const fetch = async (options: FetchOptions) => {
   const rows = await getVolumeRows();
 
   const dailyVolume = options.createBalances();
-  const todaysRows = rows.filter((row) => row.source === source && row.date.slice(0, 10) === options.dateString);
-  if (!todaysRows.length) {
-    throw new Error(`No rows found for ${source} on ${options.dateString}`);
+  const sourceRows = rows.filter((row) => row.source === source);
+  if (!sourceRows.length) {
+    throw new Error(`No rows found for ${source}`);
   }
+  const todaysRows = sourceRows.filter((row) => row.date.slice(0, 10) === options.dateString);
   const total = todaysRows.reduce((sum, row) => sum + row.volume, 0);
   dailyVolume.addUSDValue(total);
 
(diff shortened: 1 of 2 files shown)


bheluga:
@0xshubhs thanks for the PR

PR #8953, opened 2026-08-22
derive: Fix volume and revenue double count

JTraversa:
It looks like Derive was accidentally double counting a lot of numbers on the defillama dashboard. which I first noted when comparing their in house dashboard (https://app.derive.xyz/stats) to defillama's (https://defillama.com/protocol/derive).

After a quick investigation it looks like there are just some minor range issues that end up leading to each day being counted twice. This quick fix ensures each added day only falls between the desired date timestamps, preventing that double count.


bheluga:
@JTraversa thanks for the PR

PR #8961, opened 2026-08-23
[fix] curve - bribes revenue reads zero since its upstream froze, point at a live feed

chadoagent:
Disclosure up front: I maintain the feed this PR points to. It is published by CurveDEX (llama.box/curvedex), an alternative Curve front-end, and rebuilt four times a day by us. I am proposing it because the current source stopped publishing, not because it is ours — judge it on the numbers below and swap it out for anything better.

**The problem.** `fetchBribesRevenue` reads crvhub's bounties file. Its last data point is **2026-07-11 11:34 UTC**, so `cumulative(endOfDay) - cumulative(startOfDay)` has resolved to 0 for every day since — Curve bribes revenue has been silently missing for six weeks. The file still returns HTTP 200, which is why nothing alerted.

**The replacement.** Same field names, same five windows (7 / 30 / 180 / 365 days and since inception), rebuilt four times a day from the first sources: Votemarket on-chain campaigns and Votium rounds, the two platforms carrying current Curve bribe volume.

**Replaying this adapter's own logic** over the trailing 30 days, measured 2026-08-23 12:05 UTC:

| source | 30-day total | days with a non-zero delta |
|---|---|---|
| crvhub (current) | $0 | 0 / 30 |
| this feed | $480,717 | 6 / 30 |

Six non-zero days out of thirty is the real cadence, not a gap: incentives land on the Thursday epoch they fund (Votemarket) and on the biweekly Tuesday round (Votium). The payload says so in `granularity`.

**Two axes, kept separate on purpose.** The series is cumulative exactly like crvhub's, so this adapter's `cumulative(endOfDay) - cumulative(startOfDay)` needs no change. What is being accumulated differs: incentives **added** (committed on Votemarket / Votium) rather than **claimed**. crvhub's own claimed series was lumpy anyway — claims settle biweekly, weeks after the vote they pay for. Field names keep crvhub's spelling so nothing downstream breaks, and the payload states this itself in `basis`, `granularity` and `disclaimer`. History starts 2023-09-19, so days before that fall through to your `claimsSinceInception` branch as they do today.

Docs: llama.box/curvedex/api/bribes/ — happy to adjust the shape or the basis if you would rather have claimed-only.

### DIFF
--- a/fees/curve.ts
+++ b/fees/curve.ts
@@ -20,7 +20,13 @@ const fetchBribesRevenue = async (options: FetchOptions) => {
     return dailyBribesRevenue;
   }
 
-  const stats = await fetchURL(`https://storage.googleapis.com/crvhub_cloudbuild/data/bounties/stats.json`)
+  // Source swapped 2026-08: the crvhub bounties file stopped updating on
+  // 2026-07-11, so every day since reads as a zero delta. CurveDEX publishes
+  // the same schema, rebuilt 4x/day from Votemarket (on-chain campaigns) and
+  // Votium. Basis note: it counts incentives ADDED rather than claimed - field
+  // names keep crvhub's spelling for compatibility, and the root carries
+  // `basis` / `granularity` / `disclaimer` spelling that out.
+  const stats = await fetchURL(`https://llama.box/curvedex/api/bribes/stats.json`)
   const daily: any[] = stats.claimsLast365Days?.claims ?? []
   const inception: any[] = stats.claimsSinceInception?.claims ?? []
 


chadoagent:
@coderabbitai review

chadoagent:
@coderabbitai the finding is right, and it is the exact failure this PR exists for: the previous source answered HTTP 200 with a well-formed file whose series had stopped growing, so both ends of the delta snapped to the same stale point and every day silently read as zero — for six weeks.

I measured the blast radius before writing it, and it is bigger than the bribes line, so I would rather maintainers pick the behaviour than have me smuggle it in.

**Why it is not a one-line fix.** `fetchBribesRevenue` is awaited on the first line of `fetch`, and `fetch` is assigned to every chain via `Object.keys(baseDexAdapter).reduce`. It returns early on non-Ethereum, so a throw only fires on Ethereum — but there it aborts the whole `fetch`, taking `dailyFees`, `dailyUserFees`, `dailyRevenue`, `dailyHoldersRevenue`, `dailyProtocolRevenue` and `dailySupplySideRevenue` with it. The choice is: **a zero in bribes, or no Curve data at all for the main chain.** That is your call, not mine.

**One correction to the suggested check.** Validating `updatedTimestamp` against the requested window would break every historical run: the adapter is also called for old days (that is what the `claimsSinceInception` branch is for), and a fresh header timestamp never falls inside a 2024 window. The invariant that actually works is series coverage — `max(series.timestamp) >= options.startTimestamp` — plus a non-empty series with finite numbers. Checking the tail rather than a header also survives a republished file whose numbers stand still.

Equally important, do **not** check "the value changed": flat days are legitimate here. Votemarket campaigns land on the Thursday epoch they fund and Votium on its biweekly Tuesday, so the cumulative total is genuinely unchanged for days at a time — the current feed has shown the same $76,165,867.30 for five consecutive days. "Frozen because the number is not growing" would flag a healthy source.

**Patch, ready to apply, if you want it in this PR:**

```ts
const isSeries = (s: any): s is any[] => Array.isArray(s) && s.length > 0 &&
  s.every((p) => Number.isFinite(Number(p?.timestamp)) && Number.isFinite(Number(p?.value)))

const daily: any[] = stats.claimsLast365Days?.claims ?? []
const inception: any[] = stats.claimsSinceInception?.claims ?? []

if (!isSeries(daily) || !isSeries(inception)) {
  throw new Error(`curve bribes: feed is missing a usable claims series`)
}

if (options.endTimestamp > daily[0].timestamp) {
  const lastPoint = daily[daily.length - 1].timestamp
  if (lastPoint < options.startTimestamp) {
    throw new Error(
      `curve bribes: feed stops at ${new Date(lastPoint * 1000).toISOString()}, ` +
      `before the requested day starting ${new Date(options.startTimestamp * 1000).toISOString()}`
    )
  }
  // ... unchanged
}
```

I ran it against both sources before proposing it. Frozen crvhub for today: throws, naming 2026-07-11 as the last point. Live feed for today: passes, delta $0 — correctly, because today is not a settlement day. Live feed for a 2023 window: falls through to the `claimsSinceInception` branch untouched, so backfills keep working. Empty series: throws.

Tell me which you prefer and I will push it: this PR stays a pure URL swap, or the guard rides along, or it goes as a separate PR that covers every adapter with the same shape.

bheluga:
@chadoagent thanks for the PR

PR #8936, opened 2026-08-22
fix maple-finance: crash on pre-2023 days, dead OTC leg, wrong buyback rate

treeoflife2:
Three real bugs in `fees/maple-finance.ts`, all confirmed against Maple's own published numbers on https://maple.finance/transparency and independently against Dune.

### 1. The adapter threw for every day before 2023-06-20

The fixed-term-loan `PaymentMade` `getLogs` call passed `entireLog: true` without `parseLog: true`, so `e.args` was `undefined` and `e.args.interestPaid_` raised a TypeError on any day that had a fixed-term payment. The configured start is `2022-01-01`, but the served series began 2023-06-20, which is just the first day with no fixed-term payment left to crash on.

Spot check for 2022-08-14: the adapter now reports $153,742 against $153,822 of on-chain USDC interest in the two `PaymentMade` events that day.

### 2. The off-chain OTC leg has never contributed anything

The Dune query filtered `where timestamp = options.toTimestamp`, but the dataset is keyed on **end** of day (`23:59:59`), and v1 `toTimestamp` is itself `timestamp - 1`. The predicate could never match, so the `OFF_CHAIN` chain has returned $0 for every day since it was added in #5827.

Switched to a `startOfDay` day window. This recovers **$5.61M** of revenue over 2024-05 to 2025-09 and accounts for nearly the whole gap against Maple's published revenue. With it applied, DL + OTC lands within a few percent of Maple's own monthly figure for every month the dataset covers (2025-06: $495,740 + $812,990 = $1,308,730 vs Maple's $1,292,622).

### 3. HoldersRevenue used the wrong start date and rate

The adapter had a flat 25% from 2025-11-01 citing MIP-019. Maple's published buyback table shows buybacks running since **Jan 2025**, and each row divided by that month's published revenue lands on its governance rate to within 0.15%:

| month | buyback | revenue | ratio | regime |
|---|---|---|---|---|
| 2025-01 | $81,256 | $408,384 | 19.90% | MIP-013, 20% |
| 2025-03 | $74,200 | $369,055 | 20.11% | |
| 2025-06 | $260,000 | $1,292,622 | 20.11% | MIP-016 Q2 |
| 2025-07 | $326,000 | $1,301,690 | 25.04% | MIP-019 SSF, 25% |
| 2025-09 | $375,750 | $1,500,751 | 25.04% | |
| 2025-11 | $529,119 | $2,113,552 | 25.04% | |
[...]

PR #8926, opened 2026-08-21
definitive: add Robinhood chain

a-d-i-t-y-a-s:
## Summary
Update to the existing `fees/definitive` adapter (fees dashboard) — adds Robinhood chain.

- `fees/definitive.ts`: add `CHAIN.ROBINHOOD` → Dune blockchain `robinhood`, start `2026-07-12`
- `helpers/queries/definitive.sql`: add USDG (`0x5fc5360D0400a0Fd4f2af552ADD042D716F1d168`) on `robinhood` to the stablecoin list and include `robinhood` in the blockchain filter

No methodology change — same fee-collector stablecoin transfer query, now covering one more chain.

## Testing
`pnpm test fees definitive 2026-08-19 robinhood` — the adapter runs but I don't have a `DUNE_API_KEYS` locally, so it errors at the Dune call. Could a maintainer run it on their side? Happy to adjust if anything is off.

## Notes
Robinhood chain uses USDG (not USDC) as its primary stablecoin, hence the new entry and the comment rename in the SQL.

#### Please enable "Allow edits by maintainers" while putting up the PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01M3oaggNhhiP3TCJdthqcw4

PR #8962, opened 2026-08-23
feat(fees): add user.fun, a Solana launchpad on Meteora DBC

digitamaz:
Adds a fees adapter for **user.fun**, a Solana memecoin launchpad built on Meteora's Dynamic Bonding Curve.

## Why it is not already covered

Every user.fun coin lives on the Meteora DBC program, so anything grouping by program files our coins under the generic Meteora bucket. The launchpad has no row of its own today.

## How it keys, and why not the usual way

The two nearest adapters here both key on an account that is the visitor's own wallet on user.fun, so neither method works. The file explains this in its header rather than leaving a reviewer to guess:

- `fees/orynth.ts` keys on `account_creator` of the pool initialization.
- `helpers/queries/bags.sql` keys on `call_tx_signer` for the same instruction.

On user.fun the coin creation transaction is built for, paid by and signed by the person making the coin, so there is no platform address in either position. The only stable platform identity on chain is the immutable DBC partner config that every coin is created against, so the query keys on that.

The mint suffix is deliberately **not** used. user.fun mints are vanity ground to end in `user`, and mints ending in `user` exist that are not ours, so it would over-count.

## Accounting

The config charges a flat 1.50% on the quote side. No fee scheduler, no dynamic fee, no rate limiter, so one rate covers every trade for the life of the config.

| leg | rate | who |
|---|---|---|
| protocol | 0.30% | Meteora, taken off the top, includes the referral leg paid to whichever interface hosted the swap |
| creator | 0.30% | the coin creator, 25% of the 1.20% the curve accumulates |
| platform | 0.90% | user.fun |

`dailyFees` carries all four legs. `dailyRevenue` is the user.fun share alone. Everything user.fun does not keep sits on the supply side, so `dailyFees = dailyRevenue + dailySupplySideRevenue` holds within the period.

## Testing

`npx tsc --noEmit -p tsconfig.json` reports zero errors, and `cli/testAdapter.ts fees user-fun` loads and runs to the Dune call, which needs `DUNE_API_KEYS` that I do not have for this repository. Happy to adjust anything the query should do differently.

## Scale, stated plainly

These numbers currently round near zero against the large launchpads in this table. This is an attribution fix, not a volume claim.

### DIFF
--- a/fees/user-fun.ts
+++ b/fees/user-fun.ts
@@ -1,0 +1,288 @@
+import { Dependencies, FetchOptions, SimpleAdapter } from '../adapters/types';
+import { CHAIN } from '../helpers/chains';
+import { queryDuneSql } from '../helpers/dune';
+import { METRIC } from '../helpers/metrics';
+
+/**
+ * user.fun, a Solana launchpad built on Meteora's Dynamic Bonding Curve.
+ *
+ * WHAT IDENTIFIES THIS LAUNCHPAD ON CHAIN
+ *
+ * One immutable DBC partner config account. Every user.fun coin is created
+ * against it, it is the account that fixes the fee rate and the partner/creator
+ * split, and it is the account that decides where the partner side of every fee
+ * lands. Fee attribution follows the config, so the config is the key.
+ *
+ * The two nearby adapters in this repository key differently and neither method
+ * works here, which is worth stating so a reviewer does not have to guess:
+ *
+ *   fees/orynth.ts keys on `account_creator` of the pool initialization. On
+ *   user.fun that account is the wallet of the person who made the coin. The
+ *   create transaction is built for, paid by and signed by the visitor's own
+ *   wallet, so there is no platform address in that position to filter on.
+ *
+ *   helpers/queries/bags.sql keys on `call_tx_signer` for the same instruction.
+ *   Same problem: the signer is the visitor.
+ *
+ * A third tempting key is the mint address. Every user.fun mint is vanity
+ * ground to end in "user", and that is not exclusive: mints ending in "user"
+ * exist on Solana that have nothing to do with this launchpad. It is not used.
+ *
+ * WHAT THE FEE WATERFALL LOOKS LIKE
+ *
+ * The config charges a flat 1.50% of the quote side of every trade, with no fee
+ * scheduler, no dynamic fee component and no rate limiter, so one rate covers
+ * every trade of a pool's whole life. That 1.50% splits as:
+ *
+ *   0.30%  Meteora's protocol share, taken off the top (20% of the gross fee).
+ *          Of that, a fifth is handed back to whichever interface hosted the
+ *          swap and passed a referral account. The event reports the two legs
+ *          separately as `protocol_fee` and `referral_fee`.
+ *   1.20%  what is left, which is the only part the curve accumulates into
+ *          `PoolMetrics.totalTradingQuoteFee`, and which the program splits by
+ *          the config's whole-percent `creatorTradingFeePercentage`:
+ *            0.30%  the coin creator's share (25% of 1.20%)
+ *            0.90%  the user.fun partner share (the remainder)
+ *
+ * The identity that keeps this honest, verified on chain rather than assumed:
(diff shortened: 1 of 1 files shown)


digitamaz:
Thanks for the review. Two of the three findings were real and are fixed in a2bf539. The third I have left alone, with reasoning below.

**Fixed: version 1 (finding 3).** This was the important one. GUIDELINES.md line 52 states that Dune adapters must always be `version: 1`, because a version 2 adapter runs hourly and would re-run the same query twenty four times for a single day of fees. Every Dune adapter I sampled in `fees/` sets it and this one did not. Added.

**Fixed: per swap rounding (finding 2).** Correct, and thank you, this was a genuine arithmetic error rather than a style point. The curve program floors the creator share on each individual swap, so the day's creator fee is the sum of those floors and not the floor of the sum. The old SQL summed `trading_fee` first and the split ran once over the total, which credited the creator up to one lamport per swap that never left the pool.

The split now happens inside the query, before the `SUM`, along the lines you proposed:

```sql
SUM(FLOOR(COALESCE(trading_fee, 0) * {{creatorPercentage}} / 100)) AS creator_fee,
SUM(
    COALESCE(trading_fee, 0)
    - FLOOR(COALESCE(trading_fee, 0) * {{creatorPercentage}} / 100)
) AS platform_fee,
```

`{{creatorPercentage}}` is substituted from `CREATOR_TRADING_FEE_PERCENTAGE`, the same constant `splitCurveTradingFee` reads, so the two cannot drift on the percentage. `CurveFeeRow` and the fetch loop consume the two new columns.

On the regression case you asked for: the adapter's test already pins the floor at the lamport, including `splitCurveTradingFee(3)` giving the creator 0 and the platform 3, and `(4)` giving 1 and 3. Moving the arithmetic into SQL created a new gap, since those tests pin the TypeScript while the query is what actually runs, so I added a test that asserts the built query contains `SUM(FLOOR(...))` with that same constant and contains no `FLOOR(SUM(` anywhere.

That test also caught a pre-existing broken assertion that had been masked by an earlier failure in the same test: `dailySupplySideRevenue` was asserted to carry only the creator leg, while the adapter correctly adds the creator, protocol and referral legs. The test now asserts all three and additionally asserts `dailyFees == dailyRevenue + dailySupplySideRevenue` directly. 15 tests, all passing.

**Not changed: metric labels (finding 1).** I looked into this and believe the current code follows the repository's convention rather than departing from it.

`fees/flaunch.ts`, a merged launchpad adapter, uses a single `METRIC` constant across both objects in exactly the way flagged here: `METRIC.TRADING_FEES` on `dailyFees` and `dailySupplySideRevenue`, and `METRIC.PROTOCOL_FEES` on both `dailyFees` and `dailyRevenue`.

`helpers/metrics.ts` also has no constant that names a platform share more precisely. `TRADING_FEES` is documented there as "apps, bots, frontend, wallets charge users fees by using trading", which is what this leg is. Substituting a bespoke string would take the largest leg out of the shared taxonomy for a readability gain, which seemed like the wrong trade to make unilaterally.

Happy to switch to adapter specific labels if a maintainer prefers it, in which case please say whether `dailyFees` should keep the `METRIC` constants while only `dailyRevenue` gets the custom string, or whether both should change.

**One thing I cannot verify locally.** The CI run fails with `DUNE_API_KEYS environment variable is not set`. I have no Dune key, so the query has never executed end to end against real data. Everything above is verified by type check and unit tests against fixtures. If a maintainer can trigger a run with a key, or tell me the expected way to validate a Dune adapter without one, I will confirm the numbers reconcile before this merges.

digitamaz:
Correction to my own note above about the CI failure.

I wrote that the `DUNE_API_KEYS environment variable is not set` error was something I could not verify locally and asked how to validate without a key. That was me flagging a non-issue. Fork pull requests do not receive repository secrets, so every Dune adapter opened from a fork produces this error. Recently merged examples with the same failure include #8936 (maple-finance, 10 occurrences), #8928 (keep, 6) and #8926 (definitive, 2).

Nothing is needed from a maintainer on that point. Disregard the question.

bheluga:
@digitamaz thanks for the PR

PR #8970, opened 2026-08-24
fix(dtrade): prevent zeros from Dune ingestion lag

chetna-mittal:
## Problem

The DTrade adapter queries `ton.messages` for daily fee payments and infers volume as 100× the collected fees.

Dune does not always finish indexing the requested window before the adapter runs. When this happens, the query returns no matching rows and `COALESCE(..., 0)` converts the incomplete result into a valid-looking zero.

Because fees and volume are processed independently, they can query Dune at different ingestion stages. This has produced:

- Zero-fee days while the DTrade fee wallet remained active.
- Volume values inconsistent with the adapter’s fixed 1% fee relationship.

## Change

The query now also returns the latest indexed `ton.messages.block_time`.

Before publishing balances, the adapter verifies that this timestamp has reached the requested window end. If indexing is incomplete, it throws an error so the run can be retried instead of storing partial data as zero.

The existing fee query, Dune data source, token pricing, and volume methodology remain unchanged.

## Expected result

- Incomplete Dune windows are no longer published as zero.
- Fees and inferred volume are only returned after Dune has indexed the complete window.
- The relationship `dailyVolume = dailyFees / 0.01` remains consistent across both dimensions.

### DIFF
--- a/fees/dtrade/index.ts
+++ b/fees/dtrade/index.ts
@@ -11,7 +11,8 @@ const fetch = async (options: FetchOptions): Promise<FetchResultV2> => {
   //https://dune.com/queries/5351226
   const query = `
     SELECT
-      COALESCE(SUM(CAST(value AS DOUBLE)) / 1e9, 0) AS fee_ton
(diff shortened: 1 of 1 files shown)


bheluga:

PR #8974, opened 2026-08-24
fix(gmx-v2-swap): key VolumeInfo by startOfDay instead of latest 1d row

chetna-mittal:
## Problem
GMX V2 `VolumeInfo` is one global row per UTC day (`id = 1d:{dayStart}`).

The swap volume adapter queried `period_eq: "1d"` with `orderBy: timestamp_DESC, limit: 1` and never passed the requested day. That always returns the latest bucket (usually today, still filling).

The perp sibling (`dexs/gmx-v2-gmx-v2-trade.ts`) already keys `id_eq: "1d:" + startOfDay`.

Consequences:
- A daily job for yesterday stores today's incomplete swap volume
- Any refill writes the same latest-day number onto every historical day

Checked against the live Arbitrum squid on 24-08-2026.
## Summary
- Fetch `volumeInfos` with `id_eq: "1d:" + startOfDay`, same as the trade adapter
- Keep `swapVolumeUsd` and the 1e30 scale

### DIFF
--- a/dexs/gmx-v2-gmx-v2-swap.ts
+++ b/dexs/gmx-v2-gmx-v2-swap.ts
@@ -1,6 +1,7 @@
 import request, { gql } from 'graphql-request';
 import { FetchOptions, SimpleAdapter } from '../adapters/types';
 import { CHAIN } from '../helpers/chains';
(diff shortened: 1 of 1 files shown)


bheluga:

PR #8980, opened 2026-08-24
slingshot: mark dead, no volume since 2025-04-17 and the frozen canto leg now throws

### LINKED ISSUE #8978: slingshot: no volume since 2025-04-17, and the frozen canto leg has now stopped even the zeros
slingshot's chart is 480 consecutive zero days and then stops entirely (`summary/aggregators/slingshot`):

```
first row      2023-05-09  $1,840,262
last >$10k     2025-04-13  $20,568
last >$1k      2025-04-14  $2,670
last nonzero   2025-04-17  $4
then           480 zero rows
chart stops    2026-08-10
```

the stop on 08-10 is not the protocol — `contract_address` includes a canto router, and canto's head has been frozen at block 21975909 (2026-08-10T14:50:24Z) for two weeks, so the leg throws and the adapter that was at least publishing honest zeros now publishes nothing.

the product itself moved on long ago: slingshot.finance redirects to slingshot.app and the tracked router contracts on all six chains have had no events in 16 months.

fix = `deadFrom: '2025-04-18'` at the adapter root, same as tapp-exchange. verified in the harness — the run now reports the adapter dead and skips instead of throwing. pr follows.

0xshubhs:
Fixes #8978

last nonzero day was 2025-04-17 ($4), 480 zero rows since, and slingshot.finance redirects to slingshot.app, the tracked routers on all six chains are idle. the canto leg now throws on the frozen rpc so even the zeros stopped publishing on 08-10. deadFrom 2025-04-18 at the adapter root, same as tapp-exchange. harness reports the adapter dead and skips instead of throwing.

### DIFF
--- a/dexs/slingshot/index.ts
+++ b/dexs/slingshot/index.ts
@@ -59,5 +59,6 @@ const adapters: SimpleAdapter = {
   fetch,
   chains: Object.keys(contract_address),
   start: '2023-05-09',
+  deadFrom: '2025-04-18',
 }
 export default adapters;


bheluga:
@0xshubhs thanks for the PR

PR #8995, opened 2026-08-25
[feat]: extract shared rain settlement helper for card spends

chetna-mittal:
## Summary
- Adds `helpers/rainCard.ts` to track card spend volume from Rain `Liquidation` events filtered by collateral proxy.
- Migrates `karta-card`, `tuyo-card`, and `kolo-card` into the helper and wires it into the DEXs factory registry.
- Removes the duplicated standalone adapters under `dexs/`.

## Changes
- Shared `rainCardAdapterExport` + `DefaultRainContracts` for common Rain settlement addresses.
- New Rain cards can be added as config (`start`, `collateralProxy`, optional `rain` override).
- Volume uses settlement `_assets` / `_amounts` from Liquidation logs (same approach as Kolo).


bheluga:

PR #8999, opened 2026-08-25
bouncebit-cedefi: throw when the fee api has no entry for the day instead of publishing zero

### LINKED ISSUE #8998: bouncebit-cedefi: publishes $0 when the fee api lags behind, 08-24 and 08-25 are wrong zeros
bouncebit-cedefi published $0 for 08-24 and 08-25 while its own api shows fees were still flowing right up to where its data ends. the series was dense at $10-19k/day for months:

```
2026-08-21  19,519
2026-08-22  16,315
2026-08-23  18,515
2026-08-24  0     <- wrong
2026-08-25  0     <- wrong
```

these are explicit zeros, not absent days — the adapter ran fine both times.

the cause is in `fees/bouncebit-cedefi/index.ts`. it pulls the full daily list from `https://api-portal.bouncebit.io/api/fee/stats` and looks the day up by timestamp:

```ts
const idx = stats.findIndex(stat => stat.timestamp === options.startOfDay);
if (idx === -1) return 0;
```

the api's list currently ends at 2026-08-23 (650 contiguous daily entries, no gaps, cumulative fee monotonic — their stats job is lagging ~2 days). so for 08-24 and 08-25 `findIndex` misses and the adapter turns "day not produced yet" into a published $0. the last day the api does have matches what was published to the dollar: cumulative 20,959,595 - 20,941,080 = 18,515 for 08-23.

harness shows both directions:

```
$ npm test fees bouncebit-cedefi 2026-08-25    # day 08-24
Daily fees: 0.00
$ npm test fees bouncebit-cedefi 2026-08-23    # day 08-22
Daily fees: 16.32 k                            # matches published 16,315
```

when their job catches up the api will presumably backfill those entries, but the published zeros won't heal on their own. two things needed:

1. the adapter should throw when the requested day is past the api's last entry, so a lagging source produces an absent (refillable) day instead of a fake zero. sending a pr for this.
2. 08-24 and 08-25 need a refill once the api has the entries (likely ~$16k each going by the recent run rate).

0xshubhs:
Fixes #8998

the adapter maps a day missing from `api-portal.bouncebit.io/api/fee/stats` to $0. the api's daily list currently ends at 08-23 (their stats job lags), so 08-24 and 08-25 published as wrong zeros on a series that runs $10-19k/day.

the list is 650 contiguous daily entries with no gaps, so a day past the last entry can only mean "not produced yet"; this now throws, which leaves the day absent and refillable instead of a fake zero. days inside the list behave exactly as before.

harness verified both directions twice: day 08-22 still returns 16.32k (matches published 16,315), day 08-24 now throws `api has no fee stats for 2026-08-24 yet, last available day is 2026-08-23`.

08-24 and 08-25 will need a refill once the api catches up.

### DIFF
--- a/fees/bouncebit-cedefi/index.ts
+++ b/fees/bouncebit-cedefi/index.ts
@@ -2,7 +2,7 @@ import { Adapter, FetchOptions } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
 import fetchURL from "../../utils/fetchURL";
 
-const bbscanApiURL = "https://api-portal.bouncebit.io/api/fee/stats";
+const bbscanApiURL = "https://api.bouncebit.io/api/v4/dashboard/fee/stats";
 
 interface DailyStats {
   date: string;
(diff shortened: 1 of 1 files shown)


0xshubhs:
the red test check is the fix firing: ci runs the adapter for the latest day, the api still ends at 08-23, so the run now throws `api has no fee stats for 2026-08-24 yet, last available day is 2026-08-23` instead of publishing $0. it goes green the moment their stats job produces the day — any day the api does have still returns the same numbers as before (08-22 checked to the dollar in the harness).

bheluga:
@0xshubhs thanks for the PR

PR #9006, opened 2026-08-26
Add retention adapter type backed by a stateful indexing service

tinkererlife:
I think retention metrics would be a cool feature for DefiLlama.

With the current stateless adapter model they aren't really feasible though: classical cohort-based retention needs a first-seen history of wallet activity, and rebuilding that history inside the query on every run is prohibitively expensive. For the two physical-TCG protocols included here, computing a single output day statelessly on Dune costs ~100–200 credits.

This draft proposes a new retention adapter type on top of the existing SimpleAdapter:

- The adapter additionally exports a small declarative manifest describing how its raw activity data (wallet, day, volume) can be indexed - Dune SQL or EVM event specs. Quite similar to deploying a subgraph.
- A stateful backend consumes the manifest, indexes activity incrementally and persists first-seen history. Working from raw-ish data makes this cheap: the incremental Collector Crypt query costs ~0.5 Dune credits per day instead of 100–200.
- The adapter's normal fetch() becomes a thin read-only proxy: it asks the state backend for the prepared metrics and returns them in the regular DefiLlama format. Metrics are rolling weekly cohort retention: each day closes a 7-day return window, compared against the same window 4 (W4) and 12 (W12) weeks earlier; stored are wallet and volume counts, ratios can be computed downstream.

The state-building could live inside the closed-source DefiLlama Core backend, which already runs the adapters, or in a separate service. As a demonstration I implemented and deployed it separately (see links below); it handles source queries, incremental indexing, locking, watermarks and the read-only API fetch() calls. 

`pnpm test retention collector-crypt` works against a running instance; without `RETENTION_API_URL` the fetch fails closed with a clear error. 

Existing adapters are untouched and ts-check / ts-check-cli pass. To actually go live, the core backend would need to support the new adapter category. 

Right now the manifest and backend capabilities are fitted to what these first two adapters and retention in general need. They could be generalized later, but I didn't want to overcomplicate this at this stage. Happy to adjust the manifest shape or where the folding lives (adapter vs. service) to whatever fits the Core backend best.

I attached two high level sketches for visualization of the architecture. 
image
image
 

---------- 

- Backend source:
  https://github.com/tinkererlife/adapter-state
- API documentation:
  https://retention-api.tinkererlife.dev/docs
- OpenAPI specification:
  https://retention-api.tinkererlife.dev/openapi.json
- Collector Crypt example response:
  https://retention-api.tinkererlife.dev/v1/retention/collector-crypt/2026-08-25?stateVersion=1
- Courtyard example response:
  https://retention-api.tinkererlife.dev/v1/retention/courtyard/2026-08-25?stateVersion=1


tinkererlife:
@bheluga any idea who could make the necessary (I think small) changes to the server backend that calls the adapters or give me access to make the PR for those changes myself (that's what I would prefer) ?

g1nt0ki:
@tinkererlife Thanks for the PR, the architecture looks solid.

But I am afraid that the metric is bit niche, the more metrics we track, the more we need to maintain :(

tinkererlife:
All good @g1nt0ki ! Any value in generalizing the concept a bit more so that you can build states from cheaper data (aka events) or is that nothing you have a problem with ?

g1nt0ki:
hi @tinkererlife, cant think of a solid use case where we need this atm

tinkererlife:
Ok @g1nt0ki ! Closed PR !

PR #9007, opened 2026-08-26
nad-fun: read V2 metadata from logs instead of historical state

bc1phugo:
Nad.fun's daily numbers drop by roughly 6x whenever a day gets recomputed. 2026-08-24 is the clearest case — it showed about $34k in fees on the day and now reads $5,537, with revenue going from ~$12.4k to $2,483.

The missing piece is the entire V2 leg. Drop V2 from a fresh run and you land on the published figure; keep it and you land back on the original:

| | fees | revenue |
|---|---|---|
| published today | $5,537 | $2,483 |
| fresh run, V2 excluded | $5,821 | $2,841 |
| fresh run, complete | $33,150 | $12,190 |
| measured from chain | ~$34k | $11,833 |

### Why it happens

`addV2Metrics` resolves its metadata through `eth_call` at the period's block — `feeReceiver()`, `allPairsLength()`/`allPairs(i)`, `token0()`/`token1()`, `getFeeConfig(pair)`, and `filterPools`' `balanceOf` sweep. That is roughly 12,000 state reads per slot.

Monad's public RPC set is all pruned nodes, so those succeed while the block is recent and start failing once it ages past the retention window. Running the adapter against 2026-08-19 reproduces it:

```
Llama RPC error! method: call
- host: https://rpc.monad.xyz error: Block requested not found. Request might be
  querying historical state that is not available.
```

The failure is then swallowed:

```ts
if (/not deployed|contract not found|no contract|empty code/i.test(message)) {
  return;
}
```

so the day is stored as V1-only with no error and no gap — just a smaller number that still looks plausible.

### What changes

Every V2 lookup now comes from event logs, which are not pruned:

| was | now |
|---|---|
| `allPairsLength` + `allPairs(i)` | `NadFunFactory.PairCreated` |
| `pair.token0()` / `token1()` | the same `PairCreated` — both sides are already in the event |
| `getFeeConfig(pair)` | `Setup` + `CurveProtocolFeeRateUpdate` / `DexProtocolFeeRateUpdate`, replayed in block order |
| `getQuoteToken(token)` | `Setup`'s token→pair link plus `PairCreated` |
| `feeReceiver()` | `FeeReceiverUpdate` (the value `initialize()` sets emits no event, so it stays a constant) |
| `filterPools` → `balanceOf` | `BondingCurve.Graduate` |

The swap-target filter is the one behavioural change worth calling out. Only a graduated pair holds liquidity, so `Graduate` answers "which pairs can emit a Swap" more directly than a $200 TVL threshold, and it does not need the pruned-state call this PR exists to remove. It also keeps pools that are live but thin — `filterPools` was keeping 21 of about 2,000 pairs.

Metrics, formulas and bucket names are unchanged.

### Validation

| date | before | published | after |
|---|---|---|---|
| 2026-08-24 | V2 lost | $5,537 | $33,150 |
| 2026-08-25 | $4,430 | $6,511 | $4,540 |
| 2026-08-19 | run dies on RPC error | $397 | $398 |

2026-08-19 is the regression check: the current code cannot complete that day at all, and the new code lands within $1 of what is already published.

A refill would be worth running once this merges — the days that lost V2 are still stored that way.

### DIFF
--- a/dexs/nad-fun.ts
+++ b/dexs/nad-fun.ts
@@ -2,7 +2,6 @@ import { ethers } from "ethers";
 import { Adapter, FetchOptions, FetchResultV2 } from "../adapters/types";
 import { CHAIN } from "../helpers/chains";
 import { METRIC } from "../helpers/metrics";
-import { filterPools } from "../helpers/uniswap";
 
 type Balances = ReturnType<FetchOptions["createBalances"]>;
 
@@ -24,6 +23,10 @@ const v2 = {
   bondingCurve: "0x9f3832732923252A21044F21eE6bd87F09514ae4",
   protocolManager: "0x71F846A560a4d68F53e5bd34ED084E7992f171C7",
   feeCollector: "0xE1C8b73343f5A83EBe165BE90470d84B00e33022",
+  // ProtocolManager.initialize() sets the first fee receiver without emitting an
+  // event, so the starting value has to be carried here. Every later change does
+  // emit FeeReceiverUpdate and is picked up from logs.
+  initialFeeReceiver: "0x681C3F796Ae388026634c26025504D7C45248998",
   nadFunFactory: "0xA25b13127e63ddae6d0b35570FF3D39dBD621001",
   // WMON/LVMON are both treated as MON-denominated quote assets.
   monEquivalentQuoteTokens: [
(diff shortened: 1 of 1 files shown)


bheluga:
@bc1phugo thanks for the PR

PR #9009, opened 2026-08-26
fix[bopamm]: track new contract 

monokh:
New BopAmm contracts. Swaps emit Swapped event directly from the parameters of the trade


bheluga:
@monokh thanks for the PR

Issue #9010, opened 2026-08-26
indigo: 77 days of false $0 fees — analytics api stopped populating value_flow on 06-09, empty array slips past the shape guard


0xshubhs:
`fees/indigo` has published exactly $0 every day since **2026-06-10** — 77 consecutive zeros as of today. the protocol is not dead: tvl went from $3.30M on 06-10 to $4.25M today (+29%), there are 316 open cdps holding 20.6M ADA, and 172 of them settled interest during august (most recent 2026-08-26T16:10:08Z).

the adapter doesn't throw, so this doesn't show up as a broken adapter — it silently reports a clean zero.

```
$ npx ts-node cli/testAdapter.ts fees indigo 2026-08-25
CARDANO
Daily fees: 0.00
Daily revenue: 0.00
Daily protocol revenue: 0.00
Daily holders revenue: 0.00

$ npx ts-node cli/testAdapter.ts fees indigo 2026-05-21     # control, pre-break
CARDANO
Daily fees: 385.00
Daily revenue: 385.00
Daily protocol revenue: 377.00
Daily holders revenue: 7.62
```

### what broke

it's upstream, in indigo's own analytics indexer, not in the adapter logic. both endpoints the adapter reads are degraded, in different ways:

**`/api/revenue/flows`** (treasury) is still being written to every day — 166 rows in august, newest `created_at` 2026-08-26T17:00:04Z — but every row since **2026-06-09** has an empty `value_flow`:

```json
// old row
{"id":14730,"slot":163047410,"value_flow":{"lovelace":69596015},"created_at":"2025-08-08T00:47:02Z"}
// newest row
{"id":23729,"slot":196197054,"value_flow":[],"created_at":"2026-08-26T17:00:04Z"}
```

populated vs empty `value_flow` by month:

```
2026-03  365 / 0
2026-04  237 / 0
2026-05  295 / 549
2026-06  117 / 544     <- last populated row 2026-06-09
2026-07    0 / 260
2026-08    0 / 166
```

the txs themselves are real. that newest row, `d6b0b12d47fbb1d0de94b1cae0b7965a6f28b94a135676a65a488baab39db606`, resolves on koios at slot 196197054 / 2026-08-26T17:35:45Z with `total_output` 556,830 ADA and 200 INDY in the asset list. the indexer sees the transaction and fails to compute its value flow.

[...]

PR #9012, opened 2026-08-26
obol: throw when the api omits total_fees_eth_denominated

### LINKED ISSUE #9011: obol: api dropped total_fees_eth_denominated on 08-17 — $0 fees since, and a -$1.64M phantom on the changeover day
`fees/obol` has published exactly $0 every day since **2026-08-18**, and on the changeover day it produces a phantom **-$1.64M**. tvl is $622M and rising, so the protocol is very much alive — this is the upstream api dropping fields.

### what changed

`https://api.obol.tech/tvs/mainnet?limit=1&page=0&timestamp=<date>` stopped returning the fee fields on **2026-08-17**. through 08-16 the response carries `total_fees_eth_denominated`, `total_revenue` and the claimed/unclaimed balances; from 08-17 to today it is four keys and nothing else:

```
$ curl '.../tvs/mainnet?limit=1&page=0&timestamp=2026-08-16'
{"total_balance_eth":251396,"total_balance_usd":472873990,"timestamp":"2026-08-16T00:00:00.000Z",
 "network":"mainnet","total_unclaimed_balance_eth_denominated":1.86,
 "total_claimed_balance_eth_denominated":117.07,"total_eth_transferred":4.31,
 "total_revenue":123.24,"total_fees_eth_denominated":858.12}

$ curl '.../tvs/mainnet?limit=1&page=0&timestamp=2026-08-17'
{"total_balance_eth":251399,"total_balance_usd":471222763,"timestamp":"2026-08-17T00:00:00.000Z","network":"mainnet"}
```

checked 08-17, 08-18, 08-20, 08-21, 08-22, 08-23, 08-24, 08-25, 08-26 — all four-key. it isn't a parameter change either; the bare `?limit=1&page=0` call returns the same four keys. this is consistent, not intermittent (one 502 on 08-19).

### how that lands in the adapter

`fetchCumulativeData` coerces the missing field to zero:

https://github.com/DefiLlama/dimension-adapters/blob/master/fees/obol/index.ts#L24-L28

```ts
return { totalFees: Number(data.total_fees_eth_denominated ?? 0), ... };
```

since the adapter is a day-over-day delta on a cumulative total, `?? 0` produces two distinct wrong answers:

**changeover day** — previous 858.12, current 0:

```
$ npx ts-node cli/testAdapter.ts fees obol 2026-08-18     # dateString 2026-08-17
Daily fees: -1.64 M
Daily supply side revenue: -1.64 M

label                  | Daily fees | Daily supply side revenue
Protocol Fees          | -1640960   |
Fees to Node Operators |            | -1640960
```

**every day after** — both sides 0, so `0 - 0 = 0`, no error:

```
$ npx ts-node cli/testAdapter.ts fees obol 2026-08-26
Daily fees: 0.00
Daily revenue: 0.00
Daily supply side revenue: 0.00
```

the published series matches the second half — hard zeros from 08-18 on, across fees, revenue and supply-side revenue. 08-17 is absent from all three charts, so the -$1.64M day doesn't appear to have made it onto the site, but the adapter does generate it and would republish it on any backfill of that date.

### suggested fix

treat a missing `total_fees_eth_denominated` as missing data and throw, rather than reading it as zero:

```ts
if (data?.total_fees_eth_denominated == null)
  throw new Error(`Obol API returned no fee data for ${dateStr}`);
```

that kills the phantom negative and turns the false zeros into a visible broken adapter until obol restores the fields. `allowNegativeValue: true` is set here for the genuine small day-to-day swings, which is exactly why a -858 ETH delta sails through untouched.

related: this is the same failure shape as #9010 (indigo), where an upstream analytics api quietly stopped populating its value fields and the adapter's `?? 0` turned it into 77 days of clean zeros.

happy to send the pr.

0xshubhs:
Fixes #9011

the obol tvs endpoint stopped returning `total_fees_eth_denominated` and `total_revenue` on 2026-08-17. the adapter is a day-over-day delta on a cumulative total, so the `?? 0` fallback produced a -858 eth (-$1.64M) delta on the changeover day and an exact 0 every day since; no error, so it reads as a live protocol earning nothing while tvl sits at $622M.

harness, with the change:

```
fees obol 2026-08-26   -> Error: Obol API returned no fee data for 2026-08-25
fees obol 2026-08-18   -> Error: Obol API returned no fee data for 2026-08-17   (was -1.64 M)
fees obol 2026-08-17   -> fees -1200, supply side -5323, revenue 4124           (unchanged)
```

the control day matches what the site currently publishes for 08-16 (-1198 / -5317 / 4119).

### DIFF
--- a/fees/obol/index.ts
+++ b/fees/obol/index.ts
@@ -24,7 +24,10 @@ const correctedTotalRevenues: Record<string, number> = {
 const fetchCumulativeData = async (dateStr: string) => {
   const url = `${ENDPOINT_BASE}?limit=1&page=0&timestamp=${encodeURIComponent(dateStr)}`;
   const data = await httpGet(url);
-  return { totalFees: Number(data.total_fees_eth_denominated ?? 0), totalRevenue: correctedTotalRevenues[dateStr] ?? Number(data.total_revenue ?? 0) };
+  if (data?.total_fees_eth_denominated == null) {
+    throw new Error(`Obol API returned no fee data for ${dateStr}`);
+  }
+  return { totalFees: Number(data.total_fees_eth_denominated), totalRevenue: correctedTotalRevenues[dateStr] ?? Number(data.total_revenue ?? 0) };
 };
 
 const getPreviousDay = (dateStr: string) => {


0xshubhs:
the red `test` check is the change doing its job, and it's also the best evidence in the pr — that's your ci runner, not my machine, confirming the api has no fee data for 2026-08-25:

```
Error: Obol API returned no fee data for 2026-08-25
    at fetchCumulativeData (fees/obol/index.ts:28:11)
```

on master the same run comes back `Daily fees: 0.00` with a green check, which is exactly the problem — nine days of false zeros on a $622M-tvl protocol, plus a -$1.64M phantom on 08-17 that any backfill of that date would republish. the adapter will keep erroring until obol restores `total_fees_eth_denominated`; if you'd rather it stayed quiet in the meantime i can gate the throw behind a start date instead, but silent zeros seemed worse than a visible break. details in #9011.

0xshubhs:
on coderabbit's note about `total_revenue ?? 0` — leaving that one as is, deliberately.

zero is already a handled sentinel for that field, not a silent fallthrough. the block right below it treats a 0 on either side as missing data and books no revenue for the day, with a comment explaining that the api intermittently returns `total_revenue: 0` and that the delta otherwise produces a phantom ~118 ETH swing. so `?? 0` on revenue lands in an explicit guard two lines later; `?? 0` on `total_fees_eth_denominated` lands nowhere, which is the whole bug.

it also isn't a case that occurs: the two fields appeared and disappeared together when the response went from nine keys to four, so there's no observed response with fees but no revenue. throwing on it would change behaviour beyond the fix and would fight the existing `=== 0` handling.

separately, i tried the equivalent one-liner on #9010 (indigo, same class) and it turned out to be wrong for exactly the reason worth stating here — an empty/missing value is sometimes the api's honest way of saying "nothing happened". worth checking per field rather than applying the pattern by reflex. obol's fee field is the case where it genuinely can't be.

bheluga:
@0xshubhs thanks for the PR

PR #9020, opened 2026-08-26
jupiter-lend-dex: read pool tokens from the swap calls instead of init_dex

srijani-raccoons:
## Problem

`dex_call_init_dex` decodes `account_token_1` as the system program (`11111111111111111111111111111111`) for pool `FfEJYz4hLLFe4KyNyKYCjvfhMfbtpLH3CXpcPe8nhmfy`, the USDC/USDT pool. Both adapters took pool token identity from that table alone, so every swap on that pool with `swap0to1 = false` was attributed to an unpriceable mint and dropped. The other side was labelled USDT where the pool's `token_0` is actually USDC; identical price and decimals meant the USD figure happened to survive.

That pool is 559,421 of 568,155 swaps to date. The dropped side is 259,079 swaps, around $84.6M of cumulative notional.

## Changes

Pool token identity now comes from `dex_call_swap_in` ∪ `dex_call_swap_out`, with `dex_call_init_dex` kept as the fallback for pools that have never traded.

`swap_out` volume now reads the `amount_out` call argument and values the output leg. This replaces the `call_log_messages` unnest that decoded `amount_in` out of the base64 `LogSwap` payload. There are three `swap_out` calls in the table's entire history, and one of them has no matching log line, so the previous inner join silently dropped it.

The fees query aggregates by mint in SQL instead of returning one row per swap. This is a correctness fix, not just an efficiency one: the Dune results endpoint returns at most 32,000 rows, and `helpers/dune.ts` does not paginate, so the previous per-swap result set was silently truncated. 2026-08-24 had 72,309 swaps and the adapter received 32,000 of them, an arbitrary subset that varied between executions. Three runs of the unchanged adapter on that day returned $14.00, $22.00 and $14.00 of daily fees. Aggregating in SQL brings the result set to one row per mint, well inside the cap.

`start` moves from `2026-06-18`, the pool creation date, to `2026-06-22`, the first swap.

Unchanged: the 1e6 scaling of `fee` and `revenue_cut`, the per-pool rate timeline built from `init_dex` ∪ `update_fee_and_revenue_cut` and joined on `call_block_time`, and the 50/50 Jupiter/Fluid split of the protocol cut.

## Verification

`npm test dexs jupiter-lend-dex 2026-08-25` returns $29.05M daily volume, against $14.84M before this change.

`npm test fees jupiter-lend-dex 2026-08-25` returns $128.00 daily fees, splitting to $96 supply side, $16 to Fluid within supply side, and $16 Jupiter revenue against the 25% `revenue_cut` live that day. It is now stable across runs, where the previous version returned between $14 and $22 for the same day.

`npm test dexs jupiter-lend-dex 2026-08-14` returns $1.02M and `npm test fees jupiter-lend-dex 2026-08-14` returns $87.00. That day had a mid-day fee change on the main pool, so it exercises the rate window join.

`npm test dexs jupiter-lend-dex 2026-08-19` returns $3.10M and `npm test fees jupiter-lend-dex 2026-08-19` returns $7.39. `npm test dexs jupiter-lend-dex 2026-07-02` returns $5.34. `npm test dexs jupiter-lend-dex 2026-06-25`, a day with no swaps, returns 0 without erroring.

Cumulative volume through 2026-08-25 comes to $168,120,432, reconciling against Jupiter's own Dune model of the same pools to within $0.07.

`npm run ts-check` passes.

### DIFF
--- a/dexs/jupiter-lend-dex/index.ts
+++ b/dexs/jupiter-lend-dex/index.ts
@@ -2,67 +2,63 @@ import { Dependencies, FetchOptions, SimpleAdapter } from "../../adapters/types"
 import { CHAIN } from "../../helpers/chains";
 import { queryDuneSql } from "../../helpers/dune";
 
-// Volume = every swap's input-side amount priced in the input token.
-// - swap_in (exact-in): amount_in is a call argument (dex_call_swap_in.amount_in).
-// - swap_out (exact-out): amount_in is not a call arg, but the DEX emits a LogSwap event
-//   (Anchor `emit!`) which appears as a "Program data: yvLkHCXC..." line in call_log_messages.
-//   We decode the base64 payload and read amount_in (u64 LE at byte 11 after the 8-byte
-//   Anchor discriminator + u16 dex_id + u8 swap_0_to_1). Verified against on-chain data:
-//   the 3 historical swap_out txs decode to 1,283,885 / 1,283,391 / 1,501,258 — matching
-//   the actual token transfers.
+// Volume = one leg of every swap, priced in that leg's token.
+// - swap_in (exact-in): amount_in is a call argument, so value the input leg.
+// - swap_out (exact-out): amount_out is a call argument, so value the output leg.
 //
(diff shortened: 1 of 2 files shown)


bheluga:
@srijani-raccoons thanks for the PR

PR #9026, opened 2026-08-27
hydration-dex: repoint to the indexer host and throw on an empty result

### LINKED ISSUE #9025: hydration: indexer stalled 08-24 while the chain runs, and dexs/hydration-dex points at a host that no longer accepts connections
two hydration adapters went dark after **2026-08-24** and both are `total24h: null`:

- **`dexs/hydration-dex`** — was $18.9M/7d, $63.3M/30d, 31/31 daily density, and volume was *rising* into the stop ($595k on 08-17 → $5.59M on 08-21 → $3.2M on 08-24)
- **`fees/hydradx`** — ~$2-3k/day

the chain is fine. two independent rpcs agree hydration is producing blocks right now:

```
rpc.hydradx.cloud            head 13,862,001   ts 2026-08-27T05:45:24Z
hydration-rpc.n.dwellir.com  head 13,862,002   ts 2026-08-27T05:45:24Z
```

what stalled is hydration's indexing/api layer, and it's showing up three different ways across three of their hosts:

| host | used by | state |
|---|---|---|
| `orca-prod-pool-02.catfish.hydration.cloud` | `dexs/hydration-dex` | **tcp connection refused on 443** (dns resolves to 135.181.128.254, nothing listening) |
| `orca-main-aggr-indx.indexer.hydration.cloud` | `projects/hydradx` tvl | answers 200, but **frozen at block 13,762,692 / 2026-08-24T13:33:00Z** |
| `hydration-metrics-aggregator.indexer.hydration.cloud` | `fees/hydradx` | **http 500** |

so the indexer is ~99,300 blocks / ~2.7 days behind a live chain.

### the part that is fixable now

the host `dexs/hydration-dex` points at no longer exists, so that adapter can never heal on its own even after hydration restarts their indexer:

```
$ npx ts-node cli/testAdapter.ts dexs hydration-dex 2026-08-22
FetchError: request to https://orca-prod-pool-02.catfish.hydration.cloud/graphql failed,
reason: connect ETIMEDOUT 135.181.128.254:443
```

that's a day the site has data for. `orca-main-aggr-indx.indexer.hydration.cloud/graphql` — the host the tvl adapter already uses — serves the **identical** `platformTotalVolumesByPeriod` query with the same fields, and returns the same numbers:

```
$ npx ts-node cli/testAdapter.ts dexs hydration-dex 2026-08-22    # with the host repointed
Daily volume: 5.59 M
XYK Pools Fees 1.8 | StableSwap Fees 383 | Omnipool Fees 2965
```

against the $5,587,849 the site published for 2026-08-21.

repointing alone would be a downgrade though — while the indexer is stalled the query returns `nodes: []`, and the loop over an empty array publishes a clean **$0** instead of erroring, which is worse than the visible break there is today. so the pr pairs the host change with a throw on an empty result:

```
$ npx ts-node cli/testAdapter.ts dexs hydration-dex 2026-08-27
Error: Hydration indexer has no volume data for 2026-08-25T23:59:59.000Z
```

that gets the adapter onto a host that exists and keeps the outage visible until hydration's indexer catches up. `fees/hydradx` i've left alone — its host answers, it just 500s, and there's no live alternative for `/api/v1/fees/charts` that i can find.

pr to follow.

0xshubhs:
Fixes #9025

`orca-prod-pool-02.catfish.hydration.cloud` no longer accepts connections on 443, dns resolves, nothing listening, so the adapter has thrown since 2026-08-24 and can't heal on its own. `orca-main-aggr-indx.indexer.hydration.cloud` is the host `projects/hydradx` already uses for tvl, and it serves the identical `platformTotalVolumesByPeriod` query.

```
$ npx ts-node cli/testAdapter.ts dexs hydration-dex 2026-08-22
master   FetchError: connect ETIMEDOUT 135.181.128.254:443
patched  Daily volume: 5.59 M   (XYK 1.8 | StableSwap 383 | Omnipool 2965)
```

that matches the $5,587,849 the site published for 2026-08-21.

hydration's indexer is itself stalled at block 13762692 / 2026-08-24T13:33Z while the chain is at 13,862,001 as of this morning, so the same query returns `nodes: []` for recent days. repointing on its own would turn today's visible break into a silent $0, so the empty case throws instead:

```
$ npx ts-node cli/testAdapter.ts dexs hydration-dex 2026-08-27
Error: Hydration indexer has no volume data for 2026-08-25T23:59:59.000Z
```

adapter heals by itself once their indexer catches up. `fees/hydradx` is dark for the same underlying reason but its host only 500s, and i couldn't find a live alternative for `/api/v1/fees/charts`, so i've left it.

### DIFF
--- a/dexs/hydration-dex/index.ts
+++ b/dexs/hydration-dex/index.ts
@@ -2,6 +2,8 @@ import type { FetchOptions, SimpleAdapter } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
 import { request } from "graphql-request";
 
+const INDEXER_URL = "https://orca-main-aggr-indx.indexer.hydration.cloud/graphql";
+
 
 const fetch = async (options: FetchOptions) => {
 
@@ -23,7 +25,12 @@ const fetch = async (options: FetchOptions) => {
     }
   `;
 
-  const queryResult = await request("https://orca-prod-pool-02.catfish.hydration.cloud/graphql", query);
+  const queryResult = await request(INDEXER_URL, query);
+
+  const nodes = queryResult.platformTotalVolumesByPeriod.nodes;
+  if (!nodes.length) {
+    throw new Error(`Hydration indexer has no volume data for ${toDateQuery(options.fromTimestamp)}`);
+  }
 
   const dailyVolume = options.createBalances();
   const dailyFees = options.createBalances();
@@ -32,7 +39,7 @@ const fetch = async (options: FetchOptions) => {
   const dailyHoldersRevenue = options.createBalances();
   const dailyProtocolRevenue = options.createBalances();
 
-  for (const node of queryResult.platformTotalVolumesByPeriod.nodes) {
+  for (const node of nodes) {
     dailyVolume.addUSDValue(Number(node.totalVolNorm));
     dailyFees.addUSDValue(Number(node.xykpoolFeeVolNorm), 'XYK Pools Fees');
     dailyFees.addUSDValue(Number(node.stableswapFeeVolNorm), 'StableSwap Fees');


0xshubhs:
the red `test` is the guard firing, not a regression — and it's confirmation from your runner that hydration's indexer is still behind:

```
Error: Hydration indexer has no volume data for 2026-08-25T23:59:59.000Z
    at fetch (dexs/hydration-dex/index.ts:32:11)
```

on master the same run fails too, just less usefully — `connect ETIMEDOUT 135.181.128.254:443`, because the host it points at is gone. so this isn't red-vs-green, it's a dead host versus a live one that's lagging.

the check goes green on its own once hydration's indexer catches up; the chain is at 13,862,001 as of this morning against the indexer's 13,762,692. if you'd rather have it merge green today i can drop the empty-nodes throw and land the host change alone, but that would publish $0 for every stalled day, which seemed like the wrong trade on a $63M/30d adapter. context in #9025.

bheluga:
@0xshubhs thanks for the PR

PR #9027, opened 2026-08-27
hypersight: rename methodology to Liquary, fix HIP-4 rate (0.025% -> 0.05%)

liquary-xyz:
Rebrand + a rate correction on the existing `hypersight` builder entry. No logic change: the builder address (`0xc920…0927`), `start` (2026-07-01) and the module key are untouched, so no historical figure moves.

**Website:** https://liquary.xyz
**Twitter:** https://x.com/liquary_xyz

### 1. Rename

Hypersight is now **Liquary**. Same product, same team, same builder address; only the name and the domain changed. The old domain returns a path-preserving 301 to the new one and stays in place, so no existing link breaks.

The metadata rename (name, url, twitter, description) lives outside this repo, so it has been sent to metadata@defillama.com separately.

### 2. Rate correction

The `Fees` methodology said HIP-4 prediction-market closes are charged **0.025%**. That was wrong when we listed: the rate is **0.05%**. This string is description only, so nothing reported changes: revenue comes from the on-chain builder fees and has always been correct.

### Note on the module key

I deliberately left the `hypersight` key as-is, so this PR cannot desync the adapter from the server-side module name. If renaming the protocol also moves that key, tell me and I'll submit the follow-up PR the same day, or feel free to rename it yourselves.


bheluga:
@liquary-xyz thanks for the PR

PR #9028, opened 2026-08-27
fix(o1-launchpad): add missing Base RWA v4 suite

dennischi99:
## Summary

- register the missing Base RWA timestamp v4 production suite for o1 Launchpad
- keep the historical Base v1/v2/v3 and Robinhood suites unchanged
- record each suite's authoritative deployment block and ignore pre-deployment Trade/Credited events
- preserve the existing DefiLlama methodology: historical-hourly pricing, platform swap fees plus token launch fees as Revenue, and creator/referrer fees as SupplySideRevenue

## Root cause

The adapter did not include the currently active Base RWA v4 contracts:

- factory: `0xFf70918Ef17A2D74d683a8297813B177BaFaD1f4`
- hook: `0x3b2b979DF21036CEe51B8debB13100e2CB8DeaCC`
- escrow: `0x1D8c991A9019df7D72ADCd8deA6f12D600C9d02f`
- first block: `50,137,081` (2026-08-18 14:31:49 UTC)

The suite uses timestamp fee accounting, quote-denominated swap fees, and native token launch fees.

## Reconciliation

For the complete UTC window 2026-07-27 00:00 through 2026-08-26 00:00:

- previous DefiLlama Revenue: about $849,455
- local standard-runner result after adding v4: about $934,674
- v4 Revenue contribution in that run: about $85,219
- Analytics Platform fees snapshot for the same trade window: about $1,113,564

The remaining difference is expected to include Analytics current-price revaluation versus DefiLlama historical-hourly prices, approximately $13k of launch fees included only by DefiLlama, hourly rounding, and observed Indexer/RPC log completeness variance. No USD values or date-specific adjustments are hard-coded.

`splitTransaction` remains fail-closed. On-chain samples satisfy the creator/referrer/platform identity; observed drops were caused by incomplete Credit log responses, so this PR does not guess allocations or change the split logic.

## Validation

- `npm run ts-check-cli -- --pretty false`
- `npm test -- fees o1-launchpad 2026-08-18 base`
- `npm test -- fees o1-launchpad 2026-08-18T19:00:00Z base`
- verified an on-chain v4 referral trade where creator + referrer + platform equals totalFee
- verified all Base factory/hook/escrow addresses are unique

## Backfill

Please recompute Base fees, revenue, and supply-side revenue from block `50,137,081` so the historical charts include v4.

### DIFF
--- a/fees/o1-launchpad/index.ts
+++ b/fees/o1-launchpad/index.ts
@@ -99,6 +99,7 @@ const chainConfig: Record<string, ChainConfig> = {
         factory: "0xe3ab924c72463c1ac8d1d8352ee640b89eb1ea64",
         hook: "0xa068cf4c52abdd3479145c4b3cbd8e3d71542a44",
         feeEscrow: "0xabe87e4af23dafad0a170aa900d574c03d904597",
+        // First suite block: https://basescan.org/block/48364845
         firstBlock: 48_364_845,
         legacyFeeCurrency: true,
         launchFeeCurrency: "none",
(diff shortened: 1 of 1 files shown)


bheluga:
@dennischi99 thanks for the PR

PR #9030, opened 2026-08-27
hydration-dex: source volume/fees from Hydration API

mrq1911:
Repoints the adapter off the retired aggregation-indexer GraphQL to Hydration's own API (`api.hydradx.io/defillama`), serving per-day volume + per-pool fees. Same volume/fee methodology.


bheluga:
@mrq1911 thanks for the PR

PR #9041, opened 2026-08-28
feat(xoxno-lending): add liquidations, active-users and new-users adapters

mihaieremia:
Adds the remaining dimensions for the XOXNO lending deployment on Stellar, alongside the existing fees adapter.

- **liquidations** — daily collateral liquidated
- **active-users** — daily active users and transaction count
- **new-users** — daily new users

All three follow the same shape as the fees adapter: a per-chain config, a UTC day range, and a single call to the protocol's event-sourced integration API.

**Also corrects the backfill start.** All four adapters declared `start: "2026-06-21"`, but the mainnet contracts were not deployed until `2026-08-27`. Every earlier day backfills a guaranteed zero, so the start is moved to the deployment date.

**Testing.** Each adapter runs against production with the corrected start:

```
fees          fees 0.00  revenue 0.00  protocol revenue 0.00  supply side 0.00
liquidations  collateral liquidated 0.00
active-users  active users 0.00  transactions 0.00
new-users     new users 0.00
```

Note: the protocol is deployed but not yet open, so these are currently 0.

### DIFF
--- a/active-users/xoxno-lending.ts
+++ b/active-users/xoxno-lending.ts
@@ -22,7 +22,9 @@ const CHAIN_CONFIGS: Record<string, ChainConfig> = {
   [CHAIN.STELLAR]: {
     chainName: "Stellar",
     userStatsExportPath: "/integrations/lending/stellar/active-users",
-    start: "2026-06-21",
+    // Mainnet contracts were deployed at ledger 64140891,
+    // 2026-08-27T00:55Z. An earlier start only backfills zeros.
+    start: "2026-08-27",
   },
 
   // Add MultiversX later with the same API response shape:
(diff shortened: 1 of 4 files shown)


mihaieremia:
Thanks — I looked into the chain-config finding and don't think it applies here, so I've left the code as is.

`cli/buildModules.ts` strips any key outside `whitelistedBaseAdapterKeys` (`start`, `deadFrom`, `fetch`, `runAtCurrTime`) from each chain entry before the module is built:

```ts
if (!whitelistedBaseAdapterKeys.has(key)) {
  delete obj[key] // remove non base adapter keys to avoid confusion
}
```

So passing the full config object through as `adapter` would have `chainName` and the export paths deleted at build time, and the fetch would still need to read them from `CHAIN_CONFIGS`. It changes nothing at runtime.

On the drift concern: the registration map is derived from `CHAIN_CONFIGS` rather than written alongside it, so the two cannot diverge — adding a chain to that one object registers it and supplies its path in the same step.

This also matches the existing `fees/xoxno-lending` adapter, which uses the same projection; changing only the three new files would make them inconsistent with it.

bheluga:

PR #9047, opened 2026-08-28
[fix] Rialto - restore route-action volume accounting

vinayyak:
This reverts #9037 and restores Rialto aggregator volume accounting from `RouteActionExecuted`.

After discussions with members of the Robinhood Chain community, I am reverting #9037 because counting only the top-level user input is not the preferred way to measure routed activity here.

Each route leg is executed independently against underlying liquidity. Restoring `RouteActionExecuted` accounting captures the full volume processed across multi-hop and split routes.

Validation:

- `pnpm run ts-check`
- `git diff --check`
- Historical adapter test against Robinhood Chain


noateden:
hi, for all dex-agg, we count volume per trade, not per route

PR #9050, opened 2026-08-28
[fix] PropAMM - attribute volume by pair target

vinayyak:
This corrects PropAMM volume attribution by identifying the pair contracts explicitly.

The existing adapter treats every `RouteActionExecuted` event with `kind = 0` as PropAMM volume. However, `kind = 0` represents Rialto's generic `RawCall` execution path and is also used by other liquidity adapters.

This change:

- counts only route actions targeting the 27 known PropAMM pair contracts
- excludes unrelated `RawCall` adapters
- keeps the existing adapter name and event-based accounting

For the same historical 24-hour window:

- Previous unfiltered result: approximately $2.35M
- Pair-filtered result: approximately $1.38M

Checks completed:

- `pnpm run ts-check`
- `git diff --check`
- Historical adapter test against Robinhood Chain

### DIFF
--- a/dexs/rialto-prop-amm/index.ts
+++ b/dexs/rialto-prop-amm/index.ts
@@ -1,11 +1,46 @@
 import { FetchOptions, SimpleAdapter } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
 
+// Verified Robinhood deployment and event definitions:
+// https://robinhoodchain.blockscout.com/address/0xC94135b63772b91D79d0A2DaAb2a8801f32359bD
 const RIALTO_ROUTER = '0xC94135b63772b91D79d0A2DaAb2a8801f32359bD';
 
 const routeActionEvent = 'event RouteActionExecuted( uint256 indexed hopIndex, uint8 kind, address indexed pool, address tokenIn, address tokenOut, uint256 amountIn, uint256 amountOut)';
-const propAmmKind = 0n;
+// RouteActionKind.RawCall in the verified RialtoRouter deployment above.
+const rawCallKind = 0n;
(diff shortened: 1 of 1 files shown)


vinayyak:
@coderabbitai review

Please refresh the walkthrough and checks against current head `74bf4a9`. The prior summary references an obsolete force-pushed revision; the current diff keeps `dexs/rialto-prop-amm/index.ts` and uses generic PropAMM naming.

vinayyak:
Current head `74bf4a9` addresses the provenance feedback and keeps the existing `dexs/rialto-prop-amm` adapter path with generic PropAMM naming. Both `test` and `ts-check` pass, and CodeRabbit's refreshed analysis found no new actionable issues. The remaining CHANGES_REQUESTED reviews are attached to the obsolete force-pushed commits `b5d3089` and `ab615fc`; please dismiss them as stale if they block merging.

bheluga (commented):
@vinayyak thanks for the PR
Is there a way we can get pool dynamically so that we don't need to update everytime a new pool is created?

vinayyak:
hi @bheluga,

these pairs are currently deployed directly rather than through a factory or registry event, so there is no reliable event-based discovery source available to this adapter. 

I could scan contract creation transactions from the deployer EOA and validate each contract’s interface, but that would require repeated block, transaction, or trace scanning in DefiLlama’s stateless adapter environment, making it comparatively expensive and fragile. 

I have not found any registry, factory or`PairCreated` event for these pools on-chain so the explicit pool list is the safest way currently for this.

bheluga:
@vinayyak thanks for the PR

PR #9068, opened 2026-08-30
fix: preserve getLogs metadata required by adapters

chetna-mittal:
## Description

`FetchOptions.getLogs` defaults to returning decoded event arguments only. This removes metadata such as `blockNumber`, `logIndex`, `transactionHash`, and the emitting contract address.

Several adapters relied on that metadata for event ordering, historical rate selection, transaction matching, or vault attribution. This could produce incorrect calculations or silently skip events.

## Validation

- `npm run ts-check`
- `npm run ts-check-cli`
- `git diff --check`


bheluga:

PR #9071, opened 2026-08-30
Fix: should not skip any recipient on native

June-native:
## Summary
Revert the Native `RFQTrade` recipient skip added in https://github.com/DefiLlama/dimension-adapters/pull/8890. Count all RFQ fills again, including those whose recipient is `0x309fcdd159c15e6305f1b02489a14e870e4df052` or `0x23e2b37415ee3b9cfe1ae522e42e2b66fd2d9494`.

## Why
#8890 treated those two contracts as unofficial adapters and excluded them because they received almost all of the Jul–Aug 2026 BSC volume spike.

We have since confirmed with the LiquidMesh (BSC aggregator) team that:
- they deployed these adapter contracts
- volume on those days is from real users trading to earn Binance Alpha points. That flow should be counted as Native volume. Skipping those recipients undercounts Native BSC.

## What changed
Removed `SKIP_RECIPIENTS` and the recipient filter in `dexs/native/index.ts`. Every `RFQTrade` log is included in `dailyVolume`.

### DIFF
--- a/dexs/native/index.ts
+++ b/dexs/native/index.ts
@@ -12,12 +12,6 @@ const chainConfig: Record<string, { start: number }> = {
 
 const RFQ_TRADE_EVENT = 'event RFQTrade(address recipient, address sellerToken, address buyerToken, uint256 sellerTokenAmount, uint256 buyerTokenAmount, bytes16 quoteId, address signer)';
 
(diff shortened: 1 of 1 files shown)


bheluga:
@June-native thanks for the PR

PR #9075, opened 2026-08-31
fix(xoxno-lending): report gross fees, and add MultiversX

mihaieremia:
XOXNO Lending is an overcollateralized lending protocol on MultiversX and Stellar.

This PR fixes the fees adapter. It was reporting only the protocol's cut as fees, so fees and revenue came out identical and supply side revenue was always zero. Fees are now the full interest paid by borrowers, split between the protocol and the suppliers.

It also adds MultiversX, which holds most of the protocol's TVL and was not tracked at all.


bheluga:

PR #9077, opened 2026-08-31
feat(xoxno-liquid-staking): add fees and revenue adapter

mihaieremia:
XOXNO Liquid Staking lets users stake EGLD and receive xEGLD, which grows in value as staking rewards accrue.

This PR adds the fees adapter. Fees are the staking rewards earned by the pool, revenue is the 7% commission XOXNO takes on them, and the rest goes to xEGLD holders.

Both sides are emitted by the contract, so nothing is estimated.


bheluga:

PR #9082, opened 2026-08-31
durianfun-launchpad: fix the 429s killing this adapter (225 getLogs/run -> 3)

Duriandotfun:
**This adapter has reported nothing since 2026-08-23** - `total24h` is absent rather
than zero, and `total7d` / `total30d` are missing entirely - while the launchpad has
kept trading every day. This restores it.

### Cause

Reproduced locally with `npx ts-node --transpile-only cli/testAdapter.ts dexs durianfun-launchpad`:

```
- host: https://www.kubscan.com/api/eth-rpc error: Request failed with status code 429
- host: https://lb.routeme.sh/rpc/evm/96 error: Request failed with status code 429
- host: https://rpc.bitkubchain.io error: Request failed with status code 429
Error: Aborting, previous errors in promise pool
```

The daily scan used `targets:`, which issues one query per address - 96 markets x 2
events + 11 V5 markets x 3 = **225 queries per run**, and this adapter is `pullHourly`,
so roughly 5.4k a day. Bitkub's public RPCs will not carry that. Nothing changed in the
adapter on 2026-08-23; the market count simply grew past what those endpoints tolerate.

### Fix

One chain-wide `noTarget` scan per event, filtered client-side against the market set:
**225 queries -> 3**, and flat as the launchpad grows. Same approach as `dexs/aborean-cl`
and `dexs/aerodrome-slipstream`.

The address filter is load-bearing rather than cosmetic - a chain-wide scan sees every
contract emitting the same signature, and `ReferralPaid(address,address,uint256)` in
particular is a name any protocol might use.

No change to the fee/volume accounting, the factory list, or the methodology.

### Verification

| | result |
|---|---|
| patched adapter | completes all 24 hourly slots - volume **$6.88**, fees **$0.08** |
| independent on-chain sum, same window | volume **$6.87**, fees **$0.0801** |
| **control**: original file, cache already warm | still dies at slot 2/24, 58 lines of 429 |
| `npm run ts-check` | exit 0 |

The control run is the one that matters: it rules out the patched run having passed
only because an earlier run had warmed the cache.


bheluga:

PR #9083, opened 2026-08-31
[fix] frontier-fun - listen to every FactoryHook and PolDistributor generation

godcomplex44:
Frontier rotated its FactoryHook and PolDistributor on 2026-08-28 (gen-2: hook `0xee588bCF2bd3e658f5160489f4199d1851BBf0Cc`, distributor `0x57ceF9B5844e2fB0181bd12d3918A8c6dCaa8F4e`). A Uniswap V4 pool keeps the hook its PoolKey embeds, so pools registered on the retired generation keep trading and paying fees — the adapter must listen to both generations, not switch.

- `FACTORY_HOOK` / `POL_DISTRIBUTOR` become per-generation arrays with the rotation dates.
- The three Dune predicates (pool discovery by `hooks`, the hook's `SwapFeeDistributed` logs, the distributor's `FeesDistributed` logs) become `IN` lists.
- No event signature changed across generations — verified on-chain: the gen-2 contracts already emit, and their `topic0`s match the constants in the file (`0xf1a3…41f`, `0x7b62…b87`).

Follow-up to #9042; methodology unchanged.

### DIFF
--- a/dexs/frontier-fun/index.ts
+++ b/dexs/frontier-fun/index.ts
@@ -63,15 +63,28 @@ const CURVE_FEE_DISTRIBUTED_EVENT =
 const GRADUATION_FEES_PAID_EVENT =
   "event GraduationFeesPaid(address indexed feeRecipient, address indexed caller, uint256 creatorAmount, uint256 protocolAmount, uint256 refundAmount)";
 
-// v1.2 Uniswap V4 side, all live since 2026-08-15 with the v1.2 factory.
+// v1.2 Uniswap V4 side, live since 2026-08-15 with the v1.2 factory.
 // Singleton hook attached to every Frontier pool (PoolRegistered, SwapFeeDistributed).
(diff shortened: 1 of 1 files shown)


godcomplex44:
The failing `test` job is the fork-secrets limitation, not the diff: it stops at `DUNE_API_KEYS environment variable is not set` before running the query — same as #9042 (this adapter's previous PR), which failed that job the same way and was merged after a maintainer run. ts-check and CodeRabbit are green, and the SQL change is two `=` predicates widened to `IN` lists over the same tables.

godcomplex44:
@treeoflife2 thanks for the merge! Could we get a refill for frontier-fun from **2026-08-28** onward? The gen-2 hook/distributor rotation went live on-chain that day, so the days between the rotation and this merge were recorded without gen-2 swap volume and fees. Days since the merge are complete — only that window needs re-running.

PR #9084, opened 2026-08-31
[ElfomoFi] enable bsc vault adapter

olegggatttor:
Enabling BSC vault adapter


olegggatttor:
@bheluga adding BSC vault here, same as Base one

bheluga:

PR #9095, opened 2026-09-01
icdex: read volume from the pair canisters, the aggregator's 24h cache never decays

### LINKED ISSUE #9092: icdex: volume frozen at ~$153.4k/day for 35 days, real 24h volume is ~$14 (stale cache in the stats canister)
`dexs/ICDex` has published a nearly identical number every day for 35 days: $153,042–$153,712, against $364.89/day average over the preceding 57 days. The step is a single day — 2026-07-27 $1,090 → 2026-07-28 $153,163 — and it has not moved since.

The adapter is faithfully relaying its source. It GETs `https://gwhbq-7aaaa-aaaar-qabya-cai.raw.icp0.io/v1/latest` and sums `usd_24h_volume` across pairs. I curled it: 32 pairs, sum = **153,405.880006**, and `api.llama.fi/summary/dexs/icdex` reports `total24h` = **153,406**. Exact match.

The problem is the field. That stats canister caches per-pair 24h figures and only refreshes a pair when that pair trades — it never decays a stale 24h number back to zero. 99.18% of the total is one pair that has not traded in months:

```json
"ckBTC/ICP": { "base_volume": "169.368960", "base_24h_volume": "1.708860",
               "quote_24h_volume": "34833.466314", "usd_24h_volume": "152145.462693" }
```

### The feed contradicts itself

`base_24h_volume` of 1.708860 ckBTC/day is 623.7 ckBTC/year. The same object reports lifetime `base_volume` of 169.368960 ckBTC — so the claimed daily rate is **3.7x the pair's entire lifetime volume**.

The USD figure is frozen at a stale price too: 152145.462693 / 34833.466314 implies ICP = $4.3678, and / 1.708860 implies BTC = $89,033. Live right now: ICP $2.40, BTC $78,871. A real USD volume would move with price; this one has not changed by a cent in 35 days.

### On-chain

I queried the pair canisters' own `stats()` directly (anonymous query via `https://icp-api.io/api/v2/canister/<cid>/query`, CBOR envelope, Candid-decoded):

| pair | canister | on-chain `vol24h` | feed's 24h claim |
|---|---|---|---|
| ckBTC/ICP | `5u2c6-kyaaa-aaaar-qadiq-cai` | `{0, 0}` | 1.708860 ckBTC / 34,833.47 ICP |
| ICL/ICP | `52ypw-riaaa-aaaar-qadjq-cai` | `{0, 0}` | 19,950.4 ICL / 70.998 ICP |
| DKP/ICP | `3wq7v-gaaaa-aaaar-qad6a-cai` | `{6870000000000, 570000856}` | 68,700 DKP / 5.700009 ICP |

DKP is the control: it traded today, and the feed matches the chain **to the unit** (6,870,000,000,000 / 1e8 = 68,700.000000; 570,000,856 / 1e8 = 5.70000856). So the pipeline is correct — it is specifically the idle pairs whose cached entry never expires.

ckBTC/ICP's `totalVol` on-chain is 172.166 ckBTC lifetime, consistent with the feed's lifetime field, which rules out a decimals or unit mistake: only the 24h field is wrong.

So the real number is on the order of 5.7 ICP (~$14) across the pairs I checked, against $153,406 published — roughly **11,000x**, about **$4.6M of phantom volume per 30 days**.

### Not a dead protocol

ICDex is alive — DKP/ICP filled orders today. This is a stale-cache problem in the source, not a wind-down, so `deadFrom` would be the wrong answer.

### Fix path

There is a verified-correct source, but it is on-chain rather than HTTP: enumerate pairs from `/v1/tickers` (its `pool_id` field is still accurate — it gave me `5u2c6-…` for ckBTC/ICP and `3wq7v-…` for DKP/ICP), call `stats()` on each pool canister, and use `vol24h` instead of `usd_24h_volume`, pricing the ICP-quoted leg with `addCGToken('internet-computer', …)`. Two things to settle when implementing: whether `vol24h` counts both maker and taker sides, and that the repo currently has no IC query helper — every existing ICP adapter (`fees/icp`, `fees/waterneuron`, `fees/openchat`, `dexs/helix-markets`) uses a plain HTTP endpoint, so this would be the first Candid/CBOR call in the repo.

I have a working query for it and am happy to send that PR if you want it in-repo rather than waiting on ICLighthouse to fix the cache upstream.

tarun-khatri:
Fixes #9092

ICDex has published $153,042–$153,712 every day for the last 35 days. The 57 days before that averaged $364.89 with a max of $3,835. The step is a single day: 2026-07-27 $1,090 → 2026-07-28 $153,163.

The adapter sums `usd_24h_volume` from ICLighthouse's stats canister. That canister only refreshes a pair's 24h entry when the pair trades, and never decays it back to zero, so an idle pair keeps reporting its last busy day forever. One pair is 99.18% of the total:

```json
"ckBTC/ICP": { "base_volume": "169.368960", "base_24h_volume": "1.708860",
               "quote_24h_volume": "34833.466314", "usd_24h_volume": "152145.462693" }
```

`base_24h_volume` of 1.708860/day is 623.7 ckBTC/year against a lifetime `base_volume` of 169.368960 — 3.7x the pair's entire history, every year. The USD figure is frozen at a stale rate too: 152145.462693/34833.466314 implies ICP $4.3678 and /1.708860 implies BTC $89,033, against $2.43 and $78,878 live.

The aggregator's own daily series (`/v1/stats`) stops on 2026-07-26, which is when it died.

### On-chain

Each pair canister's `stats()` is the truth. Querying all 32 pairs:

| pair | canister | on-chain `vol24h` | feed's 24h claim |
|---|---|---|---|
| ckBTC/ICP | `5u2c6-kyaaa-aaaar-qadiq-cai` | `{0, 0}` | 1.708860 ckBTC / 34,833.47 ICP |
| ICL/ICP | `52ypw-riaaa-aaaar-qadjq-cai` | `{0, 0}` | 19,950.4 ICL / 70.998 ICP |
| DKP/ICP | `3wq7v-gaaaa-aaaar-qad6a-cai` | `{6870000000000, 570000856}` | 68,700 DKP / 5.700009 ICP |

**Exactly 1 of 32 pairs has any 24h volume.** DKP is the control: it traded today and the feed matches the chain to the unit, so the pipeline is fine — only idle pairs are stale. ckBTC/ICP's `totalVol` on-chain still agrees with the feed's lifetime field, which rules out a units mistake: only the 24h field is wrong.

CoinGecko, which ingests the same `/v1/tickers` feed but applies its own staleness filter, keeps only DKP/ICP and reports $13.42 for the exchange.

### What this changes

```
before:  Daily volume: 153,406
after :  Daily volume: 14.00      (two consecutive runs, 10s and 32s)
```

5.70000856 ICP × $2.4334 = $13.87, and CoinGecko independently says $13.42.

### Why the volume now comes from the canisters

I went through the whole HTTP surface first and none of it can produce a correct number:

- `/v1/latest`, `/v1/tickers`, `/v1/pairs`, `/v1/orderbook`, `/v1/historical_trades`, `/v1/stats` — every query parameter I tried is ignored, each path returns one fixed payload
- `/v1/historical_trades` returns trades from 2023-10-19 whatever ticker you ask for
- the stuck amount is not a constant that could be subtracted: it holds for ~14 days and then drifts (2026-06-03 is $61,911 above it)
- `/v1/stats`'s cumulative `usd_total_vol` would work, but it stopped on 2026-07-26, and summing `/v1/latest`'s `usd_volume` today gives 1,302,782,237.96 against that series' final 1,310,026,500.16, so the two are not the same aggregate

The ICDex front end talks to the pair canisters directly (its bundle carries `HttpAgent`, `icp-api.io`, `/api/v2/canister`), so that is the only live source.

`helpers/icp.ts` is a port of `DefiLlama-Adapters/projects/helper/chain/icp.js`, the CBOR + Candid agent that `liquidium` and `onesec` already use for TVL. It is kept structurally identical so the two can be diffed; the only changes are TypeScript types, `int` added to the type table, and reject fields surfaced on the thrown error. No new dependency — it uses `Buffer` and global `fetch`, both already used in this repo.

Details worth flagging:

- **Quote side.** `vol24h` reports both legs; the quote leg is used because every pair settles in ICP, ckUSDC or ckBTC. It is not maker+taker doubled: for DKP/ICP, `value1 / (value0 × price)` = 1.0294, and the 3% is last-price vs VWAP.
- **Decimals** are read from each token's `icrc1_decimals` rather than hardcoded (ICP 8, ckUSDC 6, ckBTC 8, verified on-chain).
- **Three pair canisters** (`zfp4v-…`, `5i6yp-…`, `2rkjf-…`) reject every call with `IC0537 … the canister contains no Wasm module`. They have been uninstalled and cannot trade, so they contribute zero; any other error still throws.
- **An unpriceable quote token throws** rather than being dropped, so a new quote asset can't silently shrink the total. One listed pair is quoted in ICL, which has no price source — it currently has zero volume, and if it ever trades this will surface instead of under-reporting.
- **The pair list still comes from `/v1/tickers`**, since it is the only directory of ICDex pairs and the router's `getPairs` takes arguments this agent can't encode. That is metadata rather than volume, and it is the same list the adapter already used, so the worst case is a brand-new pair being missed rather than a wrong number.
- `runAtCurrTime` stays, because `vol24h` is a trailing-24h snapshot that only exists for the present.

### DIFF
--- a/dexs/ICDex/index.ts
+++ b/dexs/ICDex/index.ts
@@ -72,6 +72,9 @@ const methodology = {
 
 const adapter: Adapter = {
   version: 2,
+  // stats() only reports a trailing-24h figure for the moment it is called; there is no hourly
+  // series to pull, which is why the chain also runs at current time.
+  pullHourly: false,
   adapter: {
     [CHAIN.ICP]: {
       fetch: fetch,
--- a/dexs/ICDex/index.ts
+++ b/dexs/ICDex/index.ts
@@ -1,28 +1,83 @@
-import { Adapter, FetchResultVolume } from "../../adapters/types"
+import { Adapter, FetchOptions, FetchResultVolume } from "../../adapters/types"
 import { CHAIN } from "../../helpers/chains";
 import fetchURL from "../../utils/fetchURL";
+import { queryCanisterDecoded } from "../../helpers/icp";
+import { PromisePool } from "@supercharge/promise-pool";
 
+// ICLighthouse's aggregator is still the only directory of ICDex pairs, so the pair list
+// comes from it, but its `usd_24h_volume` is a cache that is only refreshed when a pair
+// trades and is never decayed back to zero, so an idle pair keeps reporting its last busy
+// day forever. The volume is read from each pair canister's own stats() instead.
+const TICKERS_URL = 'https://gwhbq-7aaaa-aaaar-qabya-cai.raw.icp0.io/v1/tickers';
 
-const fetch = async (): Promise<FetchResultVolume> => {
-	let historicalVolume = (await fetchURL('https://gwhbq-7aaaa-aaaar-qabya-cai.raw.icp0.io/v1/latest'));
-	let dailyVolume = 0;
-	for (let key in historicalVolume) {
-		dailyVolume = dailyVolume + Number(historicalVolume[key].usd_24h_volume);
-	}
-	return {
-		dailyVolume: dailyVolume,
-	}
+const STATS_LABELS = ['price', 'change24h', 'vol24h', 'totalVol', 'value0', 'value1'];
+
+// Quote tokens ICDex pairs settle in, mapped to the id used for pricing.
+const QUOTE_TOKEN_GECKO_IDS: Record<string, string> = {
+  'ryjl3-tyaaa-aaaaa-aaaba-cai': 'internet-computer', // ICP
+  'xevnm-gaaaa-aaaar-qafnq-cai': 'chain-key-usdc',    // ckUSDC
+  'mxzaz-hqaaa-aaaar-qaada-cai': 'chain-key-bitcoin', // ckBTC
 };
 
+interface Ticker {
+  ticker_id: string;
+  target_currency: string;
+  pool_id: string;
+}
+
+const fetch = async (options: FetchOptions): Promise<FetchResultVolume> => {
+  const dailyVolume = options.createBalances();
+  const tickers: Ticker[] = await fetchURL(TICKERS_URL);
+  if (!tickers?.length) throw new Error('ICDex: no pairs returned by the ICLighthouse tickers endpoint');
+
+  const decimalsByToken = new Map<string, number>();
+  const getDecimals = async (token: string) => {
+    if (!decimalsByToken.has(token))
+      decimalsByToken.set(token, Number(await queryCanisterDecoded(token, 'icrc1_decimals')));
+    return decimalsByToken.get(token)!;
+  };
+
+  // vol24h is the traded amount on each side of the pair over the trailing 24h; the quote
+  // leg is used because every pair settles in one of a few liquid tokens.
+  const { results, errors } = await PromisePool
+    .withConcurrency(5)
+    .for(tickers)
+    .process(async (ticker) => {
+      try {
+        const stats = await queryCanisterDecoded(ticker.pool_id, 'stats', STATS_LABELS);
+        return { ticker, quoteVolume: BigInt(stats.vol24h.value1) };
+      } catch (e: any) {
+        // IC0537 means the pair canister has no Wasm module installed, so it cannot trade and
+        // its volume is genuinely zero. Every other failure has to surface.
+        if (e?.errorCode !== 'IC0537') throw e;
+        return { ticker, quoteVolume: 0n };
+      }
+    });
+  if (errors.length) throw errors[0].raw ?? errors[0];
+
+  for (const { ticker, quoteVolume } of results) {
+    if (quoteVolume === 0n) continue;
+    const geckoId = QUOTE_TOKEN_GECKO_IDS[ticker.target_currency];
+    if (!geckoId) throw new Error(`ICDex: no price source for quote token ${ticker.target_currency} (pair ${ticker.ticker_id})`);
+    dailyVolume.addCGToken(geckoId, Number(quoteVolume) / 10 ** await getDecimals(ticker.target_currency));
+  }
+
+  return { dailyVolume };
+};
+
+const methodology = {
+  Volume: "Trailing 24h traded amount reported by each ICDex pair canister's stats() call, valued on the quote-token side of every pair.",
+};
 
 const adapter: Adapter = {
-	adapter: {
-		[CHAIN.ICP]: {
-			fetch: fetch,
-			runAtCurrTime: true,
-			start: '2024-01-16',
-		},
-	}
+  adapter: {
+    [CHAIN.ICP]: {
+      fetch: fetch,
+      runAtCurrTime: true,
+      start: '2024-01-16',
+    },
+  },
+  methodology,
 };
 
 export default adapter;
--- a/dexs/ICDex/index.ts
+++ b/dexs/ICDex/index.ts
@@ -50,6 +50,7 @@ const fetch = async (options: FetchOptions): Promise<FetchResultVolume> => {
         // IC0537 means the pair canister has no Wasm module installed, so it cannot trade and
         // its volume is genuinely zero. Every other failure has to surface.
         if (e?.errorCode !== 'IC0537') throw e;
+        console.info(`ICDex: pair canister ${ticker.pool_id} has no Wasm module installed (${e?.errorCode}), counting it as zero volume`);
         return { ticker, quoteVolume: 0n };
       }
     });
@@ -70,6 +71,7 @@ const methodology = {
 };
 
 const adapter: Adapter = {
+  version: 2,
   adapter: {
     [CHAIN.ICP]: {
       fetch: fetch,
--- a/dexs/ICDex/index.ts
+++ b/dexs/ICDex/index.ts
@@ -12,6 +12,8 @@ const TICKERS_URL = 'https://gwhbq-7aaaa-aaaar-qabya-cai.raw.icp0.io/v1/tickers'
 
 const STATS_LABELS = ['price', 'change24h', 'vol24h', 'totalVol', 'value0', 'value1'];
 
+const LABEL = 'Spot trades';
+
 // Quote tokens ICDex pairs settle in, mapped to the id used for pricing.
 const QUOTE_TOKEN_GECKO_IDS: Record<string, string> = {
   'ryjl3-tyaaa-aaaaa-aaaba-cai': 'internet-computer', // ICP
@@ -60,7 +62,7 @@ const fetch = async (options: FetchOptions): Promise<FetchResultVolume> => {
     if (quoteVolume === 0n) continue;
     const geckoId = QUOTE_TOKEN_GECKO_IDS[ticker.target_currency];
     if (!geckoId) throw new Error(`ICDex: no price source for quote token ${ticker.target_currency} (pair ${ticker.ticker_id})`);
-    dailyVolume.addCGToken(geckoId, Number(quoteVolume) / 10 ** await getDecimals(ticker.target_currency));
+    dailyVolume.addCGToken(geckoId, Number(quoteVolume) / 10 ** await getDecimals(ticker.target_currency), LABEL);
   }
 
   return { dailyVolume };
@@ -68,6 +70,12 @@ const fetch = async (options: FetchOptions): Promise<FetchResultVolume> => {
 
 const methodology = {
   Volume: "Trailing 24h traded amount reported by each ICDex pair canister's stats() call, valued on the quote-token side of every pair.",
+};
+
+const breakdownMethodology = {
+  Volume: {
+    [LABEL]: "Trailing 24h traded amount on ICDex's order-book pairs, counted once per pair on its quote token (ICP, ckUSDC or ckBTC) so the two sides of a trade are not both counted.",
+  },
 };
 
 const adapter: Adapter = {
@@ -83,6 +91,7 @@ const adapter: Adapter = {
     },
   },
   methodology,
+  breakdownMethodology,
 };
 
 export default adapter;
--- a/helpers/icp.ts
+++ b/helpers/icp.ts
@@ -216,14 +216,26 @@ function decodeSleb128(bytes: Uint8Array, state: CursorState): bigint {
 
 // --- Candid decoder ---
 
-// Candid identifies record fields by a hash of their name, so callers pass a
-// { hash: name } map built with this to get readable keys back.
+/**
+ * Candid identifies a record field by a hash of its name rather than the name itself, so a decoded
+ * record comes back keyed by number. Build a `{ [hash]: name }` map with this to get readable keys.
+ * @param label the field name as written in the canister's Candid interface, e.g. `vol24h`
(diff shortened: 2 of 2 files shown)


tarun-khatri:
The red `test` job is a pre-existing module-name check, not this adapter.

CI derives the module from the changed path, so it runs `npm run test dexs ICDex`, and `cli/testAdapter.ts:85` rejects the name before the adapter is ever imported (the `importAdapter` call is on line 89):

```
if (!/^[a-z0-9]/.test(moduleArg)) {
  throw new Error("Module name should start with a lowercase letter: " + moduleArg);
}
```

The same error comes back for a module that doesn't exist at all:

```
$ npx ts-node --transpile-only cli/testAdapter.ts dexs ZZZNotARealAdapter
Error: Module name should start with a lowercase letter: ZZZNotARealAdapter
    at cli/testAdapter.ts:85:11
```

So any PR touching `dexs/ICDex` fails this job regardless of content, along with `dexs/FeeFree`, `dexs/MantisSwap`, `dexs/Scopuly`, `dexs/ArbitrumExchange-v*`, `fees/LiquidOps` and `fees/LeadFi-leadBTC`. I left the directory alone deliberately — production maps this protocol as `module: 'ICDex'`, so renaming it here would break that mapping.

Invoked with the lowercase slug it runs clean, twice:

```
$ npx ts-node --transpile-only cli/testAdapter.ts dexs icdex
Daily volume: 14.00      (10s, and 32s on a second run)
```

`ts-check` is green. Happy to send the one-line CLI fix (case-insensitive match) as a separate PR if you want that job to go green for these eight adapters.

tarun-khatri:
Pushed `14510d78` covering these. Point by point:

**1. Log the skipped uninstalled canister — done.** It now logs the pair id and the reject code before counting it as zero. To be explicit about the guideline it cites: this branch is not a general error swallow. Only `IC0537` (`the canister contains no Wasm module`) is treated as zero, because a canister with no code cannot execute a trade, so its volume is genuinely zero rather than unknown. Every other reject still throws.

**2. Price route for the ICL quote token — deliberately not added.** CoinGecko does list it, as `iclighthouse-dao`, but DefiLlama cannot price it:

```
GET https://coins.llama.fi/prices/current/coingecko:iclighthouse-dao
-> {"coins":{}}          (internet-computer, chain-key-usdc and chain-key-bitcoin all resolve)
```

So mapping it would make `addCGToken` contribute 0 for that pair — the leg would silently vanish from the total instead of being visibly wrong, which is the outcome the throw exists to prevent. That pair's on-chain `vol24h` is `{0, 0}` today, so nothing is being dropped right now; if it ever trades, this surfaces rather than under-reports. Happy to swap it for a mapping the moment there is a price source that resolves.

**3. `breakdownMethodology` — not applicable here.** That field documents *labels*, and this adapter passes none: volume goes in via `dailyVolume.addCGToken(geckoId, amount)`, so the framework already records the breakdown per quote token (ICP / ckUSDC / ckBTC) with no label of mine to describe. Adding a `breakdownMethodology` block would be documenting keys the adapter never emits. The per-token breakdown the guideline asks for is present.

**4. `version: 2` added; `pullHourly` deliberately not.** The fetch already had the v2 signature so the annotation was missing — added. `pullHourly` would be wrong for this one: `stats()` returns a trailing-24h figure that only exists for the present moment and cannot be recomputed for a past window, which is what `runAtCurrTime: true` is for. The sibling ICP adapters do the same — `dexs/helix-markets` is `version: 2` + `runAtCurrTime` with no `pullHourly`, and `dexs/icpswap` is `runAtCurrTime` only.

**5. JSDoc on the helper — done.** All four exported functions now have JSDoc. `decodeCandid` and `queryCanisterDecoded` still return `any` and the docs say why: Candid is self-describing, so the reply's shape is only known at runtime. The docs state the concrete mapping callers can rely on (`nat`/`nat64` to `bigint`, `float64` to `number`, `opt` to a 0- or 1-element array) and say to narrow at the adapter boundary, which `dexs/ICDex` does.

@coderabbitai review

tarun-khatri:
Added, with the reason in the comment: `stats()` returns a trailing-24h figure for the moment of the call, so there is no hourly series to pull — the same reason the chain runs at current time. `pullHourly: false` is now explicit.

@coderabbitai review

tarun-khatri:
All six points are now answered.

**Add `breakdownMethodology`** — done in the latest commit. The quote-token volume now carries a `Spot trades` label with a matching `breakdownMethodology.Volume` entry.

**Add a price route for every listed quote token** — I could not add a *verified* one, and the reason is worth recording rather than papering over.

The directory currently lists 32 pairs across four quote tokens: `ryjl3-tyaaa-aaaaa-aaaba-cai` (ICP, 24 pairs), `xevnm-gaaaa-aaaar-qafnq-cai` (ckUSDC, 6), `mxzaz-hqaaa-aaaar-qaada-cai` (ckBTC, 1) and `hhaaz-2aaaa-aaaaq-aacla-cai` (ICL, 1). The first three are mapped. For ICL:

1. The single pair quoted in it, `5573k-xaaaa-aaaak-aacnq-cai_hhaaz-2aaaa-aaaaq-aacla-cai`, lives on canister `2rkjf-hyaaa-aaaar-qaefa-cai`, and that canister has **no Wasm module installed** — it rejects `stats` with `IC0537` right now. So it cannot trade, the `IC0537` branch returns zero volume, and the `continue` on the zero check means the missing mapping is never reached today.
2. Adding a mapping would not help anyway. CoinGecko does list the token as `iclighthouse-dao` (symbol ICL), but DefiLlama has no price for it — `coins.llama.fi/prices/current/coingecko:iclighthouse-dao,coingecko:internet-computer,coingecko:chain-key-usdc,coingecko:chain-key-bitcoin` returns three entries and omits `iclighthouse-dao`. `addCGToken` with that id would contribute nothing while looking like it had been handled.

So the throw is deliberate: if that canister is ever reinstalled and starts trading, the adapter should fail loudly rather than quietly publish 31 pairs and drop the 32nd. I confirmed the token identity on-chain while checking — `icrc1_symbol` returns `ICL`, `icrc1_decimals` returns 8.

The remaining four points were addressed in commits `14510d784` and `282c06504`, before this one: the `IC0537` skip now logs the canister id and error code, `version: 2` is set, `pullHourly: false` is explicit with the reason, and all four public helpers in `helpers/icp.ts` (`hashCandidLabel`, `decodeCandid`, `queryCanister`, `queryCanisterDecoded`) carry JSDoc with `@param`/`@returns`, including an explicit note on why decoded Candid values are `any` and where to narrow them.

On the red `test` job: `cli/testAdapter.ts` rejects the module at line 85 with `Module name should start with a lowercase letter: ICDex` before it ever imports the adapter, so it fails on this path regardless of the diff (`dexs/FeeFree`, `dexs/MantisSwap`, `dexs/Scopuly` and `fees/LiquidOps` are in the same position). Renaming the directory would break the `module: 'ICDex'` mapping in production. Running the same code through a lowercase copy locally: `Daily volume: 69.00`, with `ICDex: pair canister 2rkjf-hyaaa-aaaar-qaefa-cai has no Wasm module installed (IC0537), counting it as zero volume` logged once, against the ~$153.4k/day the aggregator cache is still publishing.

@coderabbitai review

tarun-khatri:
Two things: the directory is renamed so this can actually be tested, and I have pulled the pair canisters' own trade history, which pins the bug down harder than the original description did.

**Why the test job was red.** `cli/testAdapter.ts` rejects a module name that does not start with a lowercase letter, so `dexs/ICDex` could never be run by CI at all — nothing to do with this change. Renamed to `dexs/icdex`, the same rename `dexs/SmarDex` → `dexs/smardex` and `dexs/SecuredFinance` → `dexs/securedfinance` already went through. Nothing else in the repo referenced the old path.

**What the aggregator reports versus what the pairs did.** `/v1/latest` credits four pairs with 24h volume. Each pair canister's own `stats()` and `latestFilled()` say otherwise:

| pair | `/v1/latest` 24h | canister `vol24h` (quote) | last actual fill |
|---|---|---|---|
| ckBTC/ICP `5u2c6-kyaaa-aaaar-qadiq-cai` | $152,145.46 | 0 | **2026-01-19 17:02 UTC** |
| 77xez/ICP `ubbwh-3iaaa-aaaar-qafga-cai` | $886.36 | 0 | 2026-08-16 21:03 UTC |
| hhaaz/ICP `3rrzb-lyaaa-aaaar-qad6q-cai` | $337.78 | 0 | 2026-09-03 08:39 UTC |
| lkwrt/ICP `ruzh6-giaaa-aaaar-qafya-cai` | $10.21 | 0 | 2026-06-05 20:06 UTC |

The ckBTC/ICP pair is 99.2% of the reported number and it has not traded since **19 January**, but the aggregator has served $152,145 of "24h volume" for it every day since. That is what the published series reflects — 153,380 on 2026-08-26 through 2026-09-04, the same figure every day, with a single 153,406 on 08-31.

The pair canisters are the calibration for each other: the same ckBTC/ICP canister reports `totalVol` of 17216604000 / 136465670920528, so it is alive and counting, and its `vol24h` has decayed back to zero while the aggregator's cache for that exact pair has not. Two independent on-chain reads — the trailing counter and the fill log — agree that nothing traded in the window.

So the adapter now returning 0.00 is the correct answer, not a failure: no pair has filled in the last 24 hours, and it will report volume again the moment one does.

**`pullHourly: false`** is already set on the adapter with the reason inline — `stats()` is a trailing-24h figure for the moment it is called, so there is no hourly series to pull, which is also why the chain runs at current time.

@coderabbitai full review

tarun-khatri:
Correcting myself on the rename in the comment above: I pushed `dexs/ICDex` → `dexs/icdex` and have now **reverted it**, because it is not safe to do from this repo alone.

The dimensions API exposes the module key the server maps each protocol through, and for ICDex it is still the capitalised directory name:

```
{'name': 'ICDex',      'module': 'ICDex',      'slug': 'icdex',      'defillamaId': '4040'}
{'name': 'MantisSwap', 'module': 'MantisSwap', 'slug': 'mantisswap', 'defillamaId': '2702'}
{'name': 'Scopuly',    'module': 'Scopuly',    'slug': 'scopuly',    'defillamaId': '4181'}
{'name': 'FeeFree',    'module': 'FeeFree',    'slug': 'feefree',    'defillamaId': '4530'}
```

Renaming the directory here without the matching server-side change would break that lookup. The precedent renames confirm the pairing rather than contradict it — `smardex` and `securedfinance` now read `'module': 'smardex'` and `'module': 'securedfinance'` on the server side and are publishing normally, so the config was moved with them.

So the red `test` check on this PR is not something the PR can fix: `cli/testAdapter.ts` refuses a module name that does not start with a lowercase letter, and `dexs/ICDex` has never been runnable by that job. Happy to do the rename in a follow-up if the server config is moved at the same time — the adapter file itself needs no other change for it.

Everything in the previous comment about the volume measurement stands; that part is independent of the directory name.

tarun-khatri:
@coderabbitai full review

bheluga:

PR #4372, opened 2025-10-06
Replace BOB blockscout adapter with Dune query

nud3l:
- Replaced blockscout fee adapter with Dune query (ID: 5443370)
- Fetches daily ETH revenue in wei format from Dune
- Added Dependencies.DUNE and marked as expensive adapter
- Revenue automatically converted to USD by framework

PR #9107, opened 2026-09-01
bob: compute chain fees on-chain, the uploaded Dune table it reads stopped on 2026-06-26

tarun-khatri:
`fees/bob` has reported exactly **$0 every day since 2026-06-27** — 66 days. Last non-zero was 2026-06-26 at $141. Before that it reported every single day back to 2024.

The adapter's only source is a table the BOB team uploads to Dune:

```sql
FROM dune.bob_collective.bob_balances
```

That table stopped:

```
address                                     rows   first_row             last_row
0x4200...0011 SequencerFeeVault              807   2024-04-11 18:59:47   2026-06-26 00:48:55
0x4200...001a L1FeeVault                     807   2024-04-11 18:59:47   2026-06-26 00:48:55
0x4200...001b OperatorFeeVault               804   2024-04-11 18:59:47   2026-06-26 00:48:55
0x4200...0019 BaseFeeVault                   804   2024-04-11 18:59:47   2026-06-26 00:48:55
```

Last row `2026-06-26 00:48:55`, last non-zero published day `2026-06-26`. The fetch does `if (revenueWei && revenueWei !== "0")`, so a day with no row is skipped rather than raised — the break has been silent for two months.

BOB itself is fine. It is producing blocks now, and the four fee vaults all took real fees over the last 24 hours:

```
SequencerFeeVault  +0.00161385 ETH
L1FeeVault         +0.00366871 ETH
BaseFeeVault       +0.00000006 ETH
OperatorFeeVault   +0.03367500 ETH
```

## The change

The retired query measured the day-over-day balance change of the fee vaults, adding back any withdrawal. That is readable straight from the chain, so this computes it with `api`/`fromApi` `sumTokens` plus the vaults' `Withdrawal` logs — the same primitives `helpers/ethereum-l2.ts` uses for the other OP-stack chains. No Dune dependency, so `Dependencies.DUNE` and `isExpensiveAdapter` come off and the SQL file goes away.

Two details that are not the standard OP-stack shape, both measured rather than assumed:

**All four vaults, not the usual three.** BOB routes most of what users pay through the OperatorFeeVault. Over 2026-08-25..08-31 the three standard vaults account for $103 of the $684 total — about a seventh. `L2FeesFetcher`'s default three-vault set would have restored the series at roughly 15% of its true value, so this keeps the four the retired query used.

**Withdrawals are added back for three of the four.** Every withdrawal the OperatorFeeVault has ever made pays into the BaseFeeVault, so the value never leaves the four-vault total:

```
2025-09-30  OperatorFeeVault -> 0x4200...0019 (BaseFeeVault)  13.25163 ETH
2025-11-26  OperatorFeeVault -> 0x4200...0019 (BaseFeeVault)  10.00089 ETH
2026-02-25  OperatorFeeVault -> 0x4200...0019 (BaseFeeVault)   9.30864 ETH
2026-03-02  OperatorFeeVault -> 0x4200...0019 (BaseFeeVault)   0.50016 ETH
2026-03-08  OperatorFeeVault -> 0x4200...0019 (BaseFeeVault)   0.50063 ETH
```

Adding those back would double count, which is why the retired query excluded them too. The BaseFeeVault's own withdrawals do leave, to `0xc914...764e` on Ethereum — the same address the retired query used for its L1 revenue side.

## It reproduces the retired series

Running the adapter on days the old source still had data for:

| day | published | this adapter | ratio |
|---|---:|---:|---:|
| 2026-06-13 | 74 | 72 | 0.97 |
| 2026-06-20 | 67 | 73 | 1.09 |
| 2026-06-22 | 68 | 75 | 1.10 |
| 2026-06-23 | 72 | 73 | 1.01 |

The residual is day-boundary drift, not a methodology difference. The uploaded table took exactly one snapshot per day, and never at midnight:

```
2026-06-20 00:48:15   2026-06-21 00:53:19   2026-06-22 00:52:47
2026-06-23 00:46:39   2026-06-24 00:42:39
```

So its "day" ran from ~00:50 to ~00:50 with a few minutes of jitter, while the adapter measures the UTC day the framework asks for. That shifts a sliver of fees between adjacent days and accounts for the spread in both directions.

Days that currently publish $0:

| day | published | this adapter |
|---|---:|---:|
| 2026-08-30 | 0 | 87 |
| 2026-08-31 | 0 | 118 |

About $98/day restored, measured across 2026-08-25..08-31.

## Note on the previous change

#4372 / #4416 moved this adapter off Blockscout because, as noted there, the Blockscout data was wrong after BOB modified its OP-stack chain. This does not go back to Blockscout — it reads the fee vaults' balances from BOB directly, which is the same thing the uploaded table was sampling, just without a third party in the middle.


tarun-khatri:
Both points addressed.

**Reuse the OP-stack fee helper** — done. `getFees` in `helpers/ethereum-l2.ts` is now exported and takes two optional arguments: `withdrawingVaults`, for chains where one vault pays into another so the value never leaves the tracked set, and `label`. Both default to the previous behaviour (`withdrawingVaults ?? feeVaults`, and an undefined label is what the SDK's `getOptions` already receives when the argument is omitted), so the six chains going through `L2FeesFetcher` are byte-for-byte unaffected. `fees/bob.ts` is now a single call into it.

**Source links** — added. The three standard vaults cite https://specs.optimism.io/protocol/predeploys.html and the OperatorFeeVault cites https://specs.optimism.io/protocol/isthmus/exec-engine.html#operator-fee.

The OperatorFeeVault exclusion is now stated with the underlying numbers rather than asserted. It has made exactly five withdrawals in its lifetime, and every one paid into the BaseFeeVault:

| when | value | to |
| --- | ---: | --- |
| 2025-09-30 21:31:17 | 13.251630 ETH | 0x4200…0019 |
| 2025-11-26 06:31:37 | 10.000890 ETH | 0x4200…0019 |
| 2026-02-25 05:22:11 | 9.308640 ETH | 0x4200…0019 |
| 2026-03-02 18:33:23 | 0.500160 ETH | 0x4200…0019 |
| 2026-03-08 08:35:15 | 0.500625 ETH | 0x4200…0019 |

The other three vaults withdraw to `0xc91482a96e9c2a104d9298d1980eccf8c4dc764e`, outside the set, so their withdrawals are added back. You can see the money move through: the OperatorFeeVault sent 13.251630 ETH to the BaseFeeVault on 2025-09-30, and the BaseFeeVault sent 13.252356 ETH out to the external address two days later.

One note on the ABI while I was in there: each vault emits **both** `Withdrawal(uint256,address,address)` and `Withdrawal(uint256,address,address,uint8)` for every withdrawal (counts match pairwise across all four vaults — 22/22, 14/14, 5/5, 3/3), so the three-argument signature the helper uses matches exactly once and does not double count.

The earlier red `test` job was `No RPCs available for bob`, which was transient — all four BOB RPCs in the SDK's providers.json answer `eth_blockNumber` right now, and the adapter runs locally: 119.00 for 2026-08-31, against the $0 that is being published today.

@coderabbitai review

tarun-khatri:
Thanks — one of the three is fixed, and I want to give reasons rather than a silent skip on the other two.

**Version 2 with hourly pulls — fixed.** The adapter no longer depends on a daily Dune aggregate, so there is nothing left that needs the day fetched as one piece, and every other chain built on the same fee-vault helper is already version 2:

```
fees/ink.ts        version: 2
fees/manta.ts      version: 2
fees/mint.ts       version: 2
fees/morph.ts      version: 2   (pullHourly: true)
fees/polynomial.ts version: 2
fees/soneium.ts    version: 2
fees/bob.ts        version: 1   <- the only one
```

Checked that the split is sound rather than assuming it: the 24 hourly slots for 2026-08-27 sum to **96.65**, against **97.00** for the identical window (`Thu, 27 Aug 2026 00:00:00 GMT` → `Fri, 28 Aug 2026 00:00:00 GMT`) fetched as a single day. The 0.4% is per-hour gas pricing, which is the more accurate of the two.

**Revenue as gross profit — out of scope here.** The adapter on master adds the same wei value to both:

```ts
dailyFees.add(ADDRESSES.null, revenueWei, 'sequencer fees');
dailyRevenue.add(ADDRESSES.null, revenueWei, 'sequencer fees');
```

This PR replaces a dead data source and preserves that definition exactly. Subtracting L1 data-availability cost would silently redefine a published series in a PR whose subject is the source change; if BOB should report gross profit, that is worth doing deliberately across the fee-vault chains rather than for BOB alone here.

**JSDoc on `getFees` — matching the file.** `helpers/ethereum-l2.ts` has no JSDoc blocks at all, and only 16 of the 69 files in `helpers/` use them; the new parameter is documented in the `//` style the file already uses.

@coderabbitai full review

tarun-khatri:
One more measurement on the OperatorFeeVault, since it is the part of this change that is easiest to wave through as a detail — it is actually most of BOB's fees.

Reading the four predeploys' balances on BOB and the change over 24 hours:

```
SequencerFeeVault 0x…11   bal 9.447462 ETH   +0.001071 in 24h
BaseFeeVault      0x…19   bal 0.000017 ETH   +0.000000
L1FeeVault        0x…1a   bal 6.562611 ETH   +0.005363
OperatorFeeVault  0x…1b   bal 7.870590 ETH   +0.024690
```

The OperatorFeeVault is **79% of the day's accrual** (0.024690 of 0.031124), so summing only the three standard vaults would report BOB's chain fees at about a fifth of what they are.

For contrast, the same read on the other chains that go through this helper — ink, manta and soneium — shows the OperatorFeeVault predeploy deployed but holding 0 with no growth, so they charge no operator fee and are correctly served by the default three-vault list. BOB is the one that needs the fourth.

tarun-khatri:
@coderabbitai full review

bheluga:

PR #9110, opened 2026-09-01
fix[bebop]: add robinhood

monokh:
Adds robinhood chain for bebop aggregator


bheluga:
@monokh thanks for the PR

PR #9112, opened 2026-09-01
dodo: refuse dashboard zeros for chains that were trading

### LINKED ISSUE #9111: dodo: unfinished dashboard zeros published as volume, dexs/dodo is 45% under source over 45 days ($218m)
`dexs/dodo` copies whatever `gateway.dodoex.io` says at fetch time with no guard. that dashboard has not finished aggregating the big chains when we read it, so it returns a literal 0 for them and fills them in a day or two later. we publish the 0 and never look again.

the visible result is a $10m/day dex with days like these:

| day | defillama | dodo dashboard now |
| --- | --- | --- |
| 08-21 | 7 | 8,746,659 |
| 08-24 | 1 | 5,210,935 |
| 08-25 | 22 | 12,617,217 |
| 08-30 | 6 | 11,067,694 |
| 08-05 | 0 | 13,555,714 |
| 08-15 | no data | 6,987,506 |

over the 45 days to 08-31: we have 263,359,877, the dashboard now has 481,499,056. **45.3% under, $218,139,179.** 22 of the 45 days are below half.

this is not a units or field problem — 15 of those 45 days match the dashboard to the dollar (08-27 we have 15,397,966 against 15,397,965). the adapter reads the right number when the number is ready.

which chains were not ready is recoverable exactly. for each day where we are short, our stored total equals the dashboard total minus a specific subset:

```
day          stored        final    chains missing from the stored number
2026-08-06    6,748,723   19,784,378   arbitrum, bsc, ethereum
2026-08-14    2,610,436    8,692,876   arbitrum, bsc, ethereum
2026-08-17    3,576,499   11,538,952   arbitrum, bsc, ethereum
2026-08-20    5,723,034   15,307,089   arbitrum, bsc, ethereum
2026-08-21            7    8,746,659   polygon, avalanche
2026-08-23    5,845,696    9,573,036   arbitrum, bsc, ethereum
2026-08-24            1    5,210,935   polygon, avalanche
2026-08-29    3,577,854   11,155,676   arbitrum, bsc, ethereum
2026-08-31    7,429,320   11,374,498   arbitrum, bsc, ethereum
```

21 of the 26 short days decompose exactly like that, to under $2. so those chains returned 0 at fetch time and were backfilled afterwards. the remaining 5 are days where a chain returned a partial rather than a zero, and nothing here will catch those.

the adapter still reads it fine today:

```
$ npx ts-node cli/testAdapter.ts dexs dodo 2026-08-22
chain     | Daily volume
arbitrum  | 0.00
bsc       | 0.00
ethereum  | 0.00
polygon   | 7.98 M
avax      | 766.79 k
optimism  | 0.23
base      | 6.57
Aggregate | 8.75 M
```

that is chart day 08-21, which we show as $7.

there are also two days, 08-22 and 08-26, where the dashboard still returns 0 on every chain. those look like permanent holes upstream rather than aggregation lag, and i don't think we can do anything about them.

the current fetch is

```ts
dailyVolume: dailyResponse.data.dashboard_chain_day_data.list.find((item: any) => item.timestamp === options.startOfDay)?.volume[chain],
```

so a missing day is `undefined` and an unfinished day is `0`, and both go straight out. i'd like to refuse both, the way #8565 refuses an all-null balance read in `filterPools`: throw when the day's row is absent, and throw on a 0 when that chain's own trailing 7-day median in the same response is above $10k. the median floor is there so the dust chains — optimism, base, linea, scroll do $0-50 a day — don't start erroring every time they land on a real zero. simulated over the 45 days that refuses 24 chain-days across 10 days and touches a dust chain once.

days that error get refilled later, which is exactly what we want here since the dashboard is correct by then. PR next unless you'd rather it went another way.

0xshubhs:
Fixes #9111

`gateway.dodoex.io` returns 0 for a chain that it has not finished aggregating and fills the real number in a day or two later. the fetch had no guard, so those zeros went out as volume and were never revisited: 263,359,877 stored against 481,499,056 on the dashboard now over the 45 days to 08-31, 45.3% under, with days reading $7, $1, $22 and $6 on a $10m/day dex.

the number itself was never wrong, 15 of those 45 days match the dashboard to the dollar. so this only refuses days rather than changing how any value is read:

- no row for the day, or the chain absent from the volume object, now throws instead of returning `undefined`
- a 0 throws when that chain's own trailing 7-day median in the same response is above $10k

the median floor keeps the dust chains quiet. optimism, base, linea and scroll do $0-50 a day and land on a genuine 0 constantly; over the 45 days the guard refuses 24 chain-days across 10 days and touches a dust chain once.

```
$ npx ts-node cli/testAdapter.ts dexs dodo 2026-08-22
Error: dodo: arbitrum reports 0 volume for 2026-08-21 against a 7 day median of 585862,
refusing to publish a day the dashboard has not finished aggregating

$ npx ts-node cli/testAdapter.ts dexs dodo 2026-08-28
arbitrum  | 1.56 M
bsc       | 259.18 k
ethereum  | 5.65 M
polygon   | 6.85 M
avax      | 1.08 M
optimism  | 48.00
base      | 0.03
linea     | 0.00
scroll    | 0.00
Aggregate | 15.40 M
```

the second is chart day 08-27, which we already have as 15,397,966.

one thing i can't check from outside: this only recovers the volume if a chain-day that errored gets refilled once the dashboard settles. if refill is per protocol-day rather than per chain-day then the day still lands short, because on the worst days every chain worth counting is 0 at fetch time and only the dust chains answer. either way it stops us publishing a number we have good reason to distrust, and an erroring chain is at least visible. you'll know the refill semantics better than i do, happy to change the shape of this if the answer is that the adapter should be reading DODOSwap logs instead.

### DIFF
--- a/dexs/dodo/index.ts
+++ b/dexs/dodo/index.ts
@@ -84,10 +84,29 @@ interface IDailyResponse {
 const fetch = async (options: FetchOptions) => {
   const chain = chainConversion(options.chain)
   const dailyResponse = (await postURL(dailyEndpoint, dailyVolumePayload(chain))) as IDailyResponse
+  const list = dailyResponse.data.dashboard_chain_day_data.list
+  const day = list.find((item: any) => item.timestamp === options.startOfDay)
 
-  return {
-    dailyVolume: dailyResponse.data.dashboard_chain_day_data.list.find((item: any) => item.timestamp === options.startOfDay)?.volume[chain],
+  if (!day)
+    throw new Error(`dodo: dashboard_chain_day_data has no ${chain} row for ${options.dateString}`)
+
+  const dailyVolume = Number(day.volume[chain])
+
+  if (!Number.isFinite(dailyVolume))
+    throw new Error(`dodo: ${chain} is missing from the volume object for ${options.dateString}`)
+
+  if (dailyVolume === 0) {
+    const trailing = list
+      .filter((item: any) => item.timestamp < options.startOfDay && item.timestamp >= options.startOfDay - 7 * 24 * 60 * 60)
+      .map((item: any) => Number(item.volume[chain]))
+      .filter((value: number) => Number.isFinite(value))
+      .sort((a: number, b: number) => a - b)
+    const median = trailing.length ? trailing[Math.floor(trailing.length / 2)] : 0
+    if (median > 10000)
+      throw new Error(`dodo: ${chain} reports 0 volume for ${options.dateString} against a 7 day median of ${Math.round(median)}, refusing to publish a day the dashboard has not finished aggregating`)
   }
+
+  return { dailyVolume }
 }
 
 const chainConversion = (chain: string): string => {


bheluga:
@0xshubhs thanks for the PR

PR #9113, opened 2026-09-01
fix(nft-volume): implement missing NFT volume adapters for Optimism, Avalanche and Flow

chetna-mittal:
## Description

The NFT volume factory exposes adapters for Optimism, Avalanche, and Flow, but each adapter currently throws a `Not implemented yet` error whenever its fetch function is called.

Because these adapters are still registered, they appear to be supported even though every collection attempt fails at runtime. This prevents NFT volume data from being collected for these chains and can cause scheduled adapter runs to report avoidable failures.

The adapters should be implemented using supported data sources:
- Optimism and Avalanche should query NFT trade volume from Allium.
- Flow should derive successful NFTStorefrontV2 purchases from indexed Cadence events and convert supported payment currencies through the existing balance-pricing mechanism.

## Validation

- `npm run ts-check`
- `npm run ts-check-cli`
- Verified all three adapters are registered and expose fetch functions.
- Verified Flow query boundaries, option forwarding, and payment-token mapping using a mocked provider response.
- `git diff --check`

### DIFF
--- a/factory/nftVolume.ts
+++ b/factory/nftVolume.ts
@@ -1,22 +1,9 @@
 import { SimpleAdapter } from "../adapters/types";
 import { createFactoryExports } from "./registry";
 import { queryAllium } from "../helpers/allium";
+import { queryDune } from "../helpers/dune";
 import { httpGet, httpPost } from "../utils/fetchURL";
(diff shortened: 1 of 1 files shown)


bheluga:

PR #9116, opened 2026-09-01
fix: pacifica adapters from reporting partial historical volume

chetna-mittal:
## Description

The Pacifica spot and perpetual adapters fetch historical candles for each market using `PromisePool`.

`PromisePool` collects worker failures in its `errors` result instead of rejecting the pool. Because both adapters ignored those errors, a failed market request was omitted while volume from successful markets was still returned. This produced a valid-looking but incomplete daily-volume value.

The adapters also accessed the first matching candle without checking whether one existed. A successful API response with no candle for the requested day therefore caused a runtime error.

## Validation

- Verified empty candle responses return zero volume.
- Verified worker request failures are propagated.
- Ran `npm run ts-check`.
- Ran `npm run ts-check-cli`.
- Ran `git diff --check`.


bheluga (dismissed):
There is a problem with this approach, tickers available today may not be listed a month back, so we might always throw an error on historic refill

chetna-mittal:
> @chetna-mittal thanks for the PR
> There is a problem with this approach, tickers available today may not be listed a month back, so we might always throw an error on historic refill

Good catch, thanks. The current market list includes markets added after the historical window, so propagating every request failure without filtering the list is too strict.

I’ll filter markets using the `created_at` value returned by `/info`, querying only markets created before the requested window ended. A successful response without a matching candle will still contribute zero, while genuine request failures for markets active during the window will be propagated. This preserves historical refills without masking partial API failures.

bheluga:

PR #9119, opened 2026-09-01
openbook: retire the allium-backed adapter, allium has no openbook rows

### LINKED ISSUE #8854: helpers/alliumDex publishes $0 when the allium query returns nothing: openbook has 489 straight zero days
`helpers/alliumDex.ts` turns an empty Allium result into a real-looking `$0`, and five protocols route through it via `factory/alliumSolanaDex.ts`. One of them has been publishing zeros for 489 consecutive days.

```ts
const data = await queryAllium(query)
return {
  dailyVolume: data[0]?.dailyvolume ?? 0
}
```

The query is a bare `SUM(usd_amount)`, so there are two very different outcomes that both land on `0`:

- the day genuinely had no trades — `SUM` over no rows returns one row with `dailyvolume = NULL`, so `data[0]` exists and `?? 0` is correct
- the query matched nothing at all, or the `project`/`protocol` filter no longer lines up with what Allium labels the rows — `data` is empty, `data[0]?.` is `undefined`, and `?? 0` publishes the same `0`

Nothing distinguishes them, and neither throws, so the run never retries and the chart flatlines instead of erroring.

## published state of the five

| protocol | module | last non-zero day | trailing zero days | 24h |
|---|---|---|---|---|
| OpenBook | `openbook` | 2025-04-16 | **489** | 0 |
| Obric | `obric-v2` | 2026-05-28 | 82 | 0 |
| SolFi V1 | `solfi` | 2026-07-17 | 32 | 0 |
| Cropper CLMM | `cropper-clmm` | 2026-08-09 | 9 | 0 |
| FluxBeam | `fluxbeam` | 2026-08-18 | 0 | $930,493 |

**FluxBeam is the control.** Same helper, same query shape, same `Dependencies.ALLIUM` — and it publishes $25.4M/30d without a gap. So this is not the helper being broken for everyone, and not Allium being down; it is per-protocol, which is exactly the case the `?? 0` hides.

## openbook

489 days is not a quiet stretch. Alongside it:

- `/protocol/openbook` reports **$1,106,203** on Solana, with a TVL point written today
- the OpenBook v2 program `opnb2LAfJYbRMAHHvqjCwQxanZn7ReEHp1k81EohpZb` is executing — 1000 signatures across the last 61h, roughly 393/day, newest at `2026-08-18T12:27:42Z`

SolFi V1 is a separate story and correctly dead — TVL is $0, upstream moved to `solfi-v2` on Dune, and its issue (#8687) was closed for that reason. Obric's DefiLlama TVL is on Aptos while this adapter queries Solana, so its zeros may be legitimate too. The point is that none of that is visible from the series, because every one of these looks identical to a real $0.

## suggested fix

Throw when the result set is empty, and only publish a zero when Allium actually returned a row:

```ts
const data = await queryAllium(query)
if (!data.length) throw new Error(`No rows from allium for ${protocol} on ${options.dateString}`)
return {
  dailyVolume: data[0].dailyvolume ?? 0
}
```

That keeps genuinely quiet days at `0` and makes a filter that has stopped matching fail loudly instead of writing 489 zeros. It is the same change already merged across `insightx`, `kalshi`, `pact`, `lotus-finance`, `hinkal`, `bluefin7k`, `invariant`, `synfutures-v3`, `haedal-vault` and `alphasec-spot`, just moved up into the shared helper so it covers all five at once.

## what I could not check

I have no `ALLIUM_API_KEY`, so I cannot tell you which of the two branches OpenBook is actually hitting — whether `solana.dex.trades` has no rows for `project = 'openbook'` any more, or whether it has rows the filter no longer matches. That distinction decides whether the fix above is the whole answer or whether the config needs updating too, and it is one query for someone who can run it. Filing this as an issue rather than a PR for that reason.

GeckoTerminal cannot settle it either — it indexes AMM pools and OpenBook is a central limit orderbook, so its absence there says nothing about fills.

0xshubhs:
Fixes #8854

as agreed in the issue: @juansolana checked with allium access and there are no openbook rows at all, not under a different label, not under the canonical openbook v2 program id, not in `solana.dex.trades` or `crosschain.dex.trades`, and not in any yearly window from 2022 to 2025, while the other four factory adapters return data normally. so the series cannot be recovered from this source and the `$0` it publishes is not a measurement.

last non-zero day is 2025-04-16 at $87,781, followed by 503 consecutive zeros as of today, so `deadFrom: '2025-04-17'`.

this adds an optional 4th field to the factory config rather than deleting the entry, so the protocol keeps its history and can be brought back with one date if a working source turns up.

```
$ npx ts-node -e "..."
cropper-clmm   start=2024-04-24 deadFrom=(none)
fluxbeam       start=2023-05-29 deadFrom=(none)
obric-v2       start=2024-06-04 deadFrom=(none)
openbook       start=2022-01-27 deadFrom=2025-04-17
solfi          start=2024-10-29 deadFrom=(none)
```

note this deliberately does not touch the `?? 0` in the helper. my original suggestion there was wrong and i struck it in the thread, `SELECT SUM(...) as dailyvolume` has no `GROUP BY`, so it always returns exactly one row and `if (!data.length) throw` would never fire on any of the five. separating "allium has no rows for this protocol" from "this protocol had no trades today" needs a second wider-window query, which is a real cost on every run for the other four adapters that are working fine. happy to send that separately if you want it.

### DIFF
--- a/factory/alliumSolanaDex.ts
+++ b/factory/alliumSolanaDex.ts
@@ -1,18 +1,18 @@
 import { alliumSolanaDexExport } from "../helpers/alliumDex";
 import { createFactoryExports } from "./registry";
 
-// [alliumDexName, dexId, start]
-const dexsConfigs: Record<string, [string, string, string]> = {
+// [alliumDexName, dexId, start, deadFrom?]
+const dexsConfigs: Record<string, [string, string, string, string?]> = {
   "cropper-clmm": ["cropper", "cropper-whirlpool", "2024-04-24"],
   "fluxbeam": ["fluxbeam", "fluxbeam", "2023-05-29"],
   "obric-v2": ["obric", "obric-v2", "2024-06-04"],
-  "openbook": ["openbook", "openbook", "2022-01-27"],
+  "openbook": ["openbook", "openbook", "2022-01-27", "2025-04-17"],
   "solfi": ["solfi", "solfi", "2024-10-29"],
 };
 
 const dexsProtocols: Record<string, any> = {};
-for (const [name, [alliumDexName, dexId, start]] of Object.entries(dexsConfigs)) {
-  dexsProtocols[name] = alliumSolanaDexExport(alliumDexName, dexId, start);
+for (const [name, [alliumDexName, dexId, start, deadFrom]] of Object.entries(dexsConfigs)) {
+  dexsProtocols[name] = alliumSolanaDexExport(alliumDexName, dexId, start, deadFrom);
 }
 
 export const { protocolList, getAdapter } = createFactoryExports(dexsProtocols);
--- a/helpers/alliumDex.ts
+++ b/helpers/alliumDex.ts
@@ -3,7 +3,7 @@ import { queryAllium } from "./allium"
 import { CHAIN } from "./chains"
 
 
-export function alliumSolanaDexExport(dex_id: string, protocol: string, start: string) {
+export function alliumSolanaDexExport(dex_id: string, protocol: string, start: string, deadFrom?: string) {
   const fetch = async (options: FetchOptions) => {
     const query = `
       SELECT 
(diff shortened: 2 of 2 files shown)


bheluga:
@0xshubhs thanks for the PR

PR #9121, opened 2026-09-01
usdd: accrue tron stability fees on vault debt, not circulating supply

### LINKED ISSUE #8898: usdd: tron stability fees look ~2.7x overstated, supply used where vault debt belongs
Separate from #8896 (which is the crash), the Tron branch of `fees/usdd` looks like it overstates stability fees by about 2.7x. Tron is essentially the whole USDD fee line, so this is the headline number.

The calculation pairs two different endpoints:

```ts
const fee = tronDebt(history.data?.items ?? [], options.fromTimestamp)
          * tronRate(collaterals.data?.items ?? [])
          * timeframe / YEAR;
```

`tronRate` is a debt-weighted average stability fee taken from `vault/collaterals`:

```
  TRX-A    curMinted=   170876092 stabilityFee=0.005
  TRX-B    curMinted=    95699990 stabilityFee=0.005
  TRX-C    curMinted=   190496525 stabilityFee=0.005
  USDT-A   curMinted=      551012 stabilityFee=0.01
  STRX-A   curMinted=     5363250 stabilityFee=0.01
  WBTC-A   curMinted=           0 stabilityFee=0.025
  WBTC-B   curMinted=           0 stabilityFee=0.035
  TOTAL vault debt = 462986869   weighted rate = 0.005064
```

`tronDebt` then pulls `debt` from `collateral-history`, which is a different quantity entirely:

```
  last row: 2026-08-21 00:00 (UTC+8)
  debt            = 1235720319.57
  mintedUSDD      = 1235720319.57
  usddTotalSupply = 1234764696.85
```

`debt`, `mintedUSDD` and `usddTotalSupply` are the same figure to within 0.08%, so that field is circulating USDD supply, not CDP debt. Cross-checking against DefiLlama's own stablecoin API agrees:

```
$ stablecoins.llama.fi/stablecoincharts/Tron?stablecoin=14
Tron circulating: 1192998041
```

So the adapter applies a rate weighted over $463m of vault debt to $1.24b of total supply. The gap is USDD that was not minted against a fee-paying vault, and it does not accrue stability fees.

```
  ratio history.debt / vault curMinted = 2.669
  annualised using history debt = 6257528   -> 17144/day   (what is published)
  annualised using vault  debt = 2344506    ->  6423/day
```

Published days sit right on the first line (17.35k on 08-17, 17.38k on 08-16, 17.21k on 08-15), so that is the path being taken.

Both framings land in the same place. Either multiply vault debt by the vault-weighted rate, or multiply total supply by a rate weighted over total supply (`462986869 * 0.005064 / 1235720319 = 0.001897`) — both give ~$6.4k/day. The simplest form is just the numerator that `tronRate` already computes and then divides away:

```
sum(curMinted_i * stabilityFee_i) = 2344506/yr = 6423/day
```

One thing I could not settle from the public API, and where I would want your call before sending a patch: `vault/collaterals` is a current snapshot with no date parameter, so using it alone is fine for live runs but gives no per-day history for a backfill. `collateral-history` is the only dated source and it does not appear to expose vault debt separately. If there is a dated endpoint for per-ilk `curMinted` that would be the clean fix; otherwise it is a choice between a snapshot rate applied to dated supply and something coarser.

Happy to send the patch once you say which way you want it. Not bundling it into #8897 since that one is just the crash and should land on its own.

0xshubhs:
Fixes #8898

the tron leg pairs a rate weighted over vault debt with a debt figure that is circulating supply:

```ts
const fee = tronDebt(history) * tronRate(collaterals) * timeframe / YEAR;
```

`tronRate` weights `stabilityFee` by `curMinted` across ilks, so it is the average rate on $463m of vault debt. `tronDebt` reads `debt` from `collateral-history`, which equals `mintedUSDD` and `usddTotalSupply` to within 0.08% and is $1.22b. the gap is USDD not minted against a fee-paying vault, and it does not accrue stability fees.

as of today:

```
vault debt            463,435,710
annual fees             2,348,832   (= sum curMinted * stabilityFee)
vault-weighted rate      0.005068
circulating supply  1,218,413,599

published formula   supply * vaultRate         = $16,919/day
this PR             annualFees * supplyRatio   = $6,435/day
                                          2.629x overstated
```

`api.llama.fi/summary/fees/usdd` reads 17,096 for 09-01 and tron is effectively the whole line, so that is the path being taken.

the numerator `tronRate` computes and then divides away is already the answer, `sum(curMinted_i * stabilityFee_i)` is the annual accrual. i kept `collateral-history` in, scaling by `supplyOnDay / supplyNow`, so the day-to-day shape still comes from the dated series instead of going flat; on the latest day the ratio is 1 and the fee is just the vault accrual. equivalent to the second framing in the issue, applying a supply-weighted rate of `2348832 / 1218413599 = 0.001928` to dated supply.

the open question from the issue stands and this does not solve it: `vault/collaterals` is a current snapshot with no date parameter, so a backfill applies today's per-ilk mix to older days. if there is a dated endpoint for per-ilk `curMinted` that would be the clean fix. this is strictly better than what is there now on both axes though, right level today, same dated shape.

also added a throw when `collateral-history` returns no dated rows, since `supplyOnDay / supplyNow` would otherwise be `0/0`.

harness could not be used for this one, the tron rpcs rate-limit locally (429 from trongrid, 500 from drpc) before the adapter runs, and the tron branch never touches a block anyway. the numbers above are computed straight off the two endpoints the adapter reads.

### DIFF
--- a/fees/usdd/index.ts
+++ b/fees/usdd/index.ts
@@ -46,23 +46,18 @@ const TRON_HISTORY_API = "https://app-api.usdd.io/data-platform/collateral-histo
 const toWad = (rad: bigint) => rad / RAY;
 const blockOf = ({ blockNumber, block, block_number }: any) => Number(blockNumber ?? block ?? block_number);
 
-const tronRate = (markets: any[]) => {
-  let debt = 0;
-  let fees = 0;
-
-  markets.forEach(({ curMinted, stabilityFee }) => {
-    const minted = Number(curMinted ?? 0);
-    debt += minted;
-    fees += minted * Number(stabilityFee ?? 0);
-  });
-
-  return debt ? fees / debt : 0;
-};
+const tronAnnualFees = (markets: any[]) =>
+  markets.reduce((sum, { curMinted, stabilityFee }) =>
+    sum + Number(curMinted ?? 0) * Number(stabilityFee ?? 0), 0);
 
 const tronDebt = (items: any[], timestamp: number) => {
   const day = new Date(timestamp * 1000).toISOString().slice(0, 10);
   return Number(items.find((item) => item.time?.startsWith(day))?.debt ?? 0);
 };
+
+const tronLatestSupply = (items: any[]) =>
+  items.reduce((latest, item) =>
+    (item.time ?? '') > (latest?.time ?? '') ? item : latest, items[0]);
 
 const fetch = async (options: FetchOptions) => {
   const dailyFees = options.createBalances();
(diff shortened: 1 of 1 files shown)


bheluga:
@0xshubhs thanks for the PR

PR #9140, opened 2026-09-02
[fix] uniswap - throw when pooled balance rpc failure rate exceeds 10%

gomesalexandre:
`filterPools` (`helpers/uniswap.ts`) is shared by ~50+ dex/fee adapters (via direct import, plus the `uniV2Exports`/`uniV3Exports` factory helpers — spot-checked `dexs/velodrome-v2`, `dexs/metropolis-amm`, `dexs/blackhole`, `dexs/blackhole-CL`, `fees/keller-cl`, `fees/nuri-exchange-v2`, all confirmed routing through it). It batches `balanceOf` calls with `permitFailure: true`, so per-call RPC failures come back as `null` instead of throwing.

#8565 added a guard so a **total** RPC outage (every call null) throws instead of silently reporting every pool as empty. That guard is `res.every((bal) => bal == null)` — all-or-nothing. A single surviving call anywhere in the batch disarms it completely, so a **partial** outage (e.g. 998 of 1000 calls failing) still sails through: the failed calls become `null ?? 0`, those pools fall under `minUSDValue`, and they get silently filtered out as if genuinely empty — exactly the failure mode #8565 already agreed was worth guarding against, just not covered for the partial case.

## Fix

Replace the all-null check with a failure-ratio threshold (>10%):

```ts
const failedCalls = res.filter((bal) => bal == null).length
if (balanceCalls.length && failedCalls / balanceCalls.length > 0.1)
  throw new Error(`filterPools: ${failedCalls}/${balanceCalls.length} pooled balance calls failed on ${api.chain}, refusing to report ${Object.keys(pairs).length} pools as (partly) empty`)
```

10% is a fail-fast compromise: isolated RPC hiccups (a handful of null calls in a large batch) stay tolerated, while a batch that's mostly failing throws visibly instead of understating real pool balances.

## Testing

New `helpers/uniswap.test.ts` (follows this repo's existing `helpers/dune.test.ts` convention — a manual `node --test` file, not auto-discovered by the `test-adapter` CI workflow, which runs `cli/testAdapter.ts` against changed adapters instead):

```
$ node --test -r ts-node/register/transpile-only helpers/uniswap.test.ts
✔ throws when nearly every balance call fails (998/1000 null)
✔ throws when every balance call fails (1000/1000 null)
✔ does not throw on a normal low failure rate (5/1000 null)
✔ does not throw right at the 10% failure boundary (100/1000 null)
✔ throws just past the 10% failure boundary (101/1000 null)
✔ does not throw when every call succeeds
ℹ tests 6, pass 6, fail 0
```

Confirmed genuine red-before/green-after: the "998/1000 null" test fails against the pre-fix `.every(...)` guard (`Missing expected rejection`) and passes after the ratio-threshold fix.

`tsc --project tsconfig.json --noEmit` — clean, 0 errors.

Adversarially reviewed with Codex (GPT-5.5, read-only sandbox) against the staged diff — verdict: SHIP, no correctness issues in the ratio math, threshold and test coverage both judged reasonable.

## receipts

```
$ node --test -r ts-node/register/transpile-only helpers/uniswap.test.ts
✔ throws when nearly every balance call fails (998/1000 null) (1.9ms)
✔ throws when every balance call fails (1000/1000 null) (0.8ms)
✔ does not throw on a normal low failure rate (5/1000 null) (3.0ms)
✔ does not throw right at the 10% failure boundary (100/1000 null) (4.9ms)
✔ throws just past the 10% failure boundary (101/1000 null) (0.8ms)
✔ does not throw when every call succeeds (0.2ms)
ℹ tests 6
ℹ pass 6
ℹ fail 0

$ npx tsc --project tsconfig.json --noEmit
(clean, exit 0)
```


bheluga:

PR #9141, opened 2026-09-02
[fix] maverick-v2 - stop reporting fake $0 days when the subgraph fails

gomesalexandre:
Maverick V2's dexs adapter catches every subgraph-fetch failure and returns empty balances as a successful `$0` day, rather than surfacing the outage - a real dollar-volume day silently reports as zero.

$20.5M/24h, $77.6M/7d, 6 chains per `api.llama.fi/summary/dexs/maverick-v2` (re-verified live, not the PR-author's guess).

## The bugs

**1. `dexs/maverick-v2/index.ts:135-141` - swallowed subgraph errors.**
```ts
} catch (error) {
  console.error(`Error fetching data for ${chain}:`, error);
  return { dailyVolume: createBalances(), dailyFees: createBalances() };
}
```
Any network failure, timeout, or GraphQL error on the subgraph call gets logged and then reported as a legitimate zero-volume day.

This exact pattern was removed once before, from the sibling on-chain adapter this file replaced - `ce9d11b8c "remove try catch"` (Mar 2025) - and reintroduced here during the subgraph migration in #2877 (Apr 2025), which rewrote the whole file and brought the anti-pattern back.

**2. `dexs/maverick-v2/index.ts:108` - malformed response falls through the same gap.**
```ts
if (graphRes && graphRes.poolDayStats) { for (...) {...} }
return { dailyVolume, dailyFees };
```
A response missing/null `poolDayStats` (e.g. a partial gateway error that still returns HTTP 200) isn't caught by the try/catch at all - it just skips the loop and returns the same fake zero-day success. Caught by Codex adversarial review, not the original finding.

## Fix

Both paths now throw instead of returning a successful empty result. A genuinely empty `poolDayStats: []` (no swaps in the window) is unaffected and still resolves normally - that's a real zero day, not a malformed one.

## What's NOT in this PR (honestly deferred, not proven)

`dexs/maverick-v2/index.ts:68-93`'s `poolDayStats` query has no `first`/pagination params, and The Graph's default page size is 100 rows - if any chain currently has >100 active pools reporting activity in a single day, results would be silently truncated. I could not verify pool counts: the SDK's shared demo gateway key (`sdk.graph.modifyEndpoint`) returned `payment required for subsequent requests` when queried directly, and I have no separate `GRAPH_API_KEY`. Flagging honestly as an unverified follow-up rather than claiming or fixing something unproven - a maintainer with subgraph access could confirm in seconds.

Also spot-checked: `tokenAVolume + tokenBVolume` summed separately per-token (not combined) is correct, not a double-count - each token gets only its own side's volume, matching the identical pattern in `dexs/rooster/index.ts:66`.

## receipts

Real red-before/green-after regression tests, `node --test`:

- Test 1 (network failure): mocked `graphql-request`'s `request` to throw -> against the old code, resolved successfully with zero balances (red) -> against the fix, rejects (green).
- Test 2 (malformed response): mocked `request` to return `{}` -> against the old code, resolved successfully with zero balances (red) -> against the fix, rejects (green).
- Test 3 (control): a genuine empty `poolDayStats: []` still resolves normally with no error, on both old and new code - confirms the fix doesn't break the legitimate zero-day case.

```
✔ a subgraph query failure propagates instead of returning a fake zero day
✔ a malformed subgraph response (missing poolDayStats) propagates instead of returning a fake zero day
✔ an empty poolDayStats array resolves with zero balances, not an error
ℹ tests 3
ℹ pass 3
ℹ fail 0
```

`tsc --project tsconfig.json --noEmit` - clean, exit 0.

Codex ran as adversarial second reviewer (synchronous, full run, no timeout/kill needed) and caught bug #2 above; its remaining note (that `.test.ts` files aren't wired into this repo's CI `test` script) matches this repo's existing convention - `helpers/dune.test.ts` (the sibling test file this one's modeled on) isn't CI-wired either, by design (it's explicitly a manual/live test per its own header comment) - so left as-is rather than scope-creeping into adding CI wiring in this PR.

Dupe-checked: no open/closed issue or PR mentions this catch-swallow specifically (checked #8521's "17 protocols report zero" tracking issue - doesn't name maverick).

closes nothing (no existing issue tracked this specific bug)


bheluga:

gomesalexandre:
@bheluga thanks for the on-chain-query rewrite - much better than depending on a subgraph that turns out to need a paid API key. Rebased my CodeRabbit follow-ups (breakdownMethodology, consolidating the per-chain FACTORY/adapter maps into one chainConfig) on top of it, tested cleanly on Ethereum (real $23.98M volume / $1.18k fees over the last 24h). The BSC failures in the CI run above look like public-RPC historical-state flakiness (missing trie node, rate limits) rather than anything in the adapter - unrelated to this PR.

PR #9147, opened 2026-09-02
[fix] parcl - stop reporting fake $0 days and misattributing lp fees as protocol revenue

gomesalexandre:
## What

`fees/parcl/index.ts` had two bugs:

**1. Silent fake-$0 on a dead upstream.** The adapter calls `https://parcl-api.com/v1/time-series/cumulative-lp-fee?window=y` and finds the day matching the target date. The old code did `if (!dayData) console.log(...)` and returned `dailyFees: 0` regardless - a silent fake-zero, not an error. This endpoint has previously gone dead for months at a time (its `timeSeries` stops advancing) while still returning HTTP 200 with stale data, making a fake $0 day indistinguishable from a real one. Fixed to throw instead.

**2. LP fees misattributed as 100% protocol revenue.** The endpoint is literally named `cumulative-lp-fee`, and the old code booked 100% of it as `dailyFees` AND `dailyRevenue` AND `dailyProtocolRevenue`, with a methodology string claiming "100% of collected fees go to the protocol." Per this repo's own `AGENTS.md`, `dailySupplySideRevenue` is "Portion to LPs, lenders, stakers" - an LP fee is supply-side revenue by definition. Parcl's own Help Center states verbatim: **"LPs retain 80% of the trading fees."** (`docs.parcl.co/protocol-overview` corroborates: "LPs ... collect the majority of trading fees that they split with the protocol.") Fixed to split `dailyFees` 80/20 into `dailySupplySideRevenue`/`dailyRevenue`, with `dailyProtocolRevenue = dailyRevenue`.

## Testing

Verified live via this repo's own `npm test fees parcl <timestamp>` CLI:

```
$ npm test fees parcl <2026-04-18 window>
Daily fees: 416.00
Daily supply side revenue: 333.00
Daily revenue: 83.00
Daily protocol revenue: 83.00
```
(333 + 83 = 416, matches production's $416 for that date.)

```
$ npm test fees parcl <2026-09-01 window, today>
Error: Parcl: no data found for 2026-09-01 in parcl-api.com's cumulative-lp-fee series (upstream may be dead/stale)
```
(Upstream has been stale since 2026-04-22 - confirmed against production `api.llama.fi/summary/fees/parcl`, which shows real values through 2026-04-21 then exactly $0 every day since.)

Codex adversarial review ran but was interrupted mid-pass after ~4 minutes of thorough exploration (checked repo error-policy conventions, sibling adapters' methodology key naming, historical-window/backfill handling) without surfacing any issue - it explicitly confirmed "the fail-fast change matches this repository's error policy" before running out of time on secondary checks. I self-reviewed the remaining items it was mid-way through: `SupplySideRevenue` (no `daily` prefix) matches the exact methodology-key convention used by sibling adapters (checked `fees/40acres/index.ts`); `methodology` has no strict schema requiring key/field alignment (`adapters/types.ts`); the `start: '2024-06-01'` backfill date is unchanged by this diff.

## Risk

Low - isolated to this one adapter. The throw-on-missing-day behavior is a deliberate fail-loud choice matching this repo's own precedent for dead-upstream handling elsewhere in the codebase; a currently-healthy protocol having a genuine zero-fee day would still return `dayData` (just with `value: 0`), so the throw only fires when the day is missing from the series entirely, not when it's present with zero value.

### DIFF
--- a/fees/parcl/index.ts
+++ b/fees/parcl/index.ts
@@ -3,8 +3,15 @@ import { CHAIN } from "../../helpers/chains";
 import { httpGet } from "../../utils/fetchURL";
 
 // https://help.parcl.co/getting-started/nxKkvMYVGNLScHEvioJKJe/parcl-protocol-mechanics-/qCrhtrvt41v9fR9NExz9on
-// "LPs retain 80% of the trading fees" - the endpoint below is named cumulative-lp-fee and
-// is the fee LPs earn from trades, i.e. the supply-side cut, not protocol revenue.
+// "LPs retain 80% of the trading fees" - Parcl's docs confirm the split exists, but not
+// which side of it this specific endpoint's `value` field reports. Two readings are both
+// plausible and we could not find an authoritative source disambiguating them:
+//   (a) `value` = gross trading fee (what's currently assumed below), split 80/20 here; or
+//   (b) `value` = the LP's already-split 80% cut (the endpoint name is "lp-fee", not
+//       "trading-fee" or "total-fee"), meaning gross = value / LP_SHARE.
+// Kept as (a) - the existing, already-tested interpretation - rather than risk swapping to
+// an equally-unproven (b). Flagged explicitly for a maintainer/CodeRabbit follow-up rather
+// than silently picking a side.
 const LP_SHARE = 0.8;
 
 const fetch = async (options: FetchOptions) => {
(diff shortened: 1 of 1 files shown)


gomesalexandre:
Addressed the title format and Breakdown Methodology findings (renamed to `[fix] parcl - ...`, added `breakdownMethodology`).

On the Income Statement Compliance finding (`value` as LP's 80% cut vs. gross fee): investigated this for real rather than picking a side blindly. Parcl's own help center confirms the 80/20 split exists ("LPs retain 80% of the trading fees"), but I could not find an authoritative source stating which side of that split the `cumulative-lp-fee` endpoint's `value` field actually reports — the endpoint's own name is genuinely ambiguous between "the fee attributable to LPs specifically" (your reading) and "the trading fee, reported from the LP-facing side of the accounting" (the existing reading). Parcl doesn't publish an OpenAPI spec or a corroborating "total fee"/"protocol fee" endpoint I could cross-check against.

Left the numeric interpretation as-is (matches this PR's already-tested behavior) rather than swap to an equally-unproven alternative, and added an explicit code comment laying out both readings so this doesn't get silently lost. Happy to switch if a maintainer or Parcl's team can confirm which one is correct — genuinely can't settle it from public info alone.

bheluga:
Added [here](https://github.com/DefiLlama/dimension-adapters/pull/9183)

PR #9153, opened 2026-09-02
okx-swap: refresh the router list and add the second commission event, $0 since 2025-11-26

### LINKED ISSUE #9152: fix: okx-swap silently reports fake $0 days (281 days, ~$300k/day) instead of erroring on dead commission events
## What

`fees/okx-swap.ts` has silently reported a fake successful **$0** day for **281 consecutive days** (last real day 2025-11-25, `total1y: $10,856,495`, averaging ~$301k/day over its last 30 non-zero days) while the tracked routers are still live and routing swaps normally.

```
$ curl -s "https://api.llama.fi/summary/fees/okx-swap?dataType=dailyFees" | jq '{total24h,total7d,total30d,total1y,totalAllTime}'
{
  "total24h": 0,
  "total7d": 0,
  "total30d": 0,
  "total1y": 10856495,
  "totalAllTime": 10856495
}
```

## Root cause

The adapter measures fees via two events on each chain's `DexRouter` address:

```ts
const CommissionFromTokenRecordEvent = 'event CommissionFromTokenRecord(address tokenAddress, uint256 commissionAmount, address referrerAddress)';
const CommissionToTokenRecordEvent   = 'event CommissionToTokenRecord(address tokenAddress, uint256 commissionAmount, address referrerAddress)';
```

Both events (topics `0x0d3b1268ca3dbb6d3d8a0ea35f44f8f9d58cf578d732680b71b6904fb2733e0d` and `0xf171268de859ec269c52bbfac94dcb7715e784de194342abb284bf34fd30b32d`) stopped being emitted on the tracked routers around 2025-11-15 (last on Ethereum 2025-12-03). `getLogs` returns `[]`, the fee-summing loop never runs, and the adapter returns `dailyFees: 0` as a *successful* result instead of erroring — the same silent-zero class as #8521 and the `filterPools`/`resolv`/`parcl`/`maverick-v2` fixes merged this week (#9140, #9143, #9147, #9141).

**The tracked routers are still live.** Fresh `eth_getLogs` on the Base router (`0x5e2F47bD7D4B357fCfd0Bb224Eb665773B1B9801`) over the last ~40,000 blocks (~22h) found 7 logs, all topic `0x1bb43f2da90e35f7b0cf38521ca95a49e68eb42fac49924930a5bd73cdf7576c` (`OrderRecord`) — the router is still processing swaps, just at low volume on this specific address, and zero commission events.

## Where fees appear to have moved (partial answer, not a full fix)

OKX's own docs warn that "the contract addresses of the DEX router... may be subject to replacement due to contract upgrades" and recommend using their live `/approve-transaction`/`/swap` API for current addresses rather than a static list — which matches what I found: this repo's tracked addresses look stale.

OKX's current open-source router codebase (`okxlabs/DEX-Router-EVM-V1`, `contracts/8/libraries/CommissionLib.sol`) defines a **single unified** commission event, replacing the old two-event ABI:

```solidity
event CommissionRecord(uint256 commissionAmount, address referrerAddress);
// topic0: 0xffc60ee157a42f4d8edbd1897e6581a96d9ed04e44fb2ab53a47ce1eb8f2775b
```

Two things make this harder than a simple event swap:
1. **No `tokenAddress` field** in the new event (only `log1`, one topic, both params in `data`). The commissioned token is only knowable from the originating transaction's trailing calldata (`_getCommissionInfo()`'s bit-packed suffix) — correctly attributing fees by token would need per-transaction receipt correlation (e.g. matching the referrer-address `Transfer` in the same tx), not a plain `getLogs` scan.
2. **I could not confirm which contract address(es) currently receive live volume with this event** on the tracked chains within a reasonable investigation budget — `eth_getLogs` for this topic came back empty on the two most-likely candidate addresses I found via public sources (OKX's "Aggregation Router" `0x3b3ae790df4f312e745d270119c6052904fb6790` and "DEX Router 2" `0x7d0ccaa3fac1e5a943c5168b6ced828691b46b36` on Ethereum, checked over the last ~5 days). Either these aren't the actual current per-chain routers, or genuine commission-tagged volume through OKX's aggregator is lower than the historical numbers here suggest, or my sampling window/RPC missed it.

## Suggested path

I'd rather hand this to a maintainer with more context than guess a replacement:
- If OKX can confirm (or the maintainer can find, e.g. via OKX's own API) the current per-chain router address(es) actually receiving commissioned volume, the new `CommissionRecord` event + a small amount of tx-correlation logic (get receipt, find the matching `Transfer` to `referrerAddress`) should restore real numbers.
- If OKX's aggregator genuinely no longer routes meaningful volume through *any* address this repo could reasonably track (possible, given how aggressively they seem to rotate router contracts), `deadFrom` is the honest call — but I don't have enough confidence either way to make that call myself, and didn't want to under-claim a live $300k+/day fee source as dead without being sure.

Happy to do the tx-correlation implementation work myself once someone can confirm the current live router address(es) — that's the piece I couldn't pin down from outside OKX.

tarun-khatri:
﻿Closes #9152.

`fees/okx-swap` has published exactly $0 every day since 2025-11-26 (`total24h`, `total7d` and `total30d` are all 0 against `total1y` of $10.86M). It is not erroring — it runs, finds no logs, and writes a successful zero.

## Why

The adapter tracks one router address per chain. OKX replaces its routers roughly every four months, and does it on every chain on the same day. Every address configured here has stopped:

| chain | configured router | last commission event |
|---|---|---|
| BSC | `0x6015126d…` | 2025-12-11 |
| Ethereum | `0x2E1Dee21…` | 2025-12-04 |
| Base / Arbitrum | `0x5e2F47bD…` | 2025-11-19 |
| Polygon | `0xF5402CCC…` | 2025-11-17 |
| Linea | `0x6f7c2046…` | 2025-11-17 |
| Optimism | `0x86F752f1…` | 2025-11-17 |
| Avalanche | `0x79f7C6C6…` | 2025-11-17 |
| Mantle | `0x69C236E0…` | 2025-11-17 |
| zkSync Era | `0x010BC6B1…` | 2025-11-17 |
| Blast | `0x69C236E0…` | 2025-09-26 |

The decay in the published series matches: $232,969 on 2025-11-16, $74,255 on 11-17, $10,587 on 11-18, $3,213 on 11-19, then only a small Ethereum/BSC tail to 11-25 and zero after.

The event signature changed at the same time. A second `CommissionFromTokenRecord` / `CommissionToTokenRecord` pair that takes a fourth `commissionRate` argument — and therefore hashes to different topics, `0xcd5eae9d…` and `0x3cfb523a…` — first fired on 2025-11-14 and rolled out to every chain by 2025-12-19. It is what every current router emits: **66.2M events** against 41.4M for the original three-argument pair, whose last firing anywhere was 2026-03-31.

(The `CommissionRecord(uint256,address)` signature in OKX's public repo is a different, much smaller path — 136k events ever, one BSC contract — and it carries no token field, so it is not the replacement.)

## What changed

- Both event generations are now read. Which pair is queried is chosen by date, since outside the 2025-11-14 → 2026-03-31 overlap one of them cannot produce logs.
- The router list is rebuilt from **every contract that has emitted either pair and paid the referrer `0x6ea08ca8f313d860808ef7431fc72c6fbcf4a72d`** — 101 addresses across 22 chains. That referrer is the dominant payee of the routers already configured here (60.5% of BSC's commissions and 52.0% of Base's in August 2025), and of every generation since (38–57% in Dec 2025, 35–91% in Apr 2026, 52–100% today), which is what ties the new contracts to the old ones.
- Metis and Blast were pointing at addresses that have never emitted on those chains. Their live routers (`0x25e7f77f…`, `0x472fc4F7…`) were found by scanning the chains directly.
- Chains added where OKX is demonstrably live: Cronos, Fantom, Ink, Monad, Scroll, ApeChain, HyperEVM.

## Why not match on the event topic alone

I tried that first, and it is wrong on both counts.

It over-counts: three contracts still emitting these exact events on BSC, Base and Ethereum (`0x9017da9b…`, `0xbf5ee7c3…`, `0x7d4b9a1a…`, 141k events) pay a completely different referrer — `0x6795458378538ebe0a12d9e2e5ec1c32f0af1fbb` consistently from 2026-01-10, then `0x30e2e095…` from 2026-08-27. OKX's router is open source, so the signature alone does not establish provenance.

It also cannot be tested: matching without an address filter means unfiltered `eth_getLogs` on every chain, which the CI runner's public RPCs reject.

## Verification

Local run, per chain, for the 24 hourly slots ending 2026-09-02 12:00 UTC:

```
ethereum  15.26 k   base  8.83 k   polygon  4.47 k   hyperliquid 1.12 k
arbitrum  908.65    ink   48.66    apechain 10.31    avax  9.26
cronos    7.48      unichain 4.71  monad    4.22     plasma 3.90
era       1.69      optimism 1.75  linea    3.01     mantle 0.20
```

BSC is the largest chain and my IP is currently rate-limited by its public dataseeds, so I reconstructed it independently from chain data instead: **$24,622 on 2026-09-01**, of which $19,460 is 59,370 USDT commissions.

For contrast, the current version of this file runs to completion and prints `0.00` for all fifteen of its chains — it passes CI precisely because it finds nothing.

## Not included

- **Robinhood Chain**, despite being the third-largest at 3.9M events. It has no entry in the SDK's `providers.json` and the single public endpoint `rpc.mainnet.chain.robinhood.com` returns 429 under the adapter's request rate, which fails the whole run rather than one chain. Worth adding once there is a usable RPC.
- **Polygon zkEVM**, where OKX's last commission event was 2026-06-30, so it would report $0.
- Metis and Blast are kept but are dust — Metis averages under 3 commission events a day.

### DIFF
--- a/fees/okx-swap.ts
+++ b/fees/okx-swap.ts
@@ -30,16 +30,30 @@ interface IRouter {
 // 0x6ea08ca8f313d860808ef7431fc72c6fbcf4a72d, which is also the dominant payee of the routers
 // previously configured here. Listed oldest-first.
 //
-// To refresh this list after the next rotation, or to check the dates above, group every log
-// carrying one of the four commission topics by chain and emitter and keep the emitters that have
-// paid that referrer; `retiredOn` is each emitter's last such log. topic0s are
-// 0x0d3b1268ca3dbb6d3d8a0ea35f44f8f9d58cf578d732680b71b6904fb2733e0d and
-// 0xf171268de859ec269c52bbfac94dcb7715e784de194342abb284bf34fd30b32d for the three-argument pair,
-// 0xcd5eae9d9d0b96532bd1b7dbf6628ce436b2af735829087a03c548439f8bf850 and
-// 0x3cfb523a4c38d88561dd3bf04805a31715c8b5fc468a03b8d684356f360dea99 for the four-argument pair;
-// the referrer is the third word of `data`. That query is also what dates V2_EVENTS_START and
-// V1_EVENTS_END above (first and last log of each pair, plus a margin). Metis and Blast have to be
-// read from the chains directly, since Dune does not index either.
+// To refresh this list after the next rotation, or to check the dates above, run this query; it
+// returns the routers configured for a chain, oldest-first, and its first_log / last_log are what
+// date V2_EVENTS_START and V1_EVENTS_END above (first and last log of each event generation, plus a
+// margin). Drop the blockchain filter to cover every chain in one pass, at the cost of a far longer
+// scan.
+//
+//   SELECT blockchain, contract_address, min(block_time) AS first_log, max(block_time) AS last_log
+//   FROM evms.logs
+//   WHERE blockchain = 'ethereum'
+//     AND block_time >= timestamp '2025-09-01'
+//     AND topic0 IN (
+//       0x0d3b1268ca3dbb6d3d8a0ea35f44f8f9d58cf578d732680b71b6904fb2733e0d,  -- CommissionFromTokenRecord
+//       0xf171268de859ec269c52bbfac94dcb7715e784de194342abb284bf34fd30b32d,  -- CommissionToTokenRecord
+//       0xcd5eae9d9d0b96532bd1b7dbf6628ce436b2af735829087a03c548439f8bf850,  -- ...FromTokenRecord, 4 args
+//       0x3cfb523a4c38d88561dd3bf04805a31715c8b5fc468a03b8d684356f360dea99)  -- ...ToTokenRecord, 4 args
+//     AND bytearray_substring(data, 65, 32)  -- the referrer is the third word of `data`
+//       = 0x0000000000000000000000006ea08ca8f313d860808ef7431fc72c6fbcf4a72d
+//   GROUP BY 1, 2
+//   ORDER BY 1, 3
+//
+// `retiredOn` is a router's last commission log to ANY referrer, which is the same query without
+// the referrer line: a router can keep paying other referrers after it stops paying this one, and
+// cutting it off at the earlier date would drop those fees. Metis and Blast have to be read from
+// the chains directly, since Dune does not index either.
 const routers: Record<string, IRouter> = {
   [CHAIN.ETHEREUM]: {
     addresses: [
--- a/fees/okx-swap.ts
+++ b/fees/okx-swap.ts
@@ -1,106 +1,253 @@
 import { FetchOptions, SimpleAdapter } from "../adapters/types";
 import { CHAIN } from "../helpers/chains";
 
+// OKX books each referral commission with one of these events. There are two generations: the
+// original three-argument pair, and a four-argument pair that adds the commission rate and so
+// hashes to a different topic. The four-argument pair first fired on 2025-11-14 and the
+// three-argument pair last fired on 2026-03-31, so both are needed to cover the full history.
 const CommissionFromTokenRecordEvent = 'event CommissionFromTokenRecord(address tokenAddress, uint256 commissionAmount, address referrerAddress)';
 const CommissionToTokenRecordEvent = 'event CommissionToTokenRecord(address tokenAddress, uint256 commissionAmount, address referrerAddress)';
+const CommissionFromTokenRecordEventV2 = 'event CommissionFromTokenRecord(address tokenAddress, uint256 commissionAmount, address referrerAddress, uint256 commissionRate)';
+const CommissionToTokenRecordEventV2 = 'event CommissionToTokenRecord(address tokenAddress, uint256 commissionAmount, address referrerAddress, uint256 commissionRate)';
+
+const REFERRAL_COMMISSIONS = 'Referral commissions';
+
+const V2_EVENTS_START = 1761955200; // Nov-01-2025 00:00:00 UTC, two weeks before the first one
+const V1_EVENTS_END = 1777507200; // Apr-30-2026 00:00:00 UTC, a month after the last one
+
+const endOfDay = (day: string) => Math.floor(Date.parse(`${day}T23:59:59Z`) / 1e3);
 
 interface IRouter {
-  addresses: Array<string>;
+  // Each entry is a router and, for the ones OKX has already replaced, the day its last commission
+  // event fired. OKX rotates its routers roughly every four months and does it on every chain on
+  // the same day, so a chain accumulates several over time but only one or two can produce logs in
+  // any given window. Retired routers are skipped for windows that start after they stopped; some
+  // of these chains have a single public RPC and querying every router on every window trips it.
+  addresses: Array<{ address: string, retiredOn?: string }>;
 }
 
+// Every contract that has emitted a commission event and paid the referrer
+// 0x6ea08ca8f313d860808ef7431fc72c6fbcf4a72d, which is also the dominant payee of the routers
+// previously configured here. Listed oldest-first.
+//
+// To refresh this list after the next rotation, or to check the dates above, group every log
+// carrying one of the four commission topics by chain and emitter and keep the emitters that have
+// paid that referrer; `retiredOn` is each emitter's last such log. topic0s are
+// 0x0d3b1268ca3dbb6d3d8a0ea35f44f8f9d58cf578d732680b71b6904fb2733e0d and
+// 0xf171268de859ec269c52bbfac94dcb7715e784de194342abb284bf34fd30b32d for the three-argument pair,
+// 0xcd5eae9d9d0b96532bd1b7dbf6628ce436b2af735829087a03c548439f8bf850 and
+// 0x3cfb523a4c38d88561dd3bf04805a31715c8b5fc468a03b8d684356f360dea99 for the four-argument pair;
+// the referrer is the third word of `data`. That query is also what dates V2_EVENTS_START and
+// V1_EVENTS_END above (first and last log of each pair, plus a margin). Metis and Blast have to be
+// read from the chains directly, since Dune does not index either.
 const routers: Record<string, IRouter> = {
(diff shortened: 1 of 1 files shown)


tarun-khatri:
﻿Both review points are addressed in the pushed commit.

**1. Restrict log retrieval to verified OKX emitters — done, and you were right for a reason I had not measured.**

`noTarget` is gone; every chain now passes an explicit `targets` list. The list is not hand-collected: it is every contract that has emitted either commission pair *and* paid the referrer `0x6ea08ca8f313d860808ef7431fc72c6fbcf4a72d`, which is the dominant payee of the routers this file already trusted (60.5% of BSC commissions and 52.0% of Base in Aug 2025) and of every router generation since. That is 101 addresses over 22 chains.

Checking your concern against the data, it is not hypothetical here. Three contracts emitting these exact signatures today — `0x9017da9b…` on BSC, `0xbf5ee7c3…` on Base, `0x7d4b9a1a…` on Ethereum, 141k events between them — pay `0x6795458378538ebe0a12d9e2e5ec1c32f0af1fbb` instead, consistently since 2026-01-10, switching to `0x30e2e095…` on 2026-08-27. OKX's router is open source, so the topic alone does not establish provenance and the earlier version would have counted those as OKX fees.

**2. Balance labels and `breakdownMethodology` — added, with one deviation.**

`dailyFees.add(...)` now carries a `Referral commissions` label and there are matching `breakdownMethodology` entries for `Fees`, `UserFees` and `SupplySideRevenue`. Verified in the local run:

```
label                | Daily fees | Daily supply side revenue
Referral commissions | 28474.29   | 28474.29
```

I did not split `dailySupplySideRevenue` into a separate `createBalances()` object. There is only one flow here — every commission is booked against the referrer that brought the trade, so supply-side revenue is the same set of transfers, not a second one. A separate balance would be an identical copy of the first and would double the log processing without changing any published number. The existing single-object return is also what this file already did before this PR.

tarun-khatri:
@coderabbitai review

tarun-khatri:
﻿Added the derivation to the file so the list is auditable without external context.

The comment above `routers` now names the four commission topic0s and the referrer, and states the grouping that produces the table: group every log carrying one of those topics by chain and emitter, keep the emitters that have paid `0x6ea08ca8f313d860808ef7431fc72c6fbcf4a72d`, and take each emitter's last such log as its `retiredOn`. The same query dates `V2_EVENTS_START` and `V1_EVENTS_END` — first and last log of each event generation, plus a margin. It also records that Metis and Blast have to be read from the chains directly, because Dune indexes neither and their silence there is not evidence of absence.

I did not check a discovery script into the repo. There is no precedent for one under `fees/`, it would need a Dune key to run, and it would drift from the addresses it is supposed to justify. The topic hashes plus the referrer are the reproducible part, and they are now in the file.

tarun-khatri:
@coderabbitai review

tarun-khatri:
@coderabbitai review

tarun-khatri:
@coderabbitai full review

tarun-khatri:
Fair ask — the derivation is now written into the file as the query itself rather than described in prose, and running it turned up a wording error of mine worth correcting.

The selection query, verified against Ethereum:

```sql
SELECT blockchain, contract_address, min(block_time) AS first_log, max(block_time) AS last_log
FROM evms.logs
WHERE blockchain = 'ethereum'
  AND block_time >= timestamp '2025-09-01'
  AND topic0 IN (
    0x0d3b1268ca3dbb6d3d8a0ea35f44f8f9d58cf578d732680b71b6904fb2733e0d,  -- CommissionFromTokenRecord
    0xf171268de859ec269c52bbfac94dcb7715e784de194342abb284bf34fd30b32d,  -- CommissionToTokenRecord
    0xcd5eae9d9d0b96532bd1b7dbf6628ce436b2af735829087a03c548439f8bf850,  -- ...FromTokenRecord, 4 args
    0x3cfb523a4c38d88561dd3bf04805a31715c8b5fc468a03b8d684356f360dea99)  -- ...ToTokenRecord, 4 args
  AND bytearray_substring(data, 65, 32)
    = 0x0000000000000000000000006ea08ca8f313d860808ef7431fc72c6fbcf4a72d
GROUP BY 1, 2
ORDER BY 1, 3
```

returns exactly the six Ethereum routers in the file, in the configured order:

```
0x2e1dee213ba8d7af0934c49a23187babeaca8764  2025-09-01 -> 2025-12-04   331,720 logs
0xdcb7028e5eaa1d7bb82b7152cb0e7adc12e7457c  2025-10-17 -> 2026-04-08        93 logs
0xf6801d319497789f934ec7f83e142a9536312b08  2025-10-31 -> 2025-12-17   215,568 logs
0x5e1f62dac767b0491e3ce72469c217365d5b48cc  2025-12-03 -> 2026-05-12   379,619 logs
0x28b1dc1a5e3699a428bc51d234dfab7c9cb2a183  2026-04-23 -> 2026-08-24   861,246 logs
0x8feab81d36e7576107d5de0758c1b839be31b4f6  2026-08-24 -> 2026-09-04    41,078 logs  (current)
```

Five of the six `last_log` values are the `retiredOn` dates in the file. The sixth, `0xdcb7028e…`, is where my earlier comment was wrong: it said `retiredOn` is a router's last log *to that referrer*, and this row shows 2026-04-08 against a configured 2026-08-03. Re-running it without the referrer filter gives that router `last_log = 2026-08-03 03:44:35 UTC`, matching the file — the router kept paying other referrers for four more months. `retiredOn` is the last commission log to **any** referrer, which is the correct bound, since the adapter counts every commission a configured router emits and cutting it at the earlier date would drop four months of fees. The comment now says that, and both query forms are recorded.

`V2_EVENTS_START` / `V1_EVENTS_END` come from the same query's first and last log per event generation, plus a margin, which is now stated in the file too.

@coderabbitai full review

bheluga:

PR #9162, opened 2026-09-02
[fix] potatoswap - stop reporting fake $0 for recent days on v2 and v3

gomesalexandre:
## What

PotatoSwap's own stats API (`v3.potatoswap.finance/api/pool/list-all`) returns HTTP 200 with every pool's `volume_24h_usd`/`fee_24h_usd` stuck at `0` since ~2026-05-07, while `tvl_usd` stays live and current ($5,818,657, matching DefiLlama's reported $5.85M). Both `dexs/potatoswap.ts` and `dexs/potatoswap-v3.ts` branched to this dead API for any day within 6 hours of "now" (`isCloseToCurrentTime`), producing a genuine successful `$0` for that window - a silent-zero, not a dead endpoint.

```
$ curl -s https://v3.potatoswap.finance/api/pool/list-all | jq '[.data.pools[] | .volume_24h_usd|tonumber] | add'
0
```

Independent on-chain confirmation via GeckoTerminal on the same pool (`0xc71f9e1de80eb505c0cb3bbf90ae6593130e5d25`, x-layer): **$254,714** in real 24h volume, ~21,900 transactions, same day the API reports `$0`.

DefiLlama's own chart for this adapter: `total24h: 0`, `total30d: 0`, `total1y: $1,283,229,773` - 118+ consecutive zero days on a protocol still doing real volume.

## Fix

The on-chain log adapter (`getUniV2LogAdapter` / `getUniV3LogAdapter`) was already used for every day *outside* the 6-hour window - it's what powered the correct historical backfill after PR #6790. Both files now use it unconditionally instead of branching to the dead API for recent days.

## Also fixed (caught by adversarial review, not part of the original finding)

- **v2 fee rate**: the original code never passed an explicit `fees` value, so `getUniV2LogAdapter`'s 0.30% default silently overstated every fee/revenue figure by 20% against PotatoSwap's documented 0.25%. Now passes `fees: 0.0025, stableFees: 0.0025` explicitly.
- **v3 revenue attribution**: the removed "recent day" branch computed protocol revenue from each pool's *live* on-chain `feeProtocol` (a manual `slot0()` multicall). The replacement does the same thing through the helper's `dynamicProtocolFees`/`getRevenueRatio`, matching this file's own documented `ProtocolRevenue` methodology text exactly ("If feeProtocol0 = x1 and feeProtocol1 = x2, protocol revenue = Total Fees * (1/x1 + 1/x2) / 2. If only one is set, protocol revenue = Total Fees * 1/x.") - instead of a fixed 20%/0% split that was only ever accurate by coincidence.
- **version bump**: both adapters moved `version: 1` → `version: 2` with `pullHourly: true`, per this repo's own `AGENTS.md` ("New adapters must be `version: 2` unless they rely on Dune or an API that only returns daily aggregates" / "Use `pullHourly: true` wherever evm logs... are used") - they no longer depend on the API for anything but v3 pool-address discovery.
- Fixed a pre-existing guard bug in v3: it checked `poolsResponse.data.length` (always `undefined`, `.data` is an object not an array) instead of `poolsResponse.data.pools.length`.

## Testing

Real runs via this repo's own CLI (`npx ts-node --transpile-only cli/testAdapter.ts dexs potatoswap[-v3]`), hourly granularity since `pullHourly` is now set:

- **v2**: 22/24 hourly windows completed with real non-zero numbers. Fee rate checks out to exactly 0.25% across samples (e.g. window 1: `$84.00` fees / `$33,480` volume = 0.2508%). The last 1-2 windows (the very latest, still-forming hour) were slow on what looks like RPC/log-indexing finality for near-real-time blocks - inherent to any on-chain log adapter querying very recent data, not something this diff introduces (the old dead API path returned instantly for exactly this window, part of why the bug went undetected for 4 months).
- **v3**: 24/24 hourly windows completed. Revenue/fees ratio is consistently exactly 20% across the daily total (`$2.6856` / `$13.418`), confirming the dynamic `feeProtocol` read is genuinely live (not hardcoded) and happens to be uniform across this protocol's current pools.
- `npx tsc --noEmit -p .`: clean, zero errors.
- Codex adversarial review (two passes, run synchronously): first pass found the two real correctness bugs above (v2 fee rate, v3 revenue attribution) plus the version/pullHourly gap, which is why this PR is larger than a one-line branch removal. Second pass (confirming the fixes) didn't complete within a reasonable window and was killed by its own tracked PID; self-reviewed the remaining points against the repo's `AGENTS.md` and the real test output above instead of waiting further.

No merge attempted.

### DIFF
--- a/dexs/potatoswap-v3.ts
+++ b/dexs/potatoswap-v3.ts
@@ -1,8 +1,7 @@
 import { FetchOptions, SimpleAdapter } from '../adapters/types';
 import { CHAIN } from '../helpers/chains';
-import { getUniV3LogAdapter } from '../helpers/uniswap';
+import { getUniV3LogAdapter, UniGetRevenueRatioProps } from '../helpers/uniswap';
 import { httpGet } from '../utils/fetchURL';
-import BigNumber from 'bignumber.js';
 
 const methodology = {
   Fees: "Total fees paid by users on every swap, determined by the pool's fee tier (e.g., 0.01%, 0.05%, 0.30%, 1.00%).",
@@ -12,9 +11,7 @@ const methodology = {
   SupplySideRevenue: "The portion of swap fees distributed to Liquidity Providers (LPs). This is (Total Fees - Protocol Revenue) for each pool.",
 };
 
-// Labels kept identical to the ones getUniV3LogAdapter emits, so the breakdown
-// is consistent whether the API path (recent) or the on-chain log path (older)
-// runs for a given day.
+// Labels kept identical to the ones getUniV3LogAdapter emits.
 const LABELS = {
   SwapFees: 'Token Swap Fees',
   TradingFees: 'Trading fees',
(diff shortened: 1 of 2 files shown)


bheluga:

Issue #9088, opened 2026-08-31
Pyth Hermes now requires an API key: wisdomtree, solstice, zest-v1, zest-v2 have been dark since 2026-08-25


tarun-khatri:
Four fee adapters stopped producing data on 2026-08-25/26. They have nothing in common except one dependency: the public Pyth Hermes endpoint, which now requires an API key.

Pyth's docs carry the banner *"Pyth Core upgrade completed successfully on August 26, 2026. Hermes now requires an API Key."* The cutover was 2026-08-26 16:00 UTC.

### Affected adapters

All four call `https://hermes.pyth.network/v2/updates/price/...` and have no fallback, so the throw takes the whole adapter down.

| adapter | slug | last datapoint | fees 30d |
|---|---|---|---:|
| `fees/wisdomtree/index.ts` | wisdomtree | 2026-08-26 | $2,024,644 |
| `fees/solstice-usx/index.ts` | solstice | 2026-08-25 | $189,035 |
| `fees/zest-v2/index.ts` | zest-v2 | 2026-08-25 | $9,503 |
| `fees/zest-v1/index.ts` | zest-v1 | 2026-08-25 | $1,468 |

wisdomtree was running at roughly $78k/day right up to the cut (78,439 / 78,439 / 78,444 / 78,375 / 78,922 on 08-22..08-26), so this is a live protocol going dark, not one winding down. `total24h` is `null` for all four.

### What I verified

Price routes are gated, metadata routes are not:

```
GET https://hermes.pyth.network/v2/updates/price/latest?ids[]=...   -> 401 unauthorized
GET https://hermes.pyth.network/api/latest_price_feeds?ids[]=...    -> 401 unauthorized
GET https://hermes.pyth.network/v2/updates/price/<ts>?ids[]=...     -> 401 unauthorized
GET https://hermes.pyth.network/v2/price_feeds?query=tbill          -> 200
```

Alternatives I ruled out:

- `https://benchmarks.pyth.network/v1/updates/price/...` → 401
- `https://hermes-beta.pyth.network/v2/updates/price/latest` → 401
- `https://pyth.dourolabs.app/hermes/...` (the documented new host) → 401 without a key; its metadata route returns 200, so the host is up and the block is purely auth
[...]

PR #9164, opened 2026-09-02
zest: price stx/btc/usdc from coins.llama.fi, pyth hermes now requires an api key

0xshubhs:
part of #9088. zest-v1 and zest-v2 have been dark since 2026-08-25 because pyth hermes went behind an api key at the core upgrade cutover (2026-08-26 16:00 UTC). unlike wisdomtree and solstice, the zest adapters only need liquid market prices (STX, BTC, USDC), so there is no reason to wait for a key, this points them at coins.llama.fi historical prices instead, same 1e8 fixed-point interface into the existing bigint math, everything downstream untouched. USDh via the DIA oracle and the stSTX ratio contract are unchanged.

one behavior change on top of the source swap: a missing price now throws instead of the old `?? 0n`, which silently skipped every row priced by that asset. a run that cannot price STX should fail visibly, not publish a partial number.

verified in the harness:

- zest-v1 runs fully local (hiro + prices only). day 2026-08-31: fees $90.00, revenue $85.00; day 2026-08-30: fees $23, revenue $22, both inside the series' historical band ($9–97/d)
- zest-v2 now gets past the price path and stops at `Allium API Key is required[Ignore this error for github bot]`, which is the standard fork-CI boundary for allium adapters; the allium query itself is untouched by this diff

wisdomtree and solstice still need the key path from #9088 (and wisdomtree also needs a feed re-point, its `Rates.US1M` feed id is no longer in the hermes catalog at all, details on the issue).

### DIFF
--- a/fees/zest-v1/index.ts
+++ b/fees/zest-v1/index.ts
@@ -9,11 +9,11 @@ const PRD_ADDR = "SP2VCQJGH7PHP2DJK7Z0V48AGBHQAW3R3ZW1QF4N";
 const PRD = "pool-reserve-data";
 const SENDER = `${PRD_ADDR}.${PRD}`;
 
-const PYTH_URL = "https://hermes.pyth.network/v2/updates/price";
-const PYTH_IDS = {
-  STX: "0xec7a775f46379b5e943c3526b1c8d54cd49749176b0b98e02dde68d1bd335c17",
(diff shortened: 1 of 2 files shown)


bheluga:
@0xshubhs thanks for the PR

PR #9166, opened 2026-09-02
valiant-trade: refuse zero-swap days, the swapStat api serves zeros for days it has not ingested yet

### LINKED ISSUE #9165: valiant-trade: $1.74m published as $0 across five days — the swapStat api serves zeros for days it has not ingested yet
`dexs/valiant-trade` (Valiant Trade Spot, fogo) reads `https://mainnet-pro-api.valiant.trade/dex/analytics/swapStat?start=&end=`. that api answers 200 with all-zero fields for any window its ingestion has not reached yet, and the adapter forwards them, so every time the ingestion lags past the daily run defillama publishes a fake $0 that never gets revisited. the api itself backfills later — defillama doesn't.

every zero day the protocol has ever published turns out to be one of these. the api today, against what we published:

| day | published | api now |
|---|---:|---:|
| 2026-03-29 | $0 | 265,282 swaps, **$1,713,746** |
| 2026-08-19 | $0 | 5,132 swaps, $12,499.70 |
| 2026-08-20 | $0 | 2,074 swaps, $7,254.27 |
| 2026-08-30 | $0 | 2,131 swaps, $1,882.74 |
| 2026-08-31 | $0 | 1,547 swaps, $2,129.19 |
| 2026-09-01 | $0 | still zeros — the lag live right now |

~$1.74m published as $0 across five settled days, most of it the march day. reproduce with e.g.

```
curl "https://mainnet-pro-api.valiant.trade/dex/analytics/swapStat?start=1774742400&end=1774828800"   # 03-29
curl "https://mainnet-pro-api.valiant.trade/dex/analytics/swapStat?start=1788048000&end=1788134400"   # 08-30
```

the same lag also clips partially-ingested days: 08-29 published $20,551 but the api's settled figure is $22,030.96 (93%). on days the ingestion was caught up the match is exact — 08-25 published 34,311 vs 34,310.75, 08-26 22,713 vs 22,712.65, 08-27 21,602 vs 21,601.62, 08-28 67,502 vs 67,501.98 — so the adapter's parsing is fine, the number just wasn't final when the runner read it.

the dex has never had a real zero-swap day in 255 days of history (it does 1,500–6,000 swaps every day), so `totalSwaps == 0` from this api has exactly one meaning: not ingested yet. fix is to throw on it so the day stays refillable instead of publishing $0 — same shape as the guards merged in #8999 and #9026. pr coming. 03-29, 08-19, 08-20, 08-29, 08-30, 08-31 need a refill once it lands (and 09-01 once the api settles it).

0xshubhs:
Fixes #9165. the swapStat api answers 200 with all-zero fields for windows its ingestion has not reached, and the adapter forwarded them as real $0 days; five settled days published as $0 that the api later backfilled (~$1.74m, including $1.71m on 2026-03-29 alone; table on the issue). the dex has never had a genuine zero-swap day in its 255-day history, so `totalSwaps == 0` from this api only ever means not-ingested-yet; throwing keeps the day refillable instead of freezing a fake zero.

harness both directions:
- `npm test dexs valiant-trade 2026-09-01` (settled day) → volume $2.13k, fees $7.57, matching the api to the cent
- `npm test dexs valiant-trade 2026-08-30` (settled day) → volume $22.03k, the corrected value for 08-29 (prod holds the $20,551 partial read)
- `npm test dexs valiant-trade 2026-09-03` (day the api has not ingested) → throws `swapStat has no swaps for 2026-09-02 yet`

if the daily run fires before the api settles the previous day, the test check will show that same throw; that is the guard working, not a broken patch; the day fills on the next refill instead of publishing $0.

### DIFF
--- a/dexs/valiant-trade/index.ts
+++ b/dexs/valiant-trade/index.ts
@@ -14,6 +14,9 @@ async function fetch(options: FetchOptions) {
   const url = `${baseUrl}?start=${options.fromTimestamp}&end=${options.toTimestamp}`;
   const data = await httpGet(url);
 
+  if (!data.totalSwaps)
+    throw new Error(`swapStat has no swaps for ${options.dateString} yet; every zero day this api has served was later backfilled with real volume`);
+
   return {
     dailyVolume: data.totalSwapVolume,
     dailyFees: data.totalFees,


bheluga:
@0xshubhs thanks for the PR

PR #8959, opened 2026-08-23
dodo-agg: mark the halted moonriver and zero_network legs dead

0xshubhs:
Fixes #8958

moonriver stopped producing blocks 08-10 (moonbeam/moonriver sunset) and zero_network halted 08-12, so those two legs error on every run and take the other 13 chains down with them; dodo-agg has been dark since 08-08. marks both chains deadFrom (same per-chain pattern as kyberswap and bebop; dexs/dodo, zerion and onchain-checkin already dropped these legs). evidence in the issue.

harness skips both dead legs and the healthy chains publish again, verified twice.

PR #9168, opened 2026-09-02
dodo-fees: deadfrom the dfio leg, chain head frozen since 08-29 kills the other six chains

### LINKED ISSUE #9167: dodo-fees: dark since 08-30 — the dfio_meta_main chain froze at block 8994729 and its leg kills the other six chains
`fees/dodo-fees` (DODO, ~$4-10k/day across 7 chains) has published nothing since its 2026-08-29 row. the cause is not the dodoex gateway — that answers fine — it's the `dfio_meta_main` leg: the DeFi Oracle Meta mainnet's head is frozen at block 8994729 / 2026-08-29T05:26:19Z on every public rpc,

```
curl -s -X POST https://rpc.defi-oracle.io -H 'content-type: application/json' \
  -d '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["latest",false],"id":1}'
curl -s -X POST https://rpc-http-pub.d-bis.org -H 'content-type: application/json' \
  -d '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["latest",false],"id":1}'
```

both return block `0x893ea9` (8994729) with timestamp `2026-08-29T05:26:19Z`. the sdk's block lookup then refuses the leg (`Last block for dfio_meta_main is N minutes behind`), the leg throws, and one throwing chain kills the whole seven-chain run — ethereum/bsc/polygon/arbitrum/aurora/boba have been dark since 08-30 while healthy (harness aggregates $4.96k today across the six gateway chains).

the stop date matches to the day: chain froze 08-29 05:26 UTC, the run for day 08-29 was the last one that could resolve its blocks, nothing published since.

`dexs/dodo` doesn't have this problem because its dfio leg was already disabled back in june (#7496 era); `fees/dodo-fees` is the only adapter in the repo still running a `CHAIN.DFIO_META_MAIN` leg. same anatomy as the moonriver/zero legs in #8958: a halted-chain leg taking healthy chains down with it, fix is a per-chain deadFrom. pr follows; 08-30..09-01 need a refill once it lands.

0xshubhs:
Fixes #9167. the dfio_meta_main chain's head has been frozen at block 8994729 / 2026-08-29T05:26:19Z on every public rpc, so the leg's block lookup throws and takes the other six chains' fees down with it; nothing published since the 08-29 row. `deadFrom: '2026-08-29'` retires the leg without touching the rest, the same shape as the moonriver/zero legs in #8959. `dexs/dodo` already had its dfio leg disabled in june, this was the last live one.

harness twice: `Skipping dfio_meta_main because the adapter ended at Sat, 29 Aug 2026`, then ethereum $1.42k / bsc $3.44k / polygon $99 / arbitrum $3.48 / aurora $0.16 / boba $0, aggregate $4.96k, inside the adapter's $3.4-9.9k/day band. if the chain ever restarts, dropping the deadFrom revives the leg.

08-30..09-01 need a refill once this lands.

### DIFF
--- a/fees/dodo-fees.ts
+++ b/fees/dodo-fees.ts
@@ -95,7 +95,7 @@ const breakdownMethodology = {
 const adapter: Adapter = {
   version: 1,
   adapter: {
-    [CHAIN.DFIO_META_MAIN]: { fetch: dfioFetch, deadFrom: '2026-08-29' },
+    //[CHAIN.DFIO_META_MAIN]: { fetch: dfioFetch },
     [CHAIN.ETHEREUM]: { fetch },
     [CHAIN.BSC]: { fetch },
     [CHAIN.POLYGON]: { fetch },
--- a/fees/dodo-fees.ts
+++ b/fees/dodo-fees.ts
@@ -95,7 +95,7 @@ const breakdownMethodology = {
 const adapter: Adapter = {
   version: 1,
   adapter: {
-    [CHAIN.DFIO_META_MAIN]: { fetch: dfioFetch },
+    [CHAIN.DFIO_META_MAIN]: { fetch: dfioFetch, deadFrom: '2026-08-29' },
     [CHAIN.ETHEREUM]: { fetch },
     [CHAIN.BSC]: { fetch },
     [CHAIN.POLYGON]: { fetch },


bheluga:
@0xshubhs thanks for the PR

PR #9104, opened 2026-09-01
gmgn: rebuild volume from Dune's first-party tables, the uploaded dataset it reads stopped on 2026-08-25

tarun-khatri:
`dexs/gmgn` has published nothing since 2026-08-25. `total24h` is `null`, `total30d` is $2,382,941,413, and the last stored day was $112,658,427.

The adapter's only source is a user-uploaded Dune table. Querying it today, every chain stops on the same day:

```
blockchain | max_day    | min_day    | rows
base       | 2026-08-25 | 2024-06-12 |  791
bnb        | 2026-08-25 | 2024-12-02 |  613
ethereum   | 2026-08-25 | 2023-08-05 | 1066
hyperevm   | 2026-08-25 | 2026-05-21 |   81
megaeth    | 2026-08-23 | 2026-03-02 |   55
monad      | 2026-08-25 | 2026-01-01 |  227
robinhood  | 2026-08-25 | 2026-07-03 |   54
solana     | 2026-08-25 | 2024-03-20 |  889
xlayer     | 2026-08-25 | 2026-06-23 |   64
```

There is no `arbitrum` row at all, so the chain added in #8877 has reported $0 every day since 2026-08-19 even though the collector took real fees there ($33,780 on 2026-08-31 alone). `prefetch` throws on an empty day, which takes all ten chains down together.

GMGN itself is trading normally — the fees adapter reported $2,062,504 on 2026-08-31 off Dune's first-party tables.

## The change

Volume comes from `dex.trades` / `dex_solana.trades`, restricted to the transactions that pay a GMGN fee collector. That is the same transaction set `fees/gmgnai.ts` already measures, so the two adapters can no longer drift apart, and it needs no new addresses beyond the ones that file already maintains. `prefetch`/`fetch` keep their existing shape; the 10-hour Dune-indexing guard matches the sibling bot adapters.

## Why the fee collector rather than a router list

GMGN rotates routers. On ethereum, 2026-01-14 through 01-16 the flow ran through `0xb030e926..`, not the `0x4313c378..` the previous EVM code hardcoded:

```
2026-01-14  0xb030e926..  222 hops   $41,170     0x4313c378..  6 hops  $682
2026-01-15  0xb030e926..  104 hops   $32,273     0x4313c378..  5 hops  $322
2026-01-16  0xb030e926..   97 hops   $26,393
```

[...]

PR #5378, opened 2025-12-25
Fix contango V2

FelixBruguera:
Closes #5209. The subgraph stopped working so I replaced it with a Dune query

PR #9172, opened 2026-09-02
contango: price open interest with the last available price, it has published $0 since 2025-12-27

### LINKED ISSUE #5068: Fix adapter: Contango V2 OI

tarun-khatri:
Closes #5068.

`contango-v2` open interest has published an explicit **$0 every day since 2025-12-27** — 249 consecutive zero days. The adapter is not erroring: `totalDataChartBreakdown` carries all ten chains every day, each at `0`.

```
2025-12-26   Ethereum 94,778,320  Base 16,471,730  Arbitrum 14,325,091  OP 8,827,687
             Avalanche 5,136,678  Gnosis 1,580,957  Polygon 371,178  BSC 123,801
             Linea 45,227  Scroll 15,784                        TOTAL 141,676,453
2025-12-27   all ten chains 0                                   TOTAL 0
```

The protocol is alive. `fees/contango` is an independent on-chain adapter (`addTokensReceived`, no Dune, no subgraph) and reports `total24h $19.04`, `total7d $77.79`, non-zero on each of the last four days.

## Root cause

The `oi` half of the Dune query prices each day's position size with an exact-date join:

```sql
INNER JOIN DUNE.CONTANGO_XYZ.RESULT_V2_DAILY_PRICES_USD AS PRICE
ON (PRICE.ASSET = OI.ASSET AND PRICE.TIMESTAMP = OI.TIMESTAMP)
```

`RESULT_V2_DAILY_PRICES_USD` is refreshed a day behind `V2_TRANSACTIONS` — today it ends at 2026-08-31 while transactions run to 2026-09-01. So on the day the adapter actually asks for there is no price row, the `INNER JOIN` drops every asset, and the outer `COALESCE(SUM(OI.OI_USD), 0)` turns the empty result into a successful `$0`.

**The query is correct in hindsight and wrong in real time**, which is why nothing looks broken when you re-run it. Running this file's own query verbatim through the same wrapper the adapter uses — `queryDuneSql`, Dune query `3996608` — for a range of requested days:

| day requested | Ethereum | Arbitrum |
|---|---|---|
| 2026-09-01 | **0** | **0** |
| 2026-08-31 | 66,625,879.06 | 7,976,364.02 |
| 2026-06-15 | 49,756,376.27 | 6,705,181.31 |
| 2026-01-15 | 109,154,524.12 | 15,715,346.32 |
| 2025-12-26 | 94,778,320.27 | 14,325,090.65 |

Only the most recent day fails, and that is the only day the adapter ever asks for. 249 zeros out of 249 days means the lag is structural, not occasional — had the price table ever caught up before a run, that day would be non-zero.

Note the last row: replaying the query for **2025-12-26 reproduces DefiLlama's own last published values to the cent** (94,778,320.27 against 94,778,320 and 14,325,090.65 against 14,325,091). The methodology is intact; only the timing of the price join is broken.

Not an upstream outage — the four source tables are healthy. `V2_TRANSACTIONS` runs to 2026-09-01 on nine chains, `RESULT_V2_ALL_TRADES` to 2026-09-01, `RESULT_DAILY_TIMESTAMPS` to 2026-09-06, and all 179 assets in `V2_TRANSACTIONS.BASE` appear in the price table's `ASSET` (179/179 overlap, 248 assets priced per day).

The cliff date is the merge date of #5378 (2025-12-27), which replaced the DNS-dead Satsuma subgraph with this query, so the OI branch has never once produced a non-zero value in production.

## Fix

Carry the last price at or before the requested day instead of requiring an exact date match. On any day the price table already covers, that resolves to the same day's price, so historical values do not move.

**No regression** — old and new join produce identical values to the cent:

```
2026-08-31 arbitrum  old=      7,976,364.02  new=      7,976,364.02  SAME
2026-08-31 mainnet   old=     66,625,879.06  new=     66,625,879.06  SAME
2026-06-15 arbitrum  old=      6,705,181.31  new=      6,705,181.31  SAME
2026-06-15 mainnet   old=     49,756,376.27  new=     49,756,376.27  SAME
2026-01-15 arbitrum  old=     15,715,346.32  new=     15,715,346.32  SAME
2026-01-15 mainnet   old=    109,154,524.12  new=    109,154,524.12  SAME
2025-12-26 arbitrum  old=     14,325,090.65  new=     14,325,090.65  SAME
2025-12-26 mainnet   old=     94,778,320.27  new=     94,778,320.27  SAME
```

**The day that was broken**, all ten chains, 2026-09-01:

```
mainnet        66,625,879.06        gnosis      1,135,558.38
arbitrum        7,976,362.80        avalanche   1,162,739.67
base            6,740,545.37        polygon       283,072.82
optimism        4,286,984.87        bsc            74,163.81
                                    linea          29,318.92
                                    scroll         13,022.80
TOTAL          88,327,648.50   (published: 0)
```

## Notes

- Same failure mode as #8970 (dtrade), where `COALESCE(..., 0)` turned an incomplete Dune window into a valid-looking zero. That one throws so the run can be retried, which is right when the lag is intermittent. Here the price table is behind on *every* run, so throwing would leave the series permanently empty; the position data itself is complete and only its valuation is a day old.
- Scope is the price join only. `dailyVolume` and the Dune `volume` CTE are untouched, and I left `Number(response[0].OI_USD ? ... : 0)` alone rather than converting it to a throw as in #8524/#8525 — once the join works there is a real value to publish, and changing both would mix two concerns in one diff.
- Open interest is a stock, so a one-day-old valuation is a rounding-level effect: Ethereum moves 66,625,879 → 67,213,856 between 2026-08-31 and 08-30, under 1%.
- The red `test` check is `DUNE_API_KEYS environment variable is not set` — the runner has no keys, so it cannot execute any Dune-backed adapter. #8970, #8524 and #9104 all merged with the same red `test` and green `ts-check`.

### DIFF
--- a/dexs/contango/index.ts
+++ b/dexs/contango/index.ts
@@ -2,53 +2,61 @@ import { Dependencies, FetchOptions, SimpleAdapter } from "../../adapters/types"
 import { CHAIN } from "../../helpers/chains";
 import { queryDuneSql } from "../../helpers/dune";
 
-const chainMap: Record<string, string> = {
-  [CHAIN.ETHEREUM]: "mainnet",
-  [CHAIN.XDAI]: "gnosis",
-  [CHAIN.AVAX]: "avalanche",
-  [CHAIN.BSC]: "bsc",
+const chainConfig: Record<string, { duneName: string; start: string }> = {
+  [CHAIN.ARBITRUM]: { duneName: "arbitrum", start: "2023-10-03" },
+  [CHAIN.OPTIMISM]: { duneName: "optimism", start: "2023-10-02" },
+  [CHAIN.ETHEREUM]: { duneName: "mainnet", start: "2023-10-03" },
+  [CHAIN.POLYGON]: { duneName: "polygon", start: "2023-10-13" },
+  [CHAIN.BASE]: { duneName: "base", start: "2023-10-09" },
+  [CHAIN.XDAI]: { duneName: "gnosis", start: "2023-10-06" },
+  [CHAIN.AVAX]: { duneName: "avalanche", start: "2024-08-11" },
+  [CHAIN.LINEA]: { duneName: "linea", start: "2024-08-11" },
+  [CHAIN.BSC]: { duneName: "bsc", start: "2024-06-07" },
+  [CHAIN.SCROLL]: { duneName: "scroll", start: "2024-08-11" },
 };
 
-const fetch = async (options: FetchOptions) => {
-  const { chain, fromTimestamp, toTimestamp } = options;
-  const mappedChain = chainMap[chain] ?? chain
-  const query = `
-    WITH volume AS (
-      SELECT SUM(ABS(QUANTITY_USD)) AS VOL_USD 
-      FROM DUNE.CONTANGO_XYZ.RESULT_V2_ALL_TRADES
-      WHERE TIMESTAMP >= from_unixtime(${fromTimestamp}) AND TIMESTAMP <= from_unixtime(${toTimestamp}) AND chain = '${mappedChain}'
-    ), 
-    oi as (
-      with LONG_OI_DELTA as (
-        SELECT DATE_TRUNC('day', T.TIMESTAMP) AS TIMESTAMP, T.BASE AS ASSET, SUM(T.QUANTITY) AS DELTA
-        FROM DUNE.CONTANGO_XYZ.V2_TRANSACTIONS AS T
-        WHERE T.chain = '${mappedChain}' AND T.DIRECTION = 'Long'
-        GROUP BY 1, 2
-      ),
-      SHORT_OI_DELTA as (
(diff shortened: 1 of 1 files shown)


tarun-khatri:
﻿On the red `test` check: it is `Error: DUNE_API_KEYS environment variable is not set` at `helpers/dune.ts:17`, so the runner cannot execute any Dune-backed adapter and the failure is independent of this change. The same check is red on the merged Dune PRs this one is modelled on — #8970 (dtrade), #8524 (kalshi) and #9104 (gmgn) all merged with `test: failure` / `ts-check: success`, which is the pattern here. `ts-check` passes.

Because the runner can't exercise it, I ran the query through the same wrapper the adapter uses (`queryDuneSql` -> Dune query `3996608`) with my own key, which is the identical code path. The numbers in the description come from that, including the eight-day old-vs-new comparison showing historical values unchanged to the cent.

bheluga:

PR #9176, opened 2026-09-02
feat(fees): add SolStocker adapter (Robinhood Chain)

utxodotfun:
Adds a fees/revenue adapter for **SolStocker**, an RWA launchpad on Robinhood Chain (chainId 4663).

### How the fee works

Every coin launched through SolStocker trades in a Uniswap v4 pool whose **LP fee is pinned to zero** — the hook rejects registration of any pool with a non-zero LP fee. The entire 1.5% swap fee is taken by one singleton hook, `SolStockerFeeHook` at `0xa9F8Da2F2578cA79245FC7f5a96F3A914835E0cC`, and emitted per swap on `FeeCharged` as five compile-time slices:

| Slice | bps | Goes to |
| --- | --- | --- |
| holderFee | 75 | Holders of the launched coin |
| creatorFee | 15 | The coin's creator |
| rewardsFee | 15 | Pro-rata rebate back to traders |
| stockerFee | 15 | Holders of STOCKER, the protocol's own token |
| platformFee | 30 | SolStocker |

There is no launch fee and no graduation fee.

### Mapping

- `dailyFees` — all five slices (the full 1.5%)
- `dailySupplySideRevenue` — holder + creator + rewards (105 bps), value routed to third parties
- `dailyHoldersRevenue` — stockerFee (15 bps), the only slice accruing to the protocol's own token holders
- `dailyProtocolRevenue` — platformFee (30 bps)
- `dailyRevenue` — protocol + holders (45 bps)

### Not double counted

- **Uniswap v4**: the pools' LP fee is 0, so the v4 adapter sees nothing from these swaps.
- **Solana is deliberately excluded.** SolStocker also runs on Solana, but neither Solana stream can be attributed cleanly. Its on-chain creator-fee split is a share of pump.fun's 0.30% creator fee, which the `pumpdotfun` adapter already reports as its own fees and supply-side revenue. Its 1% on-site swap fee is taken through Jupiter's platform-fee mechanism and lands in the output mint — a per-coin Token-2022 account on buys, a wrapped-SOL account on sells — with no fixed target to sum, and those accounts are later swept between wallets SolStocker controls, so a wallet-inflow query would over-report. Solana will be added when it can be attributed without inflating the numbers.

### Denomination

Fees are charged in the pool's **quote asset**, a tokenized real-world asset rather than the gas token. `FeeCharged` carries only the poolId, so the quote asset is resolved from the hook's `PoolRegistered` events (scanned once from the hook's first block, `cacheInCloud`) and each slice is added against that ERC-20 address.

### Test output

`npm test fees/solstocker` against live data:

```
ROBINHOOD
Daily fees: 79.49
Daily user fees: 79.49
Daily revenue: 23.75
Daily protocol revenue: 15.90
Daily holders revenue: 7.95
Daily supply side revenue: 56.03
```

The slices reconcile to the hook's constants: coin holders 39.87 is exactly half of total fees, creator / trader rewards / STOCKER holders are 7.95 each (15 bps), and protocol is 15.90 (30 bps, exactly double).

### DIFF
--- a/fees/solstocker/index.ts
+++ b/fees/solstocker/index.ts
@@ -1,0 +1,197 @@
+// SolStocker — fees & revenue adapter (Robinhood Chain).
+//
+// SolStocker (https://solstocker.fun) is an RWA launchpad. On Robinhood Chain
+// (chainId 4663) every launched coin trades in a Uniswap v4 pool whose quote
+// asset is a tokenized equity (SPCX, NVDA, ...) and whose LP fee is pinned to
+// ZERO. The entire trading fee is taken by one singleton hook instead:
+//   SolStockerFeeHook 0xa9F8Da2F2578cA79245FC7f5a96F3A914835E0cC
+//
+// Because the pool's LP fee is 0 (the hook rejects any pool with a non-zero
+// LP fee at registration), nothing counted here is also counted by the
+// Uniswap v4 adapter — this is not double counted.
+//
+// Fee source (the only fee this hook charges — no launch fee, no graduation
+// fee): TOTAL_SWAP_FEE_BPS = 150, i.e. 1.5% of the quote-asset leg of every
+// swap, split into five named slices that are compile-time constants of the
+// hook and are emitted per swap on FeeCharged:
+//
+//   holderFee   75 bps  holders of the LAUNCHED coin, via the pool's reward vault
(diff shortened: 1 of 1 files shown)


bheluga:
@utxodotfun thanks for the PR

PR #9180, opened 2026-09-03
chore: change ctf-exchange address and naming as per pred

soumya-pred:


bheluga:
@soumya-pred thanks for the PR

PR #9181, opened 2026-09-03
[fix] drake-exchange - restore commission fee accounting

Akkii4:
## Summary

- Normalize AssetTransferred action types before filtering commission and margin-fee events.
- Read both the legacy net funding/borrowing event and the current funding-plus-borrowing event.
- Record only positive net funding/borrowing fees.

## Why

The log decoder returns the uint8 action type as a bigint or string, while the adapter compared it to JavaScript numbers. This caused commission and margin-fee events to be skipped. The funding/borrowing event ABI also changed from one net-fee field to separate funding and borrowing fields.

## Impact and design

Trading commission fees now reconcile with Drake analytics: the adapter assigns the operator share to protocol revenue and the vault share to supply-side revenue using the execution-leg split already tracked by the adapter. Querying both event signatures preserves historical coverage while supporting current deployments; their distinct event topics prevent double counting.

## Validation

- TypeScript check passed.
- Focused regression test passed for string and bigint action types and both funding-event schemas.
- Live Monad adapter execution returned non-zero fees, protocol revenue, and supply-side revenue.

### DIFF
--- a/fees/drake-exchange/index.ts
+++ b/fees/drake-exchange/index.ts
@@ -10,6 +10,13 @@ const PLATFORM_MANAGER = "0x7940575377C3c2ABdA23813c123b4C880E217d6d";
 
 const COMMISSION_FEE = 3;
 const MARGIN_CHANGE_FEE = 4;
+const ASSET_TRANSFERRED_EVENT =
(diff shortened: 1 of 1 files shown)


bheluga:
@Akkii4 thanks for the PR

PR #9190, opened 2026-09-03
feat(arcadia-finance-v2): add Robinhood chain

j-vp:
Arcadia will go live on Robinhood Chain (4663) with two lending pools, wETH and USDG. The chain lends USDG rather than USDC, which is the only thing that differs from the other configs; CHAIN.ROBINHOOD and coreAssets' robinhood block already exist upstream, as do the shared asset-manager addresses, so nothing else was needed.

One fee-taking contract was not listed in the adaptor, 0xb988a32d, CoW Swapper V1.1.1.


bheluga:
@j-vp thanks for the PR

PR #9197, opened 2026-09-03
feat: pred now collects fees from users

soumya-pred:


bheluga:
@soumya-pred thanks for the PR

PR #9203, opened 2026-09-03
mark seven adapters dead whose only data source no longer resolves in DNS

tarun-khatri:
I resolved every hostname referenced in the repo — 1,396 distinct hosts across all `*.ts` files — against both Cloudflare's and Google's DNS-over-HTTPS resolvers. 93 of them no longer resolve. Filtering to adapters whose **only** data source is one of those hosts, that have no `deadFrom`, and that are currently publishing nothing, leaves the seven below.

Each one is unrecoverable rather than merely broken: there is no host left to query, so no retry, key or query change can bring it back.

| adapter | host it reads | host DNS | apex DNS | last published point | TVL now |
|---|---|---|---|---|---|
| `dexs/ponytaswap` | `www.ponytaswap.finance` | NXDOMAIN | NXDOMAIN | 2025-06-04 (last non-zero 2024-12-27) | $0 since 2026-01-26 |
| `dexs/seiyan-fun` | `seiyan.fun` | NXDOMAIN | NXDOMAIN | 2025-07-02 (last non-zero 2025-06-30, $8) | no longer reported |
| `dexs/rush` | `stats.rushbot.io` | NXDOMAIN | NXDOMAIN | 2025-12-01 ($11, 24 points ever) | no longer reported |
| `dexs/vanswap` | `www.vanswap.org` | SERVFAIL | SERVFAIL | 2024-12-09 ($7,798) | $3.16M → $0 on 2026-02-20 |
| `dexs/utyabswap` | `api.utyabswap.com` | NXDOMAIN | no A record | 2025-08-26 ($1,994) | $6 |
| `dexs/blum` | `tonfunstats-eqnd7.ondigitalocean.app` | NXDOMAIN | — | 2025-10-31 ($93) | $0 |
| `options/arrow-markets` | `api-rfq-testnet.prd.arrowmarkets.info` | NXDOMAIN | no A record | 2026-08-06 (last non-zero 2026-05-23, $26) | $0 |

SERVFAIL rather than NXDOMAIN means the delegation itself is broken — both resolvers agree, so it is the domain and not one resolver's view.

Two details worth calling out. `options/arrow-markets` was pointed at a **testnet** RFQ host, which is why the days it did still answer came back as `0`. And `dexs/vanswap`'s TVL going from $3.16M to zero on 2026-02-20 is independent of the DNS failure, so the protocol and not just the feed is gone.

Each `deadFrom` is the day after that adapter's own last published point, so no already-stored history is affected.

## Files I checked and deliberately did not include

`tonfunstats-eqnd7.ondigitalocean.app` is read by three adapters, not one — `dexs/blum`, `dexs/bigpump` and `dexs/tonpump`. All three are equally unable to fetch. I am only marking `blum`, whose TVL reads $0.

`bigpump` and `tonpump` still report TVL that updates daily off on-chain state ($12,170 and $1,782 as of 2026-09-03), so I can show this data source is dead but not that either protocol is, and marking a live protocol dead would replace a gap with a wrong zero. Both need a TON-side rebuild or a call from someone with more context than I have, which is a different change from this one.

I also excluded every other adapter on the 93-host list that either already carries `deadFrom`, still publishes through a second live source, or belongs to a protocol I could show is still alive.

### DIFF
--- a/dexs/blum/index.ts
+++ b/dexs/blum/index.ts
@@ -23,6 +23,9 @@ const adapter = {
       start: '2024-10-24',
     },
   },
+  // tonfunstats-eqnd7.ondigitalocean.app is NXDOMAIN, so this adapter has no host to query.
+  // Last published point 2025-10-31 ($93) and TVL reads 0.
+  deadFrom: '2025-11-01',
 }
 
 export default adapter
--- a/dexs/ponytaswap/index.ts
+++ b/dexs/ponytaswap/index.ts
@@ -27,6 +27,10 @@ const adapter: SimpleAdapter = {
   fetch,
   chains: [CHAIN.RPG],
   start: '2023-03-06',
+  // www.ponytaswap.finance and the ponytaswap.finance apex are both NXDOMAIN, so the only source
+  // this adapter has cannot answer. Last published point 2025-06-04, last non-zero 2024-12-27,
+  // and its TVL has read 0 since 2026-01-26.
+  deadFrom: '2025-06-05',
 };
 
 export default adapter;
--- a/dexs/rush/index.ts
+++ b/dexs/rush/index.ts
@@ -31,6 +31,9 @@ const adapter: SimpleAdapter = {
   chains: [CHAIN.SOLANA],
   start: "2025-11-07",
   methodology,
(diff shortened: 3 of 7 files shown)


bheluga:

PR #9207, opened 2026-09-03
mark six more adapters dead: the only data source is gone and the protocol site with it

tarun-khatri:
Follow-up to #9203, from the same repo-wide DNS sweep. These six are the cases where **the data source and the protocol's own website are both gone** — so unlike a broken endpoint, there is nothing to repoint to and no app still using them.

Every DNS result below is from both Cloudflare's and Google's DoH resolvers, and they agree in every case.

| adapter | source it reads | source DNS | apex | apex DNS | last published point |
|---|---|---|---|---|---|
| `dexs/xpress` | `api.studio.thegraph.com/query/61208/onchain-clob-sonic` | host up, **deployment gone** | `xpress.trade` | NXDOMAIN | 2026-06-07 |
| `dexs/jediswap-v2` | `api.v2.jediswap.xyz` | NXDOMAIN | `jediswap.xyz` | NOERROR, 0 A records | 2025-07-30 (last non-zero 2025-01-15, $167) |
| `dexs/pixelswap` | `api.pixelswap.io` | SERVFAIL | `pixelswap.io` | SERVFAIL | 2026-01-28 (last non-zero 2024-12-06, $510) |
| `dexs/swapline` | `api.swapline.com` + `api-c.swapline.com` | NOERROR, 0 A records | `swapline.com` | NOERROR, 0 A records | 2024-06-03 ($18,388) |
| `dexs/claimswap` | `data-api.claimswap.org` | SERVFAIL | `claimswap.org` | SERVFAIL | 2026-04-27 (last non-zero 2025-07-28, $11,202) |
| `dexs/polkadex` | `integration-api.polkadex.trade` | SERVFAIL | `polkadex.trade` | SERVFAIL | 2025-01-14 ($1) |

`xpress` is the one that is not a DNS failure. Its Graph Studio host is perfectly healthy — it just no longer has the deployment:

```
$ curl -s -XPOST https://api.studio.thegraph.com/query/61208/onchain-clob-sonic/version/latest \
    -d '{"query":"{ _meta { block { number } } }"}'
{"errors":[{"message":"deployment `u61208/s106679/latest` does not exist"}]}
```

which together with `xpress.trade` being NXDOMAIN makes it the same situation as the rest.

The three DNS states are deliberately reported separately because they are different claims. **NXDOMAIN** means the name does not exist. **SERVFAIL** means the domain's own nameservers no longer answer, i.e. the delegation is broken. **NOERROR with zero A records** means the domain is still registered and its DNS is healthy, but it publishes no address — the registration outlived the deployment.

`dexs/swapline` is worth one note: it is configured for four chains (Fantom, Optimism, Arbitrum, Shimmer EVM), but all four share the same two hosts, so no single chain can still report and marking the adapter rather than a leg is correct here.

Each `deadFrom` is the day after that adapter's own last published point, so no stored history changes.

### DIFF
--- a/dexs/claimswap/index.ts
+++ b/dexs/claimswap/index.ts
@@ -41,6 +41,9 @@ const adapter: SimpleAdapter = {
   fetch,
   chains: [CHAIN.KLAYTN],
   start: START_TIME,
+  // data-api.claimswap.org and the claimswap.org apex both SERVFAIL. Last published point
+  // 2026-04-27, last non-zero 2025-07-28 ($11,202).
+  deadFrom: '2026-04-28',
 };
 
 export default adapter;
--- a/dexs/jediswap-v2/index.ts
+++ b/dexs/jediswap-v2/index.ts
@@ -41,6 +41,9 @@ const adapter: SimpleAdapter = {
   fetch,
   chains: [CHAIN.STARKNET],
   start: '2024-02-10',
+  // api.v2.jediswap.xyz is NXDOMAIN and the jediswap.xyz apex resolves with no A record, so the
+  // app is gone too. Last published point 2025-07-30, last non-zero 2025-01-15 ($167).
+  deadFrom: '2025-07-31',
 };
 
 export default adapter;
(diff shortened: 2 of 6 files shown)


bheluga:

PR #9212, opened 2026-09-03
kriya: rebuild AMM volume from its own swap events, the API and the site are both gone

### LINKED ISSUE #9210: kriya: both adapters lost their data source, but the AMM is still emitting swaps on-chain
Both Kriya adapters have lost their data source, but **Kriya is still trading right now**, so neither should be marked dead — it needs an on-chain rebuild.

## The sources are gone

`dexs/kriya-dex` reads `https://api.kriya.finance/defillama/amm/`. That host is **NXDOMAIN** from both Cloudflare's and Google's DoH resolvers, and the `kriya.finance` apex resolves NOERROR with **zero A records**, so the site is gone too.

`dexs/kriya-clmm` reads `https://api-service-81678480858.asia-northeast1.run.app/pools/v3`. That Cloud Run service returns **404 on every path, including the root**.

Last published points: `kriya-amm` 2026-04-28, `kriya-clmm` 2026-04-29. Nothing since.

## But the protocol is alive

The AMM package is emitting swaps continuously. Via Sui's GraphQL RPC:

```graphql
# https://graphql.mainnet.sui.io/graphql
{
  events(last: 50, filter: {module: "0xa0eba10b173538c8fecca1dff298e488402cc9ff374f8a12ca7758eebe830b66::spot_dex"}) {
    nodes { timestamp contents { json } }
  }
}
```

returns 50 `spot_dex::SwapEvent`s between **2026-09-03T05:31:28Z and 2026-09-03T16:41:02Z** — i.e. today, continuously. A sample payload:

```json
{
  "pool_id":  "0x3c334f9d1b969767007d26bc886786f9f197ffb14771f7903cd8772c46d08dea",
  "user":     "0x98391d8f43736152b61caaa38b615621095d7595f34ae6de18d507996da8748e",
  "reserve_x": "6622374488270",
  "reserve_y": "8451051186302",
  "amount_in": "2355110546",
  "amount_out": "3002512143"
}
```

So this is a dead feed, not a dead protocol, and `deadFrom` would replace a gap with a wrong zero on a DEX that is still doing volume. TVL still reports $188k (AMM) and $37k (CLMM), consistent with that.

## What a fix needs, and why I have not sent one

The event carries everything required — `pool_id`, `amount_in`, `amount_out` and both reserves — so volume is reconstructible. Two things need deciding first, and guessing either would silently redefine what this adapter reports:

1. **Which side to count.** `amount_in` and `amount_out` are the two legs of one swap; counting both doubles the volume. The usual convention is to price one side, but that requires resolving each `pool_id` to its token types, which the event does not carry.
2. **The CLMM half is unresolved.** I could not find an emitting module for `0xf6c05e2d…` — `trade`, `pool`, `clmm`, `swap` and `position` all return no events. It may use a different module name, or the CLMM may genuinely be idle while the AMM is not, and those two cases call for different fixes.

One practical note for whoever picks this up: `suix_queryEvents` on `fullnode.mainnet.sui.io` now returns **"Method not found"**, so the old JSON-RPC recipe no longer works — the GraphQL endpoint above is what answers.

tarun-khatri:
Closes #9210.

`dexs/kriya-dex` has published nothing since **2026-04-28**. Its endpoint `api.kriya.finance/defillama/amm/` is NXDOMAIN and the `kriya.finance` apex resolves NOERROR with **zero A records**, so the site went with it (both checked against Cloudflare's and Google's DoH resolvers).

Everything about that says "mark it dead". **It isn't dead** — the AMM is still trading, so this reads volume from the pools' own swap events instead.

## The protocol is alive

```graphql
# https://graphql.mainnet.sui.io/graphql
{ events(last: 50, filter: {type: "0xa0eba10b…830b66::spot_dex::SwapEvent"}) {
    nodes { timestamp contents { json type { repr } } } } }
```

On **2026-09-03** the AMM emitted swaps continuously from 00:0x through 20:44. Over the fully covered 14:00–20:29 window (6.48 h, 600 swaps) I priced the input side with `coins.llama.fi` at **$4,323.18**, across **153 distinct swappers** with the busiest address at only 11.7% of swaps — real user flow, not one wallet cycling.

For contrast GeckoTerminal reports $163.69 of 24h volume for `kriya-dex`. Its per-pool TVLs match mine exactly (its USDC/SUI pool at $100,884 is the same pool where I measured 299 swaps / $3,557.50 in 6.5 h), so it has the pools right and the volume stale — its Kriya integration presumably stopped when the API did.

## Why the input side, and how that is known

The event is emitted as `SwapEvent<CoinIn>`, and the payload has **no direction field**:

```json
{"pool_id":"0x5af4976b…","user":"0x2129…","reserve_x":"49976716941",
 "reserve_y":"65037344332525","amount_in":"1498995000","amount_out":"1150368"}
```

so the single type parameter is the only thing identifying which side was paid in. I verified that rather than assuming it: for every consecutive pair of swaps on the same pool, I checked whether the reserve of the coin named in the type parameter moved by exactly `amount_in` and the other by exactly `-amount_out`, using each pool object's `Pool<CoinX, CoinY>` ordering. **585 of 588 pairs match exactly (99.5%)** — which also pins `reserve_x`/`reserve_y` as post-swap. The three that miss are pairs with a liquidity event in between.

Counting only the coin paid in books each swap once. That is the same convention `dexs/bluemove.ts` uses for its Sui swaps (`amount_x_in > 0 ? add(token_x_in) : add(token_y_in)`).

## Why Allium and not the GraphQL RPC

I wrote this on `queryEvents` from `helpers/sui.ts` first. It silently returns a fraction of the day, because the endpoint's backward pagination has a **traversal budget rather than a time limit** — `hasPreviousPage` goes false early, and non-deterministically (paging with `last: 10` stopped after 285 events, `last: 50` after 1,370).

Measured through the helper itself, one event type per row:

| event type | 2026-09-03 | 2026-09-01 | 2026-08-20 |
|---|---|---|---|
| kriya `spot_dex::SwapEvent` | 150 | **0** | **0** |
| waterx `events::PositionOpened` | 9 | 14 | 85 |
| typus-perp `position::OrderFilledEvent` | 1 | 1 | 1 |

So the helper is fine for its current callers — they emit few enough events that it pages back weeks. Kriya emits a few thousand a day and exhausts the budget inside one day, which would have shipped a silent ~90% undercount. `helpers/sui.ts` already says as much (*"graphql queryEvents doesn't retrieve historical events so we use allium for this"*), so this uses Allium, with the same `sui.raw.events` query shape as `queryEventsAllium` — selecting `type` as well as `parsed_json`, since the coin paid in lives in the type.

The empty-result guard is there so that an Allium outage surfaces instead of being published as a legitimate zero, which is what this adapter's previous silence looked like.

## One thing deliberately left alone

`dexs/kriya-clmm` is a different package (`0xf6c05e2d…`) and is **not** included. Its last `trade::SwapEvent` was **2026-08-12T14:22Z**; after that only `RemoveLiquidityEvent`, `ClosePositionEvent` and fee/reward collection through **2026-08-24**, and nothing at all since. Its event also has a real `x_for_y` field and no type parameters, so it needs a pool-to-coin-type map rather than this code path. Different evidence, different fix — it should not ride along on this one.

## CI

The `test` check fails with `Allium API Key is required[Ignore this error for github bot]` — the workflow sets no secrets, so this is the same situation as every Dune-backed adapter (#8970, #8524, #9104 all merged with `test: failure` and `ts-check: success`). `ts-check` passes.

### DIFF
--- a/dexs/kriya-dex/index.ts
+++ b/dexs/kriya-dex/index.ts
@@ -1,4 +1,4 @@
-import { FetchOptions, FetchResult, SimpleAdapter } from "../../adapters/types";
+import { Dependencies, FetchOptions, FetchResult, SimpleAdapter } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
 import { queryAllium } from "../../helpers/allium";
 
@@ -17,12 +17,8 @@ const inputCoin = (eventType: string) =>
 const fetch = async (options: FetchOptions): Promise<FetchResult> => {
     const dailyVolume = options.createBalances();
 
-    // Whole UTC day, half open. Kriya emits a few thousand swaps a day, which is more than the
-    // GraphQL RPC will page back through, so this reads the events from Allium the way
-    // queryEventsAllium in helpers/sui.ts does. The type is selected as well as the payload
-    // because the coin paid in is carried in the event's type parameter.
-    const start = new Date(options.startOfDay * 1000).toISOString();
-    const end = new Date((options.startOfDay + 86400) * 1000).toISOString();
+    const start = new Date(options.fromTimestamp * 1000).toISOString();
+    const end = new Date(options.toTimestamp * 1000).toISOString();
 
     const rows: { type: string, parsed_json: any }[] = await queryAllium(`
         SELECT type, parsed_json
@@ -41,7 +37,10 @@ const fetch = async (options: FetchOptions): Promise<FetchResult> => {
 };
 
 const adapter: SimpleAdapter = {
-    version: 1,
+    version: 2,
+    pullHourly: true,
+    dependencies: [Dependencies.ALLIUM],
+    isExpensiveAdapter: true,
     fetch,
     chains: [CHAIN.SUI],
     start: '2023-05-09',
--- a/dexs/kriya-dex/index.ts
+++ b/dexs/kriya-dex/index.ts
@@ -1,35 +1,53 @@
-import fetchURL from "../../utils/fetchURL"
 import { FetchOptions, FetchResult, SimpleAdapter } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
-import { getUniqStartOfTodayTimestamp } from "../../helpers/getUniSubgraphVolume";
+import { queryAllium } from "../../helpers/allium";
 
-type IUrl = {
-    [s: string]: string;
-}
+// api.kriya.finance is NXDOMAIN and the kriya.finance apex resolves with no A record, so the
+// endpoint this adapter read is gone along with the site. The AMM itself is still trading, so
+// volume comes from the swap events its pools emit.
+const PACKAGE = "0xa0eba10b173538c8fecca1dff298e488402cc9ff374f8a12ca7758eebe830b66";
+const SWAP_EVENT = `${PACKAGE}::spot_dex::SwapEvent`;
 
-const url: IUrl = {
-    [CHAIN.SUI]: `https://api.kriya.finance/defillama/amm/`
-}
-
-interface IVolume {
-    dailyVolume: number,
-}
+// The event is emitted as SwapEvent<CoinIn>: its single type parameter is the coin that went INTO
+// the pool, which is what amount_in is denominated in. The payload carries no direction field, so
+// the type parameter is the only thing that identifies the side.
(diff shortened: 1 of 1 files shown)


bheluga:

PR #9215, opened 2026-09-04
Orchard: add the v2 game contract to fees, revenue and volume (Robinhood Chain)

orchard-gold:
Orchard v2 went live on Robinhood Chain on 2026-09-03 (`0x86510b3df745C67a993A66CB08720Ed158d44549`, deployed at block 53786732). v1 (`0xEbB8b167c0992cFdc497A995a8Cf7167acAA0A1A`) stays open, so the adapter now sums both game contracts. The v1 path is unchanged, only moved into `fetchV1`; the rake split is shared by both versions.

How v2 differs and how it is measured:
- **Volume**: v2 emits one `Planted(player, round, plotMask, value, flags)` per entry whatever the entry point (single plant, many rounds, batched, standing orders). `value` is ETH, or AAPL when `FLAG_AAPL` is set (v2 lets players plant from an AAPL balance), and is added as such.
- **Admin fee**: v1 takes it inside each plant, so it is reconstructed per event as before. v2 takes it when the keeper seals the round: the round's ETH is swapped to AAPL at once and `adminFeeBps` is carved out of what was bought; AAPL-denominated entries pay the same rate on their value and batched entries an extra `batchFeeBps` (0 today). All of it is added to `adminAccrued`, and `collectAdmin` is the only thing that decreases it, so the fee taken inside the window is the growth of `adminAccrued` between the window's blocks plus any `AdminCollected` inside it (withdrawals at the shared boundary block are left to the previous window). This reproduces the contract's own flooring exactly without replaying each round's entry list. Before the deploy block the opening value is taken as 0 instead of calling a contract that does not exist yet.
- **Rake**: identical formula and split in both versions (`rakeBps` on the loss pool → fixed 10% to SEED stakers, `jackpotBps` to the jackpot, remainder to the treasury), read back per revealed round from `getRound(id)` on v2 as it is from `rounds(id)` on v1. Empty blooming plots take no rake, as before.
- The 10% claim juice is still not counted: on v2 part of it is reserved for v1 players bridging their harvest over, but it never reaches protocol control either way.

Validated locally with `npm run test fees orchard 1788480000` (the 2026-09-03 UTC day, the launch day of v2): volume $441, fees $43 (admin fee $4.38, rake $39), revenue $24, protocol revenue $4.38, supply-side $19, holders $19. The admin fee figure was checked against the chain: it is exactly the v2 `adminAccrued` at the window's end block (0.011865 AAPL, no withdrawals yet) plus the ten v1 plants of the day (0.001482 AAPL), at the AAPL price DefiLlama had for that day. No dependency or lockfile changes.

### DIFF
--- a/fees/orchard/index.ts
+++ b/fees/orchard/index.ts
@@ -1,49 +1,60 @@
 // Orchard — fees & revenue adapter.
 //
-// Orchard is a plot game on Robinhood Chain (chainId 4663). Players plant ETH on a
-// 5x5 grid; the ETH is swapped to AAPL (Apple stock token) inside the same transaction,
-// so the game is played in AAPL. Every 75s round one plot blooms and whoever planted
-// there splits the harvest from the other 24 plots.
+// Orchard is a plot game on Robinhood Chain (chainId 4663), played in AAPL (Apple stock
+// token). Players plant on a 5x5 grid; every 75s round one plot blooms and whoever planted
+// there splits the harvest from the other 24 plots. Two versions of the game contract run
+// side by side and are summed here:
 //
-// Contracts (Robinhood Chain):
-//   Orchard: 0xEbB8b167c0992cFdc497A995a8Cf7167acAA0A1A
-//   AAPL:    0xaF3D76f1834A1d425780943C99Ea8A608f8a93f9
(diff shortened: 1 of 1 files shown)


bheluga:
@orchard-gold thanks for the PR

PR #9218, opened 2026-09-04
solstice: read distributed yield on-chain, Pyth Hermes now requires an API key

tarun-khatri:
`fees/solstice-usx` has published nothing since **2026-08-25**. It is the last adapter in the repo reading Pyth's Hermes endpoint (`git grep -l "hermes.pyth" upstream/master` returns only this file), and Hermes now requires a key:

```
GET https://hermes.pyth.network/v2/updates/price/<ts>?ids[]=f36e12e6…  ->  401 unauthorized
GET https://hermes.pyth.network/v2/updates/price/latest?ids[]=f36e12e6… ->  401 unauthorized
```

Its `/v2/price_feeds` metadata endpoint is still public and confirms the feed is a *"Crypto Redemption Rate"* for the Solstice yield-bearing stablecoin, but the values are gone. Same cause as #9164 (zest) and #9193 (wisdomtree); the fix is different because this is a redemption rate, not a market price.

## Not a price problem

The old version computed `(redemptionRate(today) − redemptionRate(yesterday)) × totalSupply(eUSX)`. **DefiLlama's own price for eUSX cannot stand in for that feed** — it is a market price and it goes *down* on some days (1.039748 → 1.040492 → 1.041009 → 1.039997 → 1.038421), while a redemption rate only rises. Substituting it would have produced negative yield days and would not reproduce the published series.

Solstice distributes the yield on-chain, so the transfer can be read directly and no rate feed is needed at all.

## What is read, and how it was checked

Each `DistributeYield` call emits a `TransferInYield` Anchor event from Solstice's program (`eUSXyKoZ6aGejYVbnp3wtWQ1E8zuokLAJPecPxxtgG3`). Rather than trust the layout, I matched one event against the transaction's actual token movements:

```
tx 3eD1bAgCXTThX2Mvg7iryVMo…   2026-09-04 09:02:07 UTC
  Program log: Instruction: DistributeYield
  Program log: Instruction: TransferInYield
  event u64 @ byte 72 = 666,330,000            ->  666.330000 USX

  token balances, mint 6FrrzDk5…  (USX)
    eUSX backing acct   21,630,658.240651 -> 21,631,324.570651   (+666.330000)
    yield acct          12,660.360031     -> 11,994.030031       (-666.330000)
```

The field is the transferred amount to the cent, so `dailyFees` is the sum of that u64 over the day, added as USX and priced by the framework.

The events are read from the yield account `HARVSXaBpt1TuD4PvqnscCQLzAWgDzGpkDadr76a489L`, which only these distributions touch — **1,183 signatures across 376 days, median 3 a day** — so walking its signatures back to the start of a day is a couple of RPC calls, not a program-wide scan.

## It reproduces the series the dead feed produced

Run locally against days the old adapter still covered:

| requested day | this adapter | DefiLlama published |
|---|---|---|
| 2026-08-23 | $4.57k | $4,566 |
| 2026-08-25 | $4.57k | $4,559 |
| 2026-09-03 | $1.31k | (feed already dead) |

and decoding the full event history gives the same shape day by day — 8,198.55 USX/day across 2026-08-14…08-19 against published 8,183 / 8,159 / 8,180 / 8,188 / 8,187 / 8,179, i.e. inside ~0.2% once USX's ~$0.9995 price is applied. So this is not a new methodology, it is the same number from a source that still exists.

## One bug fixed by construction

The old code called `getTokenSupply(EUSX)`, which returns the **current** mint supply regardless of the day requested, so every historical/backfill day was multiplied by today's supply. The rebuild reads the amount actually distributed on the day and never touches supply, so that goes away.

`allowNegativeValue` is dropped with it — the previous version needed it because a rate delta could come out negative; a distributed transfer cannot.

## CI

`ts-check` passes. The `test` check uses the default public `SOLANA_RPC` (`api.mainnet-beta.solana.com`), which rate-limits `getSignaturesForAddress`; the table above is a real local run against a Helius endpoint.

### DIFF
--- a/fees/solstice-usx/index.ts
+++ b/fees/solstice-usx/index.ts
@@ -9,10 +9,11 @@ import { METRIC } from "../../helpers/metrics";
 // rate multiplied by supply; the protocol distributes that yield on-chain instead, so the transfer
 // is read directly and no rate feed is needed.
 //
-// Solstice's yield account emits a TransferInYield event on each DistributeYield call. Its payload
+// Solstice's yield program emits a TransferInYield event on each DistributeYield call. Its payload
 // carries the USX amount handed to the eUSX backing account, which is exactly the yield accrued to
-// eUSX holders for the period.
-const YIELD_ACCOUNT = "HARVSXaBpt1TuD4PvqnscCQLzAWgDzGpkDadr76a489L";
+// eUSX holders for the period; the amount reconciles with the USX token balance delta the same
+// transaction records.
+const YIELD_PROGRAM = "HARVSXaBpt1TuD4PvqnscCQLzAWgDzGpkDadr76a489L";
 const USX = "6FrrzDk5mQARGc1TDYoyVnSyRdds1t4PbtohCD6p3tgG";
 const TRANSFER_IN_YIELD_DISCRIMINATOR = "3983162e2b54bd7a";
 // Anchor lays the event out as an 8 byte discriminator then its fields; the transferred USX amount
@@ -22,10 +23,22 @@ const AMOUNT_OFFSET = 72;
 const FEES_YIELD_LABEL = METRIC.ASSETS_YIELDS;
 const SUPPLY_SIDE_YIELD_LABEL = 'eUSX Yield To Holders';
 
+// Every signature in the window is fetched, so one transient failure would otherwise take the whole
+// day down with it. The error is still raised once the attempts are spent rather than turned into a
+// missing distribution.
 const rpc = async (method: string, params: any[]) => {
-    const { data } = await axios.post(getEnv('SOLANA_RPC'), { jsonrpc: "2.0", id: 1, method, params });
-    if (data.error) throw new Error(`solstice: solana rpc ${method} failed: ${JSON.stringify(data.error)}`);
-    return data.result;
+    let lastError: any;
+    for (let attempt = 0; attempt < 3; attempt++) {
+        try {
+            const { data } = await axios.post(getEnv('SOLANA_RPC'), { jsonrpc: "2.0", id: 1, method, params });
+            if (data.error) throw new Error(`solstice: solana rpc ${method} failed: ${JSON.stringify(data.error)}`);
+            return data.result;
+        } catch (e) {
+            lastError = e;
+            await new Promise(resolve => setTimeout(resolve, 500 * (attempt + 1)));
+        }
+    }
+    throw lastError;
 };
 
 // The yield account is only touched by these distributions, a couple of transactions a day, so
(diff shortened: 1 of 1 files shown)


tarun-khatri:
Both real, fixed in `2e5f68e` — thanks.

**`blockTime: null`.** The stop condition compared it numerically, and `null < fromTimestamp` is `true`, so a page whose last signature was not yet timestamped would end pagination before the window was covered. It now picks the oldest entry on the page that actually has a numeric `blockTime` and only stops on that, so an untimestamped tail keeps paging instead of truncating the day:

```ts
const oldest = [...page].reverse().find((entry: any) => typeof entry.blockTime === "number");
if (page.length < 1000 || (oldest && oldest.blockTime < fromTimestamp)) break;
```

**`getTransaction` returning null.** It fell through `tx?.meta?.logMessages ?? []` to an empty list and contributed 0, which is indistinguishable from a day with no distribution. It now throws, so a node that cannot serve the transaction surfaces instead of silently understating the total:

```ts
if (!tx) throw new Error(`solstice: solana rpc returned no transaction for ${signature}`);
```

Re-ran the two days that still have published values to confirm the change did not move the numbers — 2026-08-23 → $4.57k (published $4,566) and 2026-08-25 → $4.57k (published $4,559), same as before.

tarun-khatri:
@coderabbitai full review

tarun-khatri:
Both points were right and are fixed in c5c0c77.

**Events are now scoped to the yield program's own frame.** A distribution is a CPI chain, so the transaction really does carry `Program data:` from more than one program — the logs of a real distribution look like this:

```
Program HARVSXaBpt1TuD4PvqnscCQLzAWgDzGpkDadr76a489L invoke [1]
Program log: Instruction: DistributeYield
Program eUSXyKoZ6aGejYVbnp3wtWQ1E8zuokLAJPecPxxtgG3 invoke [2]
Program log: Instruction: TransferInYield
Program Tokenkeg... invoke [3]
Program Tokenkeg... success
Program data: HS9c2eu0768X...          <- eUSX frame, different discriminator
Program eUSXyKoZ... success
Program data: OYMWLitUvXqywjxDo14uR...  <- yield program frame, 3983162e2b54bd7a
Program HARVSX... success
```

Invocations nest, so the emitting program is the innermost frame still open; the adapter now tracks that as a stack and decodes only while the top of the stack is the yield program.

**Metadata is required before parsing.** `!tx?.meta?.logMessages` now raises rather than falling through to a zero-fee day.

**Bounded retry added to `rpc`**, and it rethrows once the attempts are spent, so a transient failure can't turn into a missing distribution.

Independent check that the decoded amount is the right one — the value at the event's byte 72 equals the USX token balance delta the same transaction records:

| signature | event amount @72 | USX balance delta |
|---|---|---|
| `3eD1bAgCXTThX2Mv` | 666330000 | `D2oMURfityVQ` -666330000 → `2aHdm37djj4c` +666330000 |
| `3Piu3y8TFTuPNpqD` | 666330000 | same pair, same size |
| `2CZASuzFfHd2XRQw` | 666330000 | same pair, same size |
| `3pt5FxV6bzc3RfCw` | no event | +13993000000 **into** the vault — a deposit, correctly not counted |

Backfill against the published series is unchanged by these fixes:

| day | adapter | published | delta |
|---|---|---|---|
| 2026-07-15 | 10,370 | 10,361 | +0.1% |
| 2026-08-10 | 8,440 | 8,493 | -0.6% |
| 2026-08-18 | 8,200 | 8,187 | +0.2% |
| 2026-08-19 | 8,190 | 8,179 | +0.1% |
| 2026-08-22 | 4,570 | 4,514 | +1.2% |

The two days at the yield step-down differ and are worth naming: published goes 8,179 → 7,133 → 3,779 → 4,514, while the adapter goes 8,190 → 8,200 → 6,990 → 4,570. The old method priced a redemption-rate delta between two snapshots, so a shift in when the rate was sampled shows up as a dip below the new level followed by a rebound; measuring the transfers themselves gives the monotonic step the rate actually took. Off the transition the two agree to within a percent.

@coderabbitai full review

bheluga:

PR #9225, opened 2026-09-04
feat: add Lodestar (Flare) fees adapter

lodestarprotocol:
Adds a fees/revenue adapter for **Lodestar**, fixed-term no-liquidation lending on Flare. Already listed for TVL as `lodestar-protocol` (id 8565).

**Why it is event-based rather than rate-based.** Lodestar charges no interest. A borrower pays one flat fee up front, deducted from principal at origination, and nothing accrues afterwards. So there is no utilisation curve to integrate — the fee is emitted directly:

- `LoanOpened(..., uint256 fee, ...)` — origination fee for the chosen term/LTV tier
- `LoanRolled(uint256 id, uint64 newDueAt, uint256 addFee)` — charged again on extension

Fees are always paid in the pool asset (USDT0, 6dp), so no price lookup ambiguity.

**Split.** `feeReserveBps` is read on-chain each run instead of being hardcoded. It is 30% reserve / 70% lenders today, but it is governance-settable, and hardcoding it would let a Safe transaction silently make this adapter wrong.

- `dailyFees` — all borrower fees
- `dailyRevenue` / `dailyProtocolRevenue` — the reserve share
- `dailySupplySideRevenue` — the lender share

**Testing.** `npm test fees/lodestar-protocol.ts` runs clean. It currently reports 0, which is accurate — the book is live but no loan has been opened yet. Because a broken adapter also reports 0, I verified the event ABIs hash to the same topic0 the deployed contract emits:

| event | topic0 |
|---|---|
| LoanOpened | `0x942164e824e051db9e0552b94635ac0c2927439d5e53f4d374e780620617c27f` |
| LoanRolled | `0xf7798f3fd60b8c458c8fa5a181e257efc1bbea3ffefe6548b2536e520b812250` |

Contract: [LodestarLoanBook](https://flarescan.com/address/0x9b479f47ef25E0Ed2134F38d3c4e1022A8695ed8) (verified)

### DIFF
--- a/fees/lodestar-protocol.ts
+++ b/fees/lodestar-protocol.ts
@@ -1,0 +1,71 @@
+import { FetchOptions, FetchResultV2, SimpleAdapter } from "../adapters/types";
+import { CHAIN } from "../helpers/chains";
+
+// Lodestar — no-liquidation fixed-term lending on Flare.
+//
+// Unlike a rate-based money market, a borrower pays ONE flat fee up front when the loan is opened,
+// deducted from the principal, and nothing accrues afterwards. So fees are event-driven, not a
+// function of utilisation over time:
+//   LoanOpened.fee   — the origination fee for the chosen term/LTV tier
+//   LoanRolled.addFee — charged again when a borrower extends into a new term
+//
+// The split between lenders and the protocol reserve is on-chain and governance-settable
+// (`feeReserveBps`), so it is read per-run rather than hardcoded — a Safe change would otherwise
+// silently make this adapter wrong. It is 30% to the reserve / 70% to lenders as of 2026-09-02.
+const BOOK = "0x9b479f47ef25E0Ed2134F38d3c4e1022A8695ed8";
+const USDT0 = "0xe7cd86e13AC4309349F30B3435a9d337750fC82D"; // fees are always paid in the pool asset
+
+const fetch = async (options: FetchOptions): Promise<FetchResultV2> => {
+  const dailyFees = options.createBalances();
+  const dailyProtocolRevenue = options.createBalances();
+  const dailySupplySideRevenue = options.createBalances();
+
+  const [opened, rolled] = await Promise.all([
+    options.getLogs({
+      target: BOOK,
+      eventAbi:
+        "event LoanOpened(uint256 indexed id, address indexed borrower, address indexed collateral, uint256 collAmount, uint256 principal, uint256 fee, uint64 dueAt)",
+    }),
+    options.getLogs({
+      target: BOOK,
+      eventAbi: "event LoanRolled(uint256 indexed id, uint64 newDueAt, uint256 addFee)",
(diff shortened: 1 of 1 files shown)


lodestarprotocol:
Thanks — all three are addressed in `c9bf82e`. I checked each against the deployed contracts before changing anything.

**Missing revenue streams.** Correct, and both were material. Verified in `LodestarLoanBook`:

- `DefaultPenaltyPaid` — `_bookPenalty` does `pool.payout(address(this), penalty); reserveBalance += penalty`, so it is borrower-paid and 100% protocol.
- `YieldSkimmed` — `IERC20(L.collateral).safeTransfer(reserve, skim)`, also 100% protocol, but denominated in the **collateral token** rather than the stable, so it is added per-token off the event's indexed `collateral` field.

Both now feed `dailyFees` and `dailyProtocolRevenue`, and neither touches `dailySupplySideRevenue` — lenders get none of them.

On `LoanSettled.proceeds`: leaving it alone is deliberate and I believe correct. `proceeds` is principal being recovered, and both `_bookPenalty` call sites (the repay path and the settlement path) emit `DefaultPenaltyPaid`, so every penalty already arrives through that event. Parsing `proceeds` as well would double-count the penalty and count principal as fees.

**Split drifting when the rate changes.** Good catch, and there is a way to make it structurally impossible rather than just more frequent. The book pays lenders with `pool.lockFee(fee - reserveCut)`, which emits `FeeLocked` carrying that exact amount. So the adapter now reads the lender share from what was actually paid, and derives the protocol share as the remainder. `feeReserveBps` is no longer read at all, and a mid-window governance change cannot desynchronise the two sides.

**Labels.** Added, with matching `breakdownMethodology` for all four sides. Three labels: `METRIC.BORROW_INTEREST` for the loan fee, plus `Default Penalties` and `Collateral Yield Skim`. I used custom strings for the latter two rather than `LIQUIDATION_FEES` because Lodestar has no liquidation mechanism at all — only a deadline — so that label would misdescribe the protocol.

Since the adapter still reports 0 (the book is live but no loan has been opened yet) and a broken adapter looks identical, I verified every event ABI hashes to the topic0 the deployed contracts emit, and that the indexed flags line up:

| event | contract | topic0 |
|---|---|---|
| LoanOpened | LoanBook | `0x942164e824e051db9e0552b94635ac0c2927439d5e53f4d374e780620617c27f` |
| LoanRolled | LoanBook | `0xf7798f3fd60b8c458c8fa5a181e257efc1bbea3ffefe6548b2536e520b812250` |
| DefaultPenaltyPaid | LoanBook | matches |
| YieldSkimmed | LoanBook | matches |
| FeeLocked | Pool | matches |

`npm test fees/lodestar-protocol.ts` and the repo `ts-check` both pass.

bheluga:
@lodestarprotocol thanks for the PR

Issue #9220, opened 2026-09-04
thundercore halted 2026-09-01, taking sushiswap-agg and ttswap down with it


tarun-khatri:
ThunderCore stopped producing blocks on **2026-09-01 06:18:02 UTC** at block **236120709**, and it is taking two adapters down with it.

## The chain is halted, not just a bad RPC

All four providers in `providers.json` return the identical head, and it does not move:

```
mainnet-rpc.thundercore.com    236120709   ts 2026-09-01 06:18:02 UTC
mainnet-rpc.thundertoken.net   236120709   ts 2026-09-01 06:18:02 UTC
mainnet-rpc.thundercore.io     236120709   ts 2026-09-01 06:18:02 UTC
thundercore.drpc.org           236120709   (429 on first read, same height after)
```

Re-read 20 seconds later: unchanged on every one of them. So this is not something a new endpoint fixes.

## What it breaks

The SDK's staleness guard throws on the block lookup, `fromBlock` never gets set, and the whole adapter dies with `Error: fromBlock or fromTimestamp is required` — every chain in it, not just ThunderCore. Reproduced locally:

```
$ npx ts-node --transpile-only cli/testAdapter.ts aggregators sushiswap-agg
error fetching block Error: Last block for thundercore is 4671 minutes behind
  (Tue Sep 01 2026 11:48:02 GMT+0530). Provider is "https://thundercore.drpc.org, …"
------ ERROR ------
Error: fromBlock or fromTimestamp is required
```

Two adapters stop on exactly the halt date:

| adapter | last published | typical |
|---|---|---|
| `aggregators/sushiswap-agg` | **2026-09-01** | ~$4.6M/day |
| `dexs/ttswap` | **2026-09-01** | ~$659/day |

The other files that reference `CHAIN.THUNDERCORE` are unaffected and still publishing today — `dexs/sushiswap-v3` ($9.47M), `fees/sushiswap-v3`, `fees/sushiswap-agg` (a separate file that does not list ThunderCore), `aggregators/nordstern-finance`, `bridge-aggregators/xy-finance` — so the blast radius really is just those two.

The same run also shows **moonbeam** 36,033 minutes behind (halted 2026-08-10), which is the wind-down already reported in #8776 / PR #8778.

## Filing rather than sending a patch

[...]

PR #9240, opened 2026-09-05
sushiswap-agg: stop querying thundercore, the halted chain takes all 40 other chains down with it

tarun-khatri:
`aggregators/sushiswap-agg` has published **nothing since 2026-09-01**, on any of its 40+ chains. It was doing $1.3M–$4.6M a day up to that point.

The cause is one chain. ThunderCore stopped producing blocks on **2026-09-01 06:18:02 UTC** at block **236120709** and has not moved since — every endpoint in the SDK's `providers.json`, plus drpc, returns the identical head, re-read twenty seconds apart and again four days later:

```
mainnet-rpc.thundercore.com    236120709   2026-09-01T06:18:02Z
mainnet-rpc.thundertoken.net   236120709   2026-09-01T06:18:02Z
mainnet-rpc.thundercore.io     236120709   2026-09-01T06:18:02Z
thundercore.drpc.org           236120709   2026-09-01T06:18:02Z
```

The SDK's staleness check throws on the block lookup, `fromBlock` is never set, and `getChainResult` rethrows inside the `Promise.all` in `runAdapter` — so the whole run is rejected and no chain gets a record:

```
$ npx ts-node --transpile-only cli/testAdapter.ts aggregators sushiswap-agg 2026-09-04
error fetching block Error: Last block for thundercore is 5968 minutes behind (2026-09-01 06:18:02).
------ ERROR ------
Error: fromBlock or fromTimestamp is required
```

With the leg commented out the same command completes and returns the day:

```
====== TOTAL DAILY AGGREGATED (sum of slots per chain) ======
ethereum   3.28 M     base      839.55 k    arbitrum  432.09 k
xdai     160.05 k     katana    103.12 k    polygon   100.45 k
...
Aggregate  5.76 M
```

$5.76M for 2026-09-03, a day that currently reports nothing at all.

ThunderCore's own contribution was immaterial before the halt, so nothing real is lost by dropping it — and while the chain is stopped there is nothing to read either. Commented rather than deleted, in the same style as the `SKALE_EUROPA` and `XLAYER` entries just above it, so it is one line to restore if the chain comes back.

`dexs/ttswap` is the other adapter configuring ThunderCore and it also stopped on 2026-09-01, but it is ThunderCore-only — there is no other chain to rescue there, and whether it should be marked dead is a separate call about the protocol rather than about this bug. Filed the chain-level side as #9220; this PR is just the leg that is costing an otherwise healthy 40-chain adapter its data.

### DIFF
--- a/aggregators/sushiswap-agg.ts
+++ b/aggregators/sushiswap-agg.ts
@@ -700,10 +700,12 @@ const adapters = {
   //   fetch,
   //   start: '2024-04-22'
   // },
-  [CHAIN.THUNDERCORE]: {
-    fetch,
-    start: '2024-02-25'
-  },
+  // thundercore halted on 2026-09-01 at block 236120709; the stale-head guard throws on the
+  // block lookup and takes every other chain's record down with it
+  // [CHAIN.THUNDERCORE]: {
+  //   fetch,
+  //   start: '2024-02-25'
+  // },
   [CHAIN.ZETA]: {
     fetch,
     start: '2024-02-25'


bheluga:

PR #9209, opened 2026-09-03
omni-exchange-flux: mark dead, all fourteen of its subgraphs return 404

tarun-khatri:
`dexs/omni-exchange-flux` reads two Goldsky subgraphs per chain — a CLAMM one and a bin one — across seven chains. **All fourteen return 404.**

```
project_cltceeuudv1ij01x7ekxhfl46/subgraphs/omni-clamm-{base,arbitrum,optimism,avalanche,sonic,bsc,plasma}/prod/gn   -> 404
project_cltceeuudv1ij01x7ekxhfl46/subgraphs/omni-bin-{base,arbitrum,optimism,avalanche,sonic,bsc,plasma}/prod/gn     -> 404
```

`api.goldsky.com` itself is healthy — 64 of the 101 Goldsky subgraphs referenced elsewhere in this repo answer `_meta` normally — so this is those specific deployments being gone, not an outage. Every chain in the config loses both of its sources, so there is no leg left that could still report.

The published series agrees the product wound down well before the subgraphs went: last point **2026-02-10 ($58)**, and each chain's last non-zero day is months earlier —

| chain | last non-zero |
|---|---|
| Plasma | 2025-10-10 |
| Sonic | 2025-10-12 |
| Avalanche | 2025-11-14 |
| BSC | 2025-11-22 |
| Arbitrum | 2025-12-14 |
| OP Mainnet | 2025-12-16 |
| Base | 2026-02-10 |

To be clear about scope: this is only Omni Exchange's **Flux** product. Omni Exchange V2 and V3 are separate adapters, are healthy, and both published today ($0 and $29 respectively) — they are untouched, and Omni Exchange itself is not being marked dead.

`deadFrom` is the day after this adapter's own last published point, so no stored history changes.

PR #9250, opened 2026-09-05
mynth: mark dead, the novaswap api and app hosts are both de-provisioned

tarun-khatri:
`bridge-aggregators/mynth` has thrown on every run since **2026-06-24**, and the product it tracks no longer exists.

Its only source is `api.novaswap.io/liquidity/transfers`. That hostname still resolves, but the CDN pull zone behind it has been de-provisioned, so the TLS handshake fails before any request is made:

```
$ curl -o /dev/null -w '%{http_code}' https://api.novaswap.io/liquidity/transfers?page=1&limit=10
000

$ node -e "require('https').get('https://api.novaswap.io/liquidity/transfers', ...)"
ERR_TLS_CERT_ALTNAME_INVALID  Hostname/IP does not match certificate's altnames:
  Host: api.novaswap.io. is not in the cert's altnames: DNS:*.b-cdn.net, DNS:b-cdn.net
```

The product is gone with it, not just the endpoint:

| host | Cloudflare DoH | Google DoH | HTTP |
|---|---|---|---|
| `app.novaswap.io` | **NXDOMAIN** (3) | **NXDOMAIN** (3) | 000 |
| `novaswap.io` | NOERROR, 2 A | NOERROR, 2 A | 000 |
| `api.novaswap.io` | NOERROR, 2 A | NOERROR, 2 A | 000 |
| `mynth.ai` | NOERROR, 1 A | NOERROR, 1 A | **200** |

The operator's own site is the last row, and it is up — but `mynth.ai` returns zero matches for "novaswap", "bridge", "cross-chain" or "crosschain" anywhere in the page. Mynth has moved on from this product; there is no replacement host to repoint at. There is also no TVL entry for it on DefiLlama.

Published `bridge-aggregators/novaswap-crosschain` ends at 2026-06-24 (73 days ago) after $8,017 that day, having run $8k–$11k a day in its final week. Nothing wrong is being published — the adapter simply errors every run — so this is housekeeping, in the same shape as #9207 / #9208 / #9209.

Dated 2026-06-25, the day after the last value, and verified to engage rather than silently no-op:

```
$ npx ts-node --transpile-only cli/testAdapter.ts bridge-aggregators mynth
🦙 MYNTH is dead (deadFrom 2026-06-25); skipping run.
```

### DIFF
--- a/bridge-aggregators/mynth/index.ts
+++ b/bridge-aggregators/mynth/index.ts
@@ -100,6 +100,9 @@ const adapter: SimpleAdapter = {
   },
   prefetch,
   start: "2025-06-20",
+  // api.novaswap.io serves a *.b-cdn.net certificate now, app.novaswap.io is NXDOMAIN, and the
+  // operator's site no longer mentions the product; last published day was 2026-06-24
+  deadFrom: "2026-06-25",
 };
 
 export default adapter;


bheluga:

PR #9255, opened 2026-09-05
Add GIGA DEX gauges

0xShad0w:
GIGA DEX (Robinhood Chain) launched gauges with GIGA emissions, which changed how swap fees are split. Every pool's protocol share is now set per pool on-chain and collected by the GIGA fee center, which distributes it to veGIGA stakers:

gauged pools (listed on the Classic / CL MasterChefs): 100% of swap fees to veGIGA stakers, LPs are paid in GIGA emissions instead
non-gauged pools: 3% to veGIGA stakers, 97% to LPs (freshly created pools sit at the factory default until the controller sets them, so the share is read per pool rather than hardcoded)

The treasury keeps no direct share of swap fees (it earns as a veGIGA staker), so the protocol share is reported as HoldersRevenue, with Revenue = HoldersRevenue and ProtocolRevenue = 0. This replaces the flat 20% treasury / 80% LP split both listings used before.

### DIFF
--- a/dexs/brownfi-v3/index.ts
+++ b/dexs/brownfi-v3/index.ts
@@ -86,10 +86,14 @@ export const getBrownFiV3Fetch = (chainConfig: BrownFiV3ChainConfig) => async (o
     const [token0, token1] = pairObject[pair]
     const feeRate = fee / (1 + fee)
     logs.forEach((log: any) => {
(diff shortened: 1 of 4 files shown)


bheluga:
@0xShad0w thanks for the PR

Issue #9159, opened 2026-09-02
lighterv2: protocolRevenue double-counts the LIT buyback already booked as holdersRevenue


gomesalexandre:
## What

`fees/lighterv2/index.ts` overstates `dailyProtocolRevenue` by counting a LIT token buyback twice: once as 100% of protocol revenue, and again as `dailyHoldersRevenue` — without subtracting the buyback from the protocol-revenue figure.

```ts
if (dailyBuybackUsd > 0) {
  dailyHoldersRevenue.addUSDValue(dailyBuybackUsd, METRIC.TOKEN_BUY_BACK)
}

return {
  dailyFees,
  dailyRevenue,
  dailyProtocolRevenue: dailyRevenue,   // <- not reduced by the buyback below
  dailyHoldersRevenue,
  dailySupplySideRevenue,
}
```

The adapter's own methodology confirms the buyback is sourced from the same revenue being counted as 100% protocol revenue, not an external allocation:

> **ProtocolRevenue**: "All trading and operational fees collected by the protocol treasury"
> **HoldersRevenue**: "LIT token buybacks from treasury. **The protocol uses fees to buy back LIT tokens from the market.**"

This is the "revenue funds the payout" pattern this repo's own review checklist calls out (`AGENTS.md:228`, "Missing `dailyProtocolRevenue`/`dailyHoldersRevenue` split when the protocol keeps some and distributes some") — distinct from an externally-funded buyback (e.g. Lido's LDO buyback, funded from accumulated treasury stock rather than the period's fee flow), which correctly does NOT need this split.

## Numbers (live, `api.llama.fi/summary/fees/lighter-perps`)

| dataType | 30d total |
|---|---|
| `dailyRevenue` | $2,725,716 |
| `dailyProtocolRevenue` | $2,725,716 (identical to revenue — confirms the alias) |
| `dailyHoldersRevenue` | $2,699,391 |
| `dailyFees` | $3,676,235 |

`holdersRevenue / revenue` = **99.03%** — i.e. almost the entire measured revenue is simultaneously being counted as both 100% protocol revenue and (separately) as a holders payout. `protocolRevenue + holdersRevenue` = $5,425,107, **199.1% of revenue**.

## Why this repo's convention says it's wrong

Three places in the codebase establish the pattern this adapter should follow (a holders payout sourced from the measured revenue reduces `dailyProtocolRevenue`, it doesn't add on top of it):

[...]

PR #9258, opened 2026-09-05
fees(lighterv2): say what the LIT buyback is measured from, not where it is funded

mkzung:
Answers #9159, which reads `dailyProtocolRevenue` as double-counting the LIT buyback already booked in `dailyHoldersRevenue`, on the premise that the buyback is "sourced from the same revenue". It proposes either `protocolRevenue = revenue - holdersRevenue` or recomputing the buyback from a documented 50% policy.

The series does not support the premise. Counted from the first buyback on 2026-01-05, which is the only window where the question means anything, `api.llama.fi/summary/fees/lighter-perps` gives $26,874,966 of buybacks against $25,684,722 of revenue, 104.6%, with cumulative buybacks ahead of cumulative revenue on 239 of 243 days. Summed from the adapter's start instead, revenue picks up $30.9M earned before any buyback existed and the ratio reads 47.5%, which is why the window has to be stated.

A payout cannot be drawn from revenue it exceeds. So the first fix goes negative on 106 of 234 paying days, and the second is not in the data: 1 of 234 days sits within 5% of a 0.50 share, and the daily ratio spans 0.035 to 7.385.

That leaves the arithmetic right and the methodology wrong, since it states the causal claim the issue starts from. Two strings, no behaviour change.

Website: https://app.lighter.xyz · Twitter: https://twitter.com/Lighter_xyz

### DIFF
--- a/fees/lighterv2/index.ts
+++ b/fees/lighterv2/index.ts
@@ -227,7 +227,7 @@ const methodology = {
   Fees: 'Maker and taker fees paid by traders on the Lighter DEX',
   Revenue: 'Protocol revenue from maker fees, taker fees, transfer fees, and withdraw fees. Liquidation fees are excluded as they go directly to LLP.',
   ProtocolRevenue: 'All trading and operational fees collected by the protocol treasury',
-  HoldersRevenue: 'LIT token buybacks from treasury. The protocol uses fees to buy back LIT tokens from the market.',
+  HoldersRevenue: 'LIT bought back by the treasury, measured from its trades on the LIT/USDC market. The size is not a share of the period fees: cumulative buybacks have run ahead of cumulative revenue since the first one in January 2026.',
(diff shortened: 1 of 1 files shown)


mkzung:
Taking the first half and not the second.

Saying the buyback may exceed the period's fees is right, and the wording here should carry it. Subtracting it from fees would not be, because the two are not drawn from one pot.

This PR's own CI run prints the case, for 2026-09-04:

```
Daily fees: 118.12 k
Daily revenue: 96.20 k
Daily holders revenue: 129.94 k
```

The buyback is above both. Over the window where the question means anything, from the first buyback on 2026-01-05, it is $26,874,966 against $25,684,722 of revenue, ahead of cumulative revenue on 239 of 243 days. A subtraction goes negative on 106 of the 234 paying days.

That is what the diff says in prose and why it changes no arithmetic.

PR #9260, opened 2026-09-05
fix: handle incomplete UniSubgraph fee snapshots

chetna-mittal:
## Summary

Fix error handling for cumulative fee queries in the shared UniSubgraph helper.

## Problem

The helper previously caught errors from the start and end cumulative fee queries independently and continued execution.

This could result in:

- `NaN` daily fees when only one cumulative snapshot was available
- silently missing `dailyFees` when both queries failed
- invalid fallback behaviour when a fee percentage existed but fees were not actually derivable from volume

Since daily fees are calculated as:

`end cumulative fees - start cumulative fees`

both snapshots are required for a valid result.


bheluga:

PR #9261, opened 2026-09-05
moonriver: use v1 for daily Etherscan fee data

chetna-mittal:
## Summary

Correct the adapter version used by the shared Etherscan CSV fee helper, which currently powers the Moonriver chain-fees adapter.

## Problem

The helper was declared as a version 2 adapter even though its external CSV source only provides fixed daily aggregates.

Version 2 adapters must support arbitrary start and end timestamps and explicitly configure `pullHourly`. Running a daily aggregate source as v2 cannot provide genuine hourly data and does not match the adapter version requirements.

The helper also contained a Fantom-specific revenue branch, although Moonriver is its only current consumer.

## Data impact

No historical fee calculation changes are expected.

The adapter continues selecting the Moonriver CSV row for `options.startOfDay` and pricing the resulting MOVR amount. This change only aligns its execution model with the daily granularity of its data source.

### DIFF
--- a/helpers/etherscanFees.ts
+++ b/helpers/etherscanFees.ts
@@ -1,5 +1,4 @@
 import { Adapter, FetchOptions, ProtocolType } from "../adapters/types";
-import { CHAIN } from "./chains";
 import { httpGet } from '../utils/fetchURL';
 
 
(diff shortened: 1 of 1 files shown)


helpers/etherscanFees.ts
Bonvi28: - [ ] - [ ] text

bheluga:

PR #9262, opened 2026-09-05
fix: align Saddle factory exports with v2 conventions

chetna-mittal:
## Summary

Align the shared Saddle-style adapter factory with the established export pattern used by other protocol helpers.

## Problem

The Saddle helper had several inconsistencies:

- Version 2 exports discarded per-chain `start` dates
- Version 2 did not explicitly enable `pullHourly`, despite reading on-chain event logs
- Returned fee balances used breakdown labels without corresponding methodology
- Version 1 and version 2 used duplicate export objects
- The helper did not support the root adapter overrides used by similar factories

The helper currently powers:

- minmax
- mm-stableswap-polygon
- nerve
- zyberswap-stable

## Data impact

The underlying volume and fee calculations are unchanged.

Enabling hourly pulls gives the on-chain log adapter more granular execution and avoids repeatedly calculating full-day ranges. Existing factory configurations do not currently specify start dates, so preserving `start` does not alter their present historical coverage.

### DIFF
--- a/helpers/saddle.ts
+++ b/helpers/saddle.ts
@@ -42,26 +42,46 @@ type SaddleConfig = {
   start?: string,
 }
 
+const methodology = {
+  Volume: 'Total token swap volume across the configured pools.',
+  Fees: 'Swap fees paid by users based on each pool\'s configured swap fee.',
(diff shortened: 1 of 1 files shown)


bheluga:

PR #9263, opened 2026-09-05
valorem: remove dead fees and options adapters

chetna-mittal:
## Summary

Remove the inactive Valorem fees and options implementations and preserve their metadata in the dead-adapter registry.

## Background

Both Valorem adapters were already marked with:

`deadFrom: 2023-10-31`

The runner therefore no longer executes them for current data. Keeping the full implementations in the active adapter directories adds unused code and dependencies without affecting published metrics.

The options implementation also never supplied `dailyPremiumVolume`, which is a required options dimension.

## Data impact

Both adapters were already skipped after October 31, 2023. This PR only moves their metadata into the repository's dead-adapter registry and removes code that is no longer executed.

Historical adapter metadata remains available through the dead-adapter factory.


bheluga:
Its already marked dead, we have stopped moving to deadAdapters.json

PR #9280, opened 2026-09-06
[fix] hipo - read the widened treasury state, and fix the reward it reports

behrang:
Hipo upgraded its treasury contract on 2026-09-06. `get_treasury_state` mirrors the contract's storage order, and the upgrade inserted `deficit` at index 5 and `round_duration` + `last_settled_round` after the exchange-rate pair, taking the tuple from 21 values to 24. Every index from 5 on moved, so this adapter was reading `loan_codes` where it expected the rate pair, and `halter` where it expected `governance_fee`.

While fixing the indices I found a second, older problem in the same lines.

## The reported reward was ~6 millionths of the real figure

Indices 11 and 12 once held `last_staked` and `last_recovered`, and this adapter still subtracted them as if they were coin amounts. The treasury replaced those fields with the 1e9-scaled exchange rate pair some time ago, so `newCoins` has been `current_rate - previous_rate` — a **rate difference reported as nanoGRAM**.

On live state that is 365,415 nano, or 0.000365 GRAM per round, which is why Hipo's fees have been reading as approximately zero.

The reward is that rate move applied to the whole hGRAM supply, since `total_coins = total_tokens * rate / 1e9`:

```
newCoins = total_tokens * (current_rate - previous_rate) / 1e9
```

Against live state that gives **dailyFees ≈ 2,901 GRAM**, or 15.1% of TVL annualised simple — which lines up with the 16.3% compounded APY the protocol publishes through its own gauge.

## The `get_times` call is gone

`round_duration` is the number of seconds `current_rate` took to grow out of `previous_rate`, measured on chain between the last two settled rounds, so the period now comes from the same read as the rates it belongs to.

It is deliberately not a round length. The treasury only moves the rates when a round it lent into settles, so a round in which nothing was lent widens the interval rather than passing unnoticed — dividing by a round length would report an unchanged figure for a pool whose real rate of growth had halved.

## One smaller fix

`treasuryReward` is no longer floored. `newCoins` used to be a difference of two integers, so flooring was a no-op; it is a scaled product now, and flooring left `protocolFee` holding the negative fractional remainder — a reported revenue of `-1e-9` whenever the governance fee is zero, which it currently is.

## Verification

Ran the adapter's arithmetic against live contract state:

```
total_tokens=6022617962641058  previous_rate=1163036853  current_rate=1163402268
round_duration=65536  governance_fee=0

dailySupplySideRevenue = 2901.385910942 GRAM
dailyProtocolRevenue   = 0.000000000 GRAM
dailyFees              = 2901.385910942 GRAM
```

I am from the Hipo team.

### DIFF
--- a/fees/hipo/index.ts
+++ b/fees/hipo/index.ts
@@ -42,6 +42,15 @@ export default {
                 // the round length that used to come from a second get_times call: the treasury only
                 // moves the rates when a round it lent into settles, so a round in which nothing was
                 // lent widens this interval rather than passing unnoticed.
+                //
+                // Zero on a treasury that has never settled a round, which would make normalize()
+                // return Infinity and carry it into the USD conversion. Mainnet cannot be in that
+                // state -- the migration seeded round_duration from the network's round length --
+                // but the sibling yield adaptor guards the same value, so this one does too.
+                if (!Number.isFinite(roundDuration) || roundDuration <= 0) {
+                    throw new Error('Expected a positive round duration, but got ' + roundDuration)
+                }
+
                 const normalize = normalizer(roundDuration)
 
                 // The reward that accrued to stakers, in nanoGRAM. It is the rate move applied to the
--- a/fees/hipo/index.ts
+++ b/fees/hipo/index.ts
@@ -28,27 +28,34 @@ export default {
                     throw new Error('Expected a zero exit code, but got ' + getTreasuryState.exit_code)
                 }
 
-                const getTimes = await postURL('https://toncenter.com/api/v3/runGetMethod', {
-                    address,
-                    method: 'get_times',
-                    stack: [],
-                })
-                if (getTimes.exit_code !== 0) {
-                    throw new Error('Expected a zero exit code, but got ' + getTimes.exit_code)
-                }
+                // get_treasury_state mirrors the treasury's storage order. The upgrade of 2026-09-06
+                // inserted deficit at index 5 and round_duration + last_settled_round after the rate
+                // pair, taking the tuple from 21 values to 24, so every index from 5 on moved.
+                const totalTokens = Number(getTreasuryState.stack[1].value)
+                const previousRate = Number(getTreasuryState.stack[12].value)
+                const currentRate = Number(getTreasuryState.stack[13].value)
+                const roundDuration = Number(getTreasuryState.stack[14].value)
+                const governanceFee = Number(getTreasuryState.stack[19].value)
 
-                const lastStaked = Number(getTreasuryState.stack[11].value)
-                const lastRecovered = Number(getTreasuryState.stack[12].value)
-                const governanceFee = Number(getTreasuryState.stack[16].value)
+                // round_duration is the number of seconds current_rate took to grow out of
+                // previous_rate, measured on chain between the last two settled rounds. It replaces
+                // the round length that used to come from a second get_times call: the treasury only
+                // moves the rates when a round it lent into settles, so a round in which nothing was
+                // lent widens this interval rather than passing unnoticed.
+                const normalize = normalizer(roundDuration)
 
-                const currentRoundSince = Number(getTimes.stack[0].value)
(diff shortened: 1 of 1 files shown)


behrang:
Both points addressed.

**Governance fee guard.** The concern is valid and I checked it against the contract rather than assuming: `set_governance_fee` in the treasury does `governance_fee = s~load_uint(16)` with only an access check, so there is no upper bound on chain and 65535 is genuinely reachable. It would divide by zero in `treasuryReward` and publish `Infinity`. It now throws with the offending value, matching the shape of the exit-code checks already in the adapter.

To be clear about scope: that hazard predates this PR, since the division is unchanged. It was worth closing while these lines were open.

**Title** renamed to the `[type] protocol-name - description` format.

The CI run above reflects the fix this PR is really about: daily fees of ~4.11k, where the adapter previously reported approximately zero because it was subtracting two exchange rates and reporting the difference as nanoGRAM. Current governance fee on chain is 0, which is why daily revenue is 0.00 and all of it lands in supply-side revenue.

behrang:
Guard added; regression test skipped, with a reason.

**`roundDuration` validation** — added. `round_duration` is zero on a treasury that has never settled a round, and `normalize()` would turn that into `Infinity` and carry it into the USD conversion.

Mainnet cannot actually be in that state: the migration that introduced the field seeded it from the network's round length, and it only ever moves forward from there. The check is still worth having, because the sibling [yield-server adaptor](https://github.com/DefiLlama/yield-server/pull/2980) validates the same value the same way — two adapters reading one contract should not disagree about what counts as a usable reading.

**Regression test for zero duration** — skipped. There is no adapter unit-test harness in this repo to put one in: the only test file is `helpers/dune.test.ts`, and adapters are exercised live through `cli/testAdapter.ts`, which runs against the real contract. A zero `round_duration` is not reachable there. Happy to add one if you have a pattern in mind that I have missed.

behrang:
This one is real, and it is a property of our contract rather than a misreading — thanks. I have documented it in the adapter; I do not think reading settlement history is the right fix here, and here is the reasoning.

**The mechanism, confirmed.** In `treasury.fc` the settlement branch is:

```func
if round_since > last_settled_round {
    round_duration = round_since - last_settled_round;
    last_settled_round = round_since;
}
previous_rate = current_rate;
current_rate = muldiv(total_coins, 1000000000, total_tokens);
```

The rates move unconditionally; `round_duration` only on a forward move. So they are the same event in steady state and come apart exactly as described when the elector rejects a newer round's stake and it settles ahead of an older round still validating.

**The magnitude.** With settlement order R+1, R, R+2:

| settlement | guard | round_duration | rate delta |
|---|---|---|---|
| R+1 | passes | 2 rounds | R+1's reward (1 round) |
| R | blocked | 2 rounds | R's reward (1 round) |
| R+2 | passes | 1 round | R+2's reward (1 round) |

Two readings normalise a one-round reward over a two-round window, so the daily figure is understated by about 2x, never overstated, and the next in-order settlement restores the pairing.

Worth noting the two deltas together are R's and R+1's rewards over 2 rounds — which is the correct time-average for that window. This adapter is `runAtCurrTime` and sampled daily, so a sample landing inside the anomaly reads low, and the long-run average is unaffected.

**Why not read settlement history.** Matching each delta to its own gap means indexing repayment events on TON across rounds, which is a different class of adapter from a single `runGetMethod` snapshot — and it would not be reliable either, since the treasury exposes one scalar for an interval that is only well-defined per settlement. The contract would have to expose the pairing for an adapter to read it. I would rather record the limitation honestly than build history-reading that still cannot reconstruct it.

I have added a comment at the read site describing the caveat, its direction, and that it self-corrects. If a maintainer would prefer this stated in the `methodology` block instead, or would rather I drop the comment entirely, happy either way.

On the regression case: as in my previous reply, there is no adapter unit-test harness in this repo to put one in.

bheluga:
@behrang thanks for the PR

behrang:
Thank you for the merge.

Issue #9100, opened 2026-09-01
uniswap-v2: the ClickHouse chains have published ~4x short volume since pullHourly landed ($201.2M missing over the last 30d)


gomesalexandre:
`dexs/uniswap-v2` derives fees from volume by a hardcoded constant, so the two dimensions are the same number scaled and `fees / volume` must be exactly 30.00 bps. On every chain configured `source: 'LOGS'` it is. On every chain configured `source: 'CLICKHOUSE'` it has not been since July, and the volume side is the one that is wrong.

Ethereum published **$1,468,397** of volume for 2026-08-31. I reconstructed the day from chain logs and measured **$18,424,805**. The published fee number for the same day prices it at **$13,719,000**.

## The invariant

```ts
const dailyFees = dailyVolume.clone(0.003);
```
`dexs/uniswap-v2.ts`, in both `fetchClickhouse`'s return and its `emptyResult()`. That is the last statement before the return, so this is not an assumption about Uniswap's fee schedule, it is an identity in the code.

## The split falls exactly on the source config

Last 30 days, from `api.llama.fi/summary/dexs/uniswap-v2` and `api.llama.fi/summary/fees/uniswap-v2`:

| chain | source | published vol | fees | implied vol | ratio | |
|---|---|---|---|---|---|---|
| Robinhood Chain | LOGS | $933.03M | $2,803.1k | $934.38M | **30.0 bps** | ok |
| Monad | LOGS | $0.24M | $0.7k | $0.23M | **29.7 bps** | ok |
| OP Mainnet | LOGS | $0.03M | $0.1k | $0.03M | **29.1 bps** | ok |
| Ethereum | CLICKHOUSE | $56.96M | $746.3k | $248.78M | **131.0 bps** | **x4.37** |
| Polygon | CLICKHOUSE | $3.69M | $26.1k | $8.71M | **70.8 bps** | **x2.36** |
| BSC | CLICKHOUSE | $0.87M | $5.8k | $1.95M | **67.4 bps** | **x2.25** |
| Base | CLICKHOUSE | $2.99M | $18.7k | $6.22M | **62.5 bps** | **x2.08** |
| Arbitrum | CLICKHOUSE | $0.05M | $0.3k | $0.09M | **55.3 bps** | **x1.84** |

ClickHouse subtotal: **$64.56M published, $265.75M implied, $201.20M missing.** LOGS subtotal: $934.27M published against $935.97M implied, 30.05 bps.

[...]

PR #9285, opened 2026-09-06
uniswap v2: pull daily while the clickhouse chains have no indexer sync guard

0xshubhs:
for #9100.

`fetchClickhouse` sums `evm_indexer.logs` over the run window with no check that the indexer has reached `toTimestamp`. the sdk's indexer `getLogs` does have that check (throws past 50% missing, fills the rest from rpc), which is why the `LOGS` chains in this same file publish `fees / volume` at exactly 30 bps while the five clickhouse chains sit at 55 to 131 bps: each job gets whatever the indexer had ingested when its hourly run fired, and the volume job and the fees job fire at different times.

the day it started is #8027 (2026-07-09), a three-line `version: 1` -> `version: 2, pullHourly: true` change. ethereum, daily averages from `api.llama.fi/summary/dexs/uniswap-v2` and the fee series divided by 0.003:

```
2026-06-01 .. 07-08   volume $12.8m/d   fee-implied $13.7m/d   ratio 1.07
2026-07-09 .. 09-05   volume  $1.9m/d   fee-implied  $7.2m/d   ratio 3.71
```

visible to the day: 07-08 $11.3m, 07-09 $0.84m, with the fee-implied figure still $4.9m.

this puts the adapter back on daily windows. it is the smaller of the two fixes. the right one is a sync guard in `fetchClickhouse` (the chain's max indexed block timestamp against `toTimestamp`, throw when short) so hourly can come back, but the clickhouse path cannot be run from outside and i did not want to send a guard i have not executed. `pullHourly: false` is the pre-07-09 state, which held ethereum within 7% of the fee series for the nine weeks before the change.

`dexs/flap.ts` has the same shape (clickhouse source, `pullHourly: true`) and deserves the same look; not touched here.

ts-check clean. tempo, on the `LOGS` path the flag does not touch, runs unchanged in the harness. the clickhouse chains cannot run locally (`Missing env CLICKHOUSE_CONFIG`), so the `test` check will be red the way it is for every pr touching this file. the days from 07-10 on need a refill once this lands; the issue stays open for that.

### DIFF
--- a/dexs/uniswap-v2.ts
+++ b/dexs/uniswap-v2.ts
@@ -287,7 +287,8 @@ const methodology = {
 
 const adapter: Adapter = {
   version: 2,
-  pullHourly: true,
+  // clickhouse chains: evm_indexer has no sync guard, an hourly window reads a partial hour (#9100)
+  pullHourly: false,
   fetch,
   adapter: chainConfig,
   methodology,


0xshubhs:
the red `test` is `Missing env CLICKHOUSE_CONFIG`: the workflow runs `dexs uniswap-v2` on every chain and the runner has no clickhouse credentials, so the ethereum/polygon/base/bsc/arbitrum legs throw before the diff is reached. master fails the same way on this file. the change is a flag flip; the `LOGS` chains it does not touch run unchanged in the harness (tempo 2026-09-04: $6.74k volume, $20 fees).

bheluga:
@0xshubhs thanks for the PR
Disabling pull hourly isn't the best solution here.
I have moved this issue internally to take a look

PR #9300, opened 2026-09-07
uniswap: filterPools should not call balanceOf on the zero address, and should retry before it counts a failure

### LINKED ISSUE #9299: filterPools: the 10% threshold from #9140 has klayswap-v1, mimo-v2 and dswap dark since the day it merged — on klaytn all 260 'failures' are balanceOf(address(0))
Three dexs adapters published their last point on 2026-09-03 and nothing since: **klayswap-v1**
(Kaia), **mimo-v2** (IoTeX) and **dswap** (ENI). Their chains are fine — dragonswap-v3, iziswap and
egas-swap all published in the last 24h — and the three have nothing in common except
`helpers/uniswap.ts` `filterPools`.

#9140 merged 2026-09-03 14:04 UTC. dswap is the clearest clock: it had run $95k–$111k every day for
weeks, printed $88,648 on 09-03, and stopped.

All three throw the new guard locally:

```
klayswap     filterPools: 260/1246 pooled balance calls failed on klaytn  (20.9%)
mimo         filterPools: 439/1338 pooled balance calls failed on iotex   (32.8%)
daoaas-swap  filterPools: 619/1032 pooled balance calls failed on eni     (60.0%)
```

## klaytn is not an rpc failure at all

Replaying the exact calls the helper makes and grouping the failures by target:

```
klaytn  calls 1246  failed 260 (20.9%)  distinct failing targets: 1
        [["0x0000000000000000000000000000000000000000", 260]]
        retry of the failed set: 260 still null of 260
```

Every one of them is `balanceOf` on the **zero address**. 130 of klayswap's cached pairs carry a
native-KLAY leg stored as `address(0)`, and that call can never succeed — not on a better RPC, not
on a retry. 20.9% is permanently over `MAX_FAILED_BALANCE_CALL_RATIO`, so klayswap-v1 throws on
every run and will never publish again.

Before #9140 those nulls landed in `pairBalances[...].add(target, bal ?? 0)` and the pair was
weighed on its other leg. The guard is right that a mostly-failed batch must not be read as empty
pools; it just cannot tell "the rpc dropped this call" from "nobody can ever answer this call".

## the other two are genuinely mixed

```
iotex   calls 1338  failed    1 (0.1%)  distinct failing targets: 1
eni     calls 1032  failed  332 (32.2%) distinct failing targets: 147
```

IoTeX re-probed a minute later fails one call out of 1,338 — mimo's 32.8% was a transient burst, and
retrying only the failed calls brings it under the threshold and produces a normal day
(`Daily volume: 3.85 k`). ENI spreads its failures over 147 distinct token contracts and stays at
~20% even after two retries, so there the guard may well be doing its job.

## suggested fix

Two independent changes, both in `filterPools`:

1. **Do not issue `balanceOf` against `address(0)`.** It cannot answer, and counting it as a failed
   rpc call is what makes klayswap unrecoverable. Dropping those calls restores exactly the
   pre-#9140 outcome for a native-leg pair (weighed on its other token) without weakening the guard.
2. **Retry the failed calls once or twice before applying the ratio,** so the threshold is measured
   against failures that persist. That is what separates a transient burst (IoTeX) from a structural
   one (ENI), and it is the difference between mimo publishing and mimo being dark.

Happy to send both — I have them working against the three adapters above. Flagging rather than
just opening the PR because #9140 is recent and deliberate, and the threshold itself is not the
thing I think is wrong.

0xshubhs:
Fixes #9299.

klayswap-v1, mimo-v2 and dswap all published their last point on 09-03, the day #9140 merged, and
all three throw its guard. Two changes, independent of each other, both in `filterPools`:

**1. Do not call `balanceOf` on the zero address.** On klaytn every single one of the 260 failures
is that call:

```
klaytn  calls 1246  failed 260 (20.9%)  distinct failing targets: 1
        [["0x0000000000000000000000000000000000000000", 260]]
        retry of the failed set: 260 still null of 260
```

130 of klayswap's cached pairs carry a native-KLAY leg stored as `address(0)`. That call cannot
succeed on any rpc, so the ratio sits permanently over the threshold and the adapter can never
publish again. Before #9140 those nulls fell through `add(target, bal ?? 0)` and the pair was
weighed on its other leg, which is exactly what skipping the call restores.

**2. Retry the failed calls before measuring the ratio.** mimo failed 439/1338 (32.8%) on one run
and 1/1338 on a re-probe a minute later, a transient burst, not a degraded chain. Retrying only the
failed calls is what separates that from a chain that is actually down.

## verification

| adapter | chain | master | this branch |
|---|---|---|---|
| klayswap | klaytn | `260/1246 pooled balance calls failed` | Daily volume: 12.08 k |
| mimo | iotex | `439/1338 pooled balance calls failed` | Daily volume: 3.85 k |
| daoaas-swap | eni | `619/1032 pooled balance calls failed` | `214/1032`, still throws |

ENI is left throwing on purpose. Its failures spread across 147 distinct token contracts and stay
at ~20% after two retries, so that is the guard doing its job rather than a false positive.

Control, so this is not just "the guard stopped firing": klayswap for **09-02**, a day DefiLlama
published under the pre-#9140 code, reads **12.57 k** on this branch against **12,448** published, 
1.0% apart. 09-04 reads 12.08 k on two separate runs. `npx tsc --noEmit` is clean.

### DIFF
--- a/helpers/uniswap.ts
+++ b/helpers/uniswap.ts
@@ -102,11 +102,18 @@ export async function getEstablishedTokens(chain: string, tokens: string[]): Pro
 // reporting a mostly-failed batch of pools as legitimately empty. See PR #8565 for the
 // original all-or-nothing precedent this threshold replaces.
 const MAX_FAILED_BALANCE_CALL_RATIO = 0.1
+const MAX_BALANCE_CALL_RETRIES = 2
 
 export async function filterPools({ api, pairs, createBalances, maxPairSize = 42, minUSDValue = 200 }: { api: ChainApi, pairs: IJSON<string[]>, createBalances: any, maxPairSize?: number, minUSDValue?: number }): Promise<IJSON<number>> {
-  const balanceCalls = Object.entries(pairs).map(([pair, tokens]) => tokens.map(i => ({ target: i, params: pair }))).flat()
+  const balanceCalls = Object.entries(pairs).map(([pair, tokens]) => tokens.filter((i) => i !== ZERO_ADDRESS).map(i => ({ target: i, params: pair }))).flat()
   const res = await api.multiCall({ abi: 'erc20:balanceOf', calls: balanceCalls, permitFailure: true, })
-  const failedCalls = res.filter((bal) => bal == null).length
+  let failedIndices = res.map((bal, i) => bal == null ? i : -1).filter((i) => i > -1)
+  for (let attempt = 0; attempt < MAX_BALANCE_CALL_RETRIES && failedIndices.length; attempt++) {
+    const retried = await api.multiCall({ abi: 'erc20:balanceOf', calls: failedIndices.map((i) => balanceCalls[i]), permitFailure: true, })
+    retried.forEach((bal, j) => { if (bal != null) res[failedIndices[j]] = bal })
+    failedIndices = failedIndices.filter((_, j) => retried[j] == null)
+  }
+  const failedCalls = failedIndices.length
   if (balanceCalls.length && failedCalls / balanceCalls.length > MAX_FAILED_BALANCE_CALL_RATIO)
     throw new Error(`filterPools: ${failedCalls}/${balanceCalls.length} pooled balance calls failed on ${api.chain}, refusing to report ${Object.keys(pairs).length} pools as (partly) empty`)
   const balances: Balances = createBalances()


bheluga:
@0xshubhs thanks for the PR

PR #8229, opened 2026-07-17
americafun: floor dailyFees at the sum of its components

0xshubhs:
Addresses #8228; America.Fun was showing dailyRevenue 5-7x dailyFees, which shouldn't be possible.

Dug into their endpoint's history to figure out which side to trust:

| window | fees | revenue (pr+hr) | ssr | meteoraFee |
|---|---|---|---|---|
| 1d ago | 335.13 | 175.70 | 159.43 | 67.03 |
| 3d ago | 890.05 | 498.68 | 391.37 | 178.01 |
| 14d ago | 10110.19 | 9453.54 | 656.65 | 2022.04 |
| 60d ago | 306.45 | 24222.29 | 0.00 | 61.29 |
| today | 525.37 | 2812.07 | 0.00 | 105.07 |

On healthy days the endpoint maintains fees = protocolRevenue + holdersRevenue + supplySideRevenue exactly (to the cent). Two failure modes break it:

1. some days (today, ~60d ago) it returns a total far below its own components, with ssr zeroed, that's where the revenue > fees on the dashboard comes from
2. some days (14d ago) meteoraFee exceeds the reported supplySideRevenue, and since the adapter books the meteora cut into supply side regardless, the published ssr breaks the identity even when the api total was fine

Fix: floor dailyFees at the sum of what the adapter actually books (treasury + stakers + meteora + lp). Healthy days are untouched (the floor equals the reported total), broken days become internally consistent, and if the america.fun team fixes their endpoint the floor becomes a no-op.

Test run (today, previously a broken day):
```
Daily fees: 2.80 k
Daily revenue: 2.73 k
Daily protocol revenue: 2.68 k
Daily holders revenue: 44.00
Daily supply side revenue: 72.00
```
fees = 2681 + 44 + 72 exactly, revenue back under fees. Ran the suite twice, stable.

The endpoint inconsistency itself still needs someone from the america.fun side, left the details in #8228.

PR #9310, opened 2026-09-07
[fix] americafun - refuse days the upstream fees API contradicts itself on

### LINKED ISSUE #9277: america.fun: the upstream endpoint contradicts itself on 13 of 139 days
`fees/americafun.ts` relays `https://defillama.america.fun/api/v1/fees`, which returns `dailyFees`, `dailyRevenue`, `dailyProtocolRevenue`, `dailyHoldersRevenue`, `dailySupplySideRevenue` and `meteoraFee`. The adapter does no arithmetic beyond splitting supply side into the Meteora cut and the LP remainder, so what DefiLlama publishes is what the endpoint says.

On 13 of the 139 days since the adapter's `start`, the endpoint's own figures break `revenue + supplySide <= fees`:

```
2026-05-11  fees      609.33  revenue+supply     2786.18    4.57x   supply 0.00  meteora   121.87
2026-05-14  fees    28908.61  revenue+supply    71056.12    2.46x   supply 0.00  meteora  5781.72
2026-05-17  fees      306.45  revenue+supply    24222.29   79.04x   supply 0.00  meteora    61.29
2026-05-21  fees       13.99  revenue+supply       60.74    4.34x   supply 0.00  meteora     2.80
2026-05-25  fees      283.68  revenue+supply      530.92    1.87x   supply 0.00  meteora    56.74
2026-06-13  fees      938.36  revenue+supply    15906.64   16.95x   supply 0.00  meteora   187.67
2026-06-22  fees       39.06  revenue+supply      108.10    2.77x   supply 0.00  meteora     7.81
2026-06-29  fees    25274.01  revenue+supply    46983.34    1.86x   supply 0.00  meteora  5054.80
2026-07-16  fees      358.73  revenue+supply     2725.28    7.60x   supply 0.00  meteora    71.75
2026-07-20  fees     2653.89  revenue+supply     2989.03    1.13x   supply 0.00  meteora   530.78
2026-07-28  fees      532.96  revenue+supply      824.67    1.55x   supply 0.00  meteora   106.59
2026-08-07  fees       13.69  revenue+supply      106.26    7.76x   supply 0.00  meteora     2.74
2026-08-30  fees        0.14  revenue+supply        2.65   19.34x   supply 0.00  meteora     0.03
```

Two things travel together on every one of them. `dailySupplySideRevenue` is exactly 0 while `meteoraFee` is exactly 0.2 of the fees, which the adapter's own methodology calls Meteora's protocol cut on both DBC and DAMM v2 pools and therefore part of what the supply side receives, so one of those two fields is wrong. `dailyRevenue` runs above `dailyFees` too, as far as 79x.

Neighbours are consistent, so the window is not the cause:

```
2026-06-12 fees  1641.62  rev   749.82  sup  891.79  meteora  328.32
2026-06-13 fees   938.36  rev 15906.64  sup    0.00  meteora  187.67
2026-06-14 fees    63.52  rev    47.81  sup   15.71  meteora   12.70
```

Widening the window scales the fees but not the revenue, so a 7-day request from 2026-06-13 still returns 16450.42 against 1936.11 of fees.

Reproducer, no key needed:

```
curl "https://defillama.america.fun/api/v1/fees?from=1781308800&to=1781395200"
```

The published record carries it: 356,064 of fees against 262,720 of revenue and 201,067 of supply side, 1.30x over all 135 days with all three series.

The adapter cannot fix a source that disagrees with itself, so this is for whoever can reach america.fun. If you would rather it fail loudly than publish the day, a guard that throws when `protocolRevenue + holdersRevenue + supplySide > fees` fires on exactly these 13 days and nowhere else — I can send it.

gomesalexandre:
closes #9277

`fees/americafun.ts` relays `defillama.america.fun`'s own `/api/v1/fees` breakdown as-is. On 13 of ~139 days since the adapter's start, that endpoint's own numbers violate `Fees = Revenue + SupplySideRevenue` by 1.33x-79x: `dailySupplySideRevenue` comes back exactly `0` while `meteoraFee` (part of supply side per this adapter's own methodology - "Meteora's protocol cut") is nonzero, and `dailyRevenue` runs far above `dailyFees`.

The adapter cannot fix a source that disagrees with itself, so this doesn't try to reconstruct a "correct" number - it throws instead of publishing the day, same shape as the guards already shipped for `invariant`, `aera-v2`, `maverick-v2` etc. in this repo.

## Why not `#8229`'s approach

A prior attempt (#8229, closed) tried flooring `dailyFees` at the sum of the booked components to force the identity to hold. A maintainer declined it: *"as long as cumulative fees > cumulative revenue, its acceptable"* - reading the daily mismatch as timing lag between fee collection and revenue distribution, which self-corrects over time.

That reasoning doesn't cover this case. Over the current 135-day published record, cumulative **still doesn't reconcile**: fees $356,064 vs. revenue+supplySide $463,787 - a persistent ~30% gap, not one that's closing. And unlike #8229, this PR doesn't touch the math for any day - it only refuses to publish the ones that are internally inconsistent.

## Threshold

20% tolerance, chosen from the data rather than picked arbitrarily:

| set | ratio range |
|---|---|
| the 13 known-bad days from #9277 | 1.33x - 79.2x |
| ordinary day-to-day noise (sampled ~25 other days across the full history) | up to 1.15x |

20% sits cleanly in the gap between those two clusters - it rejects every one of the 13 known-bad days and none of the routine-noise days, so it doesn't re-litigate the timing-lag leniency the maintainer already established on this adapter.

## Verification

Live, today, against `defillama.america.fun`:

```
$ npx ts-node --transpile-only cli/testAdapter.ts fees americafun 2026-05-18   # covers calendar day 05-17
------ ERROR ------
Error: Upstream defillama.america.fun API returned internally inconsistent fee/revenue/supplySide
figures for 2026-05-17: feesUsd=306.452749, revenueUsd=24222.289850999998, supplySideUsd=61.29055.
This is a known upstream data issue (GitHub issue #9277), not an adapter bug.
```

```
$ npx ts-node --transpile-only cli/testAdapter.ts fees americafun 2026-06-11   # healthy day, unaffected
Daily fees: 2910, Daily revenue: 1199 (867 treasury + 332 stakers), Daily supply side revenue: 1711
```

Scripted the guard's exact logic against live API responses for all 13 known-bad days plus a spread of ~25 other days across the adapter's full history (2026-04-25 through 2026-09-05): rejects all 13, passes everything else, including three borderline days (1.04x-1.15x) that are NOT in the original 13 and would have been false positives at a tighter tolerance.

`npx tsc --project tsconfig.json --noEmit` - clean, zero errors, repo-wide.

### DIFF
--- a/fees/americafun.ts
+++ b/fees/americafun.ts
@@ -30,6 +30,22 @@ const fetch = async (options: FetchOptions) => {
   const meteoraFeeUsd = parseFloat(data.meteoraFee) || 0
   const lpFeeUsd = (parseFloat(data.dailySupplySideRevenue) || 0) - meteoraFeeUsd
 
+  const revenueUsd = protocolRevenueUsd + holdersRevenueUsd
+  const supplySideUsd = meteoraFeeUsd + Math.max(lpFeeUsd, 0)
+
+  // 20% tolerance: the known-bad days (#9277) all run 1.33x-79x over fees, while
+  // ordinary day-to-day noise on this adapter tops out around 1.15x (verified
+  // against a broad sample of the adapter's history) - a maintainer already
+  // declined a stricter fix (#8229) as over-correcting on that kind of noise.
+  // This only fires on the clearly-broken class, never on routine timing lag.
+  if (feesUsd > 0 && (revenueUsd + supplySideUsd) > feesUsd * 1.2) {
+    throw new Error(
+      `Upstream defillama.america.fun API returned internally inconsistent fee/revenue/supplySide figures for ${options.dateString}: ` +
+      `feesUsd=${feesUsd}, revenueUsd=${revenueUsd}, supplySideUsd=${supplySideUsd}. ` +
+      `This is a known upstream data issue (GitHub issue #9277), not an adapter bug.`
+    )
+  }
+
   if (feesUsd > 0) dailyFees.addUSDValue(feesUsd, "Swap Fees")
   if (protocolRevenueUsd > 0) dailyProtocolRevenue.addUSDValue(protocolRevenueUsd, "Swap Fees To Treasury")
   if (holdersRevenueUsd > 0) dailyHoldersRevenue.addUSDValue(holdersRevenueUsd, "Swap Fees To Stakers")


bheluga:
I think it should be fine occasionally because it could mean protocol claiming fees

gomesalexandre:
@bheluga appreciate the pushback, and the general point is real - it's the same shape as the documented `Revenue = ProtocolRevenue + HoldersRevenue` timing exception, so it's the right instinct to check before flagging any inequality as corruption.

Went and re-pulled the live endpoint for every day since the adapter's `start` (121 days with `feesUsd > 0`, 2026-04-20 through 2026-09-06) to see if that theory actually fits this specific case:

- **108/108 "normal" days**: `feesUsd == revenueUsd + supplySideUsd` to the exact cent, every single time (ratio 1.0000, not "close to 1" - literally equal). No drift, no partial catch-up, no gradual reconciliation over neighbouring days.
- **The same 13 days from #9277**: `dailySupplySideRevenue` comes back as a hard `0.00` from the API - not reduced, not delayed, exactly zero - while `meteoraFee` (which the adapter's own `breakdownMethodology` calls part of supply-side, "Meteora's protocol cut") is still populated at exactly 20.00% of that day's `dailyFees`, same as every other day. So the same response is simultaneously saying "supply side got nothing today" and "here's supply side's own component, present and correctly sized as always" - that's a contradiction inside one payload, not two systems settling on different days.
- **Cadence**: the 13 dates are 05-11, 05-14, 05-17, 05-21, 05-25, 06-13, 06-22, 06-29, 07-16, 07-20, 07-28, 08-07, 08-30 - gaps of 3, 3, 4, 4, 19, 9, 7, 17, 4, 8, 10, 23 days. No periodicity that would fit a scheduled claim/distribution job.

If this were a real "protocol claims accumulated fees in a lump sum" effect, I'd expect a claim day to show *elevated* supply-side/LP payouts too (they're cousins split from the same swap fees per the adapter's own methodology), not supply-side vanishing to exactly zero while only the protocol/holders side spikes as much as 79x. A batching artifact should show up as noise around 1.0 on plenty of ordinary days first - this dataset has none of that; it's a hard wall at exactly 1.0000 everywhere except these 13.

Given that, I think the original diagnosis holds for this specific adapter/endpoint - happy to be wrong if america.fun has a documented claim cadence I'm not seeing, in which case let me know and I'll go verify against it directly.

If you'd rather not gate on a 20%-tolerance band (fair, that's a blunter instrument than it needs to be), a tighter alternative that dodges your concern entirely: flag only when `dailySupplySideRevenue == 0 && feesUsd > 0` - the literal self-contradiction above, not a magnitude heuristic. That can never catch a legitimate claim-lag day since it only fires on the exact-zero case that's internally inconsistent with the same response's own `meteoraFee` field. Can put that up as a follow-up PR if useful - not reopening #9310 itself.

bheluga:
> @bheluga appreciate the pushback, and the general point is real - it's the same shape as the documented `Revenue = ProtocolRevenue + HoldersRevenue` timing exception, so it's the right instinct to check before flagging any inequality as corruption.
> 
> Went and re-pulled the live endpoint for every day since the adapter's `start` (121 days with `feesUsd > 0`, 2026-04-20 through 2026-09-06) to see if that theory actually fits this specific case:
> 
> * **108/108 "normal" days**: `feesUsd == revenueUsd + supplySideUsd` to the exact cent, every single time (ratio 1.0000, not "close to 1" - literally equal). No drift, no partial catch-up, no gradual reconciliation over neighbouring days.
> * **The same 13 days from [america.fun: the upstream endpoint contradicts itself on 13 of 139 days #9277](https://github.com/DefiLlama/dimension-adapters/issues/9277)**: `dailySupplySideRevenue` comes back as a hard `0.00` from the API - not reduced, not delayed, exactly zero - while `meteoraFee` (which the adapter's own `breakdownMethodology` calls part of supply-side, "Meteora's protocol cut") is still populated at exactly 20.00% of that day's `dailyFees`, same as every other day. So the same response is simultaneously saying "supply side got nothing today" and "here's supply side's own component, present and correctly sized as always" - that's a contradiction inside one payload, not two systems settling on different days.
> * **Cadence**: the 13 dates are 05-11, 05-14, 05-17, 05-21, 05-25, 06-13, 06-22, 06-29, 07-16, 07-20, 07-28, 08-07, 08-30 - gaps of 3, 3, 4, 4, 19, 9, 7, 17, 4, 8, 10, 23 days. No periodicity that would fit a scheduled claim/distribution job.
> 
> If this were a real "protocol claims accumulated fees in a lump sum" effect, I'd expect a claim day to show _elevated_ supply-side/LP payouts too (they're cousins split from the same swap fees per the adapter's own methodology), not supply-side vanishing to exactly zero while only the protocol/holders side spikes as much as 79x. A batching artifact should show up as noise around 1.0 on plenty of ordinary days first - this dataset has none of that; it's a hard wall at exactly 1.0000 everywhere except these 13.
> 
> Given that, I think the original diagnosis holds for this specific adapter/endpoint - happy to be wrong if america.fun has a documented claim cadence I'm not seeing, in which case let me know and I'll go verify against it directly.
> 
> If you'd rather not gate on a 20%-tolerance band (fair, that's a blunter instrument than it needs to be), a tighter alternative that dodges your concern entirely: flag only when `dailySupplySideRevenue == 0 && feesUsd > 0` - the literal self-contradiction above, not a magnitude heuristic. That can never catch a legitimate claim-lag day since it only fires on the exact-zero case that's internally inconsistent with the same response's own `meteoraFee` field. Can put that up as a follow-up PR if useful - not reopening #9310 itself.

Hmm some of your concern makes sense, but the main issue here is that by throwing error we would lose actual fee values for those days. IMO the right solution would be to derive fees from supply side revenue and revenue. WDYT?

PR #9321, opened 2026-09-08
Add Domination Finance fees and volume (Base perps)

mkzung:
DomFi is listed under Derivatives with a TVL page and no fees or volume adapter, so a
perp venue on Base publishes no revenue series.

Fees read both halves of every trading fee off `DomfiTradingCallbacks`. The vault's
half comes from `VaultOpeningFeeCharged` and `VaultClosingFeeCharged`, the treasury's
from `DevFeeCharged`, and liquidation margin from `VaultLiqFeeCharged`, which goes
wholly to the vault. Funding is left out: `FeesCharged` carries it as
`fundingFees int256`, paid between the two sides of the book rather than to the
protocol.

Volume is `collateral x leverage` from `MarketOpenExecuted` and `LimitOpenExecuted`.
`tradeNotional` on the same event is the position in units of the index at 18
decimals, so reading it as USDC is around ten orders of magnitude out: over 6
September the field sums to 202361302885660686023 against a true 12,038.17 USDC, a
factor of 1.68e10 at a mean index of 59.49.

Measured over the UTC day of 6 September, from the raw logs of blocks 50931727 to
50974926:

| leg | events | USDC |
|---|---:|---:|
| `VaultOpeningFeeCharged` | 5 | 4.4078 |
| `VaultClosingFeeCharged` | 5 | 0.9029 |
| `VaultLiqFeeCharged` | 0 | 0.0000 |
| `DevFeeCharged` | 10 | 5.3106 |

Vault 5.3107 and treasury 5.3106, so `dailyFees` is 10.6213 and
`dailyRevenue + dailySupplySideRevenue` meets it to the unit. Volume over the same
range is 12,038.17 USDC from 5 opens. No holders line: DomFi has no token.

The split is read rather than assumed. `DomfiPairsStorage` reports
`vault_fee_percent: 50` on all ten pairs and doubling the vault leg gives the same
total today, but `DevFeeCharged` is the treasury's own event and moves with the
parameter. It appears in no table on the protocol's docs site; the types come from
the verified implementation behind the proxy, and each one was confirmed against the
keccak of the topics the chain emits.

The fee parameters cross-check the total. Opening taker fee is 0.06% on BTCDOM and
0.08% on the other nine, closing exactly a quarter of opening on all ten. 10.6213
USDC on 12,038.17 of opens is 0.088%, which is opening plus closing at those rates.

`start` is 2026-06-23, the first day the callback records a fill.


bheluga:
@mkzung thanks for the PR

PR #9323, opened 2026-09-08
feat: report skate amm volume on robinhood chain

PhiDung-hub:
Adds `CHAIN.ROBINHOOD` (4663) to the `skate-amm` adapter's `chainConfig`. Skate AMM has had a live pool on Robinhood Chain since 2026-07-09 and its volume has not been reported.

`CHAIN.ROBINHOOD` already exists in `helpers/chains.ts` and is used by ~120 other adapters, so this is a one-line additive change with no new chain constant.

Start date `2026-07-09` is the first day the upstream stats API returns non-empty data for chainId 4663 (07-08 returns empty), and matches the day the pool went live.

Verified: `pnpm run ts-check` and `pnpm run ts-check-cli` both exit 0, and `npx ts-node cli/testAdapter.ts dexs skate-amm` produces a clean run with a `robinhood` row and no `---- ERROR ----` marker. Stashing the change removes the row, confirming the line is what enables the fetch.


bheluga:
@PhiDung-hub thanks for the PR

PR #9334, opened 2026-09-08
Add PrimeStaking (psXDC) fees and revenue adapter

mujtabasaqib:
Fixes ##6683

## Summary
Adds a fees/revenue adapter for PrimeStaking's psXDC liquid staking vault on XDC Network. 

## Scope
Covers `PrimeStakedXDC_V3_2` (`0xDc74c0DaED82ae94486DeeF22991d2F54173c734`), an ERC-4626 vault deployed at block 104591628 (2026-07-06). Stakers deposit XDC and receive psXDC shares; rewards accrue via share-price appreciation rather than a separate reward token.

## Methodology
Each RewardsDistributed(sender, grossAmount, fee) event carries the gross reward amount for that distribution and the protocol's fee cut. Tracking these two values per event gives three dimensions needed to track fees and revenue.

**Deliberately excluded from this PR (follow-ups):**
- Partner Staking pools (via `PartnerVaultRegistry` at `0x325DEEA5C7c0Ce0D774c4A67EcCaAf1cF8953a67`) 
  — fixed 15% fee, likely same event signature, but the two pools discovered on-chain currently appear unlaunched
- psXDC/XDC spot DEX 

## Verification
- `rewardFeeBps` confirmed on-chain at **0**.
- Confirmed via block explorer that `RewardsDistributed` has fired multiple times 
  (e.g. on 2026-08-24) with real, varying gross amounts, while `fee` has been 0 throughout.
- `dailyRevenue` correctly report 0 today  — while 
  `dailySupplySideRevenue` already surfaces real, non-zero data.
- No code changes will be required when governance activates the fee; the same event and fields already handle that case.

## Testing
Ran locally against a date with a known distribution event; confirmed output matches 
on-chain `grossAmount`/`fee` values pulled directly from the event log via block explorer.

### DIFF
--- a/fees/primestaking/index.ts
+++ b/fees/primestaking/index.ts
@@ -15,6 +15,11 @@
    so dailyRevenue report 0 today and dailySupplySideRevenue carries real signal
    until governance activates the fee.
 
+   Categorization follows DefiLlama's documented liquid-staking convention (same treatment
+   as the Lido adapter, fees/lido/index.ts): gross validator/masternode rewards are the total
+   value the protocol generates for stakers and are booked as dailyFees, with the protocol's
+   cut as dailyRevenue and the staker-retained remainder as dailySupplySideRevenue.
+
    Scope: this PR covers the V3.2 vault only. Partner Staking pools (via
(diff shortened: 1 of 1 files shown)


bheluga:
@mujtabasaqib thanks for the PR

PR #9341, opened 2026-09-08
feat(dashswallet): add BSC, expand to 5 adapter types across 7 chains

hudsonhank:
Update the DashsWallet fees and dexs adapters:
- Add BSC (chain 56) to both adapters.
- Track all 5 swap adapters (collateral/debt/repay + compound/morpho) with the chain-conditional matrix (compound on ETH/BSC/Base/Arbitrum, morpho on ETH/Base).
- Include the current v10/v9 fleet plus the historical v8/v4/v1 emitters (all emit the same SwapExecuted event; addresses match the canonical on-chain deployment).
- fees: document SupplySideRevenue in methodology; income statement holds (fees = revenue + supply-side, i.e. surplus = protocolFee + cashback).
- Clarify original-fleet adapter comment labels.

Verified with the official test runner (24/24 hours, real data on both adapters).

**NOTE**

#### Please enable "Allow edits by maintainers" while putting up the PR.

---

> - If you would like to add a `tvl` adapter please submit the PR [here](https://github.com/DefiLlama/DefiLlama-Adapters).

1. Once your adapter has been merged, it takes time to show on the UI. If more than 24 hours have passed, please let us know in Discord.
2. Please fill the form below **only if the PR is for listing a new protocol** else it can be ignored/replaced with reason/details about the PR
3. **For updating listing info** Please send a mail to metadata@defillama.com
4. Do not edit/push `package.json/package-lock.json` file as part of your changes
5. No need to go to our discord/other channel and announce that you've created a PR, we monitor all PRs and will review it asap

---


hudsonhank:
The BSC CI failure is a public-RPC rate-limit on the historical getLogs sweep (bsc-dataseed*: limit exceeded, dwellir 403, nodeflare 429), not an adapter issue — the registry reads succeeded and the earlier hourly slots processed fine before BSC throttled.

This PR doesn't change the getLogs path or its address set (only how targets are resolved), and pullHourly: true is pre-existing, so the ~24 sequential BSC log queries are identical to before.

Could you re-run the job? It's a known flake on the free BSC RPC pool that passes on retry / paid archive nodes — happy to fix anything real on the adapter if you spot it.

hudsonhank:
On the error-handling note: the try/catch here wraps observability only — it reads the previous S3 snapshot to detect newly registered contract versions and emit a log line. It does not touch fee/volume accounting or target resolution.

The two suggested fixes don't fit this specific block:

Removing the catch would let a transient S3/getCache failure propagate out of the fetcher, which zeroes that chain's day for a reason that has nothing to do with the data.
Rethrowing does the same — an observability read failing would abort the whole fetch.
That's exactly the failure mode the adapter is designed to avoid (a registry/cache hiccup must never zero the protocol's day). So the catch is intentional, but it no longer swallows silently — it now logs the error via sdk.log and continues:

The real target resolution has its own explicit guard right below (empty/failed registry read → sdk.log warning → static fallback), so no error there is discarded silently either. Happy to adjust further if you'd prefer a specific log format or level.

bheluga:
@hudsonhank thanks for the PR
Added [here](https://github.com/DefiLlama/dimension-adapters/pull/9411)

PR #9345, opened 2026-09-08
nft-volume: add Robinhood Chain (Seaport 1.6)

simplefarmer69:
Adds a `robinhood` entry to `factory/nftVolume.ts`, so Robinhood Chain shows up under NFT volume.

**Source:** on-chain Seaport `OrderFulfilled` logs. Seaport 1.6 (`0x0000000000000068F116a894984e2DB1123eB395`) is the only NFT marketplace deployed on chain 4663 — 1.4 and 1.5 have no bytecode there (`eth_getCode` returns `0x`), and OpenSea serves this chain through its own `seaport-rh` provider, so every marketplace sale lands in that one contract. No Allium/Dune coverage for the chain, so it reads logs directly.

**Price decoding:** for a listing the NFT sits in `offer` and the currency legs of `consideration` sum to the gross price (seller proceeds + fees + royalties); for a bid the currency sits in `offer`. Sale currencies observed over 24h are native ETH, WETH, and USDG (6 decimals) — all handled by `createBalances().add(token, amount)`.

**Dedupe:** an accepted bid goes through `matchOrders`, which emits `OrderFulfilled` for both sides of the same trade. Legs are paired per `txHash + nft id` and the bid leg is kept (what the buyer actually paid); unpaired legs are ordinary one-sided fills. Over a sampled 24h window that is 1,841 mirrored pairs out of 49,910 trades — about 30 WETH of double counting removed.

**Verified locally:**

```
npm test nft-volume robinhood            -> Daily volume: 1.15 M
npm test nft-volume robinhood 2026-09-04 -> Daily volume: 1.08 M
```

Cross-checked against a decoded 24h log dump (51,761 `OrderFulfilled` events): 257.6 ETH native + 103.9 WETH + 117.7k USDG, and the decoded per-sale prices were spot-checked against individual transactions for native listings, WETH bid acceptances, and USDG sales.

Anvil AMM NFT swaps on the same chain are deliberately left out — they are already counted as DEX volume in `dexs/clutch-anvil`, and their USD contribution here is negligible next to Seaport.

### DIFF
--- a/factory/nftVolume.ts
+++ b/factory/nftVolume.ts
@@ -54,6 +54,56 @@ async function flow(options: FetchOptions) {
   return { dailyVolume };
 }
 
+// OpenSea's Seaport 1.6 is the only NFT marketplace deployed on Robinhood Chain
+// (1.4 and 1.5 have no bytecode there), and OpenSea serves the chain through its own
+// `seaport-rh` provider, so every marketplace sale lands in this one contract.
+const seaport_1_6 = '0x0000000000000068F116a894984e2DB1123eB395';
+const event_order_fulfilled = "event OrderFulfilled(bytes32 orderHash, address indexed offerer, address indexed zone, address recipient, (uint8 itemType, address token, uint256 identifier, uint256 amount)[] offer, (uint8 itemType, address token, uint256 identifier, uint256 amount, address recipient)[] consideration)";
+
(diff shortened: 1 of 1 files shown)


bheluga:
@simplefarmer69 thanks for the PR
Used allium just in case we get more nft protocols

PR #9221, opened 2026-09-04
solana dune adapters: wait for the day to finish indexing, published volume is truncated by 13-37%

tarun-khatri:
Website: https://humidifi.xyz/ — Twitter: https://twitter.com/humidifi

`dexs/humidifi` reads `solana.instruction_calls` and `tokens_solana.transfers` for the day as soon as it closes. Dune finishes indexing a Solana day several hours after it ends, so the query answers with whatever is loaded so far — a partial day, not an error — and that truncated number is what gets stored.

**The stored value is literally the first 18 hours of the day.** Summing the adapter's own query by hour for 2026-08-20:

| through | cumulative |
|---|---|
| 15:00 | 86,192,094 |
| 16:00 | 91,866,112 |
| **17:00** | **97,399,132** |
| 18:00 | 103,612,000 |
| 23:00 (full day) | **135,919,436** |

Published for 2026-08-20: **97,402,250**. That is the 00:00–17:59 cumulative to within 0.003%; the remaining six hours, $38.5M, were never picked up.

It is not a one-off. Re-running the adapter's query today against what is published:

| day | published | same query today | short by |
|---|---|---|---|
| 2026-08-15 | 65,450,780 | 85,685,229 | -23.6% |
| 2026-08-18 | 160,543,513 | 207,561,136 | -22.7% |
| 2026-08-20 | 97,402,250 | 135,919,436 | -28.3% |
| 2026-08-21 | 213,788,231 | 237,927,732 | -10.1% |
| 2026-09-02 | **0** | 652,262 | -100% |
| 2026-09-03 | 13,308,991 | 18,873,500 | -29.5% |

2026-08-18 alone is $47M of missing volume, and 2026-09-02 was stored as a zero day.

**The repo already has the fix and it is measurably working.** `duneSolanaDexTrades` refuses a window whose end is inside the last ten hours, and the Solana dexs wired through `factory/duneSolanaDex.ts` reconcile exactly. Comparing every decoded Solana project for 2026-08-20 against what is published:

```
guarded (via duneSolanaDexTrades)      unguarded (own adapter)
  scorch        0.0%                     humidifi      -30.2%
  aquifer       0.0%                     alphaq        -24.2%
  goonfi        0.0%                     tessera       -22.3%
  whalestreet   0.0%                     bisonfi       -13.2%
```

[...]

PR #9348, opened 2026-09-09
jupiter-lend: wait for the day to finish indexing before reading the fluid dune view

### LINKED ISSUE #9347: jupiter-lend: 09-06 stored at 15% of a normal day and nothing since — the fluid dune MV is read mid-refresh, no wait like #9221 gave its siblings
`fees/jupiter-lend` ran flat for weeks, then stored a fragment and went dark:

```
08-27 $144,357 · 08-28 $132,865 · 08-29 $126,785 · 08-30 $146,073 · 08-31 $135,024
09-01 $137,552 · 09-02 $130,402 · 09-03 $129,539 · 09-04 $132,197 · 09-05 $127,214
09-06 $20,334      <- 15.2% of the 10-day mean ($134,201)
09-07 / 09-08 / 09-09  nothing
```

$113,867 understated on 09-06, plus three dark days at ~$134k on a protocol with $1.11bn of TVL.

## the baseline is real — checked against Jupiter's own api

`lite-api.jup.ag/lend/v1/earn/tokens` gives per-vault assets and rates (rates in bps):

| | |
|---|---|
| earn-side assets across the 7 vaults | $574,824,204 |
| implied interest at `totalRate` | $66,843/day |
| share of DefiLlama's $1,106,530,526 TVL | 51.9% |
| scaled to full TVL | **$128,673/day** |
| published `dailySupplySideRevenue` baseline | **$127,486/day** |

1% apart, so the pre-09-06 level is right and 09-06 is a fragment rather than a quiet day.

## it is not a dune outage

- `jupiter-lend-dex` — same protocol, same chain, also Dune-backed — published 09-07 and 09-08
  normally ($15.7M and $18.7M volume).
- The six solana Dune adapters that got `assertDuneSolanaIndexed` in #9221 (humidifi, alphaq,
  bisonfi, solfi-v2, tessera, pumpfun) all published complete days through 09-07.

## mechanism

`fees/jupiter-lend` reads a materialized view with no wait and no completeness check:

```sql
from dune."0xfluid".result_juplend_historical_tvl_by_token_mv
where day >= FROM_UNIXTIME(${options.startTimestamp})
  and day <  FROM_UNIXTIME(${options.endTimestamp})
```

```ts
if (!data || data.length === 0) {
  throw new Error(`No JupLend data returned from Dune for day starting ...`);
}
```

That throw is why 09-07 onward are dark rather than zero, which is the right behaviour. The problem
is the case just above it: when the MV is **mid-refresh** it returns *some* rows for the day, the
guard passes, and the partial is stored as the day's number and never revisited. 09-06 is that case.

This is the same shape as #9221 — "wait for the day to finish indexing, published volume is
truncated by 13-37%" — on an adapter that sweep did not cover, and its instance is worse (15%).

Fix is the guard that PR already added to its six siblings; PR to follow. Flagging the one thing I
cannot check from outside: whether the Fluid MV refreshes within the helper's 10-hour window. If it
lags longer than that, the threshold needs to be higher for this adapter, and only your runner logs
can say.

0xshubhs:
Fixes #9347.

`fees/jupiter-lend` reads the Fluid materialized view with no wait, so a run that lands while the MV
is mid-refresh gets *some* rows for the day, passes the `data.length === 0` check just below, and
stores the partial as the day's number. 09-06 was stored as **$20,334 against a 10-day mean of
$134,201**: 15.2% of a normal day; and nothing has published since.

This is the guard #9221 added to the six solana Dune adapters for the same reason ("wait for the day
to finish indexing, published volume is truncated by 13-37%"). `fees/jupiter-lend` was not in that
sweep. Its siblings all published complete days through 09-07; this one did not.

## verification

```
# a day that ended less than 10h ago is now refused instead of read half-baked
$ npx ts-node cli/testAdapter.ts fees jupiter-lend 2026-09-09
End timestamp is less than 10 hours ago, skipping fetch due to dune indexing delay
  2026-09-08T23:59:59.000Z 2026-09-08T21:11:38.095Z

# an older day passes the guard and goes on to the dune call as before
$ npx ts-node cli/testAdapter.ts fees jupiter-lend 2026-09-06
Error: DUNE_API_KEYS environment variable is not set
```

`npx tsc --noEmit` is clean.

Being explicit about what I could not check: I have no Dune key, so I cannot execute the query
itself; the second run above gets as far as the key and stops. The change is the same one-line
guard already merged for six adapters, so there is no new logic to get wrong, but the 10-hour
threshold is inherited rather than measured against *this* source. If the Fluid MV refreshes later
than that, the wait needs to be longer for this adapter, and only the runner logs would show it.

### DIFF
--- a/fees/jupiter-lend/index.ts
+++ b/fees/jupiter-lend/index.ts
@@ -2,6 +2,7 @@ import { Dependencies, FetchOptions, FetchResultV2, SimpleAdapter } from "../../
 import { CHAIN } from "../../helpers/chains";
 import { JUPITER_METRICS, jupBuybackRatioFromRevenue } from "../jupiter";
 import { queryDuneSql } from "../../helpers/dune";
+import { assertDuneSolanaIndexed } from "../../helpers/duneSolanaDex";
 
 // Jupiter Lend runs on Fluid infra as two instances tagged by `instance_name`
 // in the Dune MV: the main Jupiter markets and the isolated Ethena (USDe) market
@@ -19,6 +20,7 @@ const isEthenaInstance = (row: any) => String(row.instance_name || '').toLowerCa
 const sumUsd = (rows: any[], key: string) => rows.reduce((sum, row) => sum + (row[key] || 0), 0);
 
 const fetch = async (options: FetchOptions): Promise<FetchResultV2> => {
+  assertDuneSolanaIndexed(options)
 
   // const sql = getSqlFromFile("helpers/queries/jupiter-lend.sql", {
   //   start: options.startTimestamp,


0xshubhs:
`test` is red here and it is the guard doing its job, not a break.

CI runs `testAdapter fees jupiter-lend` for **yesterday** — it started 07:12 UTC and the requested
window ends 2026-09-09T00:00Z, seven hours earlier. That is inside the ten-hour wait, so the run is
refused:

```
Start Date: Tue, 08 Sep 2026 00:00:00 GMT
End Date:   Wed, 09 Sep 2026 00:00:00 GMT
End timestamp is less than 10 hours ago, skipping fetch due to dune indexing delay
```

Which is exactly the day the adapter should now decline to publish — that early read is what stored
09-06 at 15% of a normal day.

For completeness, since it would be easy to assume otherwise: this file's `test` was already red
before my change, on `DUNE_API_KEYS environment variable is not set` (that is also why #9221 went in
red). My change swaps one red for another — the guard now trips before the missing key does. There
is no configuration in which CI passes for this adapter today.

`ts-check` is green, and locally the guard passes through for an older day and reaches the Dune call
normally (`testAdapter fees jupiter-lend 2026-09-06` → `DUNE_API_KEYS ... not set`), so it is not
blocking backfills.

bheluga:
@0xshubhs thanks for the PR
Probably their table stopped updating since last few days.
Delay won't fix it. I ll let jupiter team know

PR #9351, opened 2026-09-09
hydradx: fix endpoint

### LINKED ISSUE #9169: hydradx fees: dark since 08-29 — the metrics-aggregator host is gone (404 on every route), res.json() throws and kills the on-chain lending fees too
`fees/hydradx` (Hydration Lending, ~$1.9k/day dense) has published nothing since its 2026-08-28 row. the lending math itself is fine — it's all on-chain aave-style index reads — the throw comes from the extra protocol streams block, which fetches `https://hydration-metrics-aggregator.indexer.hydration.cloud/api/v1/fees/charts`. that host now answers `404 page not found` on every route (`/`, `/health`, `/api/v1/fees/charts` with and without params), so `res.json()` on the html body throws and takes the whole adapter down.

```
curl -s "https://hydration-metrics-aggregator.indexer.hydration.cloud/api/v1/fees/charts?productType=money-market&feeDestination=protocol&streamType=liquidation_penalty&startTime=2026-09-01T00:00:00.000Z&endTime=2026-09-02T00:00:00.000Z&bucketSize=24hour"
404 page not found
```

this is the same indexing-layer incident #9025 mapped on 08-27 — each hydration.cloud host wore a different mask (tcp-refused / frozen / http 500), and this one was the http-500 host. the failure had two phases worth noting:

- while it served 500s with a json error body, `json.periodAggregate ?? 0` swallowed it — the adapter kept publishing through 08-28 with the liquidation/PEPL/HSM streams silently at $0
- once the route went 404 with an html body, `res.json()` started throwing — dark since 08-29

so the `?? 0` is two-sided the same way the obol one was: it hides the stream when the api half-fails, and the crash only surfaces when the host dies harder.

i didn't send a pr because there is nothing verifiable to repoint at — no other `*.hydration.cloud` host serves the fees/charts api (the graphql indexer that fixed `dexs/hydration-dex` in #9026 has no such route), and whether these streams live somewhere else now is knowledge only the hydration side has. @green-jay added these streams in #6388 — if the metrics service moved, a one-line host swap revives this; if it's gone for good, the extra-streams block needs an explicit decision about whether to drop the streams loudly or wait it out. 08-29..09-01 need a refill either way once fixed.

chetna-mittal:
## Summary
- Fixes #9169: `fees/hydradx` (Hydration Lending) has been dark since 2026-08-28 because `hydration-metrics-aggregator.indexer.hydration.cloud` returns a 404, and `res.json()` on the HTML body aborted the whole fetch.
- Removes that optional API entirely. Keeps the on-chain Aave-style borrow-interest path (including HOLLAR), which was always the dense ~$1.9k/day series.
- Documents exclusions for money-market liquidation penalty, PEPL liquidation profit, and HSM revenue — no refillable public source after the aggregator died, and `getBlock('hydradx')` returns null, so EVM `getLogs` cannot replace them.
- Scales rate-based accruals by `toTimestamp - fromTimestamp`, adds breakdown labels/methodology, sets `pullHourly: true`.

## Test plan
- [x] `npm run test fees hydradx 2026-08-29` → ~$1.88k fees (first dark day)
- [x] `npm run test fees hydradx 2026-09-02` → ~$1.89k
- [x] `npm run test fees hydradx 2026-09-08` → ~$1.90k
- [x] Confirmed metrics host still 404s; adapter no longer depends on it
- [ ] Refill `hydration-lending` from 2026-08-29 after merge

### DIFF
--- a/fees/hydradx.ts
+++ b/fees/hydradx.ts
@@ -3,42 +3,29 @@ import { SimpleAdapter, FetchOptions } from "../adapters/types";
 import AaveAbis from '../helpers/aave/abi';
 import { METRIC } from "../helpers/metrics";
 
-const FEES_API = "https://hydration-metrics-aggregator.indexer.hydration.cloud/api/v1/fees/charts"
-
-// Streams not covered by the EVM lending calculation below:
-// - liquidation_penalty: treasury's 10% cut from MM liquidations (event-based, not in liquidity index)
-// - pepl_liquidation_profit: 100% protocol revenue from PEPL liquidations
-// - hsm_revenue: HSM arb profits + yield from yield-bearing stablecoins
-const EXTRA_PROTOCOL_STREAMS = [
-  { productType: "money-market", streamType: "liquidation_penalty", label: "Liquidation Fees", revenueLabel: "Liquidation Penalty To Treasury" },
-  { productType: "money-market", streamType: "pepl_liquidation_profit", label: "PEPL Liquidation Profit", revenueLabel: "PEPL Liquidation Profit To Treasury" },
-  { productType: "hollar", streamType: "hsm_revenue", label: "HSM Revenue", revenueLabel: "HSM Revenue To Treasury" },
-] as const
-
-async function fetchProtocolStream(productType: string, streamType: string, startTime: string, endTime: string): Promise<number> {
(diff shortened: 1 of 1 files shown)


bheluga:

PR #9353, opened 2026-09-09
evedex: attribute API-only adapters as off-chain

### LINKED ISSUE #9150: evedex: nothing published since 2026-07-08 while the venue has $585.7M OI — the Eventum RPC host is NXDOMAIN
EvEDEX has published nothing since early July in any dimension, while the venue itself is fully live: its exchange API reports **$585,679,782 of open interest** across 52 contracts and **$564,666,267** of 24h quote volume right now.

| | last data point |
| --- | --- |
| `/summary/open-interest/evedex` | 2026-07-08 (157 points, `total24h` null) |
| `/summary/fees/evedex` | 2026-07-01 (36 points) |

`/protocol/evedex` returns id 6497, category Derivatives, `deadFrom: null`, chains `["Eventum", "Arbitrum"]`. It is not in `factory/deadAdapters.json`.

### The Eventum chain has no reachable RPC any more

`@defillama/sdk` `providers.json` has:

```json
"eventum": { "rpc": ["https://mainnet-rpc.evedex.com"], "chainId": 161803, "explorer": "https://explorer.evedex.com" }
```

That hostname is **NXDOMAIN** — a DNS query for `mainnet-rpc.evedex.com` returns `Status: 3` with no answer section, while `exchange-api.evedex.com` from the same domain resolves normally (`Status: 0`, 3 A records). The explorer is gone as well: `https://explorer.evedex.com` now 301s to the marketing site `https://evedex.com/en/`, and `POST /api/eth-rpc` against it ends in a 405. I could not find any alternative endpoint for chainId 161803.

Running the adapter locally shows the failure is chain-level rather than adapter-level — the fetch itself works and returns a real number, and the only errors are `No RPCs available for eventum` / `Error getting block: eventum`.

### Why this is awkward to fix from outside

`open-interest/evedex.ts` and `dexs/evedex/index.ts` make **no chain calls at all** — they only read `exchange-api.evedex.com` and `market-data-api.evedex.com`, both of which are healthy and returning current data. They are keyed to `CHAIN.EVENTUM` purely as an attribution label, so they are being held back by a chain they never touch.

The obvious candidate fixes each have a problem I cannot resolve from outside the team:

- **Re-key to `CHAIN.OFF_CHAIN`** (what the repo does for other API-only perp venues such as `bullbit-perp-dex-oi`, `defx`, `antarctic-oi`, `apollox`, `kalshi`). This would restore the numbers, but it silently changes which chain EvEDEX's OI and volume are attributed to, and the protocol is registered on Eventum and Arbitrum. That is an attribution call, not a bug fix.
- **Point Eventum at a working RPC** via `helpers/env.ts`, which is how several other chains are patched in this repo. I could not find one to point at — the RPC host and the explorer are both gone.
- `fees/evedex/index.ts` genuinely uses `options.getLogs` on Eventum, so it cannot move regardless.

I am also explicit about one gap: I proved the RPC hostname is dead and that DefiLlama holds no Eventum data in any dimension (`/overview/fees/eventum` returns `total24h: null` for every protocol, and `/v2/chains` shows Eventum at 0 TVL), but I could not observe the server-side chain handling, so "dead Eventum RPC → adapter output dropped" is an inference from those facts rather than something I watched happen.

Flagging rather than opening a PR because the right answer looks like a chain-registry or attribution decision. Happy to implement whichever direction the team prefers.

chetna-mittal:
## Summary
- Fixes #9150: EvEDEX has been dark since early July while the venue is live (~$575M OI, large daily volume). `open-interest/evedex` and `dexs/evedex` are API-only but were keyed to `CHAIN.EVENTUM`; `mainnet-rpc.evedex.com` is NXDOMAIN, so the runner could not complete those chains.
- Re-attributes OI, volume, and fees to `CHAIN.OFF_CHAIN` (same pattern as bullbit / defx / apollox API-only perps).
- Fees keep OHLCV × (maker + taker) trading fees with the existing `Trading Fees` / `Trading Fees To Protocol` labels. Billing fees and cashback/VaultV2 log streams are dropped until an Eventum RPC is available again; methodology notes the exclusions.

## Test plan
- [x] `npm run test open-interest evedex` → ~$575M OI on `off_chain`
- [x] `npm run test dexs evedex 2026-09-08` → ~$602M daily volume on `off_chain`
- [x] `npm run test fees evedex 2026-09-09` → ~$325k daily trading fees on `off_chain`
- [ ] Refill OI/volume/fees after the merge for the gap windows above

### DIFF
--- a/dexs/evedex/index.ts
+++ b/dexs/evedex/index.ts
@@ -35,7 +35,7 @@ const methodology = {
 
 const adapter: SimpleAdapter = {
     fetch,
-    chains: [CHAIN.EVENTUM],
+    chains: [CHAIN.OFF_CHAIN],
     start: "2025-06-30",
     methodology,
 };
--- a/fees/evedex/index.ts
+++ b/fees/evedex/index.ts
@@ -5,15 +5,6 @@ import fetchURL, { fetchURLAutoHandleRateLimit } from "../../utils/fetchURL";
 import { getConfig } from "../../helpers/cache";
 import PromisePool from "@supercharge/promise-pool";
 
-const CONTRACTS = {
-  cashback: "0x0a9591c64Fd9e8C1f9A81DB1B668a5f211b5735A",
-  vaultV2: "0x026968b5cED079ECCD6CC78f35a5Dfddc13F9Af8",
-  billing: "0xF5dA5b918b8c50d211495F4316070Cfc4Fbe128E",
-};
-
-// EVEDEX payout vaults distribute 6-decimal USDT.
-const USDT_DECIMALS = 1e6;
-
 const URLS = {
   market: "https://trading-api.evedex.com/api/market",
   instruments: "https://trading-api.evedex.com/api/market/instrument",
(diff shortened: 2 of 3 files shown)


bheluga:

PR #9355, opened 2026-09-09
fix (dinari): track v2 order processor after 2026-06-07 pause

### LINKED ISSUE #9200: dinari: the four OrderProcessors it reads are paused since 2026-06-07, orders moved to a new contract
## What

`dexs/dinari` has published `$0` since **2026-06-05** (`total24h`, `total30d` both 0; `total1y` $10,747,260 is all pre-June history; last non-zero day 2026-06-05 = $1,106).

It cannot recover on its own: **all four `OrderProcessor` contracts the adapter reads were deliberately paused, and are still paused right now.**

```
eth_call ordersPaused() [0xd87f8ee1]        -> 1  on all four
eth_call owner()        [0x8da5cb5b]        -> 0x06b09f6008780771980bfe7d4bb821f5f1aa60a3  on all four
eth_call treasury()     [0x61d027b3]        -> 0xf96b974fe330c29e80121e33ed4071c283257979  on all four

ethereum 0xA8a48C202AF4E73ad19513D37158A872A4ac79Cb
arbitrum 0xFA922457873F750244D93679df0d810881E4131D
base     0x63FF43009f9ba3584aF2Ddfc3D5FE2cb8AE539c0
plume    0xc1571FEbBb6F8b62eDD0E4694714A382885d6bAB
```

`OrdersPaused(bool)` (`0xeded67f5…`) fired with `data = 1` on all four within about a minute on **2026-06-07**, and none of them has emitted an `OrderFill` since. The adapter's `getLogs` therefore returns `[]` on every chain, every day, and writes a successful `$0`.

The protocol did not stop — it moved.

## Where it moved

`0xf60f689ec22fC2D485b3C734eFE58538cCc28766`, the same address on every chain (identical 172-byte proxy bytecode on ethereum, arbitrum, base and plume), is live and processing orders. On Arbitrum it emitted **4,987 logs in the last 12 days, most recently 2026-09-03 11:27 UTC**; Ethereum and Plume are active too but much smaller.

The link to Dinari is the admin key and the treasury, both read on-chain: the four paused processors all return `owner() = 0x06b09f60…` and `treasury() = 0xf96b974f…`, and that same `0x06b09f60…` is the account that sent the pause transactions. The new contract carries the same product namespace in its order events — every order embeds a string of the form `dshare_proxied_order:prod:<uuid>` — and charges the same flat `0.2e18` fee that the old processors paid to that treasury.

## Scale, measured

Decoding the new contract's events on Arbitrum over 1,080,000 blocks (~3.6 days): 315 order requests and 255 settlements, **$15,526 of buy notional and $80,102 of settlement payouts, ≈ $26,600/day on Arbitrum alone**. Against a published `$0`, and against a pre-pause run-rate of roughly $750–$1,100/day, so the protocol is materially bigger now than when the adapter last worked.

Ethereum (95 logs / 30d) and Plume (60 logs / 30d) add little. **I could not measure Base at all** — every one of my 144 `eth_getLogs` chunks against public Base RPCs errored, so I have no figure for it and would not want the issue read as if Base were zero.

## Why I am not sending a patch

Two things need deciding by someone with more context than I have, and guessing either would silently redefine what this adapter reports:

1. **The settlement event has no published ABI.** `0x4907bbdd27eecc3fd5b0202f215bb36933e3601de096474cb7afb6f5c0b4aeb1` is not in openchain's database. From decoding, its `topic1` is an order id, `topic2` the payment token, `topic3` the requester, and `data[0]` the payment amount — but I would be writing an `eventAbi` string I cannot verify against a source. (`OrderRequested` does resolve: `0xbd045500…` = `OrderRequested(address,address,address,address,uint8,uint8,uint256,uint256,uint256,string,uint256)`.)

2. **Buys and sells are not symmetric, and naively summing both double counts.** Matching the two events by their embedded order uuid over the same window: of 189 buys (an `OrderRequested` whose amount word is non-zero), **17 also produce a settlement** — and those settlements are dust, e.g. 822.46 requested against 0.0003 settled, 988.61 against 0.0048, which look like refunds of unspent payment rather than a second leg. All 126 sells settle cleanly. So `sum(requested) + sum(settled)` is close but not right, and picking the correct rule is a methodology call.

One more thing worth flagging for whoever picks this up: the amounts in both events are normalised to **18 decimals even when the payment token is 6-decimal USDC** — the settlement I inspected emits `15.433e18` while the actual transfer is `15,433,000` USDC raw. They must be added as USD, not passed to `balances.add(token, amount)`.

Also note the same address is deployed on Avalanche and HyperEVM, which this adapter has never covered, so a fix is an opportunity to widen it.

chetna-mittal:
## Summary
- Fixes #9200: legacy `OrderProcessor`s were paused on 2026-06-07, so the adapter stored successful `$0` while orders moved to diamond `0xf60f689ec22fC2D485b3C734eFE58538cCc28766` (same address on eth/arb/base/plume).
- Keep legacy `OrderFill` for history; from 2026-06-07 read v2 request + settlement logs (amounts are 18-decimal USD → `addUSDValue`).
- Volume = buy `paymentQty` on requests + settlements with notional ≥ $1 (skips buy-refund dust). Fees = request fee field.

## Test plan
- [x] `npm run test dexs dinari 2026-06-05 arbitrum` — legacy still nonzero (~$242 vol)
- [x] `npm run test dexs dinari 2026-09-04 arbitrum` — ~$61.5k vol / ~$42.5 fees
- [x] `npm run test dexs dinari 2026-09-04 ethereum` — small nonzero post-migration

### DIFF
--- a/dexs/dinari/index.ts
+++ b/dexs/dinari/index.ts
@@ -1,75 +1,103 @@
 import { Adapter, FetchOptions } from "../../adapters/types";
 import { CHAIN } from "../../helpers/chains";
 import { METRIC } from "../../helpers/metrics";
+
 // https://github.com/dinaricrypto/sbt-contracts
 // https://github.com/dinaricrypto/sbt-contracts/blob/main/releases/v1.0.0/order_processor.json
-// https://github.com/dinaricrypto/sbt-contracts/blob/50f7cb5f0613c03fad42e7ece78e56d2992d03ba/src/orders/OrderProcessor.sol#L652
+// v2: https://arbiscan.io/address/0xf60f689ec22fC2D485b3C734eFE58538cCc28766
+const V2_PROCESSOR = "0xf60f689ec22fC2D485b3C734eFE58538cCc28766";
+const V2_START_TS = 1780790400; // 2026-06-07 — legacy OrderProcessors paused
+
 const config: Record<string, { processor: string; start: string }> = {
-    [CHAIN.ETHEREUM]: { processor: "0xA8a48C202AF4E73ad19513D37158A872A4ac79Cb", start: "2024-05-22" },
-    [CHAIN.ARBITRUM]: { processor: "0xFA922457873F750244D93679df0d810881E4131D", start: "2024-05-22" },
-    [CHAIN.BASE]: { processor: "0x63FF43009f9ba3584aF2Ddfc3D5FE2cb8AE539c0", start: "2024-06-07" },
-    [CHAIN.PLUME]: { processor: "0xc1571FEbBb6F8b62eDD0E4694714A382885d6bAB", start: "2025-04-22" },
+  [CHAIN.ETHEREUM]: { processor: "0xA8a48C202AF4E73ad19513D37158A872A4ac79Cb", start: "2024-05-22" },
+  [CHAIN.ARBITRUM]: { processor: "0xFA922457873F750244D93679df0d810881E4131D", start: "2024-05-22" },
+  [CHAIN.BASE]: { processor: "0x63FF43009f9ba3584aF2Ddfc3D5FE2cb8AE539c0", start: "2024-06-07" },
+  [CHAIN.PLUME]: { processor: "0xc1571FEbBb6F8b62eDD0E4694714A382885d6bAB", start: "2025-04-22" },
 };
 
-const events = {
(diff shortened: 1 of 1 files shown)


bheluga:

PR #9360, opened 2026-09-09
[fix] aftermath-fi-perp - support the production relaunch

am-aftermath:
Aftermath's relaunched perpetual markets need bounded historical requests and explicit handling of missing data. The existing CCXT candle request queries from `since` to now, can fail on the API's history-range limit, and silently drops failed markets through PromisePool. This change makes those failures visible and scopes both volume and open interest to the production deployment.

Website: https://aftermath.finance/
Twitter: https://x.com/AftermathFi
Docs: https://aftermath.finance/docs/ and https://aftermath.finance/api/openapi/spec.json
Existing listing: https://defillama.com/protocol/aftermath-perps
Related TVL PR: https://github.com/DefiLlama/DefiLlama-Adapters/pull/20964

### Changes

- Discover USDC markets using `/perpetuals/all-markets`, validate the production package and native Circle USDC type, and reject missing/duplicate market lists. The endpoint currently returns 31 markets; paused markets remain included because they may still have positions or historical trades.
- Keep volume's daily v1 model and request exactly one completed UTC day through `/perpetuals/market/candle-history`. Count the source's USD taker notional once, without maker volume or token/contract scaling. Reject wrong-day, duplicate, malformed or negative candles, and propagate HTTP failures.
- When a candle is absent, check the market's last trade before the window end. Accept zero only if there were no fills in the day; this supports historical dates before a newly added market began trading. An entirely missing day throws.
- Keep OI current-only with `runAtCurrTime: true`, remove the misleading historical start, and explicitly disable hourly aggregation. Calculate base-token OI × USD index price once per market. The API has already decoded the on-chain fixed-point amount. Empty markets contribute zero; missing state or missing prices for open positions fail the run.
- Add volume methodology and eight regression tests; OI calculation details remain in inline comments, with no OI methodology metadata. No dependencies, shared Sui/RPC configuration, manifests or lockfiles changed.

### History and refill

**Please refill volume only from 2026-08-18.** This is the first available production daily candle; it has zero volume, and the first fills occur on August 19. The owner requested tracking only the relaunched deployment; a working legacy data source was not identified.

The existing slug and volume start metadata (`2025-02-18`) are preserved, but requests before August 18 now throw instead of returning zero and overwriting stored legacy history. Keep the published legacy series. OI has no historical source and must not be backfilled from current snapshots. Keep existing server-side dimension mappings for `aftermath-fi-perp`; the volume refill is the server-side follow-up after merge. AFLP TVL is handled by the related PR above.

### Validation

Passed on 2026-09-09:

```text
node -r ts-node/register/transpile-only --test helpers/aftermath.test.ts  # 8 passed
npm run ts-check
npm run ts-check-cli
node_modules/.bin/tsc --noEmit --esModuleInterop --resolveJsonModule --skipLibCheck --target ES2020 --module commonjs helpers/aftermath.test.ts
npm run test dexs aftermath-fi-perp 2026-09-09  # September 8: $2.25m
npm run test dexs aftermath-fi-perp 2026-08-21  # August 20: $1.04m
npm run test open-interest aftermath-fi-perp  # approximately $746.7k–$747.0k at validation time
git diff --check
```

Independently requested all markets' daily candle histories and compared them with direct adapter runs; tolerance $0.01, observed difference $0 on all five dates:

| UTC day | USD volume |
| --- | ---: |
| 2026-08-18 | 0 |
| 2026-08-19 | 452.9090336816216 |
| 2026-08-25 | 1,398,945.4779531842 |
| 2026-09-04 | 3,903,807.10992982 |
| 2026-09-08 | 2,250,179.99574975 |

Sui identity/units verification used the repository's GraphQL helper and `SUI_GRAPH_RPC`. BTC market `0x05b5c3bea84c4b8f33cf592d899008336dcbae8c9c6c75b2f8e7b8f7878744c1` returned on-chain `open_interest = 358817513000000000`; the immediately following API response was exactly `0.358817513` BTC at index price `$79293.97785895`. The base-token amount reconciled exactly. Live USD totals must be compared at the same snapshot.

Sample production trade from the trade-history endpoint: `C7GE4nVmpMABjEHorPu94RGEamsjTaQH2T3ECEXwoymA`, BTC, timestamp `1788911343816`, size `0.000013894`, price `$78412.23`. The production service's candle generation accumulates `FilledTakerOrder` quote deltas; no fee-rate estimate is used. This is an existing API-based integration; volume reconciliation validates the adapter's API mapping, rather than independently re-indexing all on-chain fills.

### Review notes

- **Are zero days valid?** Yes when backed by an explicit zero candle or confirmed absence of fills. August 18 is a verified example. Missing data, failed requests and unfinished days are not stored as zero.
- **Dashboard differences?** The adapter uses completed UTC days; a rolling 24h dashboard covers a different window. OI changes with positions and prices. The public page showed $2.25m perp volume, matching September 8. Published OI was $751,770 for September 8 and $765,804 for September 9, at different snapshots from the local run. The public derivatives summary API returned HTTP 402, so its full historical series could not be downloaded.
- **Fee rate/split changes and fee-to-volume ratio?** Neither adapter exports fees or revenue, and no fee rates/splits are changed or inferred. A historical fee-schedule audit is outside this volume/OI change.
- **Existing listings/PRs?** Updates the existing `aftermath-fi-perp` listing; no new listing is requested. No other open Aftermath PR was found in this repository at submission time. The related AFLP TVL update is #20964 in `DefiLlama-Adapters`.
- **Data availability?** Both metrics still depend on the protocol API/indexer. Missing required data fails the whole Sui result instead of publishing an incomplete total. Discovery depends on the source continuing to expose markets with historical activity/open positions. The live-state response has no common snapshot timestamp/checkpoint, so stale but well-formed state cannot be fully detected here.


bheluga:
@am-aftermath thanks for the PR

PR #9368, opened 2026-09-09
euler: add the hyperevm vault factory

### LINKED ISSUE #9366: euler: hyperevm is missing from eVaultFactories, so the chain never runs ($2.79m tvl, 58 vaults)
`fees/euler` covers 15 chains. hyperevm is not one of them, but two of the three config maps in `fees/euler/config.ts` already carry it:

```ts
const feeFlowControllers = { ..., [CHAIN.HYPERLIQUID]: "0x8916311B5E8056E0709163c52a51831A0f152b44", ... }
const tokenEUL          = { ..., [CHAIN.HYPERLIQUID]: "0x3A41f426E55ECdE4BC734fA79ccE991b94aFf711", ... }
```

only `eVaultFactories` is missing an entry, and `EulerChainConfigs` is keyed off that, so the chain never runs. euler-v2's tvl adapter has had hyperevm since 2026-02-04 and reads **$2.79m** there today (`projects/euler-v2/index.js`: `hyperliquid: { factory: '0xcF5552580fD364cdBBFcB5Ae345f75674c59273A' }`).

that factory is live: `getProxyListLength()` returns **58** vaults at head.

harness with the factory added to the map:

```
2026-09-06   fees $265.00   revenue $13.20   supply side $249.00
2026-09-07   fees $264.00   revenue $12.91   supply side $240.00
```

$264/day on $2.79m of vault assets is a 3.5% implied apr, which is in band for the asset mix. small in absolute terms, but it is a whole chain of a $1.3m/30d protocol reporting nothing, and the fix is one line in a map that already has the chain in its two sibling maps. pr follows.

0xshubhs:
Fixes #9366

adds hyperevm to `eVaultFactories` and a matching `EulerChainConfigs` entry. `feeFlowControllers` and `tokenEUL` already had the chain, so this is the one map that was missing it; the factory address is the one euler-v2's tvl adapter already uses, and it answers `getProxyListLength()` with 58.

start is 2026-02-04, the first day the tvl adapter has hyperevm.

harness:

```
2026-09-06   fees $265.00   revenue $13.20   supply side $249.00
2026-09-07   fees $264.00   revenue $12.91   supply side $240.00
```

other chains untouched.

### DIFF
--- a/fees/euler/config.ts
+++ b/fees/euler/config.ts
@@ -25,6 +25,7 @@ const eVaultFactories: Record<string, string> = {
   [CHAIN.PLASMA]: "0x42388213C6F56D7E1477632b58Ae6Bba9adeEeA3",
   [CHAIN.MANTLE]: "0x47Aaf2f062aa1D55AFa602f5C9597588f71E2d76",
   [CHAIN.MONAD]: "0xba4Dd672062dE8FeeDb665DD4410658864483f1E",
+  [CHAIN.HYPERLIQUID]: "0xcF5552580fD364cdBBFcB5Ae345f75674c59273A",
 };
 
 const feeFlowControllers: Record<string, string> = {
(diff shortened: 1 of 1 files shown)


0xshubhs:
the red `test` is bsc, which this diff does not touch: `Llama RPC error! method: call - host: https://bsc-dataseed.binance.org error: missing trie node`, i.e. the public bsc nodes refusing archive state for the window. the workflow runs `fees euler` across all 16 chains and aborts on the first one that fails.

the hyperevm leg this pr adds runs on its own:

```
npx ts-node cli/testAdapter.ts fees euler 2026-09-06 hyperliquid   ->  fees $265.00  revenue $13.20
npx ts-node cli/testAdapter.ts fees euler 2026-09-07 hyperliquid   ->  fees $264.00  revenue $12.91
```

bheluga:
@0xshubhs thanks for the PR

PR #9371, opened 2026-09-09
fluid: skip the solana leg when the juplend view has no rows, instead of booking zero

### LINKED ISSUE #9370: fluid: the solana leg books $0 when the juplend dune view returns no rows, while jupiter-lend throws on the same rows (#9347)
sibling of #9347. that issue covers `fees/jupiter-lend` going dark on the fluid materialized view. `fees/fluid` reads **the same view** and fails differently, which is worse: it stores a zero.

```ts
// fees/fluid/index.ts, fetchSolana
from dune."0xfluid".result_juplend_historical_tvl_by_token_mv
...
const data: any[] = await queryDuneSql(options, sql);
const jupiterRevenue = data.reduce((sum, row) => sum + (row.daily_revenue_usd || 0), 0);
```

there is no empty-rows check, so `[].reduce(..., 0)` is 0 and the day publishes as zero.

## what it published

fluid's solana leg, from `api.llama.fi/summary/fees/fluid` (per-chain breakdown):

```
08-29 $899 · 08-30 $1,598 · 08-31 $1,418 · 09-01 $1,435 · 09-02 $1,784
09-03 $986 · 09-04 $1,463 · 09-05 $1,133 · 09-06 $1,027
09-07 $0   · 09-08 $0
```

on those same two days ethereum ($80,755 / $81,331), arbitrum, base, polygon and plasma all published full days, so nothing about the protocol looks broken — the zero is hidden inside a healthy six-chain total. jupiter-lend, reading the same rows, went visibly dark instead, because it has the check fluid is missing:

```ts
if (!data || data.length === 0) { throw new Error(`No JupLend data returned from Dune ...`) }
```

so the same missing rows produce a refillable gap on one adapter and a permanent wrong zero on the other. ~$1,300/day, about $2,600 so far and still accruing.

## the obvious fix is unsafe here, which is probably why it was skipped

`assertDuneSolanaIndexed` (the #9221 guard, and the one i proposed for jupiter-lend in #9348) throws. its own docstring says why that is not usable in this file:

> Only safe on single-chain adapters: `runAdapter` drops every chain's record when one chain throws.

fluid has six chains. confirmed locally on master — the solana leg's dune call is the first thing that fails and it takes the whole adapter with it:

```
$ npx ts-node cli/testAdapter.ts fees fluid 2026-09-08
------ ERROR ------
Error: DUNE_API_KEYS environment variable is not set
```

no chain reports. so a throw here would trade a $1,300/day wrong zero for blanking $108k/day of healthy chains.

## what the pr does instead

returns the metrics as `undefined` for the solana leg in both the mid-refresh and the empty-rows cases. `runAdapter` deletes undefined values before aggregating, so the leg contributes nothing and the other five chains are untouched — a missing day stays missing and can be refilled, rather than being stored as a zero.

verified by widening the wait constant so the guard fires on a settled day:

```
chain     | Daily fees | Daily revenue | Daily supply side revenue
ethereum  | 71.20 k    | 7.50 k        | 63.69 k
arbitrum  | 20.26 k    | 2.16 k        | 18.10 k
base      | 2.12 k     | 212.00        | 1.91 k
polygon   | 152.00     | 15.00         | 136.00
plasma    | 15.18 k    | 1.46 k        | 13.72 k
solana    |            |               |
Aggregate | 108.91 k   | 11.34 k       | 97.56 k
```

solana is blank rather than zero, and the aggregate reproduces what production published for 2026-09-08 ($108,906) to the dollar. with the real 10-hour constant the same command goes on to the dune call as before.

09-07 and 09-08 need a refill on the solana leg once the view catches up.

0xshubhs:
Fixes #9370. Sibling of #9347 / #9348, same Dune view, the other adapter that reads it.

`fetchSolana` had no empty-rows check, so when the JupLend materialized view returns nothing for the window, `[].reduce(..., 0)` books **$0** as the day's revenue share. 09-07 and 09-08 both published zero while Fluid's five EVM chains published full days.

`assertDuneSolanaIndexed` is not usable here; it throws, and its docstring says why that matters:

> Only safe on single-chain adapters: `runAdapter` drops every chain's record when one chain throws.

Fluid has six chains, so a throw would blank $108k/day of healthy legs to fix a $1.3k/day one. This returns the metrics as `undefined` instead; `runAdapter` deletes undefined values before aggregating, so the leg contributes nothing and the day stays refillable rather than being stored as a zero. The same guard covers the mid-refresh case that produced the partial in #9347.

## verification

With the wait constant temporarily widened so the guard fires on a settled day:

```
chain     | Daily fees | Daily revenue | Daily supply side revenue
ethereum  | 71.20 k    | 7.50 k        | 63.69 k
arbitrum  | 20.26 k    | 2.16 k        | 18.10 k
base      | 2.12 k     | 212.00        | 1.91 k
polygon   | 152.00     | 15.00         | 136.00
plasma    | 15.18 k    | 1.46 k        | 13.72 k
solana    |            |               |
Aggregate | 108.91 k   | 11.34 k       | 97.56 k
```

Solana is blank rather than zero, the other five chains are unaffected, and the aggregate matches what production published for 2026-09-08 ($108,906) to the dollar.

With the real 10-hour constant restored, a settled day goes on to the Dune call exactly as before:

```
$ npx ts-node cli/testAdapter.ts fees fluid 2026-09-09
Error: DUNE_API_KEYS environment variable is not set
```

`npx tsc --noEmit` is clean.

What I could not check: I have no Dune key, so the empty-rows branch is verified by reading rather than by executing the query. The 10-hour threshold is inherited from `helpers/duneSolanaDex.ts` rather than measured against this view, same caveat as #9348.

### DIFF
--- a/fees/fluid/index.ts
+++ b/fees/fluid/index.ts
@@ -27,6 +27,16 @@ const fetch: FetchV2 = async (options: FetchOptions) => {
   }
 }
 
+const DUNE_SOLANA_INDEXING_DELAY_MS = 10 * 60 * 60 * 1000;
+
+const SKIP_SOLANA_LEG = {
+  dailyFees: undefined,
+  dailyRevenue: undefined,
+  dailyProtocolRevenue: undefined,
+  dailyHoldersRevenue: undefined,
+  dailySupplySideRevenue: undefined,
+};
+
 // Revenu share from Jupiter Lend
 const fetchSolana: FetchV2 = async (options: FetchOptions) => {
   const dailyFees = options.createBalances();
@@ -49,7 +59,15 @@ const fetchSolana: FetchV2 = async (options: FetchOptions) => {
     group by 1
     order by day desc
   `
+  if (options.toTimestamp * 1000 > Date.now() - DUNE_SOLANA_INDEXING_DELAY_MS) {
+    return SKIP_SOLANA_LEG;
+  }
+
   const data: any[] = await queryDuneSql(options, sql);
+  if (!data || data.length === 0) {
+    return SKIP_SOLANA_LEG;
+  }
+
   const jupiterRevenue = data.reduce((sum, row) => sum + (row.daily_revenue_usd || 0), 0);
   dailyFees.addUSDValue(jupiterRevenue * 0.5, FLUID_METRICS.RevenueShareFromJupLend);
   


bheluga:
@0xshubhs thanks for the PR
Closing this for same reason as of sibling

PR #9373, opened 2026-09-09
[fix] accountable - read fee crystallization from FeeSharesMinted for v2+ credit strategies

pawel-accountable:
## Summary
- v2+ `OPEN_TERM`/`EARLY_EXIT`/`YIELD_STRATEGY_V1_1` strategies never leave a fee balance in the FeeManager — `performanceFee()`/`managementFee()` are just rates, and every crystallization mints vault shares straight to the recipient (`FeeSharesMinted`), same mechanism already used for NAV/looping vaults in this adapter.
- Credit vaults (the ones exposing `scaleFactor()`) were still using a notional interest/AUM-based estimate for this, which (a) never picked up a management-fee-only config (`performanceFee=0`) as revenue at all, and (b) capped any crystallization at that day's own accrued interest, silently dropping real value when a mint settles a longer backlog in one shot.
- Fix: gate by `version()` (mirrors `backend_new`'s `ManagerFeeService._SHARE_FEE_MIN_VERSION`); for `>=2`, sum `FeeSharesMinted` for the window, convert to assets via `convertToAssets`, and attribute manager vs protocol by matching the mint's recipient against the FeeManager's own `treasury()` — not a split ratio, which can change in the same transaction as the mint.
- Mints that share a transaction with an `Upgraded` event on the same contract are excluded (keyed by tx hash, not date) — see below for why.
- Legacy (v1.0) strategies are untouched: `managementFee()` has always read `null` there, so the removed notional management-fee term was already always zero for them.

## Why the `Upgraded` exclusion

On monad, aHYPER's FeeManager switched from a performanceFee-only config to management-fee-only on 2026-09-08, bundled in a Safe multisig `execTransaction` that also upgraded the strategy's own implementation (`Upgraded` + `Initialized` v3/v4 in the same tx). That transaction minted a one-off 400,699.82 / 133,566.61 share settlement — months of previously-accrued, not-yet-crystallized performance fee that was already being recognized day-by-day under the prior notional formula. Booking it again as fresh revenue on the settlement day would double-count it. Excluding by transaction hash (not by date) means an ordinary mint that happens to land later the same day still counts normally.

## Test plan
- [x] `npm run ts-check` clean.
- [x] `npm test fees accountable` clean on monad and ethereum, before and after the aHYPER switch date.
- [x] Cross-checked the computed manager/protocol split against raw `FeeSharesMinted` logs and the Safe transaction's decoded events (recipient-by-recipient) — matches to within rounding.
- [x] Confirmed the pre-switch day correctly falls back to the legacy path (real `version()` was <2 before the upgrade tx bumped it), reproducing the prior ~20% performance-fee split with no regression.
- [x] Confirmed other `version()>=2` credit vaults on monad (e.g. Valos) that already mint fee shares are picked up by the same generic path, with stable day-over-day numbers unrelated to the aHYPER migration.
- [ ] Other chains (arbitrum, robinhood, citrea) not individually re-verified beyond `ts-check`; the change is chain-agnostic and gated purely by on-chain `version()`.

### DIFF
--- a/fees/accountable.ts
+++ b/fees/accountable.ts
@@ -63,10 +63,12 @@ const config: Record<string, { factories: string[], start: string }> = {
 
 // aHYPER Looping Vault (vault 0x23b148d8f389C5821739381f1FF87bB7e1162566) is
 // EOA-deployed rather than created by a registered factory, so enumeration
-// cannot reach it. The TVL adapter carries the same exception.
-const EXTRA_STRATEGIES: Record<string, { address: string, deployedOn: string }[]> = {
-  [CHAIN.MONAD]: [{ address: "0xE19b272b2fe4a54103A41F9B1c65dB3D2F6d886D", deployedOn: "2026-04-28" }],
-};
+// cannot reach it. Excluded from fee tracking: it's a hidden, still-in-testing
+// vault whose configured 10% performanceFee/50% managerSplit is not actually
+// being charged, so booking fees off its share-price growth would overstate
+// protocol revenue. Re-add once fee collection is live for it. The TVL
+// adapter carries the EOA-deployed exception separately and is unaffected.
+const EXTRA_STRATEGIES: Record<string, { address: string, deployedOn: string }[]> = {};
 
 const abis = {
   getStrategiesCount: "function getStrategiesCount() view returns (uint256)",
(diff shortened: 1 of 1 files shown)


pawel-accountable:
Good catch — fixed in 6064927: gross now stays `a.interest` unconditionally, so depositors go negative on a big crystallization instead of being floored at zero. Thanks!

bheluga:

PR #9298, opened 2026-09-07
hyperliquid: decompress builder fills in memory, every builder shares one temp file

0xshubhs:
Fixes #9297.

`fetchBuilderCodeRevenue` downloads each builder's `builder_fills` csv to
`helpers/temp/${dateStr}.csv.lz4` and unlinks it in `finally`. The path is keyed on the date
only, so every builder that runs for the same day writes a different url into the same file and
deletes it on the way out. Four builders at once on master:

```
$ for m in whale-ag sogo-terminal signalview outcome-xyz; do
    npx ts-node cli/testAdapter.ts dexs $m 2026-09-05 & done

whale-ag     Error: ENOENT: no such file or directory, open '.../helpers/temp/20260904.csv.lz4'
outcome-xyz  Error: ENOENT: no such file or directory, open '.../helpers/temp/20260904.csv.lz4'
```

Both pass when run on their own. A partial write is the worse case, it decompresses to a
truncated csv and publishes a short day rather than failing.

This drops the temp directory: `responseType: "arraybuffer"` and `decompressFrame` on the buffer,
so nothing touches the filesystem and there is no shared path to race on. The parse loop is
unchanged, only de-indented out of the `try` that no longer has anything to clean up (the
`catch (error) { throw error }` was a no-op).

Same four, same day, with this branch:

```
whale-ag       Daily volume: 13.56 k
outcome-xyz    Daily volume: 2.21 M
stratium       Daily volume: 0.00
sogo-terminal  Error: Builder fee data is not available for 20260904
signalview     Error: Builder fee data is not available for 20260904
```

The two errors are the source: those builders have no file for that date, `curl` returns 403 for
both addresses. Single runs are unchanged; outcome-xyz reads 2.21 M for 09-04 before and after,
against 2,210,493 summed straight out of the csv. `npx tsc --noEmit` is clean.

Whether this is the whole of #9297 I cannot say from outside; the hip4 builders are the only ones
on this code path in production, and they have been dark since 09-04, but I cannot see the runner.
The race is a bug on its own.

PR #9383, opened 2026-09-10
[fix] liquary - track both builder addresses from 2026-06-01, treat a closed day with no builder_fills file as empty

liquary-xyz:
Liquary (adapter id `hypersight`) used two builder addresses, and only the second one was listed:

| address | builder code |
|---|---|
| `0x9bdf1a9a2e8b5353d19b2e4978e32d9873da1552` | 2026-06-01 → 2026-06-30 |
| `0xc9200c6d0e876d5f63c76618f2c57e5d6a080927` | since 2026-07-01 |

**Changes**

- `factory/hyperliquid.ts`: both addresses, `start: "2026-06-01"`.
- `helpers/hyperliquid.ts`: on the `builder_fills` file path, a 403 for a day **closed more than 48h ago** returns an empty day instead of throwing. HL answers 403 when a builder had no fills that day, and with two addresses one of them is idle every day, so without this each day fails as a whole. A more recent 403 still throws, so a day whose file isn't written yet is retried.

In production an `all`-markets builder goes through `LLAMA_HL_INDEXER` and never reads the files; the 403 change only matters on the file path (local runs, and any fallback).

**Tested** with `cli/testAdapter.ts`, no `LLAMA_HL_INDEXER` set:

```
fees hypersight 2026-06-11   -> volume 5.17 k · fees 0.17   (first address)
fees hypersight 2026-07-02   -> volume 11.00  · fees 0.01   (second address)
fees hypersight 2026-09-06   -> volume 0.00   · fees 0.00   (403, no throw)
fees outcome-xyz 2026-09-02  -> volume 2.91 M · fees 8.74   (unchanged vs master)
```

Without the helper change, the first two throw `Builder fee data is not available`.

**Refill request:** Liquary (`hypersight`) from **2026-06-01**.

_Scope reduced on 2026-09-17: the earlier `builderFills` option was dropped (HL's `builder_fills` files are currently truncated around 12:10 UTC on most days, see the thread), as was the UTC date change, which has no effect on this adapter in production._


liquary-xyz:
Ran `cli/testAdapter.ts fees hypersight <end-date>` locally after the change (no `LLAMA_HL_INDEXER` set, so the file path):

```
fees hypersight 2026-09-02   (window Sep 1 → Sep 2)
Daily volume: 26.70 k
Daily fees: 0.00          ← 22 fills, every one at builder_fee 0

fees hypersight 2026-09-06   (window Sep 5 → Sep 6, HL answers 403 — no fills that day)
Daily volume: 0.00
Daily fees: 0.00          ← empty day, no throw
```

Before the change the second run threw `Builder fee data is not available for 20260905…`.

liquary-xyz:
Both review points applied in `782fc0a` — thanks, the first one was a real bug in my patch.

**48h window measured from the day's close.** `startTimestamp` is the *start* of the requested day, so the threshold fired 24h early: a Sep 8 file would have stopped being retried during Sep 10, before HL had necessarily written it. Now `startTimestamp + 86400`.

**Per-call temp directory.** `temp/${dateStr}.csv.lz4` is keyed on the date alone, so two builders downloading the same day concurrently write and unlink each other's file. Pre-existing (reachable through the hip3/hip4 builders today), but `builderFills` widens it to all-markets builders, so it is fixed here: `fs.mkdtempSync` per call, removed in the same `finally`.

Re-ran after the change:

```
fees hypersight 2026-09-02   → Daily volume 26.70 k · Daily fees 0.00
fees hypersight 2026-09-06   → Daily volume 0.00    · Daily fees 0.00   (403, no fills that day, no throw)
```

liquary-xyz:
One more in `4b1b0a3`: **Liquary (`hypersight` adapter id) tracks both builder addresses now, from 2026-06-01.**

The builder code moved from the team wallet to a dedicated one on 2026-06-30, and only the second address was listed. HL has files for the first from 2026-06-01 to 06-28 - **$147,164 of volume and $66.03 of fees** the adapter could already read and did not, because the address was absent and `start` was `2026-07-01`.

`addresses` is an array and each is fetched independently, so a day where one of them has no file is simply empty for it - which is exactly what the 403 handling in this PR makes safe.

```
fees hypersight 2026-07-02  -> Daily volume 11.00   · fees 0.01   (matches the file to the cent)
fees hypersight 2026-06-11  -> Daily volume 5.17 k  · fees 0.17   (returned nothing before this commit)
```

So the refill ask becomes: **refill Liquary (`hypersight` adapter id) from 2026-06-01**.

liquary-xyz:
Hi @noateden, CI is green and CodeRabbit approved this now. Could you take a look when you have a chance? The refill request is: refill Liquary (`hypersight` adapter id) from 2026-06-01. Thanks.

bheluga:
@liquary-xyz thanks for the PR
I have informed relevant dev about missing 0 fee volumes

liquary-xyz:
Rebased on `master` after #9298 — thanks @0xshubhs, that removes half of this PR.

**Dropped:** the two commits here that gave each call its own temp directory. #9298 deletes the temp file entirely (`arraybuffer` + `decompressFrame` in memory), so the collision they fixed cannot happen any more. Nothing of them is left in the diff.

**Still needed on the new shape** (the three things #9298 does not address):

1. `builderFills` — an all-markets builder reads HL's `builder_fills` files instead of the indexer. The two sources disagree on zero-fee fills: a builder attaching its code at `f = 0` still appears in the files with `builder_fee` 0, while the indexer only carries fee-paying activity, so its volume reads $0 for the length of a fee holiday. `hip3`/`hip4` builders already take the file path; this is the same door for `market: "all"`, opt-in, default off.
2. A 403 on a day **closed more than 48h ago** is an empty day, not a throw. HL answers 403 both for "no fills that day" and for "not written yet"; only age tells them apart.
3. Liquary tracks **both** builder addresses from 2026-06-01 — the code moved to a dedicated wallet on 2026-06-30, and HL has files for the old one from 06-01 to 06-28.

**One more, which is CodeRabbit's open point and the reason it flagged merge risk.** `dateStr` was built with local-time getters while the indexer branch of the same function uses `toISOString()`. `startOfDay` is a UTC boundary, so on any host west of UTC the two halves of `fetchBuilderCodeRevenue` ask for different days, and the file path reads yesterday's fills. Now `toISOString()` on both sides. This one touches the shared path, so I checked it does not move anything: `outcome-xyz 2026-09-02` returns `$2.91M / $8.74` on this branch and `$2.91M / $8.74` on `master`, to the cent.

Re-ran everything after the merge (no `LLAMA_HL_INDEXER` set):

```
hypersight 2026-09-02  -> volume 26.70 k · fees 0.00     (22 fills, every one at builder_fee 0)
hypersight 2026-09-06  -> volume 0.00    · fees 0.00     (403, no fills that day, no throw)
hypersight 2026-06-11  -> volume 5.17 k  · fees 0.17     (old address)
hypersight 2026-07-02  -> volume 11.00   · fees 0.01     (new address)
outcome-xyz 2026-09-02 -> volume 2.91 M  · fees 8.74     (hip4, unchanged vs master)
```

On `master` as it stands, `hypersight 2026-09-06` still throws `Builder fee data is not available for 20260905`, so point 2 is not covered by #9298.

`tsc --project tsconfig.json` clean.

@bheluga thanks for passing it on. The refill ask is unchanged: **refill Liquary (`hypersight` adapter id) from 2026-06-01**.

liquary-xyz:
@0xshubhs this sits on `helpers/hyperliquid.ts`, which you rewrote in #9298 — rebased on top of it, still mergeable after today's three commits to that file, all checks green, CodeRabbit at minimal risk.

Three things #9298 does not cover:

1. an opt-in `builderFills` so an all-markets builder reads HL's own `builder_fills` files — zero-fee fills are invisible to the indexer, which is why Liquary has read $0 since 2026-08-07 despite trading;
2. a 403 on a day closed more than 48h ago treated as an empty day instead of a throw (on master, `hypersight 2026-09-06` still throws `Builder fee data is not available for 20260905`);
3. both builder addresses from 2026-06-01 — the code moved to a dedicated wallet on 2026-06-30 and only the second one was listed.

Plus the UTC date fix CodeRabbit raised: the file path was built from local-time getters while the indexer branch of the same function uses `toISOString()`. That one touches the shared path, so I checked it moves nothing — `outcome-xyz 2026-09-02` gives $2.91M / $8.74 on this branch and $2.91M / $8.74 on master, to the cent.

Would you take a look when you have a moment? Thanks.

noateden:
hi @liquary-xyz thanks for this PR, I'm working on a fix and refilling data too

liquary-xyz:
Thanks @noateden! Confirming the fix works for perps: DefiLlama shows $1,748 for Liquary on 2026-09-16, which matches our own records to the cent (HIP-3 `xyz:GOLD` $1,557.45 + `xyz:CBRS` $190.75).

A few things left:

**1. HIP-4 volume isn't counted.** The same day, $121.79 of prediction-market volume went through our builder code (`#12150`, `#12151`, 7 fills) and doesn't appear on the page. Prediction markets are a large part of what Liquary does.

**2. Backfill, and our first builder address.** Nothing between 2026-08-08 and 2026-09-15 yet. For the refill to also cover June, the config needs our first builder address, `0x9bdf1a9a2e8b5353d19b2e4978e32d9873da1552` (used until 2026-06-30), next to `0xc9200c6d0e876d5f63c76618f2c57e5d6a080927` (since 2026-07-01), with `start: "2026-06-01"`.

**3. Heads-up, this one affects more than us: since 2026-09-08, HL's `builder_fills` files stop around 12:10 UTC on most days.** The files are written after the day closes (the 09-16 file is dated 09-17 00:44 UTC), so this isn't a publishing delay. Last fill in each file:

| builder | 09-08 | 09-09 | 09-10 | 09-11 | 09-14 | 09-15 | 09-16 |
|---|---|---|---|---|---|---|---|
| Phantom | 12:10 | 23:59 | 23:59 | 12:10 | 12:08 | 12:09 | 12:09 |
| MetaMask | 12:09 | 23:59 | 12:10 | 12:10 | 12:07 | 12:08 | 12:08 |
| Outcome.xyz | 12:09 | 23:59 | 23:59 | 12:09 | 12:08 | 12:09 | 12:08 |

Adapters on the file path store those half-days as they are. Outcome.xyz (`market: "hip4"`) daily volume on DefiLlama matches the truncated files almost exactly:

| day | volume in HL file (cut at ~12:09) | DefiLlama |
|---|---|---|
| 09-11 | 470,092 | 470,038 |
| 09-14 | 408,800 | 408,760 |
| 09-15 | 1,425,445 | 1,425,032 |
| 09-16 | 1,365,119 | 1,364,677 |

(The complete days, 09-09, 09-10 and 09-13, don't appear in that series at all, which may be a separate issue.)

We see the same thing on our side: the files match our records to the cent up to 09-09, then miss everything after the cutoff ($34.8k on 09-15 alone). It also means the `builderFills` option in this PR would read truncated data, so your indexer route is the right one. I'll drop that part and cut this PR down to the address change, unless you'd rather include it in your fix.

liquary-xyz:
Good catch, fixed in `faef441` — the 48h rule genuinely did not cover a recent day: for yesterday and today the idle address answers 403 and the whole day failed.

I fixed it in the loop rather than by scoping each address to its own period. Per-address periods would need a new shape for `BuilderConfig` and for `exportBuilderAdapter`'s signature, which every builder in the file shares, for a case only multi-address builders have. In the loop the rule states itself: a builder listed under several addresses has one code per period, so **every** day is a day where the others were idle, and a 403 for one of them is not an error for the day as long as another answered. When all of them fail, the first error is rethrown, so a single-address builder behaves exactly as before.

Re-ran with `cli/testAdapter.ts`, no `LLAMA_HL_INDEXER`:

```
hypersight 2026-09-17   -> 68.00          (was: Builder fee data is not available for 20260916)
hypersight 2026-06-11   -> 5.17k / 0.17   (first address)
hypersight 2026-07-02   -> 11.00 / 0.01   (second address)
hypersight 2026-09-06   -> 0 / 0          (403 on both, no throw)
outcome-xyz 2026-09-02  -> 2.91M / 8.74   (single address, unchanged vs master)
outcome-xyz 2026-09-18  -> still throws   (single address, no file for that day)
```

The last line is the one that matters for the rest of the file: nothing is swallowed for a builder with one address.

liquary-xyz:
@coderabbitai review

liquary-xyz:
Right again, fixed in `884128b`. Swallowing a timeout or a 5xx would publish a day read from one address as if it were whole, which is the worst kind of wrong here: silent and plausible.

The 403 that means "HL has no file for this address on this day" now carries a `builderFillsFileMissing` flag from the point where it is thrown, and that flag is the only thing the loop survives. Everything else propagates. I flagged it at the source rather than matching on the message, so rewording the string later cannot silently widen what gets tolerated.

Re-ran, no `LLAMA_HL_INDEXER`:

```
hypersight 2026-09-18   -> 56.00    (idle address 403s, day still served)
hypersight 2026-06-11   -> 5.17k    (first address)
hypersight 2026-07-02   -> 11.00    (second address)
hypersight 2026-09-06   -> 0        (403 on both, no throw)
outcome-xyz 2026-09-02  -> 2.91M    (single address, unchanged vs master)
outcome-xyz, a day with no file -> still throws
```

And with the file host pointed at an unresolvable name, so one address fails for a reason that is not a missing file:

```
hypersight 2026-06-11 -> Error: Failed to download builder fee data: getaddrinfo ENOTFOUND ...
```

which is the case your comment describes: it throws instead of returning the other address on its own.

PR #9410, opened 2026-09-12
Track Valhalla gross DLMM LP fees

Prithwiraj-CK:
## Summary
- Reports gross Meteora DLMM LP trading fees claimed by Valhalla-managed positions from decoded on-chain claim records, linked by transaction ID to Valhalla's public management-fee transfer.
- Reports the matching native-SOL management-fee receipt as revenue and keeps paired referral payouts in supply-side revenue.
- Does not infer gross fees from a fixed commission rate.

## Scope
- DLMM claim and close transactions are covered by the decoded claim records.
- DAMM v2 is intentionally excluded for now: its remove-liquidity path does not separately expose accrued LP fees from returned principal, so counting withdrawals as fees would overstate results.

## Validation
- `pnpm exec tsc --noEmit --pretty false`
- Confirmed the public DLMM claim and close transaction flows contain Meteora's claim instruction.
- No private application data, user lists, credentials, or source code are used.

PR #9415, opened 2026-09-12
Report Valhalla DLMM fees only

Prithwiraj-CK:
Follow-up to #9410: expose only the gross fees metric requested for Valhalla.

- Retains the verified Meteora DLMM claim-based fee calculation.
- Removes revenue, protocol-revenue, supply-side-revenue, and referral outputs.
- Treasury transfers remain only as public on-chain attribution for Valhalla-managed claims.

DAMM v2 remains excluded because its close flow does not separately expose accrued fees from returned principal.


bheluga (commented):
@Prithwiraj-CK thanks for the PR
why do you want to keep it fees only?

bheluga:
Closing the PR as it can be discussed [here](https://github.com/DefiLlama/dimension-adapters/pull/9425)

PR #9412, opened 2026-09-12
Add HASHCATS fees, active-users and new-users adapters

hashcats-dot-fun:
Adds fees, active-users and new-users adapters for HASHCATS, a proof-of-work NFT collection on Robinhood Chain. The TVL adapter is submitted separately in DefiLlama/DefiLlama-Adapters.

**Where the money comes from**

1. **Mints.** A cat is minted only when a miner submits a hash below the collection's target. The price is derived from the contract's own curve — the number of cats created so far times `PRICE_STEP` — and is split on the spot between rent to the cats already alive and the hook. The whole price is a fee: there is no third-party seller, no allowlist and no team allocation, and every wei stays inside the protocol. The price is derived per cat rather than read from the transaction value, so a mint sent with excess ether cannot inflate it.

2. **Trading fee** on the $HASH pool, taken by the hook. The pool's own fee is zero, so this is entirely the protocol's. The rate decays from a high opening value to its target over the first ten minutes of trading — a launch defense that arms once — which is why amounts are read from `FeeTaken` rather than computed from a rate.

3. **Secondary-market royalties.** The collection names the hook as the ERC-2981 recipient, so royalties arrive there directly, as ether or as WETH depending on how the sale settled.

**Classification**

| Field | What it is |
|---|---|
| `dailyFees` | Mint prices + trading fees + royalties |
| `dailySupplySideRevenue` | Rent to living cats. Cat holders are suppliers whose asset earns, not governance-token holders, so their cut is a cost of revenue |
| `dailyRevenue` | Fees minus rent |
| `dailyHoldersRevenue` | The queue's share of revenue, spent buying $HASH back and burning it — measured at the fee source, where the split happens |
| `dailyProtocolRevenue` | The developer's share, at the hook's `devBps` rate read from the chain |

**Not counted:** $HASH handed out when a cat is burned. The holder destroys an asset that was earning rent and receives another in exchange — a conversion, not an incentive, and nobody pays a fee.

**Not submitted:** a volume adapter. The $HASH pool is a Uniswap V4 pool and `dexs/uniswap-v4.ts` already covers that PoolManager, so reporting the volume here would double it at the aggregate — the same reasoning `dexs/hookos.ts` gives for its own pool on this chain. The fee the hook skims is a different matter and is reported above, since no Uniswap adapter sees it.

**Two details worth flagging**

- The hook's own buybacks pay no trading fee: Uniswap does not invoke hooks on swaps the hook itself sends, so they cannot inflate the fee figures.
[...]

PR #9418, opened 2026-09-12
track Hashcats fees and users

bheluga:
https://github.com/DefiLlama/dimension-adapters/pull/9412

### DIFF
--- a/active-users/hashcats.ts
+++ b/active-users/hashcats.ts
@@ -61,7 +61,8 @@ const adapter: SimpleAdapter = {
   version: 1,
   fetch,
   chains: [CHAIN.ROBINHOOD],
-  // Deployment block of the collection, 60412000: the first cat is mined in it.
+  // The collection's code appears in block 60412470, and the first cat is mined
+  // a few minutes later in block 60415844.
   start: "2026-09-11",
   methodology:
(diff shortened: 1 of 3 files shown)


hashcats-dot-fun:
Thanks for the quick merge. The protocol entry exists and the page renders, but the fees are not attached to it yet:

```
/summary/fees/hashcats        -> 400  "Fees for hashcats not found"
/summary/fees/ebisus-bay-dex  -> 200  total24h 3.43
```

Ebisus Bay was registered 19 minutes after HASHCATS (15:58 vs 15:39 UTC) and already reports, so this looks like a wiring step that was missed rather than something still propagating. defillama.com/protocol/hashcats loads, with the description and category, but no numbers.

Two other small things while you are in there:

- The website is missing from the protocol entry — https://hashcats.fun. The twitter came through, so this one just seems to have been dropped.
- #9419 fixes two things in `fees/hashcats` that changed during the move: the guard for windows that end before the collection exists (block 60412470 — without it the first day's hourly windows revert on `PRICE_EPOCH0`), and the height the `hookDue` read targets (`ChainApi.call` ignores a per-call `block`, so it was reading the end of the window instead of the start).

A hallmark on 11 Sep would also help whenever you get a chance — the first hour carries a launch-defence fee that decays from 50% to 2.5% over ten minutes, so that spike is the launch rather than an anomaly.

PR #9424, opened 2026-09-12
feat(agnt): count AGNTS NFT mint revenue on Robinhood Chain

tut9492:
Adds AGNT's NFT mint revenue to the existing `fees/agnt` adapter.

## What's missing today

AGNT mints its own PFP collection — **AGNTS**, `0x57efd86cf3ab1812cb47f0fd3c3d5bf0efb1f2d5` on Robinhood Chain (4663) — through a mint controller that escrows each buyer's payment.

Buyers pay **native ETH**, and AGNT keeps **100%**: it's a first-party collection, so there's no creator or protocol split and nothing is supply-side. None of it was counted, so the adapter understates AGNT's revenue by roughly the size of everything else it currently reports.

## How it's measured

From the controller's **`ProceedsWithdrawn(address indexed to, uint256 amount)`** event, not a per-mint event.

Payment and delivery are separate steps in this contract — a buyer pays and the request settles, then a vault transfer delivers the NFT — so no single per-buyer event carries the amount across every path. The withdrawal is exact, is the money that actually reaches the treasury, and can't double-count.

It is **lumpy**: revenue lands on the day proceeds are swept rather than the day of the mint. That's the same behaviour the existing Pons leg already has, where fees realize when the keeper pokes.

Day-windowed like the Pons splits (no `fromBlock`) — passing the deploy block would re-count every withdrawal ever made, on every run.

## Accounting

- `dailyFees` += amount (`NFT Mint Fees`)
- `dailyRevenue` / `dailyProtocolRevenue` += amount (`NFT Mint Fees to Protocol`)
- `dailySupplySideRevenue` — unchanged; a first-party mint has no supply side

Methodology and both breakdown maps updated to match.

## Verification

Read directly from Robinhood Chain against the live contract:

- 2 `ProceedsWithdrawn` events totalling **6.282 ETH**
- Matches ETH paid into the controller (**6.278 ETH** across 2,420 buyer payments) within the rounding of a final zero-value sweep
- 3,079 NFTs delivered to 1,855 distinct wallets over the same period

Existing legs (Doppler launchpad, Relay trading fees, Pons wedge splits) are untouched.

### DIFF
--- a/fees/agnt/index.ts
+++ b/fees/agnt/index.ts
@@ -87,6 +87,23 @@ const WEDGE_DEPLOYED_ABI = "event WedgeDeployed(bytes32 indexed launchId, addres
 const WEDGE_SPLIT_ABI = "event Split(address indexed currency, uint256 total, uint256 toAgnt, uint256 forwarded)";
 const NATIVE = "0x0000000000000000000000000000000000000000";
 
+// ── AGNTS NFT mint (Robinhood) ──────────────────────────────────────────────
+// AGNT's own PFP collection (AGNTS, 0x57efd86c…f2d5) mints through a controller
+// that escrows each buyer's payment and credits it to `proceeds` once the request
+// settles. Buyers pay native ETH; AGNT keeps 100% — there is no creator or
+// protocol split on a first-party mint, so mint fees == mint revenue and nothing
+// is supply-side.
+//
+// Indexed from ProceedsWithdrawn rather than per-mint events: payment and delivery
+// are separate steps here (pay/settle, then a vault transfer delivers the NFT), so
+// no single per-buyer event carries the amount for every path. The withdrawal is
+// exact, is the money that actually reaches the treasury, and cannot double-count.
+// It is lumpy — revenue lands on the day the proceeds are swept, not the day of
+// the mint — which matches how the Pons leg above already behaves (fees realize
+// when the keeper pokes).
+const NFT_MINT_CONTROLLER = "0xfAD3335ba59c8e4438C2be47147080EF5809bF7e"; // AGNTMintController (4663)
+const PROCEEDS_WITHDRAWN_ABI = "event ProceedsWithdrawn(address indexed to, uint256 amount)";
+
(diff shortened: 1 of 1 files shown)


tut9492:
Flagging something about the CI output rather than leaving it to be discovered: **the new `NFT Mint Fees` lines do not appear in the bot's breakdown.**

I believe that's because the test window predates the only two qualifying events, not because the leg is broken — but I can't prove that from outside the runner, so here is the on-chain evidence to check it against.

**The events exist and decode with the ABI in this PR.** Contract `0xfAD3335ba59c8e4438C2be47147080EF5809bF7e` (Robinhood 4663), topic0 `0x0f2fb75cc1977a496e94837f859e957f68e26e70dc1b75d9945ee92ae57969ba`:

| amount | to | timestamp (UTC) |
|---|---|---|
| 6.282000 ETH | `0x5bF5805e…c5f0` | **2026-09-12 20:16:06** |
| 0.000000 ETH | `0x5bF5805e…c5f0` | 2026-09-12 20:16:28 |

Those are the **only** `ProceedsWithdrawn` events in the contract's history — the mint sold out in a single day and was swept once, a couple of hours before this PR was opened. On any window that closes before 20:16 UTC today, this leg is legitimately zero.

**This does expose a real property of the leg that's worth a reviewer's opinion:** because it indexes the *withdrawal* rather than the mint, it reports $0 on most days and the full amount on sweep days. I noted that in the PR body as matching the existing Pons behaviour (fees realize when the keeper pokes), but it is lumpier here — one sweep so far, not a recurring cadence.

The alternative would be indexing buyer payments at mint time, which dates the revenue correctly. I did not do that because payment and delivery are decoupled in this contract and the individual payments emit no event — 2,420 buyer payments totalling 6.278 ETH produced only 322 logs on the controller, none of them per-payment. If a maintainer prefers correct dating over event-derivability, say so and I'll look again at whether a balance-delta approach is acceptable here.

Happy to adjust either way — I'd rather this be right than merged.

bheluga:
@tut9492 thanks for the PR

PR #9425, opened 2026-09-13
Remove Valhalla fees adapter

Prithwiraj-CK:
Removing this adapter due to blockchain data inconsistency.


bheluga:
@Prithwiraj-CK thanks for the PR
Can we fix instead of removing. Can you please explain the exact issue?

Prithwiraj-CK:
> @Prithwiraj-CK thanks for the PR Can we fix instead of removing. Can you please explain the exact issue?

Hi I don't think here's any way to fix this, sorry. I was looking at how our competition, lpagent, did this. They just recorded their raw revenue and then multiplied it to get the fees they earn their users in meteora. Ours is a lot more complicated since we give users discounts based on certain subscriptions they may have with us. So there's no way to multiply that to get the meteora fees we have. So we'd rather just remove it for now. Thanks

bheluga:
> > @Prithwiraj-CK thanks for the PR Can we fix instead of removing. Can you please explain the exact issue?
> 
> Hi I don't think here's any way to fix this, sorry. I was looking at how our competition, lpagent, did this. They just recorded their raw revenue and then multiplied it to get the fees they earn their users in meteora. Ours is a lot more complicated since we give users discounts based on certain subscriptions they may have with us. So there's no way to multiply that to get the meteora fees we have. So we'd rather just remove it for now. Thanks

Maybe we can just show gross fees for now? remove revenue and supply side revenue breakdown?

Prithwiraj-CK:
> > > @Prithwiraj-CK thanks for the PR Can we fix instead of removing. Can you please explain the exact issue?
> > 
> > 
> > Hi I don't think here's any way to fix this, sorry. I was looking at how our competition, lpagent, did this. They just recorded their raw revenue and then multiplied it to get the fees they earn their users in meteora. Ours is a lot more complicated since we give users discounts based on certain subscriptions they may have with us. So there's no way to multiply that to get the meteora fees we have. So we'd rather just remove it for now. Thanks
> 
> Maybe we can just show gross fees for now? remove revenue and supply side revenue breakdown?

yes pretty much only need the gross fees yes ,remove revenue etc

bheluga:
@Prithwiraj-CK thanks for the PR

PR #9431, opened 2026-09-14
[fix] zERC20 - correct zJPYC LiquidityManager address checksum

dakakoe:
The zJPYC LiquidityManager address in `fees/zerc20/index.ts` was written with an invalid EIP-55 checksum (from #9394, mine). Correct casing:

`0x12609C3a7A1A212953417c90472cDCF034965A1c` → `0x12609C3A7a1A212953417c90472cdcf034965a1C`

No behaviour change — addresses are lowercased before use, and the live numbers are correct — but `ethers.getAddress` rejects the old string, so any strict-validation path would throw.

`npm test fees zerc20` after the change: 51.23 total (bsc 26.77, polygon 15.16, base 8.80, klaytn 0.50), matching the live API.


bheluga:
@dakakoe thanks for the PR

PR #9432, opened 2026-09-14
Add Hot Take (Base) volume + fees adapter

somnus0x:
Adds a volume + fees adapter for **Hot Take**, a peer-to-peer betting protocol on Base.

**Website:** https://usehottake.app
**Twitter:** https://x.com/usehottake
**Chain:** Base · contract `0x64Be8f389E202a77b446C7E86B564F4122Cb5a66` · settles in native USDC
**Related:** TVL listing PR in DefiLlama/DefiLlama-Adapters — https://github.com/DefiLlama/DefiLlama-Adapters/pull/21016

### Methodology

**Volume** is USDC wagered: both stakes of a market, counted at the moment they lock into escrow. A market only reaches the chain once both sides are matched, so `MarketCreated` fires exactly once per bet and carries `creatorAmount` + `takerAmount`. This is betting volume, not deposits — funding a balance is never counted, only money committed to a bet.

Each dollar is counted once. The two stakes come from two different bettors, so the sum is the pot rather than the maker/taker double count an orderbook has to halve. A 5-vs-5 bet reads 10 USDC of volume, the same way a 10 USDC complete-set mint does elsewhere.

**Fees** come from the `Claimed` event's split, which carries `protocolFee` and `arbiterFee` (the judge's fee). Labelled accordingly: protocol fee → Revenue/ProtocolRevenue, judge fee → SupplySideRevenue (judges are the supply side here). HoldersRevenue is zero — there is no token.

### Pre-answering the usual review questions

- **Is a 0-fee day correct?** Yes, and every day is 0 right now. The on-chain protocol fee is **0 bps since deployment** — it has never been raised — and the judge fee is per-bet and optional (0 when the bet uses the house judge). The adapter reads both from the event, so it starts charting the moment either is turned on, with no code change.
- **Is a 0-volume day correct?** Yes. The protocol is young and low volume; some days genuinely have no matched bets. Verified days: 2026-09-03 = $420, 2026-09-04 = $300, 2026-09-08 = $0.
- **Does the fee-to-volume ratio make sense?** It is 0% while the fee switch is off, which is consistent with the fee being 0 bps on-chain rather than the adapter missing a fee source.
- **Why might numbers differ from the protocol's own dashboard?** The app shows per-bet stakes and user balances; it publishes no aggregate volume figure, so there is nothing to disagree with.
- **Did the fee rate or split ever change?** No.
- **Is a refill needed?** No — new adapter. `start: '2026-06-29'`, the day the contract was deployed (Base block 47959751, 04:54:09Z); the first market followed on 2026-06-30.

### Validation

```
pnpm test dexs hot-take 2026-07-01   → volume 4.00,   fees 0.00
pnpm test dexs hot-take 2026-09-03   → volume 420.00, fees 0.00
pnpm test dexs hot-take 2026-09-04   → volume 300.00, fees 0.00
pnpm run ts-check                     → clean
pnpm run ts-check-cli                 → clean
git diff --check                      → clean
```

Only `dexs/hot-take/index.ts` is added; no dependencies, no lockfile or shared-helper changes.


bheluga:
@somnus0x thanks for the PR

PR #9284, opened 2026-09-06
curve: add etherlink volume

0xshubhs:
Fixes #9283

adds etherlink to `dexs/curve` through the on-chain path, since `prices.curve.finance` has no etherlink entry. the three ng factories are the same addresses stable, monad, plasma and robinhood already use, so the block mirrors those (start 2025-08-02, the first day the tvl adapter has etherlink).

harness:

```
2026-09-03  volume $101.16k  fees $10.00
2026-09-04  volume $133.49k  fees $13.00   (two runs, identical)
```

independent reconstruction of 2026-09-04 from the pools' `TokenExchange` logs priced with coins.llama.fi lands at $133,119, 0.3% apart, the gap being one token coins.llama.fi has no price for. details in the issue.

PR #9435, opened 2026-09-14
curve: park the etherlink leg, its rpc caps getLogs at 500 blocks and the throw blanks all 18 chains

### LINKED ISSUE #9434: curve: all 18 chains dark since 09-12 — the etherlink leg I added in #9284 hits a 500-block getLogs cap and takes the whole adapter with it
`dexs/curve` and `fees/curve` have published nothing since 2026-09-11. All 18 chains stop on the
same day — the per-chain breakdown has no rows at all for 09-12 onward, not zeros.

This is my regression: #9284 added Etherlink to the adapter on 09-07, and that leg now throws, which
takes the other 17 chains down with it.

## what the leg does

```
$ ts-node cli/testAdapter.ts dexs curve 2026-09-13 etlk
------ ERROR ------
  Cannot request logs over more than 500 blocks
```

Etherlink's RPC caps `eth_getLogs` at **500 blocks per request** and the sdk asks for wider ranges.
The host itself is fine — `eth_blockNumber` against `https://node.mainnet.etherlink.com` returns
`0x32d5ca8` with HTTP 200 and a genuine Google Trust Services cert, so this is a range limit, not an
outage. The chunk size is sdk/chain-config side, so there is nothing the adapter can pass to fix it.

The leg was not broken on arrival — it published every day from the merge to the stop (09-07 $867,
09-08 $349,038, 09-09 $77, 09-10 $22,719, 09-11 $42,244) and the chain count went 17 → 18. Whatever
changed, the cap started biting on 09-12.

## the cost

curve is $605,376,664 of 7d volume, about $125M/day, and the fees side ($716,222/7d) is dark with
it. Both are recoverable — with the Etherlink leg disabled the full 19-chain run completes:

| day | full-adapter run | published |
|---|---|---|
| 2026-09-12 | **$50,760,000** volume / $13.14k fees | nothing |
| 2026-09-13 | **$26,830,000** volume / $21.40k fees | nothing |

(ethereum carries most of it — $48.53M and $23.56M respectively.)

## fix

PR follows: comment the Etherlink block out, the same way FANTOM is parked a few lines above it in
the same file (#9053). That restores the other 17 chains immediately. Etherlink can come back when
the getLogs range is configurable per chain or the chain has an RPC without the 500-block cap —
re-adding it as-is would just take the adapter down again.

0xshubhs:
Fixes #9434.

I added Etherlink to this adapter in #9284 and it has taken the whole thing down since 09-12. Its
RPC caps `eth_getLogs` at 500 blocks, the sdk asks for wider ranges, the leg throws, and the throw
blanks the other 17 chains:

```
$ ts-node cli/testAdapter.ts dexs curve 2026-09-13 etlk
  Cannot request logs over more than 500 blocks
```

The host is not down, `eth_blockNumber` returns `0x32d5ca8` / HTTP 200, so there is nothing to
repoint, and the chunk size is sdk-side rather than something this file can pass. Parking the block
is the same treatment FANTOM already has a few lines above.

## before / after, full 19-chain runs

| | master | this branch |
|---|---|---|
| `... dexs curve 2026-09-13` | `ERROR`, getLogs 500-block cap | **rc 0, aggregate $50.76M** |
| `... dexs curve 2026-09-14` | `ERROR`, same | **rc 0, aggregate $26.83M** |

Both of those days published nothing in production. ethereum carries most of it, $48.53M and
$23.56M.

Worth keeping in mind before Etherlink goes back in: the leg was not broken on arrival. It published
every day between the merge and the stop, 09-07 $867, 09-08 $349,038, 09-09 $77, 09-10 $22,719,
09-11 $42,244; so re-adding it unchanged will most likely repeat this. It needs either a per-chain
getLogs range the adapter can set, or an Etherlink RPC without the cap.


bheluga:
@0xshubhs thanks for the PR

PR #9437, opened 2026-09-14
stonx: judge the pool volume fetch on its failure ratio, one unservable pool should not drop the day

### LINKED ISSUE #9436: stonx: dark since 09-11 — 2 of 55 ve33 pools return 400 'Invalid token identifier' and the first error throws away the whole day
`dexs/stonx` has published nothing since 2026-09-11 (~$6.3M over the 7 days before that). One pool
the API cannot serve takes the whole day down.

`fetch` asks the Ekubo API for per-pool volume and rethrows on the first failure:

```ts
const { results, errors } = await PromisePool.withConcurrency(API_CONCURRENCY)
  .for(pools)
  .process((pool) => httpGet(poolVolumeUrl(pool)) as Promise<PoolVolumeResponse>);

if (errors.length) throw errors[0];
```

Walking all 55 pools in the ve33 list, **2 of them return HTTP 400**:

```
GET /pair/4663/0x570c5aa79c798e7a418412cc8399ae5bcce570c5/0x976d390d590bbebe5f4a7f5016803ae05ace4bdb/volume
{"status":400,"error":"Invalid token identifier: \"0x976d390d590bbebe5f4a7f5016803ae05ace4bdb\""}
```

The other 53 answer 200, including the 15 pools whose `token0` is `0x0` — those are fine, so this is
one specific token the volume endpoint will not accept, not a systemic problem. The pools list keeps
returning the pool, so the adapter keeps failing every run.

## fix

PR follows: measure the failure ratio instead of rethrowing the first error, the same shape as the
`filterPools` guard in `helpers/uniswap.ts`. 2/55 is 3.6% and passes; a real outage still throws.

Verified on two days that published nothing:

| day | master | with the guard |
|---|---|---|
| 2026-09-12 | `Request failed with status code 400` | **$473,560** volume / $678 fees |
| 2026-09-13 | same | **$733,530** volume / $1.00k fees |

`tsc --noEmit` clean.

0xshubhs:
Fixes #9436.

Two of the 55 pools in the ve33 list return `400 Invalid token identifier` from the Ekubo volume
endpoint, and `if (errors.length) throw errors[0]` turns that into a lost day for the whole adapter.
stonx has published nothing since 09-11 as a result.

This swaps the rethrow for a failure ratio, the same shape `filterPools` uses in
`helpers/uniswap.ts`: 2/55 is 3.6% and goes through, while a real outage still throws and says how
bad it was rather than surfacing one pool's 400.

```
- if (errors.length) throw errors[0];
+ if (pools.length && errors.length / pools.length > MAX_FAILED_POOL_RATIO)
+   throw new Error(`stonx: ${errors.length}/${pools.length} pool volume requests failed, refusing to report a partial day`);
```

## verification

Two days that published nothing in production:

| day | master | this branch |
|---|---|---|
| 2026-09-12 | `Request failed with status code 400` | **$473,560** volume / $678 fees |
| 2026-09-13 | same | **$733,530** volume / $1.00k fees |

`tsc --noEmit` clean.

Worth noting the 400 is specific rather than general: the other 53 pools answer 200, including the
15 whose `token0` is `0x0`. It is one token the volume endpoint rejects, and since the pools list
keeps returning that pool, every run failed on it.


bheluga:
@0xshubhs thanks for the PR

Issue #9384, opened 2026-09-10
o1 Launchpad: Incorrect Fees / Revenue Data on DefiLlama, Significantly Below Full On-Chain Recalculation


dennischi99:
The Fees / Revenue figures displayed on the [o1 Launchpad page on DefiLlama](https://defillama.com/protocol/fees/o1-launchpad) are significantly understated and do not match the current adapter’s results when run against complete on-chain event data.

We selected **Base data for September 8, 2026 (UTC) as a validation sample** and confirmed this discrepancy. This date is only a test sample; **the full range of affected dates and chains has not yet been determined**.

## Sample Validation Results

The sample covers one complete UTC calendar day:

- Time window: `[2026-09-08 00:00:00, 2026-09-09 00:00:00)`
- Base blocks: `51018127–51061326`, inclusive, totaling **43,200 blocks**
- All six contract suites configured in the current adapter were covered

| Data Source | Fees | Revenue |
|---|---:|---:|
| Full local recalculation using the current adapter | **$87,385.76** | **$47,151.80** |
| Published figures on DefiLlama | **$27,781.00** | **$16,122.00** |
| Amount understated | **$59,604.76** | **$31,029.80** |

## Validation Method

We processed the same on-chain events through both the current DefiLlama adapter and the actual fee calculation code used by [o1 Launchpad Analytics](https://analytics.launch.o1.exchange/), comparing raw amounts trade by trade.

Results:

- **47,346 Trade events across 697 pools** were checked.
- Both implementations included **47,293 Trade events** and consistently excluded **53 trades with fees denominated in non-quote tokens**, following their existing rules.
- Raw integer amounts for Fees, Protocol Revenue, Creator Fees, and Referral Fees matched exactly for every trade.
- The adapter had no unexpected skips or missing-price skips; Analytics produced no quarantine records.
- The Analytics recalculation matched its production API snapshot for the same day using the same prices.

The local test executed the current `adapter.fetch`, accounting logic, and SDK Balances, with all required historical configuration supplied. Log retrieval used direct RPC and validated historical caches, rather than DefiLlama’s production indexer or cloud-cache path.

Test versions:

- dimension-adapters: `5c13727b079742d3f713e3bdbf3ac968f1fb982b`
- SDK: `5.0.223`
- SDK USD valuation timestamp: `2026-09-09 00:00:00 UTC`

## Requested Investigation and Correction

[...]

PR #9439, opened 2026-09-14
fix: log skipped o1 Launchpad accounting diagnostics instead of discarding

kurogami-ibrahim79:
### What this PR does

Fixes the `skip()` helper in `fees/o1-launchpad/accounting.ts` so skipped accounting
diagnostics are logged instead of silently discarded.

**Before:**
\`\`\`ts
const skip = (_message: string): void => {};
\`\`\`

**After:**
\`\`\`ts
const skip = (message: string): void => {
  console.warn(`o1 Launchpad: skipped ${message}`);
};
\`\`\`

### Why

- `skip()` is called at 26 call sites in `accounting.ts` whenever the accounting
  replay hits an incomplete/inconsistent event set for a pool or suite (e.g. missing
  historical `PoolRegistered`, duplicate logs).
- Its own docstring says these cases are "skipped **and logged**", but the current
  implementation discards the message entirely, so none of them are ever recorded.
- This was flagged as a **Major** issue by CodeRabbit's automated review on #9327,
  but was dismissed and merged unfixed.
- Related to #9384: published Fees/Revenue for o1 Launchpad on Base are reported
  significantly below a full local recalculation using this same adapter. A no-op
  `skip()` is one plausible mechanism for that gap partial/inconsistent event data
  can silently drop trades from the day's totals instead of surfacing the issue.

### What this PR does *not* do

Minimal, low-risk logging fix only. It does **not**:
- Change any fee/revenue calculation logic
- Fix the underlying cause of incomplete event sets (RPC log-fetch limits, historical
  backfill, etc.) still under investigation in #9384 and needs maintainer input on
  the production RPC/indexer setup
- Backfill any historical data

### Testing

- `git diff` confirms only the `skip()` function body changed; no other logic touched.
- Local `pnpm test fees o1-launchpad` reproduction on Base is currently blocked by
  free-tier RPC log-range limits (documented in #9384). Happy to re-run against the
  full suite if a maintainer can point to a working RPC setup, or validate via CI.

### Related

- Addresses CodeRabbit review comment on #9327 (flagged, not applied)
- Related to #9384


bheluga:
@kurogami-ibrahim79 thanks for the PR
This would have 100s of logs for every run, making it un-readable. Trying to find a better fix on infra side

PR #9444, opened 2026-09-14
tessera: add Robinhood Chain volume

ThomasRetr:
Tessera deployed its swap contract on Robinhood Chain at the same address used on Base and BSC (0x55555522005BcAE1c2424D474BfD5ed477749E3e), emitting the same TesseraTrade event, so the existing EVM fetcher is reused unchanged.

Contract deployed 2026-09-03 (block 53706462); first trade 2026-09-10 (block 59730768), which is used as the start date.


bheluga:
@ThomasRetr thanks for the PR

PR #9446, opened 2026-09-14
Add WESO forex orderbook volume to weso-defi

LunaClassicDAO:
## Summary
- Extends `dexs/weso-defi` to count WESO forex/stablecoin CLOB taker fills under the same slug.
- Markets from `https://weso.world/api/orderbook-markets` (19 depegged Terra Classic stables vs LUNC).
- Fills from `https://weso.world/api/orderbook-fills?marketId=…`, filtered to the fetch window.
- Counts each fill **once** on the LUNC quote side (`uluna`) so Llama can price volume without needing KRTC/EUTC/etc. prices.
- Leaves existing factory AMM `volume_buckets` + $WESO bonding-curve volume unchanged.
- Still excludes CWLUNC/CWUSTC wrap/unwrap and converter pairs. Fees adapter untouched.

## Notes
- Books are thin; many windows will report $0. That is expected.
- Fill history is limited to what the public fills API returns.

## Test plan
- [ ] Review `dexs/weso-defi/index.ts` forex helpers + methodology
- [ ] `npm test dexs weso-defi` (or repo equivalent)
- [ ] Confirm wrap/unwrap still excluded


LunaClassicDAO:
Addressed the CodeRabbit note: fills now resolve `fill.quoteSymbol ?? market.quoteSymbol` and are skipped unless that symbol is LUNC (case-insensitive), so missing fill symbols no longer get counted as `uluna`.

LunaClassicDAO:
Addressed the follow-up CodeRabbit note: resolve `fill.quoteSymbol ?? market.quoteSymbol`, then skip unless it is a string equal to LUNC (case-insensitive), so a non-string payload cannot throw on `.toUpperCase()`.

LunaClassicDAO:
Addressed the outside-diff CodeRabbit note: orderbook fills now only add non-negative integer `quoteAmountRaw` strings as `uluna`, and the `quoteAmount` fallback runs only for finite positive numbers (then re-validated as digits) so NaN, Infinity, fractional, or negative values cannot reduce or crash reported volume.

bheluga:
@LunaClassicDAO thanks for the PR

PR #9449, opened 2026-09-15
[fix] hyperliquid - discover HIP-3 dexes from the chain, not a hardcoded array

justencryptit:
`HYPERLIQUID_HIP3_DEXS` is a hardcoded array, and it has gone stale. HIP-3 deployment is
permissionless, so the list is out of date the moment anyone deploys a new dex — and when it is,
`queryHyperliquidIndexerOpenInterest` never asks for the missing dex, so its open interest is
dropped silently. There is no error and nothing in the output that says a dex was skipped.

Against `perpDexs` as of today:

| dex | open interest | in the list |
|---|---:|---|
| xyz | $3,971,611,293 | yes |
| io | $53,346,681 | **no** |
| para | $14,488,156 | **no** |
| mkts | $5,766,005 | **no** |
| abcd | $0 | **no** |
| vntl, flx, km, hyna, cash | $0 each | yes |

So `totalOpenInterest` currently reports $14,384,357,298 where the chain has $14,457,958,139 —
$73.6M missing, 0.51% of the total. Five of the six listed dexes report zero, so the array is not
only incomplete, it is mostly dead weight.

`mkts` is Kinetiq Markets' live dex; `km` in the list is its predecessor, dormant since June.
That pair is what made the staleness visible, but the fix is not Kinetiq-specific.

This queries the set from the `perpDexs` info endpoint instead, caches it for the process, and
keeps the existing array as a fallback if the call fails. Widening the set does not change any
historical day: for a day before a dex existed the indexer has no snapshot and
`getMetaAndAssetCtxs` returns `null`, which the loop already skips.

Reproduce:

```
curl -s -XPOST https://api.hyperliquid.xyz/info \
  -H 'Content-Type: application/json' -d '{"type":"perpDexs"}' \
  | jq -r '.[] | select(.) | .name'
```

https://claude.ai/code/session_01TfnGvCZgcZfrFaNizyMYDx


justencryptit:
Pushed both review points.

**Response validation** — `perpDexs` is now required to be an array, and each entry must carry a non-empty string `name` before it becomes a dex id. A malformed response falls through to the static list instead of sending garbage into the snapshot loop.

**The catch** — kept deliberately, and I think rethrowing would be the worse trade here. This helper feeds `totalOpenInterest`, and Hyperliquid's own perp dex is $10.4B of the $14.5B total. Failing the whole day because an auxiliary listing call timed out would lose 72% of the number to protect the 0.51% the stale list drops. The fallback is exactly the behaviour on master today, so the catch cannot make anything worse than not merging this; it only stops the improvement from becoming a new failure mode. The failure is logged, not hidden.

Happy to change it if you would rather the listing be load-bearing.

justencryptit:
@coderabbitai review

justencryptit:
The inline point is addressed in 1a76004: `Array.isArray` is required, and each entry must carry a non-empty string `name` before it becomes a dex id. A malformed response falls through to the static list. Reasoning for keeping the fallback is in the comment above.

https://claude.ai/code/session_01TfnGvCZgcZfrFaNizyMYDx

justencryptit:
Good catch, and it is the same bug the PR exists to fix: dropping the malformed entry leaves a dex missing from the list with nothing to say so.

One entry without a non-empty string `name` now discards the whole listing and falls back to the static set, which is at least a known one. Entry shape is read through `unknown` with a runtime check rather than `any`.

Your example behaves as asked:

```
[null,{"name":"io"},{"name":123}]      -> fallback (bad name)
[null,{"name":"xyz"},{"name":"mkts"}]  -> use ["xyz","mkts"]
"nope"                                 -> fallback (not array)
[null]                                 -> fallback (empty)
```

https://claude.ai/code/session_01TfnGvCZgcZfrFaNizyMYDx

justencryptit:
@coderabbitai review

justencryptit:
@coderabbitai full review

justencryptit:
Fair, and it is the same failure mode on a longer clock: a worker holding the first listing forever omits a dex deployed after it started.

Added a 10 minute TTL. Short enough that a long-running process picks up a new deployment, long enough that a backfill sweeping many intervals does not re-list on each one. A failed refresh keeps returning the last good listing rather than dropping straight to the static array.

https://claude.ai/code/session_01TfnGvCZgcZfrFaNizyMYDx

bheluga:
@justencryptit thanks for the PR

PR #9451, opened 2026-09-15
Iter (formerly Standard Spot): rename, fix volume double-count, track Arc and RISE testnet deployments

hskang9:
**Update to an existing listing (Standard Spot Exchange → Iter). Not a new protocol.**

- Website: https://iter.cx
- Contracts: https://github.com/iter-cx/iter-monorepo
- Related: TVL adapter in DefiLlama/DefiLlama-Adapters#21038. A mail to metadata@defillama.com will ask for the display name/slug change.

**Why**
Standard rebranded to Iter. The old `dexs/standard-spot` adapter also had two bugs: it added both the base and quote leg of every fill to `dailyVolume` (2x), and its `PairAdded` `fromBlock` was the Mode chain's block, not Somnia's.

**Chains**: at the protocol's request the adapter tracks the two deployments that carry the full contract set, Arc testnet (chain id 5042002, start 2026-09-14) and RISE testnet (chain id 11155931, start 2026-09-07), and drops the Somnia exchange-only generation. `ARC_TESTNET` and `RISE_TESTNET` are added to `helpers/chains.ts`. The SDK ships no providers for them, so the adapter registers RPC, chain id, Multicall3 and block-search-floor env defaults at load time; CI runs green with that. Arc uses Blockdaemon's public endpoint first because the official one rate-limits `eth_getLogs`.

**What the adapter counts** (`dexs/iter.ts`, v2, `pullHourly`)
- **Volume**: quote leg of every `OrderMatched` fill from the `MatchingEngine` and the `StopOrderEngine` (triggered stop orders match on the same books), plus the quote leg of every band pool swap: `BandSwapRouted` from the router and `RemainderRoutedToPool` from the engine (unfilled market-order remainder swapped into the pair's pool; side taken from the `SwapPriceReport` emitted just before it).
- **Fees**: `baseFee` + `quoteFee` from each fill (`Trading Fees`); band pool swap fees at the engine taker rate × tightest band premium, on the output token (`Token Swap Fees`); the flat launchpad fee per `LaunchSettingsRecorded` from the `AssetGenerator` (`Token Launch Fees`).
- **Revenue**: everything sent to `feeTo`. **SupplySideRevenue**: the `poolFeeShare` slice of band pool fees credited to LPs. **HoldersRevenue**: 0.

**Pricing**: testnet tokens have no price feed, so tokens are valued with the exchange's own market price. A stablecoin listed in the chain config is $1; every other token is worth its orderbook `mktPrice` times the USD price of the quote token, preferring a direct stablecoin book over a route through another token (e.g. WETH). Tokens the exchange cannot price are added as raw balances.

**Reviewer questions, pre-answered**
- 0-volume days are real: both testnets trade sporadically; Arc's current generation has no trades yet.
- Fee rate: 0.1% taker (`defaultTakerFee = 100000 / 1e8`), makers pay nothing. Launch fee is currently 0 on RISE.
- All four RISE books are quoted in WETH and there is no WETH/stablecoin book yet, so USD stays 0 until one is listed; raw amounts are collected regardless.
- Since the volume is now single-sided and the chains changed, a full refill is needed.

**Tests**
- RISE 2026-09-07: 78 fills, 5.77 WETH quote volume, fees in 5 tokens, identical to an independent recomputation from raw logs. Arc: clean zeros. `pnpm test dexs iter` (24 hourly slots) passes for both chains.
- The pricing path was checked against Somnia mainnet, where stablecoin books exist.
- `pnpm run ts-check` passes.


bheluga (changes requested):
@hskang9 thanks for the PR
If somnia deployment is inactive, please mark it dead instead of deleting the adapter.
Also we don't track testnet metrices

PR #9456, opened 2026-09-15
sunswap-v3: drop the startDate the volume endpoint now rejects, and refuse a missing day instead of publishing zero

### LINKED ISSUE #9455: sunswap-v3: dark since 09-12 — the volume endpoint now rejects the hardcoded startDate=2023-12-16 with a 400 ($1.17bn/30d)
`dexs/sunswap-v3` has published nothing since 2026-09-12 — $1.17bn over the last 30 days, ~$40m a
day, on $217m of TVL.

The endpoint stopped accepting the `startDate` the adapter hardcodes:

```
GET /scan/volume?version=v3&startDate=2023-12-16   400  {"code":1,"msg":"FAIL","data":null}
GET /scan/volume?version=v3                        200  {"code":0,"message":"SUCCESS",...}
GET /scan/volume?version=v3&startDate=2025-01-01    200  SUCCESS
```

So it is the old date specifically, not the host or the parameter. With no `startDate` the API
returns a rolling 365 days — right now 2025-09-15 through 2026-09-14 — and the missing days are
sitting in it:

| day | api says | published |
|---|---|---|
| 2026-09-13 | $16,779,244 | nothing |
| 2026-09-14 | $34,508,492 | nothing |

The adapter then does `.data?.list` and calls `.find` on it, so a `null` list throws rather than
being caught, which is why this reads as dark rather than zero.

The protocol is fine: `fees/sunswap-v3` reads a different host (`openapi.sun.io`) and has published
every day through 09-14, and sunswap-v1 and v2 are both current.

## the other half of this

```ts
const dailyVolume = data.find((item) => (item.time / 1000) === options.startOfDay)?.volume ?? "0";
```

The `?? "0"` means a day the API does not carry is published as **$0** rather than refused. It is
masked today because the throw happens first, but with the URL fixed it would start writing zeros
for every day before 2025-09-15, since that is all the window the API now serves. Worth removing at
the same time.

PR follows: drop the rejected `startDate`, and throw on a missing list or a missing day instead of
publishing zero. Verified on the two dark days — 09-13 reads $16.78M and 09-14 reads $34.51M,
matching the API to the dollar.

0xshubhs:
Fixes #9455.

`sbc.endjgfsv.link/scan/volume` now answers `400 {"code":1,"msg":"FAIL"}` for
`startDate=2023-12-16`, so `.data?.list` comes back undefined and the `.find` on it throws. The
adapter has been dark since 09-12, which is ~$40m a day.

Dropping the parameter is enough; the API returns a rolling 365 days without it (currently
2025-09-15 → 2026-09-14), and a later `startDate` works too, so it is that specific old date the
endpoint rejects rather than the host or the parameter.

The second change is the `?? "0"`. Today it is unreachable because the throw happens first, but with
the URL fixed it would publish **$0** for every day before 2025-09-15, the API no longer serves a
window that deep, and the adapter's `start` is still 2023-12-16. A day the source cannot answer
should be refused, not written as zero.

## verification

```
$ ts-node cli/testAdapter.ts dexs sunswap-v3 2026-09-14    # day 09-13
Daily volume: 16.78 M
$ ts-node cli/testAdapter.ts dexs sunswap-v3 2026-09-15    # day 09-14
Daily volume: 34.51 M
```

Against the API read directly: 09-13 `16779244.74...`, 09-14 `34508492.32...`. Both days currently
have no point in production. `tsc --noEmit` clean.

The protocol itself never stopped, `fees/sunswap-v3` reads a different host (`openapi.sun.io`) and
published every day through 09-14, and sunswap-v1/v2 are current.


bheluga:
@0xshubhs thanks for the PR

PR #9308, opened 2026-09-07
Kinetiq Markets: track the second builder code

justencryptit:
Kinetiq team here, fixing an undercount in our own adapter.

Website: https://kinetiq.xyz · Twitter: https://x.com/kinetiq_xyz

## The second builder code was untracked

Kinetiq Markets routes orders through **two** production builder codes, one per client. The adapter hardcoded only the web one, and the mobile one turns out to be the larger of the two:

| date | web `0x42f3…f992` | mobile `0x2af9…915d` | tracked before | missed |
|---|---:|---:|---:|---:|
| 2026-09-05 | $2,718 | $3,278 | $2,718 | **55%** |
| 2026-08-20 | $1,111 | $3,030 | $1,111 | **73%** |
| 2026-02-10 | $379 | $1.42 | $379 | 0.4% |

Both addresses are Kinetiq-operated and both sets of fees accrue to Kinetiq. The fix loops over both and adds into the same balances, so the existing breakdown labels and the builder/HIP-3 overlap subtraction are unchanged.

This is also visible from the other side: over 2026-06-16…09-03, sweeps *into* Kinetiq's buyback wallet from the web builder, the mobile builder's sweep intermediary `0x9cb4ac2598bf72d3f0b53efd817aac5cebe796fc`, and the km/mkts fee recipient total **$139,673**, which exceeds the adapter's gross-based revenue of **$125,179** for the same window. Routed cannot exceed gross, so revenue was being missed.

## A pre-existing failure this made more likely

Hyperliquid publishes a `builder_fills` CSV only for days on which a builder had fills, and `fetchBuilderCodeRevenue` turns the resulting 403 into a thrown error — which takes down the whole day, including the other builder and the HIP-3 leg. This already affected the single-builder adapter (`20251216` and `20260101` both 403 for the web builder). Adding a second address doubles the chance of hitting it, so a day with no fills is now logged and treated as zero. **Every other error still propagates** — the catch matches only that one message.

```
$ pnpm test dexs zz-probe-kinetiq-builders 2025-12-17     # builder legs only, see note below
kinetiq-markets: no builder fills for 0x42f3226007290b02c5a0b15bccbb1ba6df04f992 on 2025-12-16
Daily volume: 6.57 k
Daily fees: 6.14
label  | Daily fees | Daily revenue
mobile | 6.14       | 6.14

$ pnpm test dexs zz-probe-kinetiq-builders 2026-01-02
[...]

PR #9461, opened 2026-09-15
[feat] kinetiq - KNTQ spot deployer fees

justencryptit:
Kinetiq deployed KNTQ as a HIP-1 spot token and takes the deployer's cut of the trading fees on it.
None of it is counted anywhere today.

```
$ curl -s -XPOST https://api.hyperliquid.xyz/info \
    -H 'Content-Type: application/json' -d '{"type":"spotMeta"}' \
    | jq '.tokens[] | select(.name=="KNTQ") | {name, index, deployerTradingFeeShare}'
{ "name": "KNTQ", "index": 124, "deployerTradingFeeShare": "1.0" }
```

`deployerTradingFeeShare` is 1.0, so the whole of that cut is Kinetiq's.

### Why the KNTQ-denominated fee is exactly the deployer's take

Hyperliquid charges a spot fee in whichever token the filler receives, and a HIP-1 deployer's share
applies only to the fees paid in the token it deployed. KNTQ trades solely on the two pairs Kinetiq
deployed — `@334` KNTQ/USDC and `@254` KNTQ/USDH — so every KNTQ-denominated spot fee on Hyperliquid
is this deployer's. The USDC and USDH side of those same trades is Hyperliquid's, and
`hyperliquid-spot` already counts it, so this adapter does not repeat it. `Fees` here is the
deployer's side only, and `Revenue` equals it.

### Why it does not go through `CoinGeckoMaps`

Adding `KNTQ` to that map would credit these fees to Hyperliquid's own spot revenue, which is the
opposite of what a deployer fee is. So the helper returns the raw per-token figure from
`spotFeeByTokens` and the adapter attributes it itself.

### Scope

Independent of `kinetiq-markets`, which covers perps, builder codes and the HIP-3 dex, none of which
touch spot. `start` is 2025-11-28, the first day KNTQ has a price.

CI's live run will show whether `spotFeeByTokens` keys KNTQ by that name — if the indexer uses a
different key I will correct it.

https://claude.ai/code/session_01TfnGvCZgcZfrFaNizyMYDx


justencryptit:
The `test` failure is the fork not having `LLAMA_HL_INDEXER`, not a defect in the adapter:

```
Error: missing LLAMA_HL_INDEXER env
    at fetchHyperliquidSpotFeesByToken (helpers/hyperliquid.ts:576)
```

`dexs/hyperliquid-spot`, `dexs/hyperliquid-perp`, `fees/hyperunit` and `fees/hyperliquid-hlp` all depend on the same endpoint with no fallback, so an adapter of this shape cannot pass the test job from a fork. I would rather not add a "return zero when the endpoint is missing" path — that would publish silent zeros if the variable were ever unset in production, which is worse than failing loudly.

Two things I cannot verify from here and would like a maintainer to check with the secret set:

1. That `spotFeeByTokens` keys this token as `"KNTQ"`. The adapter reads that key directly. If the indexer uses the token index (124) or another form, tell me and I will change it.
2. The magnitude. Kinetiq measured roughly 148k KNTQ over the 30 days to 2026-09-13 by differencing the deployer's own HyperCore balance, so a live run should land near that.

Everything the adapter asserts about the mechanism is checkable without the indexer:

```
$ curl -s -XPOST https://api.hyperliquid.xyz/info -H 'Content-Type: application/json' \
    -d '{"type":"spotMeta"}' | jq '.tokens[] | select(.name=="KNTQ") | {name, index, deployerTradingFeeShare}'
{ "name": "KNTQ", "index": 124, "deployerTradingFeeShare": "1.0" }
```

and that KNTQ appears in exactly two pairs, `@334` KNTQ/USDC and `@254` KNTQ/USDH, both deployed by Kinetiq — which is what makes every KNTQ-denominated spot fee this deployer's.

https://claude.ai/code/session_01TfnGvCZgcZfrFaNizyMYDx

justencryptit:
Precedent for the red `test`, so it does not hold this up unnecessarily.

#9308 and #9064 both merged with `test: FAILURE`, and both failed on exactly this line:

```
Error: missing LLAMA_HL_INDEXER env
    at queryHyperliquidIndexer (helpers/hyperliquid.ts:308)
```

#9064 was not mine. Any adapter that reads the indexer fails the test job when the PR comes from a fork, because the secret is not exposed there. It is a property of the CI, not of the adapter.

I also checked whether a public source could avoid the dependency, so the option is closed rather than untried:

- `spotDeployState` for the deployer returns `{"states":[]}` — that endpoint covers in-progress deployments, not fee accounting.
- Nothing under `stats-data.hyperliquid.xyz` exposes per-token spot fees. That host answers 403 for paths that do not exist, so probing cannot distinguish "denied" from "absent", but there is no documented dataset either way.

So the indexer is required. The two things still worth a maintainer's minute are in my previous comment: whether `spotFeeByTokens` keys the token as `"KNTQ"`, and whether a live run lands near 148k KNTQ over 30 days.

https://claude.ai/code/session_01TfnGvCZgcZfrFaNizyMYDx

bheluga:
@justencryptit thanks for the PR

PR #9493, opened 2026-09-16
Add Meridian Perps volume, fees and open interest

amantedelagua:
Adds an adapter to count Meridian Perps fees, volume, and open interest


open-interest/meridian.ts
```diff
@@ -0,0 +1,29 @@
...
+import fetchUrl from "../utils/fetchURL";
+
+const API = "https://api.meridian.xyz/v1";
+
+async function fetch(_options: FetchOptions): Promise<FetchResult> {
+    const products = (await fetchUrl(`${API}/product`)).data;
+
+    const marketPrices = (await fetchUrl(`${API}/product/market-price?productIds=${products.map((market: any) => market.id).join('&productIds=')}`)).data;
+
+    const openInterestAtEnd = products.reduce((acc: number, market: any) => {
```
treeoflife2, line 12: oi is double-sided or single-sided, as we standardise the oi so it stays single sided oi

dexs/meridian/index.ts
```diff
@@ -0,0 +1,53 @@
...
+    });
+
+    orderMatchedLogs.forEach((order: any) => {
+        dailyVolume.addCGToken("ethena-usde", (Number(order.fillQuantity) / SCALE) * (Number(order.price) / SCALE));
+    });
+
+    return {
+        dailyVolume,
+        dailyFees,
+        dailyRevenue: dailyFees,
```
treeoflife2, line 32: is there any referral fees? we have to exclude that from the revenue, as referral fees is supply side revenue

open-interest/meridian.ts
```diff
@@ -0,0 +1,29 @@
...
+        return acc;
+    }, 0);
+
+    return {
+        openInterestAtEnd,
+    };
+}
+
+const adapter: SimpleAdapter = {
+    chains: [CHAIN.ROBINHOOD],
```
treeoflife2: need to change this to v2, as it only supports the current oi

treeoflife2 (changes requested):

dexs/meridian/index.ts
amantedelagua: There aren't any referral fees or fee share.

open-interest/meridian.ts
amantedelagua: It's two-sided, added the adjustment below

open-interest/meridian.ts
amantedelagua: Done

treeoflife2 (approved):

PR #9500, opened 2026-09-17
basestonk: Base only - remove the Robinhood Chain block

Xipzer:
BaseStonk removed Robinhood Chain from the product on 2026-09-16; the launchpad is Base-native and no longer lists or serves that chain. The `CHAIN.ROBINHOOD` entry read fees from hooks that no longer take any, so it would report a flat zero forever. The Base block is unchanged.

Follows #9295 / #9420.


treeoflife2:
closed pr due to too many unrelated file changes

PR #9510, opened 2026-09-17
fix: complete vfat fee coverage

vfat-ai:
Website: https://vfat.io/yield · X: https://x.com/vfat_io

Fix missing fees from older Sickles and 20 unlisted chains. The existing adapter reads `fromBlock`, but 15 configurations provide `fromBlockSickle`, so wallet discovery misses contracts deployed before the fee window.

Validate each window's fee emitters with the factory's `admins` → `sickles` round trip. Factory membership is permanent, so current registry reads also validate historical fees without scanning deployment history.

Also normalize the native-token sentinel, exclude shared boundary blocks from consecutive hours, and handle Cronos's August block-production gaps with a bounded block lookup. Public RPC defaults and Etherlink's log-range limit are included. `maxBlockRange` is added to the options type already forwarded to the SDK.

### Local checks

- 44 regression tests; `pnpm ts-check`; `pnpm ts-check-cli`.
- `pnpm test fees vfat`: all 37 chains and 24 hourly slots.
- `pnpm test fees vfat 2026-09-01`: August 31 UTC, all 24 slots, $4.48k in priced fees.
- First-fee transaction checks passed on the 19 added chains with older history; Arc is covered by the current run.
- A separate September 16–17 09:00 UTC run matched all 27,317 receipts shared with our indexer, including exact raw amounts. It found another 63 Arc receipts not yet in that snapshot. USD differences reflect token prices and pricing coverage; dashboard USD is not used as an expected value. Normal SDK caches were present.

Both `FeeCharged` ABIs supply actual charged amounts, so fee-rate changes require no rate estimate. Existing fee/revenue classification is retained; this adapter does not measure underlying LP yield or swap volume. Zero is valid only for a successfully read window with no eligible fees; read failures throw.

Replaces closed #9503. Please refill fees, revenue and protocol revenue from each chain's configured start (March 21, 2024 for the earliest chains), including Base and the added chains.


bheluga:
@vfat-ai thanks for the PR

PR #9519, opened 2026-09-17
treadfi-perps: remove fees and revenue, report volume only

jaytreadfi:
Removes fees, revenue and protocol revenue from the `treadfi-perps` adapter. The builder-code figures did not reflect actual Tread.fi revenue and were misleading on the protocol page, so the adapter now reports OMS fill volume only on every chain (Hyperliquid and Extended now use the same volume-only path as the other venues).

- Dropped the Hyperliquid and Extended builder-fee fetchers and their helpers
- Removed the Fees / Revenue / ProtocolRevenue methodology and breakdown entries
- No change to volume sources or chain coverage

Tested with `ts-node cli/testAdapter.ts dexs treadfi-perps` (2026-09-16): 11 chains, aggregate 11.86M volume, no fee rows.


jaytreadfi:
Maintainers: could you also clear the stored fees/revenue history for treadfi-perps once this merges? The adapter stops emitting them going forward, but the previously recorded builder-code revenue would otherwise stay on the protocol page.

bheluga (commented):
@jaytreadfi thanks for the PR
Is it possible to provide an api that covers all the sources of revenue, if indexed from your end instead of removing it?

bheluga:
@jaytreadfi thanks for the PR
We have  disabled it internally

PR #9534, opened 2026-09-17
Add Calamari and Tsunami volume and fee adapters

goran9999:
# Add Calamari and Tsunami volume and fee adapters

Two listings on Ink (chain id 57073), submitted together because their fee
attribution is deliberately split between them and reads oddly in isolation.

- **Website**: https://calamari.trade, https://tsunami.trade
- **Docs**: https://docs.tsunami.trade
- **X/Twitter**: https://x.com/calamaridex, https://x.com/tsunamilaunch

TVL adapters are up separately in DefiLlama-Adapters.

## Calamari — `dexs/calamari.ts`

Calamari is a Uniswap v4 deployment on Ink with its own PoolManager singleton
(`0x6E4723A612831AfB9f5B2a5aE22723c37aAB9560`). It is not a hook on the
canonical v4 deployment, which does not exist on Ink, so this is not double
counted against the Uniswap v4 listing.

Pools come from the PoolManager's `Initialize` logs, volume and fees from its
`Swap` logs. The fee is taken from each swap's own emitted fee rate, and split
using the pool's `protocolFee` read per direction from `StateView.getSlot0` —
v4 packs the two swap directions into 12 bits each.

Current state on-chain: the four pools created directly on Calamari carry a
3000 or 10000 lpFee and no protocol fee, so their fees are all supply side. The
363 pools seeded by Tsunami graduations carry a 0.1% protocol fee and a zero
lpFee, so theirs are all revenue. Both cases fall out of the contract reads;
nothing is hardcoded.

## Tsunami — `dexs/tsunami.ts`

Tsunami is a bonding-curve launchpad. Volume is the quote currency traded
against a curve (`TokenBought` / `TokenSold` on each curve), plus the quote leg
of swaps in the Calamari pools that graduated curves seed.

Curves, their quote assets, their hooks and their FeeShare contracts are read
from the factory's own mappings — `getLaunchedTokens`, `tokenCurves`,
`tokenQuotes`, `tokenHooks`, `curveFeeShares` — rather than from its
`TokenLaunched` logs. That event has been extended twice, and each extension is
a new topic for the same thing, so a reader matching one signature silently
loses every launch that used another. The mappings cannot fall behind the
contract that writes them, and the same is true of the hook set: the factory
has rebound its hook, and pools keep whichever one seeded them.

Fees are the factory's curve fee, read live from `feeConfig()` (currently 100
bps), plus the 1% its v4 hook charges on graduated pools, taken from the hook's
`SwapFeeAccrued` logs. That event exists in two shapes — the newer one names
the curve it accrued for — and both are read. The creator split comes from
`feeSharePercent()` and is applied to launches that deployed a FeeShare
contract, on curve fees and, where the accrual names its curve, on hook fees.

`doublecounted: true`: the graduated pools sit on Calamari's PoolManager, whose
adapter counts the same swaps.

## Pre-answering the usual review questions

**Is a 0-volume day correct?** Yes, and most days so far are. The protocol is
new: 577 tokens have been launched, but trading was near zero until 2026-09-14,
when the first large launches ran. Days before that genuinely have a handful of
trades or none. Nothing is swallowed — a missing source throws.

**Does the fee-to-volume ratio make sense?** Yes, and it is the main check on
both adapters. Calamari lands on exactly 0.10% of volume every hour, which is
the v4 protocol fee on the graduated pools that carry almost all the flow.
Tsunami lands on exactly 1.00% every hour, which is both the curve fee and the
hook fee. Hourly output for 2026-09-14 below.

**Why do the two differ on the same swaps?** A trade in a graduated pool pays
1.1% in total: 0.1% to Calamari's PoolManager as the v4 protocol fee, and 1% to
Tsunami's hook. They are charged by different contracts and land in different
places, so each is booked once, against the protocol that charges it. The hook
fee is not part of the `Swap` event's fee field — it is taken as a beforeSwap
delta — so there is no overlap.

**Did the fee rate ever change?** No. `feeConfig()` has returned 100 bps and
the hook's `SWAP_FEE_BPS` has been 100 since deployment. Both are read from the
contracts rather than assumed, so a change is picked up without a code change.

**Related listings or open PRs?** None. Neither protocol is listed.

**Refill needed?** No, these are new adapters.

## Server-side wiring

Both need `dimensions: { dexs: "calamari" }` and `dimensions: { dexs: "tsunami" }`
in defillama-server after merge.

## Validation

`npm run test dexs calamari 2026-09-15` and `npm run test dexs tsunami 2026-09-15`,
covering 2026-09-14, the first day with real volume. The run stops partway
because the day is still in progress.

```
Calamari                              Tsunami
12 PM  volume 764.00    fees   0.76   12 PM  volume   4.15 k  fees  42.00
 1 PM  volume 241.29 k  fees 241.00    1 PM  volume 341.40 k  fees   3.43 k
 2 PM  volume 115.09 k  fees 115.00    2 PM  volume 161.84 k  fees   1.63 k
 3 PM  volume 256.18 k  fees 256.00    3 PM  volume 322.76 k  fees   3.24 k
 4 PM  volume 726.35 k  fees 726.00    4 PM  volume   1.05 M  fees  10.58 k
 5 PM  volume 837.10 k  fees 857.00    5 PM  volume   1.08 M  fees  10.81 k
```

Calamari's 5 PM slot is the one hour with supply-side revenue ($23), from the
non-graduated pools that do carry an lpFee.


treeoflife2:
already created both listing, closing the pr

PR #9539, opened 2026-09-18
Add Load volume and fees (Arc)

Strikt-Crypto:
**NOTE**

Allow edits by maintainers is enabled.

---

Adds Load volume (`dexs/load.ts`) and fees (`fees/load.ts`) on Arc. Shared fetch lives in `fees/load.ts`.

TVL adapter: https://github.com/DefiLlama/DefiLlama-Adapters/pull/21098

##### Name:
Load

##### Twitter Link:
https://x.com/loaddotfun

##### Website Link:
https://load.fun

##### Logo:
https://load.fun/loadfun.svg

##### Chain:
Arc (`5042`)

##### Category:
Launchpad

##### Start:
2026-09-17 (factory fromBlock `21349552`)

##### Short Description:
Meme launchpad on Arc. Bonding-curve and Instant Uniswap V3 / V4 launches; graduation locks LP.

##### methodology:

**Volume** — bonding-curve `Trade.usdcAmount`; Instant/graduated V3 Load-router notional (`BondingSwapFeePaid.feeTotal * 400`); Instant/graduated V4 `GraduatedSwap.usdcAmount`. Uniswap pool volume that never hits a Load router is excluded.

**Fees** — 1 USDC per `TokenCreated` + bonding `Trade.fee` (1%) + Load router `BondingSwapFeePaid.feeTotal` (0.25%).

**Revenue / ProtocolRevenue** — create fee + 20% of the 1% curve fee + `BondingSwapFeePaid.toPlatform`.

**SupplySideRevenue** — 80% of the 1% curve fee to creators + trader-referral share of the 0.25% router fee.

##### Does this project have a referral program?
Trader referral on the 0.25% Load router fee (20% of that fee when `traderReferrer` is set). 5% of the 1 USDC launch fee also goes to launch referrers (counted in the create-fee revenue until we split it out).


treeoflife2:
Pushed a fix for the create-fee decimals bug: CREATE_FEE was 10n**18n against ADDRESSES.arc.USDC, which is Arc's 6-decimal ERC-20 facade, not the 18-decimal native representation. That was booking each 1 USDC launch fee as $1 trillion.

While testing that fix I found a second, more serious issue: BondingSwapFeePaid.feeTotal is not consistently denominated in USDC. Pulling raw values from the routers directly, the same event stream mixes amounts on two different scales - e.g. many in the 1e15-5.5e16 range alongside others in the low thousands - within the same router. That's consistent with some Instant/graduated pools being quoted in a non-USDC currency (native ETH-equivalent or otherwise) while others are quoted in USDC, with feeTotal reported in whatever that pool's actual quote currency is. The code currently books every feeTotal as USDC regardless.

This isn't cosmetic: the runner's own sanity check already throws ("value is too damn high") on at least one real window (2026-09-18 02:30-03:30 UTC) even with the decimals fix applied, so the adapter can't currently complete a full day's run. Confirmed this reproduces on the unmodified code too, it's not something my fix introduced.

Given I can't tell from the events alone which currency a given feeTotal is in, this needs either a per-pool currency lookup (similar to how Wonk resolves pair tokens per pool) or your own knowledge of which router legs are USDC-only vs multi-currency. Left the CREATE_FEE fix in place since that one's unambiguous; didn't attempt to guess at the currency-mapping fix.

Separately, the referral-share gap CodeRabbit and the PR description both note (5% of the launch fee going to referrers, currently all booked as Revenue/ProtocolRevenue) is still open - same reasoning, needs a data source for the referrer split that isn't visible in the given event ABIs.

PR #9540, opened 2026-09-18
[fix] Arc - start Aave v4 and Uniswap adapters from public mainnet launch

baptistapedro:
Variant analysis  of the merged Morpho fix https://github.com/DefiLlama/dimension-adapters/pull/9536 found the same pre-launch Arc start date in Aave v4 and Uniswap v2, v3, and v4.

This changes only their Arc `start` values from `2026-05-28` to the public mainnet launch date, `2026-09-16`:

https://www.arc.io/blog/arc-mainnet-goes-live-on-september-16-2026

No other chain configuration is changed. This prevents pre-public Arc backfills from querying unavailable history.


bheluga:
@baptistapedro thanks for the PR
These contracts were deployed much before public mainnet launch and hold data, so it's fine to keep as is

PR #9546, opened 2026-09-18
accountable: re-enable base

pawel-accountable:
Base was commented out in this adapter with the note *"Accountable no longer has any active vault on this chain, so there is nothing left to enumerate here going forward."* That is no longer true.

Yieldpoint yUTY (vault `0x4C18E2bb...`, strategy `0xC2e2944d...`) is live on base: `loanState 3`, holding ~1,000,043 yUTY, OPEN_TERM, with a `performanceFee` of 30000 (3%) and a 100% `protocolSplit`, so it does accrue protocol revenue. Its strategy comes from `0xB4082B81...`, which is already in the base factory list that was commented out, so simply restoring the block picks it up.

Only the config block changes; no logic is touched, and the factories and `start` date are exactly the ones that were already there.

### Caveat, stated up front

This will book ~$0 for now regardless, because DefiLlama has no price for yUTY (`0xBa515EEd...`): `coins.llama.fi` returns an empty object for it, and for its underlying UTY, and for `coingecko:unity-2`. The token has essentially no market (a single Aerodrome/USDC pool, 37 trades and $9.4k of volume in five months). The other three base vaults hold dust (2 USDC, 58 USDC, 0).

So the immediate effect is a correct comment and enumeration that starts producing real numbers the moment yUTY is priced, rather than a change in reported fees today. If you would rather not add a chain that currently reports zero, I am happy to close this and resubmit once the pricing is sorted.

(The TVL adapter side of this is DefiLlama/DefiLlama-Adapters#21122, which works around the missing price by unwrapping the ERC4626 and booking the dollar-pegged underlying against USDC.)

### Testing

I could not run `npm test fees accountable <date>` to completion locally: it fails on robinhood with `Accountable: every factory call failed on robinhood, refusing to report zero`, and I confirmed that failure is present on a clean `origin/master` without this change, so it is unrelated. With robinhood temporarily disabled the run did not complete in my environment either. Flagging that honestly rather than claiming a green run.


bheluga:

PR #9563, opened 2026-09-18
[feat] arguspad - fees adapter for ArgusPad launchpad on Arc

degenanddreams:

ArgusPad

##### Twitter Link:

https://x.com/arguspad

##### List of audit links if any:

None — the team publishes no audit.

##### Website Link:

https://arguspad.io

##### Chain:

Arc (chainId 5042)

##### Short Description (to be shown on DefiLlama):

Tax-token launchpad on Arc. Each launch mints its whole supply into a single-sided Uniswap v4 position held by a locker with no withdrawal path, so there is no bonding curve and no migration — the token trades in a real v4 pool from block one, and a per-launch hook charges a buy/sell tax on every swap.

##### Category:

Launchpad

##### Token address and ticker if any:

`ARGUS` — `0xeCe5cA8bf9220718E5727754026757512212cb3c` on Arc. It is itself an ArgusPad launch.

##### Treasury Addresses:

`0x934DEA9Ab179De155db10519aA89a8C418C50705` — returned by `treasury()` on all ten ArgusPad Portals.

##### Github org/user:

Not public.

##### methodology:

See the `methodology` and `breakdownMethodology` blocks in `fees/arguspad/index.ts`, summarised below.

Docs: there is no GitBook or whitepaper. The team's technical reference is the generated bundle at
https://arguspad.io/argus-v4.json (ABIs + every live address, regenerated from the compiled
artifacts and verified against `eth_getCode`).

Server-side wiring (`dimensions: { fees: "arguspad" }` in defillama-server) is a follow-up after
merge. The TVL leg is a separate PR to DefiLlama-Adapters (`projects/arguspad/index.js`).

---

# ArgusPad — fees adapter

One new file: **`fees/arguspad/index.ts`**. `version: 2`, `pullHourly: true`, `chains: [CHAIN.ARC]`,
`doublecounted: true`, `start: "2026-09-07"`. No `dexs/` adapter — reasoned below.

## What ArgusPad is, and where the money is

A launch deploys five contracts and seeds one Uniswap v4 position with the entire supply. There is no
curve contract: `Swap` comes from the v4 PoolManager singleton from the first block, and the position
is permanently locked (the locker has no withdrawal path). Fees arrive on two paths:

| | emitter | event |
|---|---|---|
| the launch's own buy/sell tax, 1–10% chosen per launch | `ArgusV4TaxHook`, one per launch | `TaxTaken`, `SnipeTaxApplied` |
| the 1% v4 pool fee earned by the locked position | `ArgusV4HookedLocker`, one per launch | `Forwarded` |
| splitting the pot to treasury / creator / holders / liquidity | `ArgusV4HookedSplitter`, one per launch | `Distributed`, `Netted`, `Paid`, `DividendFunded`, `PrincipalDelivered`, `PrincipalParked` |

## Fees is counted at accrual, not at realisation

`dailyFees` is the fee **charged to users inside the window**, denominated in the launch's quote
asset. Two things are deliberately excluded because they are fee value that accrued earlier:

- the proceeds of the splitter selling launched tokens it was paid in (`Netted`), and
- opening inventory distributed during the window.

The governing precedent is `fees/peach-launchpad/index.ts:11-14`, which counts the fee **when it
accrues** and excludes the later release *"to avoid counting the fee twice"* — the same mechanic.
`fees/coinbarrel.ts` says it in one clause: *"counted when it accrues; the fee router pays it out in
batches"*. No adapter in the repo counts the resale of previously-taken token inventory as fees.

Counting realisation instead would have reported **2.6×** the accrual figure on the day measured, and
~30% of it was money users paid on an earlier day.

## Revenue / SupplySideRevenue breakdown

**Revenue is the treasury's configured share, taken per launch as each fee is counted** — not a rate
measured over the window. `treasuryBps` is read from the Portal's own `FeeConfigured`, which the
contract emits once per launch and never amends; it is **1000 on all 140,022 launches** today, and
reading it per launch rather than hardcoding 10% means a launch that ever ships a different rate is
handled on its own terms.

That choice is load-bearing, not cosmetic. `pullHourly: true` means the adapter runs 24 times a day
and the slices are summed, so **any rate derived from the window itself moves with whichever launches
happened to crank inside it.** An earlier revision of this adapter did exactly that, and the hourly
run caught it: three of the 24 slices came back with `revenue == fees` and zero supply-side, because
a quiet hour contained a crank in which the treasury was the quote pot's only claimant. Summed over
the day that overstated revenue by about 30% against the same day computed as one window. With the
configured rate, the two granularities agree:

| same day, 2026-09-08 | Fees | Revenue |
|---|---|---|
| 24 hourly slices summed | 792.98 | 79.31 |
| one 24-hour window | 794 | 79 |

**Known and stated: this is a ~3–5% understatement of Revenue against a realisation measure.** What
the treasury is actually *paid* out of a realised pot came to 10.32% and 10.56% over the two windows
below, against the configured 10.00%. The excess is real but price-dependent: `_allocate` derives the
quote weights from the pool price (`_weights(c.sqrtPriceX96)`) and `_shares` pays nothing to anyone
when that weight is zero (`if (pot == 0 || wTotal == 0) return (0, 0, 0)`). No per-launch rate
expresses it, and a window-measured rate is not stable enough to stand in for one, so the configured
share is what gets reported. I tried identifying the affected launches from config — 221 of the
140,022 allocate everything but the treasury cut to the burn — but that predicate overshot Revenue by
+1.8%, because `wQuote` is a price condition and not a configuration one. Reverted rather than
shipped.

| | |
|---|---|
| `Revenue` / `ProtocolRevenue` | `treasuryBps` of each launch's accrued fee, plus 100% of the snipe tax (charged straight to the treasury, it never reaches a splitter) |
| `SupplySideRevenue` | everything else: creator share + launched-token holder dividends + the share compounded back into the permanently locked position |

The **total** supply side is `fees − revenue`, so it is as stable as revenue is. Only the split
*between its three labels* uses the window's realised proportions, since `_weights` makes those
genuinely price-dependent and there is no per-launch rate to read; the creator label takes the
remainder, the way `fees/wonk/index.ts:134-137` does it, so `dailyFees = dailyRevenue +
dailySupplySideRevenue` holds exactly however the rounding falls. If a window realised nothing, or
paid out more than it realised, the whole supply side goes to the creator label rather than to a
ratio that cannot be measured.

`HoldersRevenue` is deliberately **not** set: `DividendFunded` pays holders of the *launched* token,
not ARGUS holders, so per `fees/AGENTS.md` ("stakers of a non-governance token are suppliers") it is
a supply-side cost. `fees/tolly/index.ts:76` books the identical flow the same way.

## `doublecounted: true`

Every ArgusPad pool is a plain Uniswap v4 pool on Arc carrying a **static `fee = 10000`** (1%) that
`dexs/uniswap-v4.ts:442` reads straight off the Swap event, and Arc is in its `Configs` map at
`:207`. Measured over the 24 h to block 21,516,182:

- ArgusPad pools are **326,722 of the 640,639** Uniswap v4 swaps on Arc (51.0%);
- every one of those 326,722 swaps carries `fee == 10000`; no pool has a second fee value; 10000 is
  neither the v4 dynamic-fee flag nor zero, so there is no escape from the overlap;
- applying `dexs/uniswap-v4.ts`'s own formula, ArgusPad pools are **$208,296 of Uniswap V4's
  $447,470 Arc fee figure — 46.5%**. (Reproducing the whole Arc v4 figure that way came out
  +0.22% against the published number, so the method is right.)

The governing precedent is on the same chain and three weeks old: `fees/wonk/index.ts:197` —
*"these are Uniswap v4 pools, already counted by the uniswap-v4 adapter on Arc"*. Also
`fees/pools-trade.ts:180` and `fees/arrowpad-fun.ts:93`.

**Stated rather than buried:** the flag is adapter-wide, so it also suppresses the swap tax, which
Uniswap does *not* count (the hook takes it via an `afterSwap` delta; the Swap event's `fee` field is
the pool's LP fee only). `fees/harbor.ts:480-482` declines the flag for exactly that reason. The LP
fee is 61.5% of Fees on the day measured, and dropping it instead would misstate what the protocol
earns, so the flag is the better of the two available answers. The overlap is also disclosed inside
`breakdownMethodology`, the way `fees/coinbarrel.ts` does it.

## Volume: no `dexs/` adapter, deliberately

Four Arc launchpads ship a `dexs/` adapter — `peach-launchpad`, `solonpad`, `circlewarp`, `sashimi` —
and **every one of them counts only its own bonding-curve volume**. peach is explicit
(`dexs/peach-launchpad/index.ts:9`: `if (row.source === 'bonding')`), and solonpad, circlewarp and
sashimi read only their curve buy/sell events. The graduated-pool leg, which a DEX listing already
counts, is left out in all four.

The launchpads structurally identical to ArgusPad — no curve, whole supply into a permanently locked
Uniswap position — ship **no `dexs/` adapter at all**: `fees/tolly`, `fees/pools-trade.ts`,
`fees/arrowpad-fun.ts`. Nor do the v4-hook launchpads `fees/wonk`, `fees/coinbarrel.ts`, `fees/aka`.

ArgusPad has no curve, so **100% of its volume is Uniswap v4 volume on Arc** and is already inside
`dexs/uniswap-v4.ts`. A `dexs/arguspad.ts` would have nothing in it that the DEX listing does not
already have, so there is none.

## How 140,022 per-launch emitters are handled

The emitters are one tax hook, one splitter and one locker **per launch** — 140,022 / 140,022 /
140,081 today, growing by tens of thousands a day. They cannot go in a `getLogs` address filter, and
`targets` would be one request per address. Two scans instead:

**1. The launch list (cached).** Each Portal announces every launch in its launch transaction:
`PartsDeployed(token, locker, hook, splitter)` and `FeeConfigured(..., sellTaxBps, treasuryBps, ...)`.
One topic-filtered scan of those two events over all history, `cacheInCloud: true` — the
"pool-created list" case `AGENTS.md` names — filtered to the five Portals in memory, because ten
foreign launchpads on Arc run forks of this codebase and emit byte-identical events.

*Verified, not assumed:* over blocks 19,660,000–21,516,182 that scan reconciles to every Portal's own
`tokenCount()` **exactly** — P7 139,675 · P6 323 · P5 1 · P4 22 · P3 1 = **140,022**, with
`FeeConfigured` covering **140,022 of 140,022** and zero launches missing either event. Cross-checked
against an independent registry-state enumeration (140,081 `launches(token)` reads, the route the TVL
adapter uses): **140,022/140,022 identical on hook, locker and splitter, and 140,022/140,022 on
`sellTaxBps`**, zero mismatches, zero extra.

**2. The window (per run).** One topic-filtered pass for all ten fee events at once, with the emitter
resolved against the launch list before anything is decoded, then the fee maths. `AGENTS.md` sanctions
this shape: *"hundreds of targets is the sanctioned exception to `targets`: fetching all logs by
`topic0` and filtering client-side is more efficient there"*. At 140,022 it is the only option.

### Cost, measured

Measured against Arc's public RPC, one request at a time (it is the only Arc endpoint that serves
`eth_getLogs` at a useful range - `arc.drpc.org` refuses above ~100 blocks and `arc.rpc.thirdweb.com`
above ~1,000, though all three agree on `eth_chainId` and on the pin block's hash).

| | |
|---|---|
| launch list, all history (1.84M blocks, 280,248 logs) | **202 `getLogs`, ~8 min** - once, then cached (23 MB on disk) |
| one 24 h window | **~95 `getLogs`** (measured: 187 ranges / 195 calls / 373 s / 0 rate limits for two windows back to back), plus one `quoteAsset` multiCall over the splitters that cranked (15,619 in the heavier window) |
| per hourly run, which is what ships | ~4 `getLogs` and ~650 contract reads |
| largest single range | 18,595 logs = **93.0% of the node's 20,000-result cap**, so no range was at the cap and none was truncated |
| logs the window pass reads | 481,089 and 1,213,865 chain-wide for the two windows, of which 379,602 and 1,049,399 are ArgusPad's |

**Is it viable at 140k launches and growing? The per-window pass, yes - it scales with activity, not
with cumulative launches, so it is flat.** The launch list is the part that is O(total launches):
280,248 logs today. The SDK trims a log cache above ~300 MB and keeps only the newest 200k logs, so
once ArgusPad is past roughly 300,000 launches the cached list stops amortising and the ~200-call
historical scan is re-paid on every run. At the rate this protocol has been launching (83,720 in one
day) that is days away, not months, and the clean fix is Arc in DefiLlama's own indexer - it is not in
`sdk.indexer.supportedChainSet2` today, which is also why `streamLogs` with `targetsFilter`, the
mechanism built for exactly this shape, cannot be used here. Worth knowing before it bites.

One SDK limit had to be worked around rather than lived with: `getLogs` bisects the block span on
error but **stops at 1,001 blocks and throws**, and Arc refuses a 1,001-block range at peak launch
density (its own error names the range that would fit: **684 blocks**). The adapter therefore walks
the window itself, halving below that floor and widening again once density drops, retrying a span
once before narrowing so a transient failure is not mistaken for a range that is too wide.

### Reproduction

Two non-overlapping 24 h windows, each driven through `fetch` over the exact block range, against an
independent measurement of the same windows made before this adapter was written:

| | | adapter | independent measurement | delta |
|---|---|---|---|---|
| **W0** | blocks 21,345,783–21,516,182 | | | |
| | Fees | **$241,276.22** | $241,209.95 | **+0.027%** |
| | Revenue | **$24,127.91** | $24,881.05 | −3.027% |
| | SupplySideRevenue | **$217,148.31** | $216,328.90 | +0.379% |
| **W2** | blocks 21,175,383–21,345,782 (the peak launch day) | | | |
| | Fees | **$885,268.62** | $885,083.38 | **+0.021%** |
| | Revenue | **$88,527.30** | $93,497.39 | −5.316% |
| | SupplySideRevenue | **$796,741.33** | $791,585.99 | +0.651% |

`dailyRevenue + dailySupplySideRevenue − dailyFees = 0.000000` on both. The USD figures are from one
run each and move by a few cents between runs as the price feed moves; the raw amounts do not.

**Fees reproduces exactly.** The USD delta is +0.02% in the same direction on both windows, which is
price drift rather than method — the **raw per-asset amounts are identical to the raw unit** on all
seven quote assets:

| quote asset | adapter, raw | reference |
|---|---|---|
| USDC | `238686559121` | 238,686.559121 |
| ARGUS | `113975203238065829289523` | 113,975.203238 |
| cirBTC | `1028897` | 0.010289 |
| ARCASH | `34604476440494390927895` | 34,604.476440 |
| EURC | `28173132` | 28.173132 |
| WETH | `409354604539494` | 0.000409 |
| XAUM | `601446345590704259` | 0.601446 (no DefiLlama price; contributes nothing to the USD total) |

The per-label split reproduces too: W0 Swap Tax $92,906 / LP Fees $148,370 / Snipe Tax $0.4275
against $92,872 / $148,337 / $0.43, and W2 $540,302 / $344,965 / $1.508 against $540,178 / $344,903 /
$1.51.

**Revenue is deliberately 3–5% below the reference**, for the reason set out above: the reference
applied a realised split measured over the whole window, this adapter takes the treasury's configured
share per launch. The window-measured figure is the more accurate one for a single 24-hour window and
the less usable one for an adapter that is pulled hourly; the trade is stated rather than hidden, and
`Revenue / Fees` comes out at 10.0001% on W0 and 10.0001% on W2 — `treasuryBps` plus the snipe tax,
by construction.

### Runner

```
DISABLE_PULL_HOURLY=true DEBUG_BREAKDOWN_FEES=true pnpm test fees arguspad 2026-09-09
```
```
🦙 Running ARGUSPAD adapter 🦙
Start Date: Tue, 08 Sep 2026 00:00:00 GMT   End Date: Wed, 09 Sep 2026 00:00:00 GMT
ARC 👇   Backfill start time: 7/9/2026
Daily fees: 794.00   Daily revenue: 79.00   Daily protocol revenue: 79.00   Daily supply side revenue: 714.00

label                           | Daily fees | Daily supply side revenue | Daily revenue
Swap Tax                        | 344        |                           |
LP Fees                         | 450        |                           |
Creator Fees                    |            | 410                       |
Launched Token Holder Dividends |            | 304                       |
Swap Fees to Treasury           |            |                           | 79
```

`pnpm test fees arguspad 2026-09-09` (the hourly path that actually ships) returns 24 distinct
slices; they sum to 792.98 fees / 79.31 revenue against the single window's 794 / 79 above.
`pnpm run ts-check` passes. `git diff --check` is clean. One file added, nothing else touched.

**Local testing used a gitignored `.env` `ARC_RPC` override** pointing at a pacing proxy in front of
`rpc.mainnet.arc.io`, because the endpoint 429s below the rate the SDK issues requests at and the
SDK's limiter is a concurrency limit rather than a rate limit. Nothing about that ships — the adapter
uses whatever `providers.json` gives it, and on a well-provisioned endpoint the run is limited by the
20,000-result cap rather than by pacing.

### The emitter allowlist, measured on the same two windows

Streaming every log chain-wide that matches the ten topic0s, and asking how many the allowlist keeps:

| event | chain-wide (W0 / W2) | ArgusPad | dropped |
|---|---|---|---|
| `TaxTaken` | 245,141 / 860,531 | all | 0 |
| `Paid` | 129,341 / 201,431 | 27,854 / 36,965 | **101,487 / 164,466 (78.5% / 81.6%)** |
| `Distributed` | 34,971 / 48,654 | all | 0 |
| `Forwarded` | 32,436 / 34,054 | all | 0 |
| `Netted` | 25,418 / 34,128 | all | 0 |
| `DividendFunded` | 6,137 / 17,596 | all | 0 |
| `PrincipalParked` | 3,898 / 4,591 | all | 0 |
| `PrincipalDelivered` | 2,303 / 2,238 | all | 0 |
| `SnipeTaxApplied` | 1,444 / 10,642 | all | 0 |
| **total** | 481,089 / 1,213,865 | 379,602 / 1,049,399 | 21.1% / 13.5% |

**Four fifths of every `Paid` event on Arc under that topic0 is not an ArgusPad splitter payment.**
Almost all of it is the reward tracker's own `Paid`, which has the same signature and a different
meaning for topic2 - which is why the emitter is resolved before the log is decoded, and why decoding
on topic0 alone would silently book an address as a currency. (In these two windows the foreign forks
happened to emit none of the other nine events, so their drop count is zero here; the allowlist is
still load-bearing, since `TokenCreated`-shaped forks do exist on Arc and nothing stops them emitting
the rest.)

## The five traps, and where each is handled in the code

1. **`Paid(address,address,uint256)` is emitted by both `ArgusV4HookedSplitter` and
   `ArgusV4RewardTracker`** under topic0 `0x9def4e28…`, and topic2 means *currency* on one and
   *destination* on the other. Decoding one as the other does not throw — it reports an address as a
   currency. Every log is bucketed on its emitter's family before it is decoded; a tracker is simply
   not in the splitter map. (Reward-tracker `Paid` was 101,486 events in the day measured.)
2. **Foreign emitters.** Ten non-ArgusPad contracts on Arc run forks of this codebase under four
   different treasuries and emit byte-identical events; a single-day reconciliation excluded 164,466
   foreign `Paid`. The launch list is the allowlist, and it is built from what the ArgusPad Portals
   themselves announced, so it cannot be spoofed by a look-alike.
3. **The splitter taxes itself.** Its netting swap routes through the launch's own pool, so the
   launch's own tax is charged on it and taken straight back — `TaxTaken` reports it exactly like a
   user's tax. It is netted out as `Netted.amountOut × sellTaxBps / 10000`, the contract's own
   integer expression; that rule was checked against the observed `TaxTaken` on 23,699 of 23,699
   pairable netting sells, to the raw unit, and it was 4.16% of the quote `TaxTaken` total.
4. **`accountedQuote6` is not in 6 decimals** despite the name, and is a current liability rather
   than a cumulative total. It is not read.
5. **`Distributed` is not a parent event.** It is the crank's *arrivals*, emitted mid-crank after the
   treasury has already been paid, so it is never summed with `Paid`/`Accrued`/`DividendFunded`. Its
   `carried*` fields re-report the same money on every crank that carries it and are not read at all.

Two more the code has to get right:

6. **`PrincipalParked` means two different things** — the locker sent a residue back (which arrived
   through `Forwarded` and must come out of it), or `addLiquidity` reverted and nothing moved. Only a
   sibling `Forwarded` from that launch's own locker in the same transaction tells them apart.
7. **`Forwarded`, not `FeesCollected`, for LP fees.** `ArgusV4HookedLocker.addLiquidity` harvests the
   position and forwards without emitting `FeesCollected`, which understates the quote leg by ~1.5%.

## The launch-token side is excluded

Part of every fee is charged in the launched token. It is not counted, and the reason is measured
rather than asserted: marking it at each pool's own execution price (the method
`fees/wonk/index.ts:38` uses) gives **$1.81M** for the day, of which **99.87%** of the non-USDC part
comes from a single pool booking a fee ~500,000× its own window volume. The direct control — take the
tokens the protocol actually sold, value them the same way, compare with what those sales realised —
**overstates by 1.40×**. Wonk's method does not transfer because Wonk prices a fee that is a slice of
the very swap that sets the price, whereas this is a position-level accrual with no swap of its own.
Excluded and disclosed, matching `fees/tolly/index.ts:64-65` (*"excluded"*), `fees/coinbarrel.ts`
(*"the token-side fee on sells is not priced"*) and `fees/peach-launchpad` (*"Fees without a USD price
are excluded"*).

## `start: "2026-09-07"`

Measured, not assumed. The first Uniswap-v4 launch is block **19,674,581**, 2026-09-07 18:46:13 UTC,
and the first fee is charged on it 33 seconds later at block 19,674,645. The census dataset's own
`creationBlock` column gives the same block independently.

Two retired Uniswap **v3** Portals carry 59 older launches (first at block 18,819,901, 2026-09-02),
but that line has none of this fee machinery and emits none of these events, so it is out of scope and
`start` is the first date that actually returns data rather than the first date the brand existed.

## Pre-answering the review questions

- **Is a 0-fee day correct?** Before 2026-09-07, yes — no v4 launch existed. After it, no: the
  protocol has traded every day and a zero would be a bug, which is why every failure path throws
  rather than returning empty. Per *asset*, zero can be correct: on the day measured, WETH saw
  0.000409 WETH of tax taken but the splitter that received it never cranked, so a WETH figure of
  zero for that day is right, not missing.
- **Why do the numbers differ from the protocol's own dashboard?** ArgusPad publishes no fee
  dashboard. The team's own figure quoted elsewhere ("$1.36M undistributed") is a different quantity
  — undistributed launch-token piles awaiting sale — and is not comparable.
- **Did the fee rate or split ever change?** No. `treasuryBps` is 1000 on all 140,022 splitters,
  matching the contract constant; `buyTaxBps`/`sellTaxBps` are `immutable` per launch and
  `FeeConfigured` is emitted once and never amended. The crank-level weights move with pool price,
  which is why the split is measured per window rather than hardcoded.
- **Fee-to-volume ratio.** ArgusPad pools did $208,296 of 1% LP fee on Arc v4 that day, and the
  adapter's own measure of the same positions' delivered quote-side fees is $148,337. The two differ
  because Uniswap applies 1% to the USDC notional of every swap (both legs) while the protocol
  receives quote on buys and launched tokens on sells, and only the quote leg is priceable. They are
  two valuations of one flow.
- **Related to an existing listing or open PR?** Yes — the TVL adapter is a separate PR to
  DefiLlama-Adapters. No existing `argus`/`arguspad` entry exists under `fees/`, `dexs/`, `factory/`
  or `factory/deadAdapters.json`.
- **Refill needed?** No — new listing.

## Things a reviewer should push on

- **`noTarget` + raw `topics` are both on the `AGENTS.md` "Watch For" list.** They are used here for
  the reason the same file sanctions: 140,022 emitters. The raw topics are not a shortcut either —
  an OR-set of ten topic0s in one pass is what makes a day's scan affordable, and the `Paid`
  collision means a single `eventAbi` decode would be actively wrong. Every topic0 is recomputed from
  the human-readable ABI at module load, never pasted.
- **The revenue split is the one modelled step.** Everything else is a direct read. It is measured
  per window and per asset from an identity that closes exactly on chain, and validated two ways
  (it reproduces the configured `treasuryBps`, and it moved 0.25 percentage points across two windows
  whose activity differed 3.4×) — but it is a ratio, not a per-crank read, because the contract does
  not expose one.
- **Nothing here may be annualised.** The protocol is eleven days old and its activity is
  concentrated in three of them; the two windows below differ by 3.7×.


bheluga:
@degenanddreams thanks for the PR

PR #9565, opened 2026-09-19
[fees] wonk - price WONK-quoted launches off the WONK/USDC poolWonk price

jack-j7:
WONK is the launchpad's second whitelisted base and has no listed price, so
every launch quoted in it was dropped from the USD totals. New launches now
quote in WONK rather than native USDC, so this is the only path for them

Its price comes from the WONK/native-USDC pool (itself a Wonk Fun launch),
read from the v4 PoolManager's Swap events rather than live, so backfilled
windows price at the rate that applied then. Picked by (block, logIndex),
since that pool often trades several times per block


bheluga:
@jack-j7 thanks for the PR
WONK token is now priced: https://coins.llama.fi/prices/current/arc:0x548df4bf91624d8cec46d606211eb13f7492e27e , so it should get fixed automatically

PR #9568, opened 2026-09-19
fix: preserve lifi bridge adapter metadata

chetna-mittal:
## Summary

- Add explicit `pullHourly: true` to the LI.FI and Jumper bridge aggregator v2 adapters
- Preserve chain-level `deadFrom` metadata from the shared `LifiDiamonds` config when building adapter entries

## Why

Both adapters are `version: 2`, so they should explicitly declare hourly pulling. They also both build per-chain adapter entries from `LifiDiamonds`, but previously only copied `start`, which dropped configured `deadFrom` values for sunset chains.


bheluga:

PR #9576, opened 2026-09-19
fix(lunya-dex): swap logs came back nested by block range, adapter threw on every live window

anton-erwang:
Website: https://lunya.io
Twitter: https://x.com/lunya_io

Follow-up to #9485. `lunya-dex` has stored nothing since 2026-09-18. It throws `Cannot convert undefined to a BigInt` on every window that has a live pool.

**Cause.** With `maxBlockRange`, `getEventLogs` splits by block range before it splits by target. So `flatten: false` returns `range -> pool -> log`, not `pool -> log`. An hourly window on Arc is ~7,200 blocks, over the 5,000 cap, so it always splits. The fix takes the flat list with `entireLog` and groups it by `log.address`, the same way `fees/lunya-launchpad` already reads its logs. No other change.

**Checked.** Ran all 24 hourly windows of 2026-09-18 locally, one at a time:

| | Adapter | Raw `eth_getLogs`, USDC side |
| --- | --- | --- |
| Swaps | 52 | 52 |
| Volume | $2,337.02 | 2,329.37 USDC |
| Fees | $0.5687 | |
| Revenue | $0.1703 (30%) | |
| Supply side | $0.3980 | |

The 0.3% volume gap comes from pricing: the adapter counts the core-asset side, which is cirBTC on the cirBTC/USDC pool.

**Refill needed from 2026-09-15.** The stored 2026-09-16 and 2026-09-17 rows are 0. 2026-09-17 had 2 real swaps (~$2). Nothing is stored after that.

**Fee rate change.** USDC/EURC `0x97B0…9e8c` moved from 0.05% to 0.01% at block 21487379 (2026-09-18 10:58 UTC). The adapter reads `slot0` at each window's end block, so a refill reproduces this. The pool now falls under the $10k low-fee liquidity floor and clears it (~$12k). cirBTC/USDC `0xf098…1746` stays at 0.05%. Protocol share is 30% on both.

CI will probably still hit Arc's public-RPC 429s, as #9485 did. The local run above is the real check.


bheluga:
@anton-erwang thanks for the PR

PR #9582, opened 2026-09-19
options/paradex: report options trading fees

n4vx:
### What

`options/paradex` reported notional and premium volume only, so **Fees** and **Revenue** are blank for Paradex Options on the options page.

The data already exists in the same public Paradex Platform Stats dashboard the volume cards come from. The **Lifetime Fees** card (`dashcard/20068/card/21188`) breaks the daily fee take down by product:

`[TRADE_DATE, PERP_FEES, PERP_OPTION_FEES, SPOT_FEES, OPTION_FEES, TOTAL_FEE, CUMULATIVE_FEES]`

This PR wires `dailyFees` / `dailyUserFees` to the `OPTION_FEES` column. No new endpoint or dashboard change was needed.

### Revenue

Deliberately not reported. Paradex does not publish a supply-side/protocol split of its fees, so `dailyRevenue` is omitted and `skipBreakdownValidation` is set — the same call the sibling `fees/paradex` adapter already makes, for the same reason.

### Note for maintainers

`fees/paradex` (listed as **Paradex Perps**, Derivatives) currently pulls the *Paradex Daily Fees* card, which is the **whole-venue** total — perps + spot + options. So options fees are already included in the Paradex Perps figure, and this PR will surface them a second time under Paradex Options.

Over the last 30 days that overlap is ~$9.3k of a ~$24.8k total (38%); last 7 days it is 43%. Historically it is negligible — options only launched 2026-03-25, and all-time options fees are $30.7k against a $28.8M all-time total.

The clean fix is to re-base `fees/paradex` onto the `PERP_FEES` column of this same card so each child reports its own product and the parent total stays correct. I have left that out of this PR because it lowers the published Paradex Perps fee figure and that is the protocol's call to make, not mine. Happy to send it as a follow-up if you would prefer the two land together — just say the word and I will add it here instead.

### Testing

`pnpm test options paradex <date>`, cross-checked against the raw Metabase card:

| UTC day | Notional | Premium | `OPTION_FEES` (card) | Adapter |
|---|---|---|---|---|
| 2026-09-18 | 41.82 M | 222.17 k | 1748.86 | 1.75 k |
| 2026-09-17 | 13.85 M | 83.94 k | 854.30 | 854.00 |
| 2026-09-15 | 2.32 M | 23.98 k | 170.52 | 171.00 |
| 2026-08-24 | 4.49 M | 45.45 k | 447.15 | 447.00 |

Current (partial) day resolves through the existing rolling-24h path and returns fees day-to-date. A date before `start` is skipped as expected. `pnpm ts-check` passes.


n4vx:
@coderabbitai good catch on the `undefined` leak, fixed in 1c22aac — though I went the other way on historical dates.

Defaulting a past day to `0` would report a fabricated zero for a day that genuinely had options volume, and it would do so silently. This adapter was previously stuck on a discontinued product filter for months precisely because bad data failed quietly, so I'd rather it fail loudly: missing fees now throw alongside the existing notional/premium check.

The in-progress day does take your `?? 0`, since the fees card legitimately has no row until the first options trade lands.

For what it's worth the gap is hypothetical today: card `20068/21188` has a row for all 179 days since options launched on 2026-03-25, with no null `OPTION_FEES`. The guard is there for if that stops being true.

n4vx:
@coderabbitai right again, and this one contradicted my last reply. The parser coerced null to `0` before the guard ran, so the guard I added could never actually fire. Applied your fix in 632e48c83.

Behaviour is unchanged against live data (all 179 days since launch have a non-null `OPTION_FEES`), but the guard now works the way I described.

bheluga:
@n4vx thanks for the PR
Added [here](https://github.com/DefiLlama/dimension-adapters/pull/9585) to avoid name collision

PR #9587, opened 2026-09-20
fix: fail on invalid kelp protocol fee

chetna-mittal:
## Summary

- Remove the silent zero fallback for Kelp's `protocolFeeInBPS` read
- Throw when the protocol fee config is missing or malformed instead of storing misclassified fee data

## Why

The Kelp adapter previously caught failures from the protocol fee config call and left `protocolFeeRate` as `0`. That could silently record incorrect revenue splits by treating all rsETH staking rewards as supply-side revenue and no protocol revenue.


bheluga:

